PR #11, opened 2014-06-27
Add audio notifications

F1ynn:


tux3:
Small bug: It sets the volume to 100% and destroys my ears everytime. But maybe that's just my nondeterministic sound driver, can you reproduce ?

F1ynn:
Yikes! That doesn't happen on either of the two PCs I tested it on. Let me look up the docs for QSound and I'll see what I can do

tux3:
It's probably just my computer then, it's fine.

F1ynn:
Okay, interesting. I checked the docs and tested it on a third computer, didn't hear anything odd.

PR #26, opened 2014-06-28
Fix audio notifications, horizontal scrollbars

F1ynn:


tux3:
I still have the scrollbar squares, even with 13e24f2
I already cherrypicked 5986f88 and c43b612, not sure what e6fc25a is doing here.

ProMcTagonist:
Can we have subtle sounds for sending and receiving messages in an active window? Not the same as the alert noises for out of view chats. This would solve having to deal with window obstruction detection on multiple platforms, I think.

F1ynn:
ProMcTagonist: Sure, do you have free-as-in-freedom sound samples for us?

PR #28, opened 2014-06-29
don't upload binary to repo

benwaffle:
if you really want to use the releases page

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.pro.user*
+toxgui


tux3:
But why ?

benwaffle:
repo is for code

tux3:
Oh damn sorry I just understood. I though you were saying not to use the release pages or something.
Yes of course, you're right.

I just noticed that I uploaded a binary, but it was a mistake. Thanks.

PR #36, opened 2014-07-01
Disabled custom window borders, fixed minimized window alerts, re-added dark groupchat icon

F1ynn:


widget/widget.cpp
```diff
@@ -492,7 +618,7 @@ void Widget::onGroupWidgetClicked(GroupWidget* widget)
     {
         g->hasNewMessages = 0;
         g->userWasMentioned = 0;
-        g->widget->statusPic.setPixmap(QPixmap(":/img/status/dot_groupchat.png"));
+        g->widget->statusPic.setPixmap(QPixmap("img/status/dot_groupchat.png"));
```
tux3: All the images/ressources need to be accessed from the .qrc, the img/, audio/ and ui/ folders will not be shipped with the binary

widget/widget.cpp
```diff
@@ -12,13 +12,132 @@
...
+#include <QDesktopWidget>
+#include <QCursor>
+#include <QSettings>
 
 Widget *Widget::instance{nullptr};
 
 Widget::Widget(QWidget *parent) :
     QWidget(parent), ui(new Ui::Widget), activeFriendWidget{nullptr}, activeGroupWidget{nullptr}
 {
+    //qApp->installEventFilter( this );
```
tux3: Can be removed (?)

widget/widget.cpp
```diff
@@ -12,13 +12,132 @@
...
     ui->setupUi(this);
+
+    QSettings settings("windowSettings.ini", QSettings::IniFormat);
```
tux3: You can't just store/read files in the working directory, use QStandardPaths::writableLocation(QStandardPaths::ConfigLocation) instead

widget/widget.cpp
```diff
@@ -12,13 +12,132 @@
...
+    if (useNativeTheme)
+    {
+        ui->titleBar->hide();
+        //setWindowFlags(windowFlags() & ~Qt::FramelessWindowHint);
+        this->layout()->setContentsMargins(0, 0, 0, 0);
+
+        QString friendListStylesheet = "";
+        try
+        {
+            QFile f("ui/friendList/friendList.css");
```
tux3: please use the .qrc, don't access the files directly

widget/widget.cpp
```diff
@@ -12,13 +12,132 @@
...
+        foreach (QWidget *widget, widgets)
+        {
+            widget->setMouseTracking(true);
+        }
+    }
+
+    isWindowMinimized = 0;
+
+    centralLayout = new QHBoxLayout(ui->centralWidget);
+    centralLayout->setContentsMargins(9,9,9,9);
```
tux3: Isn't 9,9,9,9 the default ?

widget/widget.cpp
F1ynn: Oops, sorry about that. I merged the files by hand in gedit, apparently I missed a line.

widget/widget.cpp
```diff
@@ -557,3 +683,409 @@ bool Widget::isFriendWidgetCurActiveWidget(Friend* f)
...
+
+        if (moveWidget)
+        {
+            inResizeZone = false;
+            moveWindow(e);
+        }
+        else if (allowToResize)
+            resizeWindow(e);
+        //right
+        else if (xMouse >= wWidth - PIXELS_TO_ACT or allowToResize)
```
tux3, line 704: Is or supposed to mean || ?

widget/widget.cpp
F1ynn: On the two linux systems and one Win8 machine I tested, yes. The events are surprisingly buggy, using the standard functions for detecting window state doesn't work quite right.

widget/widget.cpp
F1ynn: You lose the one thing CSS is useful for (easy edits), but okay. Will do.

widget/widget.cpp
F1ynn: I believe so. I'm not changing those margins anywhere else any more so I can delete that.

widget/widget.cpp
tux3: The problem is that it makes packaging/installation harder, because we need special cases for every OS, and it breaks when people don't understand that the working directory has to be the git root (aka most people who compile themselves)

F1ynn: Good point, fixed now. As soon as I finish the settings path code I'll push the update.

widget/widget.cpp
F1ynn, line 704: Yup.

PR #60, opened 2014-07-03
Simple script for bootstrapping

msteinbeck:
This commit adds scripts for Windows and Linux/Mac OS X to provide an easy to use bootstrapping.

The Windows script will:
1. Create the directory "libs"
2. Download the latest libtoxcore version from: http://jenkins.libtoxcore.so/job/libtoxcore-win32-i686/lastSuccessfulBuild/artifact/libtoxcore-win32-i686.zip into the just created directory libs
2. Extract the zip archive
3. Remove the zip archive

The Linux/Mac OS X script will:
1. Create the directory "libs"
2. Clone the current master of libtoxcore into the just created directory libs
3. Run autogen.sh, configure with install prefix to libs, make -j 2 and make install
4. Remove the directory used for cloning libtoxcore

Unfortunately  Windows does not provide usable download and unzip shell commands. This commit also adds "wget" and "unzip" from http://gnuwin32.sourceforge.net/ with all its dependencies (some dll). These tools are stored in the directory "tools" and will automatically used by the bash script.

The scripts take care of "location independent invocation". So it doesn't matter where the scripts are started.

Last but not least this commit modifies the pro file using the fresh installed library at libs.

I think this makes things easier, since a developer can now just run:
1. git clone .... && cd ...
2. ./bootstrap.(bat|sh)
3a. mkdir build && cd build && qmake ..
4a. make
3b. Open the project and press compile


msteinbeck:
These scripts will also be useful for other external dependencies like hunspell for spellchecking...

msteinbeck:
This PR somehow depends on #59, since the current master of libtoxcore added a const correctness commit, which makes the current source not compilable.

PR #67, opened 2014-07-03
Fixed path to static libs

msteinbeck:
Somehow I missed to commit this line. Now Windows builds will work on every machine


msteinbeck:
This commit might break Jenkins again, because the path was resolved as follow: -ls/lib/... sorry for that stupid bug

msteinbeck:
As far as I can see irungentoo fixed this bug by overwriting INCLUDEPATH and LIBS with:
- echo 'INCLUDEPATH += ./libs/include/'
- echo 'LIBS = ./libs/lib/libtoxav.a ./libs/lib/libvpx.a ./libs/lib/libopus.a ./libs/lib/libtoxcore.a -lws2_32 ./libs/lib/libsodium.a /usr/i686-w64-mingw32/sys-root/mingw/lib/libwinpthread.a'

after he adds the current build of libtoxcore to libs by its own.
I'm really confused why using relative paths doesn't work inside the pro file.
Nevertheless I think (but not completely sure) that merging this commit will not break Jenkins build

PR #74, opened 2014-07-04
translations/ru: minor improvements

suhr:
Subj.


tux3:
Just so you know, you need to update the .qm file too, it's what is actually used at runtime.
Use the Release button in Qt Linguist.

But I'll just do it, this time.

PR #106, opened 2014-07-06
Added GPL header to all source files

dubslow:
It's plausible I missed a file or two…..


tux3:
Did you just give copyright for all the repo's files to the Tox project ?

dubslow:
You can of course change it as you wish. (sed is useful in these sorts of situations :P)

PR #114, opened 2014-07-06
Use a non locking circular ringbuffer for audio output

msteinbeck:
Based on: http://www.codeproject.com/Articles/43510/Lock-Free-Single-Producer-Single-Consumer-Circular
- The performance of the buffer is very good since it does not need any locking
- This buffer provides a push and pull api
- To get maximum performance use a size to the power of 2. Nevertheless it can be used with any sizes.

On the next days I will try to use this buffer for audio input as well. Furthermore I will add the acquire-release memory model to increase the performance.

Tested on Linux and Windows 7. As far as I can see the number of buffer underflows is lower compared to the old buffer but not completely removed.


tux3:
It doesn't work well for me when calling qTox <-> µTox
I've got 1-2 seconds of crackling then only white noise at full volume. 

Edit: But it does work when calling qTox <-> qTox, so this could be completely unrelated

Edit2: So yeah, in master qTox <-> uTox calls work, I can't merge that until it's fixed.

msteinbeck:
I tried several combinations, windows <-> linux, qTox <-> uTox.... and was unable to reproduce this bug. However I will postpone this PR for some days because the current solution works fine. Maybe someone else could test this branch and leave a comment whether it works or not. I'm really confused why this audiobuffer works fine with qTox but does not with uTox :/

PR #126, opened 2014-07-11
Add the files transfered Widget

dubslow:
For now it's quite simple, and it's lacking a bit of functionality: the ToxFile struct doesn't have any path information, so clicking on the files fails to open them (excepting the hack for downloads, which would be broken if the client allowed custom download locations).

So, if anyone knows more about file transfers and can somehow locate the full path for the widget, that would be great.

Also Flynn (or whoever does these things), a better checkmark icon wouldn't be bad (and I'm not even sure it works currently).

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -563,6 +563,17 @@ void Core::removeGroup(int groupId)
     tox_del_groupchat(tox, groupId);
 }
 
+QString Core::getUsername()
+{
+    int size = tox_get_self_name_size(tox);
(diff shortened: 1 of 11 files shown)


tux3:
You added img/checkmark.png to the .qrc, but you didn't commit it, so I can't compile.

And there's already a checkmark at ui/acceptFileButton/default.png if that's what you were looking for.

dubslow:
...how about now? lol

tux3:
Works for me :)

PR #136, opened 2014-07-13
mac os build settings, audio io refactoring initial stage

skotopes:
- Mac OS build settings. Most significant change is that dependencies provided by brew grabbed from /usr/local. Plus dock icon.
- Audio Input/Output Proxy with ring buffer. Prototype. I'm still working on it.


msteinbeck:
Is this buffer thread safe for single producer single  consumer?

skotopes:
Memory ring itself is 100% thread safe. I'm using it in my projects for a long time.

There is one potential problem(caused by Qt way of stream handling) with pushing data into almost full buffer.
In this case 1-byte shift could occur in stream. I'm working on clean version now.

msteinbeck:
> Memory ring itself is 100% thread safe. I'm using it in my projects for a long time.

Very nice. I worked on a non locking ringbuffer for audio myself, but tux3 said it produces random noise which I was not able to reproduce. Maybe the reason is that you use different proxies for input and output. It would be very nice if yours is working.

Why did you remove the line

 LIBS += -L$$PWD/libs/lib/ -ltoxcore -ltoxav -lsodium -lvpx

for Mac OS X in the projects .pro file? Some people may use the bootstrap.sh script which installs the right version of toxcore and sodium to $$PWD/libs/lib/. As far as I can see your commit breaks this feature. I think it would be useful if you extend the LIBS path with /usr/local/lib instead of completely replace it.

PR #145, opened 2014-07-15
Add tooltips to user-provided text (helps to accommodate too-long names)

dubslow:
Ideally this would be done only if the widget is too small for the message, but
that can be quite difficult

Additionally, it doesn't cover the case of having a Tox ID for the name... eh

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -23,7 +23,7 @@
 QT       += core gui network multimedia multimediawidgets
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
-TARGET    = toxgui
+TARGET    = qtox
 TEMPLATE  = app
 FORMS    += widget.ui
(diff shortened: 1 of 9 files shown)


dubslow:
This is now two independent commits in one request

qtox.pro
```diff
@@ -23,7 +23,7 @@
 QT       += core gui network multimedia multimediawidgets
 greaterThan(QT_MAJOR_VERSION, 4): QT += widgets
 
-TARGET    = toxgui
+TARGET    = qtox
```
tux3, line 26: I need to make sure this won't break everything before merging. (spoiler: it will)

dubslow:
This is now three independent additions in one request. The third fixes #151. Note that the original reason for not merging has been (supposedly) fixed by nurupo.

PR #167, opened 2014-07-27
Randomized the bootstrap table, to distribute load away from the first 5

dubslow:
Also added a tooltip


dubslow:
This is now two independent pull requests, but because GitHub's PR web interface is limited, we are stuck with one thread for both.

Issue #172, opened 2014-07-29
Bundle default smiley pack


msteinbeck:
qTox should bundle a smiley pack which is enabled by default.

PR #180, opened 2014-08-05
Smileys

krepa098:
A default set of smileys. This should fix issue #172.


tux3:
Are you sure it's okay to overwrite the copyright ?
Did you ask nurupo ? He might sue us.
Note that I don't know the first thing about licenses and copyrights.

dubslow:
Those two files were written exclusively by him; he merely copied the wrong header from other files. Here he corrected it to the right header.

(That is to say, nurupo never had anything to do with these particular files.)

krepa098:
Exactly. I just copied the wrong header by mistake.

tux3:
My bad !
All is well.

PR #196, opened 2014-08-19
Volume control

mdnight:
There was implemented microphone muting for calls. Volume control is not ready yet


tux3:
Why are the icons clipped / not rounded on your screenshot ?

Your branch also appears to have conflicts with master, could you please rebase it on master ?

PR #225, opened 2014-08-31
Fixed build for OS X

### LINKED ISSUE #1: Error compiling qtox with homebrew
https://gist.github.com/3a35560668e2f74e67e5.git

Following directions here:

https://github.com/tux3/qTox#osx-easy-install

ReDetection:
This will also fix ReDetection/homebrew-qtox#1

### DIFF
--- a/core.h
+++ b/core.h
@@ -20,8 +20,13 @@
 #include <tox/tox.h>
 #include <tox/toxav.h>
 
-#include <AL/al.h>
-#include <AL/alc.h>
+#if defined(__APPLE__) && defined(__MACH__)
+ #include <OpenAL/al.h>
+ #include <OpenAL/alc.h>
+#else
+ #include <AL/al.h>
+ #include <AL/alc.h>
+#endif
 
 #include <cstdint>
 #include <QDateTime>


tux3:
And bam : Broken jenkins.

ReDetection:
Just found jenkins. :)
I don't understand what you mean. Jenkins os x build is OK now (and was OK when there was include issue o_O ), win builds and linux builds are failing for a long time. o_O 

It seems, Jenkins are working on the OS X too.. I do not know too much about cross-compiling from OS X. I thought it should include AL/al.h when building for them, shouldn't it?

tux3:
> It seems, Jenkins are working on the OS X too.. 

Yeah, I'm surprised it didn't break. But I'm not complaining!

> win builds and linux builds are failing for a long time. o_O 

You can go yell at NikolaiToryzin on #tox-dev @ irc.freenode.net if you want them fixed, I'm sure he'll appreciate it :)

ReDetection:
Oh, I misunderstood :)
And also it seems like I misunderstood tox's jenkins setup. And now I see that `tux3-toxgui` task just downloads sources and triggers build for each platform. Task `ToxGUI OSx X` did break but after some failures it began to build (but with failures in log, seems like there is a bug in the buildscript but I can't see buildscript). I can't see differences between failed and passed build so I think someone just created symlink OpenAL → AL.

Issue #237, opened 2014-08-31
OS X not working (not connecting to DHT)


Tamriel:
After adding a friend, nothing happens. He is shown offline, although he is online. Can anybody confirm qTox is working on OS X right now?

tux3:
Are you shown online ? (is the dot near your name green and not red ?)

Tamriel:
On startup, it's red, but then I click on it to go online. Then it's green.

tux3:
You shouldn't do that. There is no "safeguard" in qTox yet, but when you do that it doesn't actually do anything. It should become green on it's own, or else it means your computer couldn't connect to the Tox network.

If you try restarting and not touching the red circle, does it go green eventually ?

Tamriel:
I waited 10 minutes, it did not turn green.

tux3:
Ok, I'll be working on a fix. Thanks for reporting the bug.

dubslow:
@Tamriel, it's probably worth trying again after that commit -- but you'll need to delete your settings.ini file for the change to take effect.

I can't say for sure if it will help or not, but I believe it will.

Tamriel:
I can't find that settings.ini file.

dubslow:
I believe it's in $HOME/.config/tox.

Tamriel:
It's not :/

mouseym:
Looks like this bug is being reported more and more often.

It certainly works for me on 10.9.4, initially it took a long time to connect but after that one time it now connects immediately each time i start qTox.

davydotcom:
We had a one direction connection issue. I sent a request, It was received but user was not shown as online nor could we communicate. We dropped the request and he requested me. He got a "Error while looking up DNS" error. Then finally it just kicked in.

tux3:
This should be fixed in the latest qTox.

mouseym:
connects almost immediately now, can't seem to successfully add groupbot or audiobotze however, groupbot is reported to be online by others...

Edit: no longer connects to the network... :(

zetok:
@mouseym audiobotze was mostly offline due to constant crashes, sorry

Tamriel:
Any progress on this? 
I still can't find my settings.ini.

tux3:
The settings.ini should be in "~/Library/Preferences/Tox"

Tamriel:
Thanks! I deleted that folder. Tox still does not connect for me (version of today).

mouseym:
Still an issue alright.

[...]

Issue #221, opened 2014-08-29
Can't connect to DHT on Mac OSX


tux3:

tux3:
Can't reproduce / Works On My Machine™ / Not enough information

https://i.imgur.com/MINKYlY.png

tux3:
Reopening since more people are affected.

tux3:
Log : http://pastebin.com/LsqBUBPF

tux3:
Fails even with IPv6 on.

mouseym:
I have managed to bootstrap off a node (eventually) on OSX 10.9, no idea what version of qTox (its one made after these changes were made).

tux3:
Ok. qTox was trying to bootstrap way too many nodes at once, now it's a bit more gentle.

PR #240, opened 2014-09-01
Update bootstrap.sh, match nodes to wiki.tox.im

dubslow:
This might help #221 or #237, several of the old nodes are probably offline


dubslow:
Here's my log that caused me to update the node list:

https://gist.github.com/dubslow/3582eba9e42ee6bce3b6

Note that ChauffeR is no longer on the wiki, yet toxcore was apparently able to successfully connect to the node ([otherwise](https://github.com/tux3/qTox/blob/master/core.cpp#L783) the log would have the error message instead).

The same goes for astonex on the retry; his ip and id are different now, yet apparently toxcore was able to connect to the old node.

This suggests that toxcore doesn't sufficiently ensure that nodes are operating properly (i.e., up to date).

dubslow:
Wait, even this won't help, because it doesn't overwrite the old settings file...

msteinbeck:
According to: https://github.com/irungentoo/toxcore/blob/master/INSTALL.md libsodium in version 0.5.0 should be used. Maybe the documentation is outdated or it doesn't matter, but I decided to take this tag, because the instructions do so.

dubslow:
No, several weeks ago (the time of the first commit here, which I proceeded to forget about) I asked irungentoo, and he (and some others) said he used 0.6.1, so I updated bootstrap here. INSTALL.md is just as out of date as bootstrap.sh currently is.

tux3:
Just because qTox doesn't immediatly connect to the DHT doesn't mean that the nodes are offline, as far as I know. It's probably a good idea to match the wiki tho.

Issue #204, opened 2014-08-27
No new features until all bugs are fixed.


tux3:
This ticket will serve as a list of gaping flaws to fix and the directions to go to get them fixed. No major features or new bugs should be introduced until the codebase is sane again.
- The file-transfer code doesn't manage ownership of the currently transfered file in a sane way. The objects avoid use after free and leaks only 'thanks to' bad hacks and heuristics. The file transfer's metadata should perhaps be owned by the file transfer widget. Find if the problems with large files happen due to overflows, mem management issues, threading and signaling problems, bad interpretation of the protocol, or just all of the above. <b>Fixed</b>
- The chat box/view/form is completely broken. Both implementations are ridiculous, one is spamming layouts and elements, making the GUI painfully slow and making it impossible to select text accross message boundaries. The other is a disgusting pile of hack on top of other hackets that create visual glitch and are very much using features in a way explicitely not intended by the author. <b>Fixed</b>
- Audio has unsolvable bugs. Need to come up with a solution, perhaps use a decent audio library instead of Qt's barely documented/tested audio that requires packages nobody has at runtime on *nix and has weird segfaults involving a pulseaudio plugin running in the wrong thread. And fix the broken audio scheduling and threading. Just get rid of Qt/PA audio, get OpenAL-soft. I'm done with Qt's audio.<b>Fixed</b>
- Video is broken and innefficient probably due to the custom video conversion fonctions. Consider using a real library, and not using Qt's completely broken and not even finished Video API (crash due to missing internal functions on windows, difference in support accross platforms, makes it painfull to share video output between objects, ...). Use a real library, unless there is absolutely positively nothing better for video acquisition. Conversions will require a lib anyway, I'm not going to write high perf implementations for all possible formats your webcam can decide to support. <b>Probably fixed</b>
- Groupchats being nondeterministic. I still have no idea if it's toxcore being toxcore, or some mis-alignment of the stars that make it work _sometimes_ and not work other time with no observable external difference in behavior, except for a blatant lack of acceptance from the group. Perhaps auto-resending a couple of time after a failure to join would be an acceptable work around. <b>Core issue, can't fix</b>
- Death by a thousand cut. qTox is not even remotely polished. There isn't enough attention to details and little-things-that-make-the-difference. See the various other tickets and go use/fix it while being critical of every little thing until it's just comfortable to use. And then some.

zetok:
About group chats, current ones are simply broken, and being broken includes not being able to work over TCP.

[...]

PR #254, opened 2014-09-02
package version is more appropriate

dubslow:


debian/changelog
```diff
@@ -1,4 +1,4 @@
-qtox (1.0-1) UNRELEASED; urgency=medium local package
+qtox (0.01pre-alpha-1) UNRELEASED; urgency=medium local package
```
tux3: > 0.01pre-alpha

Well, this is technically a valid upstream_version, but :
- 01equals 1, so we should write 0.1pre-alpha instead of 0.01pre-alpha
- It's a good idea to use ~ instead of -, because - is allowed only if there's a debian_revision (there is currently, the -1), and ~ sorts before anything, even the empty part, so 1.0~prealpha is a lower version than 1.0 for debian. That's not true with -.
- qtox is not _that_ buggy! I like my 1.0.

tux3:
Can you send the new debian/control description in another PR ? I'd like to merge than in and we can keep bikeshedding about version numbers in this issue if you want.

msteinbeck:
> I like my 1.0

I think qTox is ready for 1.0 if #204 is fixed.

tux3:
Let's call it 1.0~beta1 or something then. Then we can make as many betas as necessary until #204 is fixed.

dubslow:
I don't mean that it's buggy (though there are some pretty big bugs, e.g. your big issue), but rather that things are still very subject to change. 

Even if it was provably entirely bug free (hypothetically speaking), I still wouldn't call it "stable".

1.0 to me, implies any sort of stability, which I think at this point is a lie. toxcore's API is still liable to change, much less our UI. Frankly versioning itself is perhaps a bit premature imo, at least until toxcore starts doing it -- but we need something, so I changed it to something I thought more appropriate. At any rate, I believe that at a minimum the version should be below 1.0.

(As a side note, using two decimal places (e.g., 0.52->0.53, 1.22->1.23) instead of more than one decimal (3.3.2 -> 3.4.0) is a reasonably common practice, but it's utterly unrelated to the primary discussion.)

Edit: Fixing _all_ of #204 would make it worthy of alpha or maybe even beta status (at which point versioning starts to become meaningful).

Edit 2: No toxcore\* or related software (like qTox) should be called anything beta or better until feature parity with Skype is met -- i.e., rewritten groupchats, as well as group audio/video.

tux3:
In my opinion 1.0 means that the project is finally usable, and you can start distributing shamelessly. All the important features are in, and it's what it's supposed to be : an IM client. Just because there's a 1.0 doesn't mean it stops there, not even close. GCC is talking about a version 5.0 after 4.9 (instead of 4.10), Firefox is at version 32 or so, etc.

In the "traditional way" major number changes are supposed to mean breaking changes, API incompatibilities, etc. And <1.0 is highly unstable alpha software. But imho we've passed this point, the GUI won't change much even if toxcore does, and new features shouldn't break anything anymore.

> (As a side note, using two decimal places (e.g., 0.52->0.53, 1.22->1.23) instead of more than one decimal (3.3.2 -> 3.4.0) is a reasonably common practice, but it's utterly unrelated to the primary discussion.)

Yeah, but usually you start with one decimal and switch to two when you reach 10. E.g. linux 2.X and 3.X versionning.

> Edit 2: No toxcore\* or related software (like qTox) should be called anything beta or better until feature parity with Skype is met -- i.e., rewritten groupchats, as well as group audio/video.

I agree that 1.0 without feature parity with Skype isn't quite right, but if qTox 1.0 is defined as a client using all the features of tox{core,av} 1.0 without bugs, we're in for months or years of pre alpha. Toxcore moves slowly, imho qTox can still be a useful and (mostly) stable tox client, even if toxcore doesn't have video groupchats with personalized avatars quite yet.

dubslow:
Fair enough. At any rate, I did modify this PR to be just the description.

PR #264, opened 2014-09-07
ChatForm

apprb:
Reimplemented version of chat widget. Instead of placing QLabel into the chat area all messages presented in HTML format as well as file transfer widgets. Interaction with transfer widget implemented through catching mouse clicking on button images by the chat form widget and then passing this events to corresponding file transfer widget.
The things that should be corrected:
1. A message format is hardcoded. Eventually should be read from file (XML, or smth. like that) in order to allow change it by the user.
2. Word wrap problem: veeeery long words are wrapped, but username and timestamp become squeezed. Probably program should forcefully break long words by placing &#8203;.
3. Filetransferwidget: progress bar should be added, minor resizing problems.


dubslow:
:+1: :+1: :+1:

PR #265, opened 2014-09-07
Documentation for Windows

msteinbeck:
I Added the file INSTALL.md which contains a howto for Windows and Linux. Unfortunately I have no Mac and was therefore not able to write documentation for it. Maybe someone could complete the section for Mac OS X. Furthermore I fixed the pro file for Windows.


dubslow:
Why list Debian and Ubuntu seaparetly when their instructions are identical?

Also, Clang compiles qTox just fine; gcc isn't technically a requirement.

msteinbeck:
> Why list Debian and Ubuntu seaparetly when their instructions are identical?

When I started to write this howto I thought Debian and Ubuntu needs different packages. I think it's ok to leave it as it is, because technically Debian and Ubuntu are different distributions with different repositories.

> Also, Clang compiles qTox just fine; gcc isn't technically a requirement.

Currently GCC feels like the default compiler for the most Linux distributions. For example GCC is part of the metapackage build-essentials, Clang is not. Furthermore MinGW for Windows is a port of GCC and thus I think it is more consequent to use GCC in this howto. At least writing a howto for different compilers is more complex :).

tux3:
This PR was manually merged in, since there were conflicts.

PR #284, opened 2014-09-12
"rebased" Settings Dialog

msteinbeck:
same description here


tux3:
I merged it, but the video test is horribly broken : It shows a still frame of garbage (weird), the button says "enable video" wich isn't quite right (makes it sound like if you don't click on it video will be disabled), and closing the settings with the camera enabled doesn't disable it.

msteinbeck:
> I merged it, but the video test is horribly broken

I tested it on my device without any issues. It just uses the SelfCamView

> the button says "enable video" wich isn't quite right (makes it sound like if you don't click on it video will be disabled)

Yeah that's right, not the best choice :)

>  and closing the settings with the camera enabled doesn't disable it.

I use the close method of SelfCamView. On my notebook this closes the webcam :(

tux3:
> I tested it on my device without any issues. It just uses the SelfCamView

Yes, I'm not sure why it's doing that. I'll test with an older commit see if it's my linux failing.

> I use the close method of SelfCamView. On my notebook this closes the webcam :(

Oh you're right, it works now that I tried again. I must have dreamt it, sorry :)

tux3:
Ok, I can confirm that on my computer the video fails even before merging this, now I just need to figure out why.

PR #294, opened 2014-09-15
Newsettings infrastructure done

dubslow:
Please create a new branch (and delete the 4 others that are no longer needed). The infrastructure is complete, all that remains is to add options (and wire those options to settings), and make it look pretty (which it certainly does not yet). 

The header space can have the new images that retuxx added (haven't gotten around to adding it myself), and as far as general settings goes, I would move that the redundant and useless name and status line edits be dropped. (That is something I did long ago in my profiles branch, but that obviously hasn't made it to master yet.)

For some reason the last commit causes a double free or other such corruption in Widget::~Widget(), but the total effect of that is to spew some console output upon closing qTox.


msteinbeck:
> I would move that the redundant and useless name and status line edits be dropped

I would not, some users may not know how to change name/status in main ui. The settings widget should contain all settings.

dubslow:
> some users may not know how to change name/status in main ui

I'd like to meet those users....

iShift:
@retuxx +1

tux3:
> For some reason the last commit causes a double free or other such corruption in Widget::~Widget(), but the total effect of that is to spew some console output upon closing qTox.

Nope, so much nope. I'm fixing this before it goes into master, even if it's just in a destructor.
Also, your PR conflicts with master, I'll fix it if I have the time tommorrow or you can do it now.

dubslow:
Like I said, I was expecting this to go into a new branch and get cleaned up a bit graphically before going to master. I was hoping for help, because the last thing I know how to do is make QWidgets behave. 

As for the destructor, well... even with debugging symbols I have no idea what it's cleaning up that it shouldn't. The backtrace has two frames in ~Widget(): one line number of the opening paren, and the line number of the close paren.

tux3:
> As for the destructor, well... even with debugging symbols I have no idea what it's cleaning up that it shouldn't.

I haven't looked at the code, but remember that QWidgets clean up their children automatically, you only delete the parents and never delete the children, the parent does it for you.

dubslow:
> I haven't looked at the code, but remember that QWidgets clean up their children automatically, you only delete the parents and never delete the children, the parent does it for you.

I didn't put a single delete statement anywhere in this entire PR.

tux3:
Oh you used that "form" thing I did instead of a QStackedWidget, that's pretty hard to get right and I'm not sure it was such a good idea in the first place. I remember having the same crashes when I did it.

IIRC the problem is that your QWidgets live on the stack as part of another object, but have a QWidget parent at the same time, so they are owned by two things (at the same time!). The parent will delete it along with all it's children recursively in it's dtor, and the object it's in will also delete it in it's own dtor, hence the double free.
So this crash would happens whenever you delete a "form", not necessarily on exit (it could happen when you close the settings dialog for exemple, if we were to free the widgets to save some RAM).

TL;DR: I'm rambling, just put those widgets on the heap!

tux3:
Rebased on master and pushed into <a href="https://github.com/tux3/qTox/tree/newsettings">newsettings</a>

Issue #130, opened 2014-07-12
Force video output to follow aspect ratio


tux3:
As requested by Roman

apprb:
Can be closed, I guess.

PR #297, opened 2014-09-15
fix: Force video output to follow aspect ratio #130

apprb:


tux3:
This needs testing / profiling, I'll do it later.

PR #302, opened 2014-09-21
Split message

apprb:
This function splits a long message in the nearest space character (if possible), or just in the middle of a word (as uTox).


msteinbeck:
nice

Issue #306, opened 2014-09-23
Notifications when message not sent


mouseym:
When sending a message and having connectivity issues there is nothing to show that the message didn't go through.

This was found yesterday when Groupbot wouldn't invite me to the chat, only to discover that my first two messages hadn't actually gone through.

This is on OSX by the way.

tux3:
Part of the problem is that qTox doesn't stop you from sending messages even if your contact is disconnected.
But I also don't remember seeing a way in toxcore to know if a message failed

mouseym:
void tox_callback_read_receipt(Tox _tox, void (_function)(Tox *tox, int32_t, uint32_t, void *), void *userdata);

would a read receipt work?

apprb:
uint32_t tox_send_message(Tox *tox, int32_t friendnumber, const uint8_t *message, uint32_t length)
return 0 if it message wasn't successfully put put into the sending queue. I don't know how many cases it may cover, but if you or your friend aren't online this function return 0 (just tested).

As for group messages:
     /\* send a group message
      \* return 0 on success
      \* return -1 on failure
      */
     int tox_group_message_send(Tox *tox, int groupnumber, const uint8_t *message, uint32_t length);

This may help.

tux3:
Yes, checking that would be a good start. I'm not sure what happens if the network or peer goes down just after the message was added to the queue though, is there some sort of faux-offline messaging ensuring it'll be received ? Will toxcore give up and if so can it be checked ?

But yeah, checking this and disabling the buttons when the peer is offline first.

Issue #53, opened 2014-07-02
Give a text notification in chat box when file transfer fails / friend isn't online to receive messages?


TLausmaa:
I think it would be good for the client to react to these cases and give the user some info.

tux3:
This is planned. It shouldn't be possible to send anything to someone that isn't online (until offline messaging works), and file transfer is being fixed.

dubslow:
can be closed

PR #307, opened 2014-09-23
Notifications

apprb:
Notifications (only error messages as for now) in the chat area are implemented.
This patch covers #53  and partially #306


apprb:
Yes, I love tables (SystemMessageAction) ;-)

PR #317, opened 2014-09-25
Progress bar in file transfers (fix #309)

### LINKED ISSUE #309: [Regression] Missing progress bar in file transfers

apprb:
Default width now isn't full line.


apprb:
Probably it is better to draw whole filetranfer widget as an image, I don't know.

PR #321, opened 2014-09-26
improved avatars

krepa098:
Avatar rendering now matches the mockup


tux3:
I'm experiencing some bugs with this PR.

Avatars in the contact list break partially or disappear completely until you hover on them, then break again if you scroll or wait.

pic19
Notice how Root and Impyy's avatars are broken and zetok's avatar is cut at the top.

krepa098:
Sorry, I fixed it :)

tux3:
Now the pic in the chat form header looks strange. It's not just the corners being rounded, but it looks like the whole pic is rounded.
And in the contact list the rounded corners are barely visible, unlike on the mockup.

pic20

Sorry for being nitpicky :)

krepa098:
Well your friend's avatar is too narrow. The only solution I can think of is adding some sort of background. Any ideas?

It works using sufficiently large avatars:
snapshot4

I think I'll remove the mask for groupschats though.

tux3:
According to ItsDuke's Tox-UI (is that the latest mockup ?), the image in the chat form header and in the contact list should be the same.

In your pic the image in the chat header is completely round, instead of just having round corners.

krepa098:
I was looking at the mockup on tox.im but it seems like it's not up-to-date.

That's how it looks now:
snapshot5

tux3:
Well, this time you broke transparency !
But otherwise everything looks fine.

pic21

krepa098:
That's how it looks now:
snapshot6
snapshot7

tux3:
It still doesn't handle transparency, those white bands around some avatars shouldn't be there.
And the default contact icon looks a bit less clean now.

pic22

krepa098:
> And the default contact icon looks a bit less clean now.

Yes, there is no dark version of avatar.png. contact_dark.png is blurry (the same applies to group.png and group_dark.png).

> It still doesn't handle transparency, those white bands around some avatars shouldn't be there.

That was actually the intended behavior for avatars not having the right size. I just changed scaling to KeepAspectRatioByExpanding:

snapshot9 to 
snapshot8

tux3:
Do you mean that the white bands are the intended behavior ? Because imo it doesn't look so good.
And if the avatar is stretched with the ratio ignored, it can also look weird.

Can't we just replace the white bands by transparency, like in the first pic I posted ?

krepa098:
> And if the avatar is stretched with the ratio ignored, it can also look weird.

No it's not. It's more like zoomed in. http://qt-project.org/doc/qt-5/qimage.html#scaled

> Can't we just replace the white bands by transparency, like in the first pic I posted ?

I personally don't like that because you get no round corners and no uniformly sized avatars.  I think the current code should look ok, even for poor _Root_ ;). Maybe we can get more opinions about that?

tux3:
Oh KeepAspectRatioByExpanding, right, sorry.

Then that's perfect, it's just that when I compiled it looked like KeepAspectRatio, with white stripes.

Issue #324, opened 2014-09-27
File transfer widget looks bad when file name is long


zetok:
![2014-09-10 qtox long messages not displayed correctly](https://cloud.githubusercontent.com/assets/3148759/4430710/b276bf32-4640-11e4-82fd-e7b17d64f525.png)

As a bonus, there's problem with timestamp (#323).

dubslow:
also closeable

zetok:
Still present on newest qTox when picture is received (3rd file transfer), also causes scroll bar to appear and "hide" timestamp:
![2014-10-01 steam borderlands 3 pucharse](https://cloud.githubusercontent.com/assets/3148759/4514853/1675276e-4b94-11e4-82e5-9459fe6541f3.png)

apprb:
Problem from issue #372

zetok:
Seems to be now fine, except for missing image miniature

PR #328, opened 2014-09-27
long names

apprb:
Long name in filetransfer widget: fixed
Word wrap in the chat form: time stamps: OK, names: partial

### DIFF
--- a/filetransferinstance.cpp
+++ b/filetransferinstance.cpp
@@ -34,6 +34,12 @@ FileTransferInstance::FileTransferInstance(ToxFile File)
     remotePaused = false;
 
     filename = File.fileName;
+
+    QFont font("sans");
+    font.setPixelSize(10);
+    QFontMetrics fm(font);
+    filenameShort = fm.elidedText(filename, Qt::ElideRight, 250);
+
     size = getHumanReadableSize(File.filesize);
     speed = "0B/s";
     eta = "00:00";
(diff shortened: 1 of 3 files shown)


filetransferinstance.cpp
```diff
@@ -34,6 +34,12 @@ FileTransferInstance::FileTransferInstance(ToxFile File)
     remotePaused = false;
 
     filename = File.fileName;
+
+    QFont font("sans");
```
tux3, line 38: What if "sans" isn't available ? What if it's available but not what you expect it to be ? 

Is this just supposed to load Helvetica ? 
http://qt-project.org/doc/qt-4.8/qfont.html#fontmatching

filetransferinstance.cpp
```diff
@@ -41,9 +47,12 @@ FileTransferInstance::FileTransferInstance(ToxFile File)
     {
         QImage preview;
         File.file->seek(0);
-        if (preview.loadFromData(File.file->readAll()))
+        if (File.file->size() <= 1024*1024*25)
```
tux3, line 50: Well, that's not a bad idea, but I'd appreciate if you mentioned it in the PR description or even just in the commit message. That's completely unrelated to long names and fixing #324.

apprb, line 50: My bad. Completely forgot that I've done this fix too.

filetransferinstance.cpp
apprb, line 38: I looked through code and it looks like just `Qfont font` is enough (for example: addfriendform.cpp line 32).

apprb:
So it is better to make this PR again than try to correct this one.

PR #342, opened 2014-09-29
implement avatar hash caching instead of getting full avatar each time

dubslow:
Jfreegman can attest that I tested this


dubslow:
I added two small commits here, they're very small

dubslow:
Fixed the file move, should compile now

tux3:
Yeah I was looking at the code just before you fixed the includes, I was a bit confused.

misc/ isn't a very descriptive name, but it beats nothing at all I guess.

Issue #351, opened 2014-09-30
File transfers exceed physical limitations


zetok:
When receiving file from older µTox version:
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4462671/f0712f76-48c4-11e4-9609-03dc22300462.png)
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4462672/f224e5b0-48c4-11e4-81dd-f3d49b3a0452.png)

It speed up to over 1.5 GB/s...

dubslow:
My bad, I can't debug right now but I'll get around to it within a day hopefully

apprb:
I'll do it.

dubslow:
Twas https://github.com/tux3/qTox/commit/a4d42a713621344a58b1cd2aad02aface0dd422d

zetok:
There's still similar behaviour after resuming file transfer

Issue #266, opened 2014-09-08
Ridiculously big CPU and RAM usage


zetok:
![2014-09-08 qtox ram usage messages](https://cloud.githubusercontent.com/assets/3148759/4189135/442c05d0-3779-11e4-89e4-aa12e7628e4f.png)

Apparently qTox doesn't handle well receiving a lot of messages.

I've sent thousands of lines of text to qTox, probably not more than 100k.

During, and after qTox remained hanged.

Tested on https://github.com/tux3/qTox/commit/cfb3313821e28a3f40c9712c5f7b675f796e5e7f

tux3:
That probably has to do with the layout/widget spam, it'll be fixed Soon™

zetok:
After #264 got merged RAM usage dropped significantly (only few MB increase during test), but unfortunately CPU usage stays the same, and after receiving 10k lines sample qTox stays hanged.

tux3:
Do you use some sort of script I could use to send the tens of thousand of lines ?

It should be easy to run that in a profiler.

zetok:
Nah, I'm just using Toxic.

Copy content of some big file and paste into Toxic. Since new versions support faux offline messaging and read recipes, one knows whether messages were really received.

tux3:
Profiling shows that there is a massive slowdown in chatareawidget.cpp, updateChatContent()
The call to setHtml() alone is responsible for >60% of the CPU used by qTox over the whole duration of the test.

tux3:
Please comment if you still have this bug and I'll reopen. 
If you notice any strange behavior of the qTox chat after the commit that closed this issue, please also report it since I was particularly tired when I wrote the fix and can't exclude the possibility that I may have introduced new bugs.

zetok:
Situation improved, after sending 10k lines of text qTox only has increase of RAM usage up to 419 MB. While receiving for a short moment became unresponsive.

I'm going to test how well it'll behave with 10 million lines sample..

dubslow:
10K lines of text taking up 419MB is still rather a lot...

tux3:
That last patch kind of broke everything. I'll make a real fix.

dubslow:
I think the issue is [getHtmledMessages](https://github.com/tux3/qTox/blob/737e76c72fdd966014744b17b80360d9e6258c71/widget/chatareawidget.cpp#L74). Everytime it's called (i.e. everytime updateChatContent is called), it constructs the entire chat history via message-by-message string addition. 

[...]

PR #352, opened 2014-09-30
Fix: #351 (File transfers exceed physical limitations)

apprb:
Also resolve hight CPU usage problem during file transfer (was mentioned in #266).


apprb:
Oh, no. Wrong branch.

Issue #334, opened 2014-09-28
[Regression] "Default" avatars look disgusting


zetok:
Current ones:
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4433983/89f5ef3c-46fa-11e4-9db3-f1ac3ccb738d.png)

Previous:
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4433984/a7af360a-46fa-11e4-8d2d-a152a48a3161.png)

zetok:
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4434036/66621a3e-46fd-11e4-9d6e-fd061f872aa6.png)
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4434040/725d9638-46fd-11e4-9207-9429b8cf1179.png)

dubslow:
closearooni macroni

zetok:
Is it me, or it looks even worse on fd36b6cfaadd8126faffa3f32a4ad0c404bb2014 ?
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4435303/253df8fc-473e-11e4-857d-706ac7068fd2.png)

dubslow:
It doesn't look worse, but I did notice that audiobotze, which doesn't have an avatar, now has the white square instead of what is supposed to be the default image. The chatarea picture is still the default, but its friends list image is the square of pure white. Edit: This isn't always the case, but it shouldn't be able to happen at all...

This should be temporarily reopened.

dubslow:
It happened to another friend, I think maybe it's second on friend's list? But still not consistently... hopefully @krepa098 knows what's up...

krepa098:
Seems like something went wrong during the merge. Can you please apply the patch above and give it a shot? I can't seem to add audiobotze.

tux3:
I cherry-picked your patch in 67ffbfcb6ef6f3c52103f8fbdabbf35f21fb9d2c

zetok:
Group chats are still visually inconsistent:
![2014-09-28 hexafluoride bully](https://cloud.githubusercontent.com/assets/3148759/4449627/a66b884c-481d-11e4-93f6-4bcc6d216622.png)

krepa098:
I agree, although that's how it's done in the mockup (see https://github.com/ItsDuke/Tox-UI)
[...]

PR #362, opened 2014-10-01
Style

krepa098:
- background of MaskablePixmapWidget now depends on the average color of the pixmap (the goal is not to please everyone's flavor!)
- issue #334 a)
- extended Style to allow symbols like @mediumGrey -> #c1c1c1
- removed some cases where QPalette was used for styling (styling should only be done in .css files, some platform-styles don't even use palettes)
- some changes to the chatlog
- updated .css files (more to go)

Hopefully this doesn't cause any regressions :smile:


tux3:
> Hopefully this doesn't cause any regressions :smile:

pic32
Speak of the devil ... now the background color is broke!

Edit: Also the average color doesn't seem to make a lot of sense, Impyy doesn't have a single blue pixel in his avatar, and he gets a blue background. Maybe the median color would make more sense ? 
(or just transparency, really ..)

krepa098:
Hmm... did you do a clean rebuild?
snapshot13

tux3:
Well, I'll try again.

tux3:
I'm still having the light grey background after a make clean && qmake && make

pic33

I can git bisect it if you want, since you don't seem to be affected.

tux3:
I bisected it back to 214e428b7ea8efb6e40b3df5c6ed7fbb924bd611

krepa098:
Does activating/hovering change the color? I just tried a fresh git clone & make from my repo and it worked just fine.

Could be due to the removal of setAutoFillBackground(true); in widget/groupwidget.cpp:34 although it shouldn't be needed using css.

tux3:
If I hover or click in/out the background doesn't change at all.

tux3:
Nope setAutoFillBackground doesn't help.

I'm trying to fix it on my end, by nothing seems to work. It's like the CSS background-color just doesn't apply to the parent's bg for some reason.

tux3:
Wait, something's weird. It looks like my local copy of the PR doesn't have the same style.cpp as your branch.
EDIT: Nope, still doesn't work.

tux3:
It looks like background-color just doesn't work.

With the changes I made it should be completely red, but the background stays undefined no matter what.
pic34

krepa098:
Seems like Oxygen (the widget style) is the problem :anguished: I'm using Qt Cruve. Well at least I can reproduce it now...

tux3:
Maybe there's a workaround, surely we're not the first project using a background color on KDE/Oxygen ?

krepa098:
It's fixed now, I think.

tux3:
Compiling right now.
Edit: Alright, it works for me, thanks !

dubslow:
This causes the ChatArea to not have bold nicknames, instead they look the same as text. Previously the other nick was bold black, my nick bold gray, and chat nonbold. Now all three are nonbold.

PR #364, opened 2014-10-01
Add actions

### LINKED ISSUE #258: Missing support for /me (action) messages

dubslow:
~~Second commit added~~

Fixes #258.

I got bored and this changed from changing debian versions to adding actions. https://debian.org/stable/allpackages (<-- literally all packages, lotsa text)

The actions look like they do in my [IRC client](http://quassel-irc.org/).

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,7 +7,7 @@ Build-Depends: debhelper (>= 9), cdbs, qt5-qmake, libopenal-dev (>= 1:1.15.1), l
 
 Package: qtox
 Architecture: any
-Depends: libc6 (>= 2.17), libgcc1 (>= 1:4.1.1), libgl1-mesa-glx | libgl1, libopenal1 (>= 1.14), libopencv-core2.4, libopencv-highgui2.4, libopus0 (>= 1.0), libqt5core5a (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.0), libqt5widgets5 (>= 5.2), libqt5xml5 (>= 5.0), libstdc++6 (>= 4.9), libvpx1 (>= 1.0.0)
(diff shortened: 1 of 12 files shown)


debian/control
```diff
@@ -7,7 +7,7 @@ Build-Depends: debhelper (>= 9), cdbs, qt5-qmake, libopenal-dev (>= 1:1.15.1), l
 
 Package: qtox
 Architecture: any
-Depends: libc6 (>= 2.17), libgcc1 (>= 1:4.1.1), libgl1-mesa-glx | libgl1, libopenal1 (>= 1.14), libopencv-core2.4, libopencv-highgui2.4, libopus0 (>= 1.0), libqt5core5a (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.0), libqt5widgets5 (>= 5.2), libqt5xml5 (>= 5.0), libstdc++6 (>= 4.9), libvpx1 (>= 1.0.0)
+Depends: libc6 (>= 2.13-38+deb7u4), libgcc1 (>= 1:4.1.1), libgl1-mesa-glx | libgl1, libopenal1 (>= 1.14), libopencv-core2.3, libopencv-highgui2.3, libopus0 (>= 0.9.14+20120615-1+nmu1), libqt5core5a (>= 5.2), libqt5gui5 (>= 5.2), libqt5network5 (>= 5.0), libqt5widgets5 (>= 5.2), libqt5xml5 (>= 5.0), libstdc++6 (>= 4.7.2-5), libvpx1 (>= 1.0.0)
```
tux3: This is not going to work.
OpenCV 2.3 is ancient, its last version was released on 2011-09-12, most distributions have OpenCV 2.4 instead, while OpenCV 3 is in alpha. If we depend on 2.3, a system with 2.4 will not be able to run qTox.

Besides, the next debian release is getting close, soon 2.4 will be in stable.

Can you resend the PR without changing the packages dependencies ?
Maybe we can find a way to link opencv statically instead, to just avoid this issue.

dubslow:
Mostly it was an attempt to fix 365, and since stqism builds on stable I thought it might be better. I'll revert it shortly.

PR #381, opened 2014-10-05
Newsettings v2

apprb:
Tried to rewrite newsettings form with form-designer assistance as was offered by @krepa098.
At least segfault issue has been fixed, as well as selfcam test.
I now this implementation is not perfect but I want to hear your opinions.


dubslow:
It _looks_ great, which is a lot better than mine could say :)

And dang, no merge conflicts with master, that's sweet :)

krepa098:
Yes, that's what I'm talking about :smile:. Well done. Btw. you can use the form-designer do create slots (right-click on a widget -> go to slot). That way you don't have to write connect() all the time (see QMetaObject::connectSlotsByName).

PR #384, opened 2014-10-05
Add proxy support based on newsettings_v2

dubslow:


dubslow:
Something about the merge fucked over this pr, let me make a new one with only two commits

PR #385, opened 2014-10-06
Simplify groupchat names

dubslow:
two other minor tweaks :Kappa:

Oh yeah, also all actions are rendered, regardless of core.

If you decide to merge newsettings_v2, do that first because I expect this merge conflicts.

PR #386, opened 2014-10-06
relative to system wide default font sizes in css

apprb:
Font sizes should be relative to the sizes in the system environment, as @krepa098 said.
Also changed own status font size.

PS: this PR in conflict with PR #385 (commit: 200c6d3).


dubslow:
The conflicting commit has been reverted.

PR #405, opened 2014-10-08
Move code into src/

dubslow:
"cleanup on aisle source"


msteinbeck:
One of the best commits :D. This is something that annoyed me all the time. In general we should do some more cleanup :).

PR #419, opened 2014-10-10
preview of emoticons

agilob:
qtox-emot


krepa098:
Beware: some smiley packs don't use :) :( ;) etc. Maybe you could just show the first 5 smileys?

agilob:
I noticed similar problem in `emoticonswidget.cpp`, dynamic list of of emotions doesn't work properly sometimes:
missing emoticons

krepa098:
Can you please provide a link to the smileys used in that pic?

agilob:
ok, i fixed the problem with unavailable graphical icon for emotion and I'll implement your suggestion now o/

agilob:
@krepa098 done.

qtox-emo-preview
qtox-emo-preview1

// smilies are from [standard KDE pack](https://github.com/Fat-Zer/kdeartwork/tree/master/emoticons/greggman.com), they're built-in on Arch KDE 4.13

tux3:
I'm getting some strange results here.
img1
img2
img3

agilob:
@tux3 do you want them aligned to left?

src/widget/form/settings/generalform.cpp
```diff
@@ -119,3 +121,29 @@ void GeneralForm::onUseProxyUpdated()
...
     Settings::getInstance().setUseProxy(state);
 }
+
+void GeneralForm::reloadSmiles()
+{
+    QList<QStringList> emoticons = SmileyPack::getInstance().getEmoticons();
+    QStringList smiles;
+    smiles << ":)" << ";)" << ":p" << ":O" << ":'("; //just in case...
+    
+    for(int i = 0; i < emoticons.size(); i++)
```
krepa098: I think this should be something like

```
for(int i = 0; i < emoticons.size(); i++)
    smiles.push_back(emoticons.at(i).first());
```

Because you can have multiple keys for the same smiley (like :) :-) 😊)

tux3:
The actual emoticons shown don't really make sense, maybe since most emoticons pack seem to not handle them.
For exemple I'm getting 4 times times a pig then a smiley or two 8) and three >:) for some reason.

Maybe use more simple/common emotes than ":)" << ";)" << ":p" << ":O" << ":'(" ?
Or maybe it's just me having bad emoticon packs, but I'm not getting what you show in the OP.
img4

EDIT: What @krepa098 says!

agilob:
Ok, got it. with push_back i had one empty emoticon, with push_front it's fixed.

PR #420, opened 2014-10-10
Keyboard shortcuts to clear chat area

agilob:
clearing chat area


apprb:
Potential bug: what happen if you clear chat area with active file transfer?
Since FileTransferAction updates itself by QTextCursor pointed to the certain plaice in ChatAreaWidget.
Probably not a problem, but anyway it is not very good to get rid of active filetransfers.

If you don't against it I'll reimplement this in my history branch (already partially done in apprb@6860790 but without shortcuts for now) on slightly different way.

agilob:
Status of active file transfer reappears.

Bug: paused file transfers don't reappear.

tux3:
I can't merge this just yet if it's still buggy, sorry.

agilob:
No problem, I don't know if I should fix that bug now or wait for fix #266 (last 3 comments)

PR #421, opened 2014-10-10
Qt style selector

agilob:
Unfortunately Qt developers removed some old native skins from support, only 3 skins are supported on Linux.
Number of native skins depends on OS and Qt built system.


agilob:
merged with tux3/master to remove conflicts.

tux3:
Merged manually after fixing some remaining conflicts in ff12166753ba64ccfe642cc6e911ec93bb2dd6c4

PR #422, opened 2014-10-10
Status change messages in chat window

agilob:
and possibility to enable in settings

status changes
status settings

also, removed one emit friendStatusChanged signal in Core.cpp, it was emitted twice per one status change.
Created new SIGNAL contactSignIn, it will be used later for sound notification and pop-up widget.


dubslow:
Maybe remove the "has come online" message, because it will always be followed by a status change and corresponding message. Also, these messages should only be made for chats that are already in use.

agilob:
> Maybe remove the "has come online" message, because it will always be followed by a status change and corresponding message.

good point.

> these messages should only be made for chats that are already in use.

that was happens, sceenshot is... misleading :<

I'll fix it

dubslow:
@tux3 this should be pretty easy to rebase on master; after doing that and then applying https://github.com/dubslow/qTox/commit/9c4bb134bea03e53bb332ece49b6463a42b19add (cleans up the sign in change as unnecessary), I would recommend this for merging. It's quite a nice feature to have.

tux3:
I'll test this later today when I have some time

tux3:
Rebased and merged in a0600d1ce46977b512cb70dd644e3ebb05de11ca, I also cherry picked dubslow@9c4bb134bea03e53bb332ece49b6463a42b19add

PR #423, opened 2014-10-11
History

apprb:
Log history into a file. Currently without encryption.
Notable changes:
1. ChatAction's setup(...) function doesn't clear stored content in the class instance after call (for MessageAction, SystemMessageAction, etc). It is much easier to operate with chat area using this class' instance (past history, for example) in cost of memory use.
2. Clear chat area (like in PR #420) with correct unfinished downloads handling but without shortcuts.

Things left to do:
~~1. Finish infrastructure for saving history as an encrypted file.~~
~~1.1 Encrypted file read/write~~
~~2. Implement correct actions saving.~~
~~3. Don't mix history from different profiles (when associated PR will be accepted).~~
4. Groupchats history: as for now it is no way to distinguish one group chat from another. I mean chat names or IDs are not persistent (probably won't be a problem after "New Group Chats" introduction).

update: 14a71a0 changed database structure


dubslow:
> <widget class="QCheckBox" name="cbEnctyptTox">

:')

apprb:
I misused "encrypt" word? Or something like "cbEncryptToxData" is more suitable for this?

dubslow:
I'm pretty sure it's not spelled "Enctypt" :)

apprb:
You are totally right :)

agilob:
@apprb  could chat log table also include boolean column 'read', indicating if the messaged was delivered? This could be useful to handle offline messages, when a contact logs-in, check if there are any unread messages, if so, send them.

apprb:
Need help :) 

I have no idea how to encrypt file message log in the right way.
Currently encrypted message log file contains copy of INSERT and CREATE sql queries passed to the db. So, write to file only on close isn't good idea, as well as append to file queries encrypted separately. Appending by encrypted chunks of constant size is the best way (as I think) but I completely don't know how to handle unfilled chunks (can't understand how to rewrite unfilled chunk at the end of file).

> @apprb could chat log table also include boolean column 'read', indicating if the messaged was delivered? This could be useful to handle offline messages, when a contact logs-in, check if there are any unread messages, if so, send them.

I thought it should be core feature, not client's. If you insist I'll do it, but bit later, encryption first :)

dubslow:
Before hand caveat: I know very little about SQL, and thus only have a vague understanding of your paragraph.

Having said that, is it possible to flush everything to the file on every write?

If being able to selectively overwrite a certain part of the file at the end would suffice, that should certainly be easily doable with either stdlib functions or Qt functions.

apprb:
Finally made it work. Code cleanup is left.
@dubslow, why tox_derive_key_from_pass(...) from toxencryptsave generates different keys on each execution from the same password?

Update: 
tox_pass_encrypt(..) works pretty slow in comparison with tox_pass_key_decrypt(..)

dubslow:
@apprb It requires a salt, and because I'm not one who is willing or able to decide otherwise, I'm not using a zerod-salt instead of a proper random salt. The other functions in the module are still capable of working with it, however.

tox_pass_encrypt() is as slow as tox_pass_decrypt(), and tox_pass_key_encrypt() is as fast as tox_pass_key_decrypt().

Additionally, I meant for you to use the corencryption branch I have -- I'll update it for how you want to use it, and then merge it into your branch.

apprb:
Update:
TODO:
1. Change password for encrypted history.
~~2. Test tox datafile encryption.~~
~~3. Handle all wrong password cases.~~

apprb:
Another update.
Encryption of tox data is working.
History and its encrypted version are working too.

Definitely not all required features are implemented, like converting encrypted history to plain and vice versa, merging separated histories, history password changing, etc. But I think this is the next step.

PS: need assistance to test base functionality, I'm pretty sure that I missed something :)
PS2: encryption is slow (toxcore problem).

dubslow:
I can work on at least PS2 hopefully tomorrow evening

zetok:
Well, if it'll apply cleanly on master, I'm going to test it at least a bit tomorrow evening (if there still will be need for testing)

agilob:
@apprb client crashes after encrypting log database and starting tox. Application quits after providing correct and incorrect password. No error message.

tux3:
@agilob weird it works for me, can you create an issue and post a log/backtrace ?

EDIT: Oh wait, it does crash for me too if I give a wrong password and hit try again.
EDIT2: Found it I think "QWidget::repaint: Recursive repaint detected"

dubslow:
Uh, this probably should not have been merged yet. The encryption isn't up to snuff yet, nor is the UI for it...

Among other things, there's a PR in toxcore that's related to this that needs merging (or at least it did some hours ago, still going through notifications).

tux3:
I disabled the checkboxes for encryption in the settings for the moment.

It was my understanding that this part was about ready after talking with @apprb and minimal testing, sorry.

dubslow:
It's _nearly_ ready, after I fix the encryption speed (which would include the password UI and such).

I'll need a few hours of spare time after it's merged though... lol

PR #428, opened 2014-10-12
Get basic core encryption funcionality up, still needs to be hooked up to UI and actual usage

dubslow:
Good basis for e.g. the history branch #423 to use (though is independent of that PR)


dubslow:
It's in a commit note, but this also depends on irungentoo/toxcore#1114 being merged. Should probably wait until then for this.

dubslow:
I'm a bit hesitant about all those memory deep copies with returning raw QByteArrays everywhere, but it might be better than ensuring the caller cleans up memory allocated in the functions... but then again maybe not... thoughts are welcome.

Ooops, misclick trying to comment...

krepa098:
QByteArray is copy-on-write. You can also use QByteArray to allocate memory and pass QByteArray::data() to tox.

> but it might be better than ensuring the caller cleans up memory allocated in the functions

Yes.

dubslow:
Misclicks of the century with this PR

dubslow:
I'm closing this since it's intertwined with and improved in the history branch

PR #429, opened 2014-10-12
hide moc and .obj files on *N*X systems

agilob:
generated files will be placed in hidden directories .moc and .obj

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -32,6 +32,9 @@ FORMS    += \
     src/widget/form/settings/identitysettings.ui
 CONFIG   += c++11
 
+MOC_DIR = .moc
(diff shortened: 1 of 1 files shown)


apprb:
This is why I always build every project in a separate folder.

dubslow:
No rest for tux3 lol, if he merges even 1 PR, another pops up to take it's place :')

tux3:
That sounds like a Bad Idea that could confuse some scripts, can't you just build into a separate folder or make clean ?

Plus, you didn't add those folders to the .gitignore, so if we start doing that chances are I'll merge a PR with ".moc/moc_*" files by mistake eventually...

Issue #235, opened 2014-08-31
Copy friend ID no worky properly


mouseym:
Audiobotze's ID according to toxme.se:  BF9A09F70D6C0C3424472B740E9F70878A72CECCB30AC396D60D8160BB6C2E22E6EA2917FFE2

Audiobotze's ID according to copy friend ID: 
BF9A09F70D6C0C3424472B740E9F70878A72CECCB30AC396D60D8160BB6C2E22

tux3:
The problem is that Toxcore doesn't remember the full address (ID+Nospam+Checksum), just the ID, so I have to build a separate database of friend IDs instead of just the tox_save.

Note that you'll need to re-add your friends so that Copy Friend ID can learn their full address. If the full address isn't found, it'll just give you the id (like it does now).

Since that's the second time someone reports that, I'll fix it right now. Expect a fix in ~15 minutes.

zetok:
I don't see what it "fixes".

@mouseym You realize that what qTox gave is correct public key?

Perhaps it's a bit wrongly named, but nonetheless it is a correct public key of audiobotze.

Clients are **supposed** to offer a way of changing `nospam`, and there is no way, nor will be, to announce changed `nospam` to to friends.

It means, that if you'll save anything except for public key from Tox ID, and your friend changes `nospam`, currently you'll only get not working, 76 long string of characters.

@tux3 Be prepered for users' question regarding why they have this long string of chraracters, and how come so that it doesn't work, and why only part of it is valid.

Anyway, whole "idea" of being able to just "copy" friend's Tox ID to add it in other Tox instance is dumb and won't work.

tux3:
@zetok I agree with you, except that people seem to be expecting it to work this way. And if someone changes it's nospam on purpose, then maybe it's his job to explain to his friends while they can't add him anymore ?
I'm not seeing a better solution that doesn't involve removing the "copy friend id" feature completely.

zetok:
> I'm not seeing a better solution that doesn't involve removing the "copy friend id" feature completely.

Um, that's what I would like to suggest :3

Thing is, user shouldn't be really encouraged to copy Tox ID from client to client, but there should be a way for user to check friends' public key.

IMHO best approach would be to morph `Copy ID` into `Show public key`, and after clicking `Show public key` there would be pop-up window with public key, and something like question mark somewhere, which would lead to explanation of how things are supposed to work, and why copying public key from client to client isn't supposed to work.

[...]

Issue #84, opened 2014-07-05
``Copy friend ID'' doesn't copy a usable friend ID


F1ynn:
It only returns their public key, leaving out the nospam and checksum bits.

Will fix soonish.

msteinbeck:
Do we have any progress here?

F1ynn:
Haven't had time yet, sorry. Based on a brief glance a few days ago, the nospam isn't stored in the standard tox data file so it might be somewhat interesting fixing this.

lehitoskin:
The Nospam value isn't saved in Tox. Unless you keep a separate DB of Tox IDs, there's no way to retain it.

msteinbeck:
> Unless you keep a separate DB of Tox IDs

Do we have any plans to do this?

Issue #44, opened 2014-07-01
Friend ID copying


suhr:
Add “Copy friend's ID” item to context menu.

F1ynn:
Done.

tux3:
Fixed by Flynn's PR, closing

PR #434, opened 2014-10-14
Fix wording for copying friend public key

zetok:


dubslow:
Technically the key is the ID; the key+ns+chk is called the address.

zetok:
Key is only a part of ID, thus current wording isn't accurate and causes confusion

Edit: https://libtoxcore.so/core_concepts.html#the-tox-id

dubslow:
https://github.com/irungentoo/toxcore/blob/master/toxcore/tox.h#L43

tux3:
We actually try to give back the full Tox ID not just the pubkey (in Core::getFriendAddress(int)) so the wording should be okay.
Plus, I'd rather not confuse the users with too many technical terms if it can be helped. For all intents and purposes the Tox ID is the bunch of letters and numbers you copy paste to add someone, no crypto knowledge required.

zetok:
k, just don't be surprised when users will come screaming that copying ID is broken, as I've highlighted it in #235

tux3:
I haven't had any report of confused users yet, but I remember several people asking for a way to copy their friend's ID, and even making issues on github (three times! #235, #84 and #44).

I know that it's not a very good solution, but I don't have anything better, and if people want this feature I'm not going to hide it from them like I know better.

If you have a good solution I'll gladly consider it, I'm always open to discussion. That said I don't think that changing the wording like this is right, we actually try to return a working friend ID right now, not just the pubkey, so this PR would make it :
- A bit more confusing, this is the only place qTox mentions a public key, eveywhere else we handle Tox IDs
- Wrong most of the time, we normally do return a Tox ID.

jbshirk:
copy friend ID does nothing on OS X, nothing in clipboard. How do I get it?

tux3:
> copy friend ID does nothing on OS X, nothing in clipboard. How do I get it?

If you have absolutely nothing in the clipboard, then that's a bug. 
Does it never work, or just sometimes ?

EDIT: I just tried copying GroupBot's ID on OS X and it worked for me. 
Can you give me more details ? Are you sure you're running the latest version of qTox ?

pic11

PR #439, opened 2014-10-15
Fix 436, and also fix a threading issue WRT profiles

dubslow:
Turns out that directly calling functions on objects owned by a different thread is A Bad Idea™. It's a minor miracle qTox didn't crash whenever any profile switch occurred.

### DIFF
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -1184,7 +1184,7 @@ void Core::saveConfiguration(const QString& path)
     }
 }
 
-void Core::switchConfiguration(QString profile)
+void Core::switchConfiguration(const QString& profile)
 {
     if (profile.isEmpty())
     {
@@ -1204,7 +1204,7 @@ void Core::switchConfiguration(QString profile)
         tox = nullptr;
     }
     emit selfAvatarChanged(QPixmap(":/img/contact_dark.png"));
-    Widget::getInstance()->clearContactsList(); // we need this to block, so no signals for us
+    emit blockingClearContacts(); // we need this to block, but signals are required for thread safety
     
     loadPath = QDir(Settings::getSettingsDirPath()).filePath(profile + TOX_EXT);
     Settings::getInstance().setCurrentProfile(profile); 
--- a/src/core.h
+++ b/src/core.h
@@ -53,7 +53,6 @@ public:
 
     void saveConfiguration();
     void saveConfiguration(const QString& path);
-    void switchConfiguration(QString profile);
     
     QString getIDString();
     
@@ -70,6 +69,7 @@ public slots:
     void start();
     void process();
     void bootstrapDht();
+    void switchConfiguration(const QString& profile);
 
     void acceptFriendRequest(const QString& userId);
     void requestFriendship(const QString& friendAddress, const QString& message);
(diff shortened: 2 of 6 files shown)


tux3:
Sadly I really don't have the time to go looking through the at the moment code, sorry. Maybe a simple moveToThread() would do the trick ?

Your fix looks ok so I'm merging it in the meantime.

dubslow:
I'm pretty sure the threading fix is absolutely fine.

What I was asking about is checking if there are any calls active. I wrote a three line function that loops over core->calls[i].active, but was wondering if there's a simpler way (i.e. no looping). Perhaps some sort of sound variable and/or variable that is active when any call is?

At any rate it doesn't matter, it's purely a question of elegance.

tux3:
Oh sorry I misread that.

I'm pretty sure there's no easy way to check if any calls are active right now, and I'd rater avoid weird hacks relying on some obscure side effects.
You could simply maintain a (atomic?) number_of_calls_in_progress and check that instead of checking all the calls. Especially if you're doing that checking on another thread right now (I haven't checked, but if so it seems like an obvious race condition).

dubslow:
I'm pretty sure it is another thread (in IndentityForm.cpp), so yes theoretically the race condition could occur... but frankly, the user has to manually accept any calls anyways; the time between the user clicking "Load" and the event executing/checking for calls is a few milliseconds tops, far shorter than any time user could accept a call in, so I'd be shocked if it actually ever happened.

tux3:
> so I'd be shocked if it actually ever happened.

If there is a race condition somewhere, then @zetok WILL run into it. That's probably a law of the universe.

But more seriously this race condition can actually happen even if it's unlikely :
- User starts a call with a contact (note that the call is NOT enabled until the contact accepts it)
- In the meantime (~15s timeout) the user decides to change profile for some misguided reason
- anyActiveCalls() returns false
- We get the contact's reply and start the new call immediately
- Boom, profile switch with an active call

> a few milliseconds tops, far shorter than any time user could accept a call in

Except if it's the remote contact accepting the call, and the local user changing profile.
But yes, it's unlikely.

dubslow:
Ah yes, attempting to switch before remote accepts; a simple solution would be to add another bool "attempting" or some such to the struct.

And you are certainly right, @zetok manages to find every bug... who would even try to switch to the current profile...

tux3:
> Ah yes, attempting to switch before remote accepts; a simple solution would be to add another bool "attempting" or some such to the struct.

Yes, that would make the race extremely unlikely. Maybe change the "active" flag into an enum (Not active, attempting, active) since it's taking a full byte anyway.

zetok:
> a few milliseconds tops
> 
> extremely unlikely

You're forgetting about given scenario:
> clients in future support auto-accepting call
> call is being made between clients on LAN

Does that count as <few milliseconds ?

Edit: also, consider that one of ends could "hang" for like ~0.5s, which is quite a lot of ms..

tux3:
I haven't looked at the code so I'm probably completely wrong, but why wouldn't we just do the actual profile switching in the Core thread instead of a different thread like you said in the OP ? 

As long as the core thread isn't calling qApp->processEvents(), then we should be safe, yes ?
Even if the GUI thread decides to call a Core or CoreAV slot, it'll get queued and no race can happen. Same with toxcore events, unless tox_do is asynchronous, they'll be delayed.

Issue #107, opened 2014-07-06
[optional] Change status to “away” after being inactive for a while


suhr:
Add to user settings the option and the delay value.

dubslow:
Skype at least is able to detect activity on a computer-wide basis, though I have no idea how.

ProMcTagonist:
Skype doesn't just yellow you after you leave it alone for a while? It detects if you're using other applications? That's pretty interesting.

tux3:
Fixed by #447

PR #447, opened 2014-10-16
UI for autoaway

agilob:
addition to https://github.com/tux3/qTox/pull/441

autoaway

### DIFF
--- a/src/widget/form/settings/generalform.cpp
+++ b/src/widget/form/settings/generalform.cpp
@@ -112,7 +112,9 @@ void GeneralForm::onStyleSelected(QString style)
 
 void GeneralForm::onAutoAwayChanged()
 {
-    Settings::getInstance().setAutoAwayTime(bodyUI->autoAwaySpinBox->value());
+    int minutes = bodyUI->autoAwaySpinBox->value();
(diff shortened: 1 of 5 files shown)


tux3:
It doesn't actually reset the timer, does it ?

So if I set the auto-away to 2376 minutes by mistake then back to 1 minutes, it'll stay locked at 2376 until the next restart (or after 2376 minutes of inactivity!)

agilob:
Should do the job now, @dubslow, would you like to add anything?

And.. I set maxmimum value to 600 minutes, I can change it to INT_MAX if you want, this artificial limit is useless I think but qtcreator made me to set anything there.

dubslow:
FWIW, one day is a... _plausible_ auto away time, and that's somewhat more than 600, but I won't be miffed unless someone complains about it.

I'd be more likely to complain about the checkbox and int box not being on the same line, wasting oh-so-precious vertical space.

@tux3, there's at least one old issue about this that can be closed.

tux3:
@dubslow thanks for the heads up, I closed #107

dubslow:
Wait what? There isn't even a checkbox, I'm nuts.

Also, the minimum should be zero so it can be disabled. (The code already handles it.)

PR #452, opened 2014-10-17
More work on camera

krepa098:
Big change: NetCamView makes use of the VideoSurface class.
Video chats are working fine for me but it would be great if someone else could give it a try before merge.
- lots of small changes (see commits).

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -19,7 +19,7 @@
 #include "ui_avsettings.h"
 
 AVForm::AVForm() :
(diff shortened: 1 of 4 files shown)


zetok:
It doesn't look good. Current view isn't good, but this is IMHO worse.
Is there a **real** reason why audio and video should be put together in 1 tab?

dubslow:
Making another tab is always in the books, there's no reason it has to be four.

krepa098:
Well creating another tab for 2 volume sliders doesn't make sense either ;). The idea was to make effective use of horizontal space as most people are using widescreens nowadays. I agree that there is quite some room for improvements here.

PR #464, opened 2014-10-18
version in debug message based on git HEAD

agilob:
might be usefull to track version and issues sent by non-technical users

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -44,6 +44,9 @@ TRANSLATIONS = translations/de.ts \
 
 RESOURCES += res.qrc
 
+GIT_VERSION = $$system(git rev-parse HEAD)
+DEFINES += GIT_VERSION=\\\"$$GIT_VERSION\\\"
+
 contains(JENKINS,YES) {
 	INCLUDEPATH += ./libs/include/
 } else {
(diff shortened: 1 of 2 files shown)


zetok:
Doesn't that assume that there's git present in system?

What about cases where someone would just download zip from github? https://github.com/tux3/qTox/archive/master.zip

agilob:
> Doesn't that assume that there's git present in system?

Yes. 

> What about cases where someone would just download zip from github?

If build system has .git in the folder, commit hash will be hardcoded, you don't need git on your client machine. Will work, if jetkins/other publisher has git installed.

tux3:
Will it fail horribly on a Windows environment (or anywhere without Git really), or will it just result in something like <code>GIT_VERSION='git' is not recognized as an internal or external command operable program or batch file</code> ?

If the compilation fails because git can't be found that's not okay, if it just results in a bad version that's alright (although I'd prefer a clean fallback).

agilob:
I can't tell what will happen on Windows, but on Linux when I change command to: `meh rev-parse HEAD`
make prints warning: `/bin/sh: meh: command not found`
and no commit hash is available in qDebug:

```
built on:  11:28:47 Oct 18 2014
commit:   
```

PR #471, opened 2014-10-18
all forms are translatable now

agilob:
changed type of main tabs to pointers to fix that

before this commit only SettingsTab was translatable, now tabs 'Uploaded/Downloaded', 'Friend requests' are in selected language too.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -54,7 +54,7 @@ Widget::Widget(QWidget *parent)
     : QMainWindow(parent),
       ui(new Ui::MainWindow),
       activeChatroomWidget{nullptr}
-{
+{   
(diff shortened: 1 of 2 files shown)


src/widget/widget.cpp
```diff
@@ -54,7 +54,7 @@ Widget::Widget(QWidget *parent)
     : QMainWindow(parent),
       ui(new Ui::MainWindow),
       activeChatroomWidget{nullptr}
-{
+{
```
tux3, line 57: Trailing spaces make git complain, it's not all that important but please try not to do it!

agilob, line 57: sorry, it happens when I click that line qtcreator removes or adds them :<

PR #472, opened 2014-10-18
changed qstringlists to qmap

agilob:
does anyone else think qmap is better than two qstringlists?


tux3:
I don't think it's really better. The map is not going to be any faster (probably much slower actually, even tho it's negligible), and it adds 13 more lines to do the same thing.

dubslow:
The problem is going from the name back to locale, and that's easier with the num rather than a reverse lookup in QMap (or QHash). I spent some time thinking about this before deciding on parallel lists.

PR #475, opened 2014-10-18
fix #457, styles in mainContent for qtextedit, qspinbox and more

### LINKED ISSUE #457: [Regression] dark system theme broken again
@krepa098 change you made in c4529d4 in CSS overwrites fixes I wrote several days ago for dark themes 

tox
tox1
tox2

I can fix this if you work on light themes.

agilob:
before merging I would like @krepa098 to check if you still have 'Win98 rendering' problem, this commit fixes my problem with dark system theme.


krepa098:
There are still a few issues

snapshot19

snapshot18

CSS doesn't apply to QScrollArea although it does after switching from Fusion to Qt-Curve back to Fusion. But I guess that's not related to this PR. The scrollbar and the arrow buttons need theming too.

agilob:
it's my second CSS code ever o/
So there is progress and two bugs were eliminated, I think I can finish it tomorrow (or soon).

agilob:
@krepa098 could you try it now?

I don't have any arrow button, nor scrollbar Oo

krepa098:
> I don't have any arrow button, nor scrollbar Oo

Only A/V settings have it. The other tabs may get it because of:
snapshot20

I think you can remove the workaround by using this css file https://gist.github.com/anonymous/12fbf8d33e02da7c7d89
That way we have the same scrollbars everywhere and the black color in combination with QScrollArea is gone (you can remove the scrollbar css code from ui/chatArea/chatArea.css). Another thing I spotted is the styling of the tray menu... I don't think we should do that. It feels really strange and there's no need for it.

snapshot21

tux3:
Is it ready to merge ?
It does work for me, if that's any help.

agilob:
I think so.

PR #488, opened 2014-10-20
Add tab completion to groupchats

dubslow:


src/widget/form/genericchatform.h
```diff
@@ -22,7 +22,7 @@
 #include <QDateTime>
 
 // Spacing in px inserted when the author of the last message changes
-#define AUTHOR_CHANGE_SPACING 5
+#define AUTHOR_CHANGE_SPACING 5 // why the hell is this a thing? surely the different font is enough?
```
tux3, line 25: IIRC it was added to follow the mockup. I don't mind changing that, but cursing about it in source code comments isn't really helpful :)

dubslow, line 25: Most of the reason I don't leave qTox running all the time is because its chat area is so much suckier than my irc client, like for the same space quassel gets twice as many lines (and I can paste things from it too lol).

So, in this case at least, the mockup is wrong. :) Frankly, is there any other chat program at all on this planet that does that?

tux3, line 25: Yeah, I agree that we're wasting a lot of space there, let's change that if you want. I went with the safe choice of blindly following the mockup at first since I'm so bad at design, but I guess it's alright to take some 'artistic freedom' now :)

dubslow, line 25: I'm all for it, it really needs to be more compact.

dubslow:
Okay, all the files I meant to be here are actually here now...

tux3:
Yes, it makes more sense now.

tux3:
> I copied this from Quassel ;)

Careful not to copy paste too much or we might run into licensing hell. It says they are GPL too, but IANAL.

dubslow:
> Careful not to copy paste too much or we might run into licensing hell.

Nope, they are GPL 2+. Read the headers in tabcompleter.*

tux3:
My eye is trained to skip copyright headers :)
And sorry, I edited my post just as you were replying.

dubslow:
> My eye is trained to skip copyright headers :)

Me too lol, but I took the 5 seconds to make sure this was okay.

> And sorry, I edited my post just as you were replying.

Edit: crosspost

PR #494, opened 2014-10-20
Update russian translation

suhr:
Subj.


Ilya87:
For future translation update please .ts file with -no-osolete option to lupdate. Keep obsolete items is a waste of space.

Ilya87:
Also term «автопринятие» is kind of Newspeak. Try to avoid using it.

PR #519, opened 2014-10-24
Update refined mannol translation

zetok:
I went through it and tried to do something about problematic parts. I can't assure that now everything is good in it, but should be ~ok.
If there are still things that should be changed for it to be accepted, please outline them.


translations/mannol.ts
```diff
@@ -457,174 +595,230 @@ Do you want to erase this profile ?</source>
...
+        <translation type="unfinished"></translation>
+    </message>
 </context>
 <context>
     <name>MainWindow</name>
     <message>
         <location filename="../src/mainwindow.ui" line="20"/>
         <source>qTox</source>
-        <translation type="unfinished"></translation>
+        <translation>uTox</translation>
```
tux3, line 821: 10/10 I'm merging it

PR #525, opened 2014-10-24
Camera changes

krepa098:
Just a bunch of fixes, cleanup and a little refactoring. Nothing new in terms of features.

Reverts 2d4329767e5e5b75e29ff009909eaa3b21488fc9 as it should not be needed anymore.

@master-passeli you can get started now :)


tux3:
> it should not be needed anymore.

Why is that ?
Edit: Never mind, found it.

PR #529, opened 2014-10-24
disable switching buttons during call

dubslow:
also add new reconnect button, since we can do that now

### DIFF
--- a/src/coreav.cpp
+++ b/src/coreav.cpp
@@ -128,8 +128,8 @@ void Core::answerCall(int callId)
 void Core::hangupCall(int callId)
 {
     qDebug() << QString("Core: hanging up call %1").arg(callId);
-    toxav_hangup(toxav, callId);
     calls[callId].active = false;
+    toxav_hangup(toxav, callId);
 }
 
 void Core::startCall(int friendId, bool video)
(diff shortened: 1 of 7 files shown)


dubslow:
I guess I should really disable the reconnect button too... but I just told myself I'm done coding for the day...

src/coreav.cpp
```diff
@@ -157,20 +157,20 @@ void Core::startCall(int friendId, bool video)
...
     qDebug() << QString("Core: cleaning up call %1").arg(callId);
-    calls[callId].active = false;
     disconnect(calls[callId].sendAudioTimer,0,0,0);
     calls[callId].sendAudioTimer->stop();
     calls[callId].sendVideoTimer->stop();
     if (calls[callId].videoEnabled)
         Camera::getInstance()->unsubscribe();
     alcCaptureStop(alInDev);
+    calls[callId].active = false;
 }
```
tux3, line 174: That looks like a race condition waiting to happen. If you start cleaning up before notifying that the call is inactive, then sendCallAudio and sendCallVideo could segfault. 

It's probably safe right now since the A/V is single threaded and as long as nobody uses qApp->processEvents() from another thread, but why not disable first like we used to ?

dubslow, line 174: Oh.

I think the former way might also be the source of race conditions the other way... but that's somewhat more serious.

PR #544, opened 2014-10-26
Reducing memory consumption

apprb:
No real reason to store all ChatAction objects in the memory.


tux3:
Refcounting, why not.

Have you measured what difference that makes ?

apprb:
To be honest, no.
The main reason I wrapped it into shared pointer: to prevent possible memory leakage, because objects are being created and deleted in many places in the code.

agilob:
avg. 71MB less here.

PR #572, opened 2014-10-29
Update settings.cpp

TUSF:
Make portable the default, so information is saved to the excutable directory. This allows for better usability for people operating qTox out of a removable device.

### DIFF
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -28,7 +28,7 @@
 #include <QList>
 
 const QString Settings::FILENAME = "settings.ini";
-bool Settings::makeToxPortable{false};
+bool Settings::makeToxPortable{true};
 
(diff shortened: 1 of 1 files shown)


tux3:
I think that most people just want to install Tox on their computer and use it like they would use Skype, so the option is there but I don't think that's a good default.

This would have unexpected results if people install qTox packages or use the Windows installer with portable on by default, it would add config files on the desktop or whatever directory qTox was started from.

tux3:
That said, you could still redistribute a 'portable qTox' without having to change the code. Just ship a settings.ini with portable enabled along with qTox.

(When qTox starts it looks in the current directory for a settings.ini, and if the "Make tox portable" option is enabled in that settings.ini, then qTox will use it.)

Issue #596, opened 2014-11-02
[Feature request] Use qTox's emoticons as fallback ones


zetok:
Emoticons bundled with qTox should be always used as fallback ones, since set of emoticons user may chose may not have some icons, thus UX wouldn't be complete, and users may be confused when receiving some `:kappa:` instead of emoticon.

Fallback should be also used when it comes to emoticon packs that come with qTox, unless all sets of emoticons in qTox would have equivalent emoticons, which isn't the case..

PR #608, opened 2014-11-02
default smiley pack is cylgom

agilob:
Before merging please test it on WINDOWS with NEW, RAW profile. Not sure if this string will work on not Unix-like systems.

somehow solves #596?


zetok:
@agilob it doesn't solve it, since user on default pack still wouldn't see `:Kappa:` emoticon, and after switch to other pack wouldn't see `:terminal:` emoticon

agilob:
OK, get it now.

dubslow:
Similar strings are used all over the place in qTox. The Qt file functions are platform independent.

Issue #341, opened 2014-09-29
When editing one's username or status, font color changes to match the background


dubslow:
Making it almost impossible to see what you're typing

dubslow:
Hmm, this should be changed to the font color changes on editing, which it shouldn't.

tux3:
What OS / WM were you using to reproduce ?

dubslow:
I still have this issue on LMDE/Cinnamon 2.0

tux3:
I can't repro on Debian sid/Cinnamon 2.2.16

![image3](https://cloud.githubusercontent.com/assets/5155966/4877116/afa7be78-62dd-11e4-965f-1facf2edd7ff.jpg)

Is this still an issue ?

dubslow:
Yes, but maybe it's a Cinnamon 2.0 thing

agilob:
I'm sure it's related to system theme. Looks like this on my dark theme:
![status](https://cloud.githubusercontent.com/assets/2544251/4877362/a3028ee8-62e5-11e4-99c5-65e7a6800636.png)

Will fix in a second.

PR #611, opened 2014-11-02
style for qlineedit in status panel

agilob:
damn, I'm quick #341 

before:
status

after:
status1

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -397,7 +397,7 @@ QSplitter:handle{
           </property>
           <layout class="QGridLayout" name="gridLayout_2">
            <item row="0" column="0">
-            <layout class="QHBoxLayout" name="horizontalLayout_3">
+            <layout class="QHBoxLayout" name="myProfile">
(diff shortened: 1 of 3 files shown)


dubslow:
Er... I don't see the difference? How is this an improvement?

agilob:
I defined background colour for qlineedit, it won't be inherited from system theme, the same with font colour, border was removed.
Build commit to see the different on your config.

dubslow:
Hmm, it works (though I still can't tell the difference between your pics :D)

agilob:
I have dark theme on KDE, Firefox, QtCreator and everything else, so my system theme is extremely similar to the one I defined ;)

agilob:
Modified it a bit. I think it's better now.

indicate

ghost:
Corners are inconsistent :^3

agilob:
Now.
indicate1

ghost:
Better :3

agilob:
:*

PR #612, opened 2014-11-03
Call duration label

agilob:
call3
call4

I also added some qDebug messages, please don't remove them, I'll need them soon.

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -59,6 +59,9 @@ ChatForm::ChatForm(Friend* chatFriend)
 
     headTextLayout->addWidget(statusMessageLabel);
     headTextLayout->addStretch();
+    callDuration = new QLabel();
+    headTextLayout->addWidget(callDuration, 1, Qt::AlignCenter);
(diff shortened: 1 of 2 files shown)


src/widget/form/chatform.cpp
```diff
@@ -689,3 +730,57 @@ void ChatForm::onLoadHistory()
         chatWidget->verticalScrollBar()->setValue(savedSliderPos);
     }
 }
+
+void ChatForm::startCounter()
+{
+    if(timer)
```
dubslow: Perhaps you mean `if(!timer)`?

agilob: @dubslow I have thought I meant that, but when it check for (!timer) qtox segfaults. Try it yourself.

krepa098: Yes because your timer variable is uninitialized (ie. has some random value != nullptr). Add timer(nullptr) to the constructor.

agilob: Ok, thanks @krepa098 o/

PR #630, opened 2014-11-05
Settings: autosaveDir, useEmoicons, selectTimestamp

agilob:
This is in settings only, does not include support in ChatForm or global autosave, only timestamping is supported globally.

menu


dubslow:
Auto save of both kinds already has a GUI, it does _not_ need to be in the settings dialogs....

dubslow:
And there's also now duplicate code in misc/settings.\* ...

agilob:
Oh now I see that parameter ;_;
It's not really user friendly to hide it under right click, I think.

agilob:
@dubslow I will remove that global auto-accept from context menu if you don't mind. I would never guess this setting can be there.

dubslow:
Yeah, I guess that makes sense. But then maybe make a note about the per-contact setting in the global interface.

PR #641, opened 2014-11-06
Add identifier, prepare for code signing on OS X

stqism:


dubslow:
brah that indentation

PR #649, opened 2014-11-06
highlight tox:// in chat

agilob:
#648


suhr:
Need to highlight tox:‹address› as well.

Issue #653, opened 2014-11-07
Rename friends


stollr:
Currently it seems to be impossible to rename a friend, but sometimes it could be helpful, to change friends nicknames to their real names or for grouping them.

agilob:
duplicate #335

Issue #335, opened 2014-09-28
Support for aliases


zetok:
User should be able to set custom aliases for contacts

PR #662, opened 2014-11-07
Aliases feature

apprb:
relates to #653 and #335.

### DIFF
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -722,18 +722,7 @@ void Core::requestFriendship(const QString& friendAddress, const QString& messag
         emit failedToAddFriend(userId);
     } else {
         // Update our friendAddresses
-        bool found=false;
-        QList<QString>& friendAddresses = Settings::getInstance().friendAddresses;
-        for (QString& addr : friendAddresses)
-        {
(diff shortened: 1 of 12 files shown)


zetok:
Finally \o/

I wonder though, is saving aliases per profile, or global?

Edit: r-click.. how hard would it be to make it changeable by clicking on name in chat window?
Also, could it be possible to display contact's own nick in parentheses right to set alias, and in italics? (that's how Venom does it)

apprb:
Globally.
Don't enough energy left after long working week to extract profile related data to separated configure files.
May be later.

zetok:
After setting a lot of text as alias, tooltip showing nick is of enormous size, and from log:

```
QXcbShmImage: shmget() failed (22) for size 151145592 (13093x2886)
```

zetok:
Also after changing alias of selected contact title of window doesn't change until switched to other contact
alias

apprb:
You always help to find problems with big data processing :)

apprb:
Newer look at the window title before, probably because I'm using awesome WM.

zetok:
Contacts that haven't accepted FR/don't have name now don't even show public key:
void

PR #664, opened 2014-11-07
Threadify file logger, handle save files form cmdline

### LINKED ISSUE #638: EXEC_CRASH on invite from groupbot when already in a split groupchat
So... this happened http://tox.lighthouseapp.com/projects/123754-tox/tickets/10-qtox-crash-on-invite

dubslow:
fixes #638

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -152,7 +152,8 @@ HEADERS  += src/widget/form/addfriendform.h \
     src/misc/flowlayout.h \
     src/ipc.h \
     src/widget/toxuri.h \
-    src/toxdns.h
+    src/toxdns.h \
+    src/widget/toxsave.h
 
 SOURCES += \
     src/widget/form/addfriendform.cpp \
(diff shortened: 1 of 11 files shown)


src/main.cpp
```diff
@@ -39,7 +42,8 @@ void myMessageHandler(QtMsgType type, const QMessageLogContext& ctxt, const QStr
             && msg == QString("QFSFileEngine::open: No file name specified"))
         return;
 
-    dflt(type, ctxt, msg);
+    dflt(type, ctxt, msg); // this must be thread safe, otherwise qDebug() would never ever work
+    QMutexLocker locker(&mutex);
```
tux3, line 45: It probably works, but if we're going to bother locking anyway, why not guard the entire function just to be safe :) ?

qtox.pro
```diff
@@ -214,4 +215,5 @@ SOURCES += \
     src/misc/flowlayout.cpp \
     src/widget/toxuri.cpp \
     src/toxdns.cpp \
-    src/ipc.cpp
+    src/ipc.cpp \
+    src/widget/toxsave.cpp
```
tux3, line 219: You didn't include those in the PR (?)

tux3:
Thanks for the copyright headers, I always forget those!

qtox.pro
dubslow, line 219: Every time, I swear...

PR #675, opened 2014-11-08
Core update

ghost:
Wait until my core patch is merged

### DIFF
--- a/src/coreav.cpp
+++ b/src/coreav.cpp
@@ -18,6 +18,8 @@
 #include "video/camera.h"
 #include <QDebug>
 #include <QTimer>
+#include <QtConcurrent/QtConcurrent>
+#include <functional>
 
 ToxCall Core::calls[TOXAV_MAX_CALLS];
 const int Core::videobufsize{TOXAV_MAX_VIDEO_WIDTH * TOXAV_MAX_VIDEO_HEIGHT * 4};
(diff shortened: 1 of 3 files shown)


src/coreav.cpp
```diff
@@ -250,16 +270,22 @@ void Core::sendCallAudio(int callId, ToxAv* toxav)
...
+    vpx_image_t* images [100]; // I dunno how much images you'll usually need
+    int size;
 
-    calls[callId].videoSource.pushVPXFrame(img);
+    while (calls[callId].active && calls[callId].videoEnabled)
+        if ((size = toxav_recv_video(toxav, callId, images, 100, -1)) >= 0)
+            for(int i = 0; i < size; i++) {
+                calls[callId].videoSource.pushVPXFrame(images[i]);
+                vpx_img_free(images[i]);
+            }
```
tux3, line 286: Isn't 100 a bit much ?

That could be up to 10s of delay on a bad webcam/connection. If we have to wait 10s for a call to end, that may be a problem.

EDIT: Or will toxav_recv_video return immediately when a call ends ?

Have you tested it ?

zetok, line 286: Why would waiting 10s be a bad thing? Considering that user should have a way to stop call at any moment..

tux3, line 286: > Considering that user should have a way to stop call at any moment..

The user would have to wait 10s, unless toxav_recv_video can magically return early when we stop a call, which it probably can now that I think about it. Can it ?

If you disregard that and decide not to wait the 10s, then there are probably many ways to make it crash, like loading a new profile while the call is not-quite-dead-yet.

EDIT: I've done my homework, and toxav_recv_video will indeed return immediately when we toxav_kill_transmission or toxav_kill, so my point is moot.

ghost, line 286: toxav_recv_video returns as soon as there is a packet ready or when error occurred so the number of max images (100) is just an upper limit to avoid overflow and will not cause any kind of delay.

tux3, line 286: On the other hand when starting a new call, the user won't see the first frame until the next 99 are received, or there's an error, yes ? 

That could be a long time, are there any downside you're aware of to using a lower upper limit ?
Have you tested it ?

ghost, line 286: It doesn't fill up the whole buffer instead, it returns size of received images as soon as there is a packet ready.

Also, yes, I have tested it and it worked.

tux3, line 286: Okay, then. Thanks.

dubslow:
How come this doesn't need to call `toxav_do()`?

ghost:
@dubslow Ah yes! `toxav_do()`... I forgot to add it here... The reason this worked for me is that I didn't expect the timeout that `toxav_do()` is responsible for. Thank you for reporting, I'll uptade the patch later this evening.

ghost:
New changes should come soon, this is obsolete.

PR #678, opened 2014-11-09
93a4d30 (Add boolean 'sent_ok' to new history DBs) fix: default value

apprb:


agilob:
Should we remove databases manually now?

tux3:
Yeah, unless someone knows a way to update them automatically. That's probably possible with some clever SQL, @apprb  ?

agilob:
Not really in SQLite3, you need to copy content of a table to another table, remove original table, re-create it, and copy again.

tux3:
Well, then ..

apprb:
This is the reason why history marked unstable now.
I think the best way to handle this situation in future is create a new table for history after each update (something like "history_v1", "history_v2", etc).
But for now, everything will work fine even with old table layout.

agilob:
I think I had outdated information:

> SQLite supports a limited subset of ALTER TABLE. The ALTER TABLE command in SQLite allows the user to rename a table or to add a new column to an existing table. 
> https://www.sqlite.org/lang_altertable.html

I used to create extra tables like 'future_feature1', 'future_feature2' and when I had to add another table I just renamed tables.

PR #691, opened 2014-11-10
Autoshow

dubslow:
A few minor ui things as well


tux3:
Hm, it doesn't work when qTox is minimized or just hidden behind another window.

Is that by design? Because the checkbox is a bit confusing then. Otherwise it's good.

dubslow:
I don't know. @darkcoconuts said it worked for him, and... actually it just occurred to me that I think I misinterpreted the minimize to tray code.

dubslow:
fixed

tux3:
Merged manually in f9cb19a157e962f314cb110cad74026e8edcddf4 after a painful rebase.

PR #723, opened 2014-11-11
fix minor switching bug introduced by 07ea2e

dubslow:


src/widget/widget.cpp
```diff
@@ -838,10 +838,10 @@ void Widget::removeFriend(Friend* f)
         activeChatroomWidget = nullptr;
         onAddClicked();
     }
-    FriendList::removeFriend(f->getFriendID());
-    core->removeFriend(f->getFriendID());
+    FriendList::removeFriend(f->getFriendID(), fake);
+    core->removeFriend(f->getFriendID(), fake);
     delete f;
-    if (ui->mainHead->layout()->isEmpty())
+    if (ui->mainHead->layout()->isEmpty()) // tux3: this should have covered the case of the bug you "fixed" 5 lines above
```
tux3, line 844: It doesn't.

What is this line supposed to do ? If that was it's only purpose, then let's remove it because that didn't work.

dubslow, line 844: I dunno, but it's unrelated to this PR

tux3, line 844: Pls. Don't add meta comments in the code, do that on github.

dubslow, line 844: Yeah... I removed it in the rebased group titles branch

PR #725, opened 2014-11-12
fixes #207

### LINKED ISSUE #207: Command line options
Basically, like Toxic and Venom 

https://img.bi/#/WpzHYKR!BjebnxxQL9NeUXPiNe7bkwUTbgU0IK5Q4pTVPsDd

<blockquote><pre>
~ venom --help
Usage:
venom [OPTION...]
 
Help Options:
-h, --help Show help options
 
Application Options:
-f, --file=<file> Set the location of the tox data file
-l, --loglevel=<loglevel> Set level of messages to log
--offline Start in offline mode
--textview Use textview to display messages
-V, --version Display version number
</pre></blockquote>

Zer0-One:
Adds command-line argument/flag parsing

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -21,11 +21,12 @@
 #include "src/widget/toxsave.h"
 #include "src/autoupdate.h"
 #include <QApplication>
-#include <QFontDatabase>
-#include <QDebug>
-#include <QFile>
-#include <QDir>
+#include <QCommandLineParser>
 #include <QDateTime>
+#include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QFontDatabase>
 #include <QMutexLocker>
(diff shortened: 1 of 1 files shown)


dubslow:
Debug logging needs to be on by default, and as far as I can tell the option doesn't actually do anything.

tux3:
Yes, and we're not going to wrap every qDebug in if { } blocks :)

dubslow:
That's what the message handler is for :P

Zer0-One:
I believe it's good now

dubslow:
Better.

The main reason I didn't do this myself is that the docs say that qt options are parsed before this is even constructed, meaning things like "-p" wouldn't be parsed here...

PR #742, opened 2014-11-13
I implemented a new setting, which toggles the visibility of the system tray icon.

ghost:
Background: On Ubuntu 14.04, system trays in Qt5 don't integrate into Unity.


ghost:
By the way... This is a GREAT Tox client! :+1:

PR #756, opened 2014-11-15
Things

dubslow:
Three bug fixes

### DIFF
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -113,7 +113,7 @@ void Settings::load()
     s.beginGroup("General");
         enableIPv6 = s.value("enableIPv6", true).toBool();
         translation = s.value("translation", "en").toString();
-        showSystemTray = s.value("showSystemTray", false).toBool();
+        showSystemTray = s.value("showSystemTray", true).toBool();
         makeToxPortable = s.value("makeToxPortable", false).toBool();
         autostartInTray = s.value("autostartInTray", false).toBool();
         closeToTray = s.value("closeToTray", false).toBool();        
--- a/src/friend.cpp
+++ b/src/friend.cpp
@@ -46,7 +46,7 @@ void Friend::setName(QString name)
         chatForm->setName(name);
 
         if (widget->isActive())
-            Widget::getInstance()->setWindowTitle(name + " - qTox");
+            Widget::getInstance()->setWindowTitle(name);
     }
 }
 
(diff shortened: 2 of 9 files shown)


src/misc/settings.cpp
```diff
@@ -113,7 +113,7 @@ void Settings::load()
     s.beginGroup("General");
         enableIPv6 = s.value("enableIPv6", true).toBool();
         translation = s.value("translation", "en").toString();
-        showSystemTray = s.value("showSystemTray", false).toBool();
+        showSystemTray = s.value("showSystemTray", true).toBool();
```
tux3, line 116: It's still broken on Ubuntu, I don't think potentially buggy features should be on by default, even if it's Unity's fault.

dubslow, line 116: It's trivial to disable, and I doubt Unity is the majority of users.

dubslow:
@mwuttke97 Unfortunately I don't see line numbers by your comments, so it's hard for me to locate the context for them.

src/misc/settings.cpp
tux3, line 116: It's also trivial to enable.

But people shouldn't have to go change options to un-break things. 
Ubuntu is not a small distro, it's the major Linux distro. We need sensible defaults that work for most people.

Skype isn't broken by default.

ghost:
@dubslow You can look at
https://github.com/dubslow/qTox/commit/fcc8257924ee5340677a60b82c8678261b020ec3#commitcomment-8585652

tux3:
I cherry picked all but fcc8257 for now.

src/misc/settings.cpp
dubslow, line 116: Unity is still a minority, and we shouldn't disable something nice for the majority because of the minority, when the minority can easily fix it with one click. But I guess we'll just have to agree to disagree :)

tux3, line 116: > Unity is still a minority

As far as I know, Unity is the default on Ubuntu, and Ubuntu is the largest Linux distro, is that right? I don't have stats on what distro/OS our users actually have, tho.

I think that we could enable it by default on Windows and Mac at least, and if we had a way to detect Unity or detect that it doesn't work, enable it everywhere else. But it should absolutely not be show by default on Ubuntu imo.

ghost, line 116: I think so, to. It could be disabled on Linux by default. I read that on some distributions it works and on some other distributions (e.g. on Ubuntu with Unity) not.
I recommend a source code like this:

bool iconDefault;
#if  Q_OS_LINUX
iconDefault=false; // we guess, that it doesn't work
#else 
iconDefault=true; // Windows® or Max, on which it should work
#endif

PR #762, opened 2014-11-15
Fixed some bugs...

ghost:
The main window, will never be visible, when the user unchecked 'Show system tray' and checked 'Start in tray'.
EDIT: Furthermore the window will not close, even if it was visible.


ghost:
@tux3  Sorry for "fixing" #772. I reverted it. You did it better then me.

tux3:
I cherry picked c2e93ce and 6778e4d, the rest was already fixed. Thanks!

PR #793, opened 2014-11-17
The showSystemTrayIcon's default value is false on linux and else true.

ghost:
Hi,
there was a little discussion on https://github.com/tux3/qTox/pull/756 (https://github.com/dubslow/qTox/commit/fcc8257924ee5340677a60b82c8678261b020ec3) and a think I found a reasonable soulution for both parties, see the code below:

``` c++
#ifdef Q_OS_LINUX
#define SHOW_SYSTEM_TRAY_DEFAULT (bool) false
#else   // OS is not linux
#define SHOW_SYSTEM_TRAY_DEFAULT (bool) true
#endif
```

``` c++
showSystemTray = s.value("showSystemTray", SHOW_SYSTEM_TRAY_DEFAULT).toBool();
```

I found a fix on http://stackoverflow.com/questions/17193307/qt-systray-icon-appears-next-to-launcher-on-ubuntu-instead-of-on-the-panel. It is not hard to implement, but we'll depend on gtk.
What do you think about it?


ghost:
Sorry for the 12 other commits. I opened a new branch, though they merged in accidentally.
But it seems to don't causes conflicts.
EDIT/PS: Git is verry new to me. I read the book "Pro Git" a little. I hope, I got it. :smile:

tux3:
> Sorry for the 12 other commits. I opened a new branch, though they merged in accidentally.
> But it seems to don't causes conflicts.

Wow, I'll just be cherry picking your new commit again, it looks cleaner in the history ! Try to start a new empty branch maybe :)
(for exemple with <code>git fetch upstream master:newCleanBranch</code>)

> I found a fix on http://stackoverflow.com/questions/17193307/qt-systray-icon-appears-next-to-launcher-on-ubuntu-instead-of-on-the-panel. It is not hard to implement, but we'll depend on gtk.
> What do you think about it?

Well, that's probably our best solution. That question on stackoverflow has been a problem for more than a year, so it's unlikely it'll be resolved soon.
I think that it's a good idea.

ghost:
@tux3 Thank you for your tip!
Though, it took me some hours to do so.
I folowed the instruductions on http://scribu.net/blog/resetting-your-github-fork.html.

PR #824, opened 2014-11-20
New OS X updater

stqism:
The new updater is super nice, it closes a currently running qTox, updates the files from the update folder with a proper admin permissions prompt, wipes the folder, and relaunches qTox magically.

Keep in mind that I did write it in Go so binaries we'd deploy of it in the app would be fully statically linked and work ootb everywhere, but take up 2 mb.


iShift:
Admin permissions? Why?

PR #848, opened 2014-11-24
Update de.ts "calling" will be "ruft an"

rubo77:
"Wird angerufen" would be "is being called"

### DIFF
--- a/translations/de.ts
+++ b/translations/de.ts
@@ -235,7 +235,7 @@ Ignoriere diesen und nutze eine direkte Internetverbindung?</translation>
     <message>
         <location filename="../src/widget/form/chatform.cpp" line="254"/>
         <source>%1 calling</source>
-        <translation>%1 wird angerufen</translation>
+        <translation>%1 ruft an</translation>
     </message>
     <message>
         <location filename="../src/widget/form/chatform.cpp" line="323"/>


translations/de.ts
```diff
@@ -235,7 +235,7 @@ Ignoriere diesen und nutze eine direkte Internetverbindung?</translation>
     <message>
         <location filename="../src/widget/form/chatform.cpp" line="254"/>
         <source>%1 calling</source>
-        <translation>%1 wird angerufen</translation>
+        <translation>%1 ruft an</translation>
```
ghost, line 238: ACK.

cebe, line 238: Shouldn't the original be %1 is calling?

ghost, line 238: Am 25.11.2014 um 10:44 schrieb Carsten Brandt:

> In translations/de.ts:
> 
> > @@ -235,7 +235,7 @@ Ignoriere diesen und nutze eine direkte Internetverbindung?</translation>
> >      <message>
> >          <location filename="../src/widget/form/chatform.cpp" line="254"/>
> >          <source>%1 calling</source>
> > -        <translation>%1 wird angerufen</translation>
> > -        <translation>%1 ruft an</translation>
> 
> Shouldn't the original be %1 is calling?

Would be better.

## 

PGP-Schlüssel:
0xD2A804E4

cebe, line 238: https://github.com/tux3/qTox/pull/856/files#diff-60595a64f8caa4731410f791ff3e8bbaL254

PR #851, opened 2014-11-25
added opencv imgproc to dependencies

cebe:
```
$ ./qtox 
./qtox: error while loading shared libraries: libopencv_imgproc.so.2.4: cannot open shared object file: No such file or directory
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,7 +12,7 @@
 | Qt           | >= 5.2.0    | core, gui, network, widget, xml, sql |
 | GCC/MinGW    | >= 4.8      | C++11 enabled                        |
 | Tox Core     | most recent | core, av                             |
-| OpenCV       | >= 2.4.9    | core, highgui                        |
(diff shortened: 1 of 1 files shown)


cebe:
May need to adjust the debian dependencies also but I could not figure out where they are defined...

dubslow:
Where are you getting this binary from? @stqism

cebe:
https://github.com/tux3/qTox#downloads

"linux binary download"

running on debian wheezy, adding missing packages incrementally, formerly was complaining about the other two opencv libs.

cebe:
I would fix this and https://github.com/dubslow/qTox/commit/3127e023d867e0650108ebf977a9252e81f5a46d myself and send a PR, but I am not able to find the tool that is used to generate the debian packages... Can you give me a hint?

It needs to be adjusted in the [control](https://github.com/tux3/qTox/blob/master/debian/control) file but you have a custom `${shlibs:Depends}` in there which I can not figure out where it comes from.

PR #854, opened 2014-11-25
Activate existing instance

ghost:
This patch adds activation of existing qTox instance when existing instance is running in the background and new instance is started without any command line arguments. This is useful in cases where user has qTox pinned in taskbar/sidebar/panel and if window is hidden application can be quickly revealed by using that pinned icon. This is standard behaviour of pidgin/skype.

Also command line argument -P <profile> is added. It can be used to avoid activation of existing instance and loading specific profile.


ghost:
Works fine.

PR #857, opened 2014-11-25
Focus the text input area when clicking on the text area

titeuf87:
This pulls request focuses the text input area whenever the user clicks on the chat area. 
This is useful when you have selected text, copy-pasted it, and then easily want to get back to the chat: you get a huge area to click on now where you can start typing right away instead of having to click on the much smaller input box.

### DIFF
--- a/src/widget/chatareawidget.cpp
+++ b/src/widget/chatareawidget.cpp
@@ -87,6 +87,7 @@ void ChatAreaWidget::mouseReleaseEvent(QMouseEvent * event)
             }
         }
     }
+    emit onClick();
 }
 
 void ChatAreaWidget::onAnchorClicked(const QUrl &url)
--- a/src/widget/chatareawidget.h
+++ b/src/widget/chatareawidget.h
@@ -41,6 +41,7 @@ public slots:
 
 signals:
     void onFileTranfertInterract(QString widgetName, QString buttonName);
+    void onClick();
 
 protected:
     void mouseReleaseEvent(QMouseEvent * event);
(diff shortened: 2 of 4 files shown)


zetok:
From your description it would seem that it introduces regression to PPT[1] functionality, triggered by `P` key[2] when chat area is selected, but not when input field is selected.

Is that right?

[1] Push to talk
[2] key triggering PPT is supposed to be changeable by user later in future

titeuf87:
@agilob Double clicking to select text still works.

@zetok You're right, this would break PPT. Maybe only allow focus change when no audio call is in progress?

tux3:
> Maybe only allow focus change when no audio call is in progress?

That seems a bit too confusing, people might wonder why sometimes it works and sometimes not.

titeuf87:
Yeah :(
I'm not sure what else is possible.

I've used Pidgin for years. There whenever you start typing the input box gets the focus, which makes switching windows a lot easier.

TheLastProject:
I personally feel PTT should be moved to some other hotkey (Alt+P perhaps, something you wouldn't type in chat) and typing should focus the chat box for the current conversation. It is pretty annoying that you have to click it again.

zetok:
@titeuf87 Instead of some workaround there could be made a proper implementation of PTT.

PTT should have a global detection (maybe with a way to disable it) bound to some key that isn't used by default (like ``` or `CAPS LOCK`). Also user would have a way to set another key or combination of them.

This way you would have a proper support for PTT and input filed autofocus.
That probably will be considerably more work to do, but you at least would get best you could without breaking UX.

Also, when global PTT would be disabled in favour of chat window PTT, autofocus on input field when clicking in chat window would too be silently disabled.

On a side note, I wonder how behaviour of PTT should be defined when it'll be global and there will be more than 1 call 'active' (should 'on hold' be considered?). I don't think that it should send sound to all calls, but rather to the last one on which user was focused (or maybe an option?).

Hope it helps.

tux3:
I changed push-to-talk to CTRL+P and merged this manually in 983c0b7e6b06744e44c6bc859745811c4719a2f6

PR #861, opened 2014-11-26
Group chat button at the bottom of contact list resized

ghost:
Image that was used for group chat button icon was bigger than other button images. While on some systems/desktop environments it looked good (windows or cinnamon) on others (at least ubuntu unity) it was slightly too big. This must be Qt quirk. Now group chat button looks of same size on all platforms/systems.


tux3:
Merged in 1b7193fb1f8bc4d9362febe29e76bf790d7288d6.

PR #862, opened 2014-11-26
Global auto-away

ghost:
Implements desktop-wide auto-away (like in skype/pidgin/others).

Be aware:
MacOS code was not tested, i would love if anyone could test it before merging!
X11 part works ok for me.
Windows part not tested but there isnt much room for mistakes.

Also someone knowledgeable please verify qtox.pro, i am not sure i added dependencies properly. No idea what to link for macs, not sure if x11 deps are in correct spots, not sure if user32 is linked in auto or we need to add dependency manually.

Also now timer checks for system idle time every 10 seconds. It works but if anyone have better ideas i would love to hear them.

I would love some feedback on all of this ;)


dubslow:
Interesting.

Note that on my Cinnamon installs (one LMDE/Cinnamon 2.0, the other LM17/Cinnamon 2.2), Skype instantly detects any activity whatsoever, so I doubt they're only checking the idle time every so often.

Additionally, I anticipate the need for further unrelated platform specific code (e.g. for screenshots/desktop streaming), so I'd prefer to organize it differently: a new `src/platform/` folder, which contains something like:

```
winbloze/
osx/
x11/
idletimer.h
screenshot.h
desktop.h
```

The top level headers would then conditionally include files from the subfolders as appropriate. The subfolders would of course each contain a `idletimer.h idletimer.cpp screenshot.h screenshot.cpp` etc

stqism:
Your defines for OS X are horribly broken, lemmie shoot you a PR

stqism:
Cool, fixed a linking issue and the define issue with a PR against this branch, https://github.com/novist/qTox/pull/2

tux3:
I'm not too fond of putting all the non-portable code in a platform.{cpp,h} either, that would become messy fast.

I'd rather like @dubslow's solution or something similiar. Keep filenames meaningful and if possible abstract the 'public API' away from the platform-specific details.

I still haven't tested it on various systems but the code does look good otherwise :)

dubslow:
@novist I know you've said Pidgin is a source, are there any other sources?

ghost:
Ill move around platform-specific code then.

@dubslow only mac code came from pidgin (GPLv2), rest of it comes from me.

EDIT:
how about

```
├── ./platform
│   ├── ./platform/platform.h
│   ├── ./platform/osx
│   │   └── ./platform/osx/timer.cpp
│   ├── ./platform/win
│   │   └── ./platform/win/timer.cpp
│   └── ./platform/x11
│       └── ./platform/x11/timer.cpp
```

My reasoning for single platform.h is that it would contain only few declarations anyway. Creating separate files seems unnecessarily bulky. Cpp files on the other hand usually contain more code so they can be separate.

dubslow:
The headers should all be separate so that they can be included individually as needed (i.e. call it timer.h instead). Be sure to ifdef guard the cpp files as well.

ghost:
At the moment i used qmake variable and included src like this:

```
src/platform/$${PLATFORM_DIR}/timer.cpp
```

Is that ok or too hacky?

EDIT:
actually nvm, #ifdef is better because then its possible to easily edit files of other platforms.

But qmake complains about being multiple timer.o due to multiple timer.cpp in different dirs. So it can be either

```
├── ./platform
│   ├── ./platform/timer.h
│   ├── ./platform/osx
│   │   └── ./platform/**osx/osx_**timer.cpp
│   ├── ./platform/win
│   │   └── ./platform/**win/win_**timer.cpp
│   └── ./platform/x11
│       └── ./platform/**x11/x11_**timer.cpp
```

But marked parts i dont like with **..** because adding prefix defeats purpose of dir.
Or:

```
├── ./platform
│   ├── ./platform/timer.h
│   ├── ./platform/osx_timer.cpp
│   ├── ./platform/win_timer.cpp
│   └── ./platform/x11_timer.cpp
```

I am leaning towards later variant. Even if all files would be in same dir prefix should provide enough of grouping for them to not be a mess.

dubslow:
I'll try and take a whack at it

tux3:
> I am leaning towards later variant.

```
├── ./platform
│   ├── ./platform/timer.h
│   ├── ./platform/osx_timer.cpp
│   ├── ./platform/win_timer.cpp
│   └── ./platform/x11_timer.cpp
```

That's pretty much what Qt itself does internally, I think it's a good solution.

dubslow:
Hmm, perhaps reverse the file name and platform name? Though that's extreme pedantry lol.

tux3:
> Hmm, perhaps reverse the file name and platform name?

Agreed :)

PR #868, opened 2014-11-27
Increase buffer for slow connections like satellite

cjpit:
Just re-adding this fix in, since audio got pushed across to a seperate thread.  PS first github pull request :) let me know if I'm doing anything wrong - Ah i think i was meant to do this in a separate branch

### DIFF
--- a/src/audio.cpp
+++ b/src/audio.cpp
@@ -188,7 +188,7 @@ void Audio::playAudioBuffer(ALuint alSource, const int16_t *data, int samples, u
         alDeleteBuffers(processed - 1, bufids + 1);
         bufid = bufids[0];
     }
-    else if(queued < 16)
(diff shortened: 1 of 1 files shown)


dubslow:
Generally a non-master branch yes, but for small things doing it on master is usually fine.

tux3:
@irungentoo any opinion on this ?

irungentoo:
The reason why the buffer size is limited is so the delay doesn't increase in cases where audio is sent slighty faster than the receiver plays it (it's a simple method that works pretty well).

Packets usually don't arrive in exact intervals, on bad internets they sometimes arrive in quick bursts which seems to be the issue here.

32 is probably the most you can increase it to before some people start complaining about delays.

It's something that needs to be fixed in core.

PR #863, opened 2014-11-26
Option for tray icon to display user status

ghost:
Setting added for tray icon to be displayed as user status. Default value is **off** (original tox icon as it is now). Useful when user wants to see his current status in tray. Especially handy when auto-away kicks in, we immediately can see if we are away or not without a risk of breaking away status.

PR #870, opened 2014-11-27
Option for tray icon to display user status

ghost:
Cleaned up version of original PR https://github.com/tux3/qTox/pull/863

Setting added for tray icon to be displayed as user status. Default value is off (original tox icon as it is now). Useful when user wants to see his current status in tray. Especially handy when auto-away kicks in, we immediately can see if we are away or not without a risk of breaking away status.

### DIFF
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -125,6 +125,7 @@ void Settings::load()
         makeToxPortable = s.value("makeToxPortable", false).toBool();
         autostartInTray = s.value("autostartInTray", false).toBool();
         closeToTray = s.value("closeToTray", false).toBool();        
+        trayShowsUserStatus = s.value("trayShowsUserStatus", false).toBool();
         forceTCP = s.value("forceTCP", false).toBool();
         useProxy = s.value("useProxy", false).toBool();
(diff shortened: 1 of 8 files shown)


dubslow:
Much cleaner lol. Two things:

1) The window geometry save fixes, which are independent of the feature this PR is about, should really be in a separate commit.

2) There are a couple of deletions at the top of generalform.cpp which look like a mistake to me

ghost:
Damn it i thought i got rid of window geometry saving code. Ill work on that. U mean these deletions?

```
- bodyUI->transComboBox->setCurrentIndex(locales.indexOf(Settings::getInstance().getTranslation()));
- bodyUI->cbMakeToxPortable->setChecked(Settings::getInstance().getMakeToxPortable());
```

dubslow:
Yep those are the ones

ghost:
K now its ready i think. All were patched before your comment heh. Besides i opened another PR https://github.com/tux3/qTox/pull/871 with window saving stuff. Now all should be golden.

dubslow:
@ItsDuke doesn't seem to be available right now, so I'll merge and then later we can update the images and remove the settings option.

PR #871, opened 2014-11-27
Improved window and splitter geometry saving

ghost:
At the moment window size and splitter position is not always saved. One instance being abnormal application termination. This patch implements saving of window and splitter states when they change.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1014,17 +1014,6 @@ void Widget::removeGroup(Group* g, bool fake)
     contactListWidget->show();
 }
 
-void Widget::saveWindowGeometry()
-{
-    Settings::getInstance().setWindowGeometry(saveGeometry());
-    Settings::getInstance().setWindowState(saveState());
-}
-
-void Widget::saveSplitterGeometry()
-{
-    Settings::getInstance().setSplitterState(ui->mainSplitter->saveState());
-}
-
 void Widget::removeGroup(int groupId)
 {
     removeGroup(GroupList::findGroup(groupId));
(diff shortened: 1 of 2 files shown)


dubslow:
You chose some odd locations to put new functions in widget.cpp lol. It's definitely not very well organized, but putting the two new functions between `removeGroup(Group* g, bool fake)` and `removeGroup(int groupId)` seems overly weird.

I'll fix it myself and merge.

dubslow:
This isn't actually working for me... qTox always starts with the same geometry, no matter what. I'm just closing it in the regular manner too.

https://github.com/dubslow/qTox/commits/pr871 Note that in addition to moving the functions I had to rebase on master in order for the av to compile.

ghost:
Thanks for rebasing \o/ As for the weird placement - header is a mess so dont blame me for cpp :p QtCreator put those functions there hehe. See in header private section:

```
void removeGroup(Group* g, bool fake = false);
void saveWindowGeometry();
void saveSplitterGeometry();
```

And thats where functions in cpp went. Anyway.

As for it not working - window geometry restoring should have been before showing window. Fixed that.

This rebasing messed things though i think :|

dubslow:
Hmm now I see 14 commits here, not 1 or 2...

ghost:
Yeah i merged your pr871 branch. Maybe i should not have done it?

dubslow:
Much better, only the two commits now. Now it needs be fixed lol

ghost:
Fixed? You mean instance where it did not work for you? Because i fixed one instance where it didnt work if "start in tray" was unchecked. Basically problem was restoring geometry after window is shown where it should happen before. Thats fixed already ;)

dubslow:
Ah ok, my laptop is dead again so I'm basically out of commission till Sunday evening at the earliest lol

src/widget/widget.cpp
```diff
@@ -73,7 +73,12 @@ Widget::Widget(QWidget *parent)
 void Widget::init()
 {
     ui->setupUi(this);
-    
+
+    //restore window state
+    restoreGeometry(Settings::getInstance().getWindowGeometry());
+    restoreState(Settings::getInstance().getWindowState());
+    ui->mainSplitter->restoreState(Settings::getInstance().getSplitterState());
+
```
dubslow, line 81: Why did you move this chunk?

https://github.com/tux3/qTox/commit/f8063fc809a93c362944deaf0cffc46f6ca17fa8

Edit: Is this the primary fix of the pr?

ghost, line 81: https://github.com/tux3/qTox/pull/871#issuecomment-64906517

dubslow:
Twas rebased before merging cause core api change

Does anyone know if/how I can mark this as merged rather than closed?

Issue #515, opened 2014-10-23
Rejecting an incoming call


tux3:
Maybe right clicking could reject an incoming call. Or maybe another button/a menu on hover?

agilob:
I think I can do that in notification, but this will have to wait, I got a lot of work from univ.

darkman088:
I think that adding a new red button, as on all the cell phones will be the most elegant solution.
By right clicking, it might be the easiest to implement, but it won't be intuitive - one would need to know
about the feature, otherwise it's quite hard to discover... And for the Mac users you can forget the right click for sure :-D

TheLastProject:
What about a drop down menu on the "incoming call" button, giving you an option to reject?

A bit like this but much less ugly: 
![snapshot69](https://cloud.githubusercontent.com/assets/1885159/4769355/ba4d73ac-5b78-11e4-9201-6775e22d7503.png)

Of course, a reject button in the notification would also be good.

darkman088:
I do not like the drop down menu option. It is too cumbersome... I have seen software like that, I do not remember the exact example but I do not like it...

suhr:
cast @ItsDuke

ItsDuke:
@darkman088 I'm inclined to agree. Ultimately the user should be able to accept or deny a call with just one click. Making them go through a dropdown just means extra work for the user, and we don't want that.

The main question I have is, are you currently viewing this person's chat window and they initiate a call, or are you in a chat with someone and a completely different contact initiates a call? Only asking because each one needs a different handling.

agilob:
Let me show my vision:
![stage1](https://cloud.githubusercontent.com/assets/2544251/4885855/8d064890-6377-11e4-9a78-3f02305fe0f6.png)
![stage2](https://cloud.githubusercontent.com/assets/2544251/4885854/8ce460b8-6377-11e4-9d49-85ec9444967b.png)

After call is ended it comes back to the first stage where both buttons are green.

I have it implemented and it works but needs CSS some polishing and more opinions.

ItsDuke:
How do folks feel about something like this?

![ring](https://cloud.githubusercontent.com/assets/7714365/4890594/6b18824c-63a0-11e4-9b2d-bc875be41a3d.gif)

tux3:
I like the idea. 
It looks like a tooltip, so does it only appear when the mouse is over the call button ? Should it stay here until you dismiss it ?

darkman088:
@ItsDuke:
[...]

PR #873, opened 2014-11-28
Rejecting call

agilob:
It's not best, looks (very) ugly, but well... it's something. Ideas how to improve it are welcome.

reject button

Technically, it can be merged if you're not disgusted with that.


agilob:
reject button1

Why are all SLOTs called som many times?
https://gist.github.com/agilob/92ef79f4feed0c0383a4

zetok:
Sorry, but it looks confusing and disgusting.

agilob:
Yup, that's what I wrote ;) How would you make it?

zetok:
I wouldn't since I cannot into coding :P
But rejecting call in the way @ItsDuke [proposed](https://github.com/tux3/qTox/issues/515#issuecomment-61551871) seems to be the best way IMO.

agilob:
I cannot into css and design.

tux3:
I also think that it doesn't look very good :)

Closing for now.

PR #874, opened 2014-11-28
Spanish Translation Created

ACKino:


translations/es.ts
```diff
@@ -0,0 +1,1639 @@
...
+    <message>
+        <location filename="../src/widget/form/addfriendform.cpp" line="38"/>
+        <source>Message</source>
+        <comment>The message you send in friend requests</comment>
+        <translation>Mensaje</translation>
+    </message>
+    <message>
+        <location filename="../src/widget/form/addfriendform.cpp" line="39"/>
+        <source>Send friend request</source>
+        <translation>Mandar solicitud de Amig@</translation>
```
dubslow, line 101: Dat typ@ though

Edit: oh snap is this a common thing on the internet?

translations/i18n.pri
```diff
@@ -1,6 +1,7 @@
 # For autocompiling qm-files.
 
-TRANSLATIONS = translations/bg.ts \
+TRANSLATIONS = translations/es.ts \
```
zetok, line 3: Should be in alphabetical order

dubslow, line 3: The build file order doesn't matter, but it does matter in the UI list.

zetok, line 3: @dubslow I beg to differ. Keeping things organized helps maintain whole thing in good shape, beside making it easier to navigate and look through. Once it starts getting messy…

translations/es.ts
ACKino, line 101: Yes. It avoids being gender specific. I figured tox users might appreciate that.
Alternatives would be to write amigo/amiga or avoid the word friend altogether and use contact which only translates as contacto, contacta being the verb.

translations/i18n.pri
ACKino, line 3: will fix build file and UI list on next translation update, if nobody else has by then,

PR #875, opened 2014-11-29
Implemented filter audio

ghost:
Hi,
as echo cancellation was needed I tried to integrate the library as provided by @irungentoo into qTox.
I also implemented a new audio filter setting - the default value is **false**.
## TODO:
- Makefiles has to be customized for Windows and Mac.
- The libfilteraudio-installer has to be customized for Mac and to be written for Windows.
- It has to be tested.

### DIFF
--- a/src/audiofilterer.cpp
+++ b/src/audiofilterer.cpp
@@ -1,0 +1,52 @@
+/*
+    Copyright (C) 2014 by Project Tox <https://tox.im>
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    This program is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+    See the COPYING file for more details.
+*/
+
+
+#ifdef QTOX_FILTER_AUDIO
+
+#include "audiofilterer.h"
+extern "C"{
+#include <filter_audio.h>
+}
+
+void AudioFilterer::startFilter(unsigned int fs)
+{
+    closeFilter();
+    filter = new_filter_audio(fs);
+}
+
+void AudioFilterer::closeFilter()
+{
+    if (filter)
+        kill_filter_audio(filter);
+}
+
+
+void AudioFilterer::filterAudio(int16_t* data, int framesize)
+{
+    if (!filter)
+        return;
+
+    filter_audio(filter, (int16_t*) data, framesize);
+}
+
+
+AudioFilterer::~AudioFilterer()
+{
+    closeFilter();
+}
+
+#endif // QTOX_FILTER_AUDIO
--- a/src/audiofilterer.cpp
+++ b/src/audiofilterer.cpp
@@ -26,7 +26,7 @@ void AudioFilterer::startFilter(unsigned int fs){
 }
 
 void AudioFilterer::closeFilter(){
-    if (filter != nullptr)
+    if (filter)
         kill_filter_audio(filter);
 }
 
--- a/src/audiofilterer.cpp
+++ b/src/audiofilterer.cpp
@@ -17,7 +17,7 @@
 #include "audiofilterer.h"
 
 extern "C"{
-#include "filter_audio/filter_audio.h"
+#include <filter_audio.h>
 }
 
 void AudioFilterer::startFilter(unsigned int fs){
--- a/src/audiofilterer.cpp
+++ b/src/audiofilterer.cpp
@@ -1,0 +1,31 @@
+#include "audiofilterer.h"
+
+#ifdef __cplusplus
+extern "C"{
+#endif
+
+#include "filter_audio/filter_audio.h"
+static struct Filter_Audio * filter;
+
+#ifdef __cplusplus
+}
+#endif
+
+void AudioFilterer::startFilter(unsigned int fs){
+    if (!filter)
+        kill_filter_audio(filter);
+    filter = new_filter_audio(fs);
+}
+
+void AudioFilterer::closeFilter(){
+    if (filter != 0)
+        kill_filter_audio(filter);
+}
+
+
+void AudioFilterer::filterAudio(const int16_t *data, int samples){
+    if (!filter)
+        return;
+
+    filter_audio(filter, (int16_t*) data, samples);
+}
--- a/src/audiofilterer.cpp
+++ b/src/audiofilterer.cpp
@@ -31,11 +31,11 @@ void AudioFilterer::closeFilter(){
 }
 
 
-void AudioFilterer::filterAudio(const int16_t *data, int samples){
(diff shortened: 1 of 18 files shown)


src/audiofilterer.cpp
```diff
@@ -0,0 +1,44 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+    See the COPYING file for more details.
+*/
+
+#include "audiofilterer.h"
+
+extern "C"{
+#include "filter_audio/filter_audio.h"
```
zetok: Why there's a hard dependency on `filter_audio.h` being in a subdirectory, instead of relying on what system provides in `/usr/include`, especially when considering that you include script installing `libfilteraudio.so` system-wide?

ghost: @zetok You are right, I'll fix that.

src/coreav.cpp
```diff
@@ -65,6 +68,13 @@ void Core::prepareCall(int friendId, int callId, ToxAv* toxav, bool videoEnabled
         calls[callId].sendVideoTimer->start();
         Camera::getInstance()->subscribe();
     }
+
+    if (Settings::getInstance().getFilterAudio()){
+        Core::filterer[callId] = new AudioFilterer();
```
dubslow, line 79: Before allocating a new filterer, you should check if there already is one, and if so, garbage collect it. (callId can be reused, so this is necessary.)

Edit: Ah, I see, you GC in onAvCancel, but I don't believe we can assume that it will be properly called to clean up the call every time.

src/coreav.cpp
```diff
@@ -65,6 +68,13 @@ void Core::prepareCall(int friendId, int callId, ToxAv* toxav, bool videoEnabled
...
+        filterer[callId]->startFilter(48000);
+    } else{
+        filterer[callId] = nullptr;
```
dubslow, line 86: Ditto, have to delete before forgetting.

(See above edit as well)

src/coreav.cpp
```diff
@@ -246,8 +256,13 @@ void Core::sendCallAudio(int callId, ToxAv* toxav)
             return;
         }
 
-        if((r = toxav_send_audio(toxav, callId, dest, r)) < 0)
+        if (filterer[callId])
+            filterer[callId]->filterAudio((int16_t*) buf, framesize);
+
+        if((r = toxav_send_audio(toxav, callId, dest, r)) < 0){
             qDebug() << "Core: toxav_send_audio error";
+            return;
```
dubslow: Why?

src/coreav.cpp
```diff
@@ -321,6 +336,12 @@ void Core::onAvCancel(void* _toxav, int32_t callId, void* core)
 
     calls[callId].active = false;
 
+    if (filterer[callId] != nullptr){
```
dubslow: Why not just `if (filterer[callId])` like you have above? Be consistent.

dubslow:
The code seems largely fine. I assume you've tested it?

Additionally, please re-style everything to qTox standards. Although I prefer K&R myself, qTox has managed to maintain consistently Allman style bracing. Also, though this is somewhat less consistent than the bracing, I prefer and qTox mostly (but not entirely) uses `type* var;` style pointers, as opposed to `type * var;` or `type *var;`. I'd appreciate it if you could change those as well.

ghost:
@dubslow I pointed out inconsistence in `void Core::playCallAudio(void* toxav, int32_t callId, const int16_t *data, uint16_t samples, void *user_data)`. It's also in `void Core::playCallVideo(void*, int32_t callId, const vpx_image_t* img, void *user_data)`.

src/coreav.cpp
```diff
@@ -294,14 +316,16 @@ void Core::sendCallVideo(int callId)
...
         calls[callId].muteMic = !calls[callId].muteMic;
     }
 }
 
 void Core::volMuteToggle(int callId)
 {
-    if (calls[callId].active) {
+    if (calls[callId].active)
+    {
         calls[callId].muteVol = !calls[callId].muteVol;
```
ghost, line 335: Allman :)

src/coreav.cpp
```diff
@@ -294,14 +316,16 @@ void Core::sendCallVideo(int callId)
 
 void Core::micMuteToggle(int callId)
 {
-    if (calls[callId].active) {
+    if (calls[callId].active)
+    {
         calls[callId].muteMic = !calls[callId].muteMic;
```
ghost, line 327: Allman :)

ghost:
@dubslow I have tested it -- but only on **one** Notebook (two instances of qTox).
I have no other PC which can run qTox on Linux and I don't know -- and don't want to know -- how to compile on _Windoze_. I have (payed an academic version of) MS VS 2010 (naively, when I was younger and still had been using Windows), but qTox is written in C++11 which is not supported by VS 2010.

ghost:
@tux3 When you merge it, please don't forget to reconfigure your Jenkins server. Does it run on Linux or on Microsoft Windows or on something dubious else?

I intended to don't disorder the bootstrap script. So I added a new simple and customizable script which downloads, compiles and installs `filter_audio`, assuming you are on Linux. Could someone rewrite it to bat for Windows?

ghost:
@dubslow Which C++ code style should I apply in QtCreator: `Qt [built-in]` or `GNU [built-in]`?

dubslow:
I've never used QtCreator. :P

excalibur1234:
there is a audio filter written for utox: https://github.com/irungentoo/filter_audio
it is intended to be used by other tox clients: https://github.com/notsecure/uTox/issues/227

zetok:
@excalibur1234 learn to read.

dubslow:
That's exactly what this is. I've been meaning to get around to this, but I'm more hesitant to touch audio code than other parts of qTox. Right now my focus is on encryption.

tux3:
> @tux3 When you merge it, please don't forget to reconfigure your Jenkins server. Does it run on Linux or on Microsoft Windows or on something dubious else?

@mwuttke97 It's a Linux box, don't worry I'll take care of Jenkins.

dubslow:
@mwuttke97 If you can, please rebase on master, and additionally guard everything here by a compile time define (e.g. `QTOX_ECHO_CANCEL` or `QTOX_FILTER_AUDIO` or the like). I'll try and get to it This Weekend™, and you doing either or both of those would make my part easier.

ghost:
I'll open a new PR soon.

dubslow:
That probably wasn't necessary, you could have force pushed to the same branch on Github. Oh well lol

Edit: Didn't really mean for you to squash it either... oh well haha

ghost:
@dubslow Force pushing didn't work because I reinstalled Ubuntu (two times at two days), so my local git-repository got lost. After pulling my `orgin filter_audio` branch and merging/rebasing it from/to the (up-to-date) local `master`, `git push -f orgin filter_audio` said already up-to-date. So I had to rebase it and I pushed my local branch to a new `orgin filter_audio2` branch. I opened a new PR. There were some conflicts so the process was quiet tricky.
@dubslow I thought, one conflict-free single-context-related commit could be better then some ugly and conflicted 14 ones.

dubslow:
Yeah, in this case I don't think it's worse, though generally I'm a fan of keeping history and small commits.

Also, I force push with `git push mine +filter_audio`. Is that different than `-f`?

PR #881, opened 2014-11-30
Update Polish translation

zetok:


tux3:
Merged in 2d0fb3595b58fa67850b6b698c70fd5d9587dbb4

zetok:
Why did you commit it with your signature?

dubslow:
Twas cherry picked (don't know why), and HEAD should always be signed.

It's possible he cherry picked it to reduce the commit count by one (by eliminating the separate merge commit).

zetok:
I guesses that it was due to cherry picking and a need for head to be signed, but I don't really like my commits having signature other than my own.

dubslow:
Hmm... did you sign your original commit yourself?

This is a bit of an issue with cherry picking things, but at the end of the day it has to be signed by tux3 or me in some way; when we sign a merge commit that's effectively (but not actually) signing every commit in the merge.

I suppose it wouldn't hurt though to have your signature as well... I guess I'll try and avoid cherry picking your commits, but no guarantees and I can't speak for tux3.

zetok:
Yep, commit was signed by me. Since I've learned how to sign commits all of them are signed, except for like 4 commits, which aren't going upstream anyway.

I realize that merging too signs all that has been merged, but it doesn't remove signature, as it would seem to be the case with cherry picking, judging by the output of `git log --show-signature`

What I want to say is: if you can, please avoid removing signature. If that can't be done for some reason, so be it…

cebe:
The reason for this is how git works, when you sign, you always sign the commit hash which is unique for the commit itself including its whole history which it is based on.
If you merge, the commit stays exactly as it was so the signature can remain. Another commit (merge commit) is generated which is signed by the person who did the merge.
Cherrypicking is creating a new commit with another parent an thus a new history + new commit hash. Signature can not remain as you have signed another commit hash with yours.

PR #894, opened 2014-12-04
Cleaning up of Emote button and new images

Xylemon:
I personally thought the the mic and volume buttons were very out of place compared to the rest of the GUI. I created some new ones and fixed up the emote button so it looks properly round on the smile. Criticism is welcomed.


dubslow:
Muchhh better lol

PR #898, opened 2014-12-05
Italian translation: update

Ansa89:


dubslow:
I'm gonna wait with this one for a few days, encryption should be coming Soon™ with some major string changes far more worthy of an update than one line.

Encryption works, I just have to work out a GUI threading issue.

Ansa89:
Leave this here and add a comment when the major update is pushed.

PR #930, opened 2014-12-13
Use new qtox_sudo to do updates as System Administrator

stqism:


dubslow:
Your indentation is wayyyy [out of whack](https://github.com/stqism/qTox/blob/patch-8/osx/qtox_sudo.m#L72-L102); it looks like you're mixing spaces and tabs. Get a real text editor :3

(And when you fix it, please fix it to Allman style. qTox is pretty consistently Allman style.)

PR #933, opened 2014-12-14
filter_audio (rebased)

ghost:
Added some `#ifdef QTOX_FILTER_AUDIO`s.
Conflicts:
- qtox.pro
- src/coreav.cpp

### DIFF
--- a/install_libfilteraudio.sh
+++ b/install_libfilteraudio.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+CURRENT_DIR=$DIRSTACK
+SOURCE_DIR="filter_audio/"
+SOURCE_PATH="./../"
+LIB_DIR="/usr/local/lib/"
+INCLUDE_DIR="/usr/local/include/"
+
+echo "Clone filter_audio from GitHub.com"
+cd $SOURCE_PATH
+git clone --quiet https://github.com/irungentoo/filter_audio.git $SOURCE_DIR
+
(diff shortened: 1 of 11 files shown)


ghost:
Here you have one conflict-free rebased, squashed commit.

apprb:
Is there a particular reason to compile irungentoo's filter_audio code as standalone library? 
Probably using `git submodule` is more suitable for this purpose.

ghost:
I don't know how to use `git submodule`.

dubslow:
It's also true that jenkins.libtoxcore.so doesn't use git (at least currently).

zetok:
@apprb isn't that a standalone library though?

ghost:
@apprb I read a little about `git submodule`s. I come to the conclusion that using `git submodule` would be IMHO **inconsistent**, because we also have no _submodule_ for `libtoxcore`. We also have to install `libtoxcore.so` on UNIXes. IMO, treating `filter_audio` as a separate library is here good.
Furthermore, adding the C-sources and C-headers to `qtox.pro` can cause some problems, because `filter_audio` is written in C and qTox is written in  C++. There are some differences between C and C++, so it is IMO correct to separate them using shared objects (`so`s on UNIXes and `dll`s on Windows'es).

dubslow:
@mwuttke97 is there a particular reason for the `unix|win32` in `unix|win32: LIBS += -lfilteraudio` ?

Issue #864, opened 2014-11-26
Small visual issues


TheLastProject:
Yes, these are indeed very minor, but I wanted to document them regardless.
- [x] Changing group title/user alias by clicking on nick causes edit bar to touch the microphone button

![snapshot123](https://cloud.githubusercontent.com/assets/1885159/5205384/03af67b6-7597-11e4-89fe-2df773506f26.png)
- [ ] Line below person/group chatting with does not extend all the way left, but does extend all the way right
- [x] Contact info is aligned slightly below own info, causing the line below the contact info to not align nicely

![snapshot124](https://cloud.githubusercontent.com/assets/1885159/5205414/453acfae-7597-11e4-8d85-cafe9c101be1.png)

suhr:
@agilob

TheLastProject:
Sorry for bumping this, but I wanted to make sure that, despite the pull request made for this, the second issue still exists:

![linemargin](https://cloud.githubusercontent.com/assets/1885159/6766981/00c83358-d01d-11e4-9f48-dbc736a78262.png)

agilob:
This issue can't be fixed, that white space is QSplitter (separator), just to show you I changed its color from white to green:
![addons2](https://cloud.githubusercontent.com/assets/2544251/6767074/108b719c-d018-11e4-8c1c-79a9ae657d48.png)

ghost:
Remove the line; problem solved :^)))

agilob:
Line removed in my PR with themes. It's not worth writing size-independent style for qsplitter just to add a few pixels.

zetok:
@agilob but your PR hasn't been merged yet.. #1882

tux3:
It would be hard to make the line cross over the QSplitter, what if we made the line have the same margin on the right as on the left?

![snap8](https://cloud.githubusercontent.com/assets/5155966/12470404/1ff30a60-bff4-11e5-913d-18c40af8071e.jpg)

It's only a couple lines changed, but I think it looks better.

zetok:
I guess that adding a margin could work, given that it would seem like a by-design UI, rather than a bug.

PR #938, opened 2014-12-15
removing border from contact name linedit

agilob:
@TheLastProject  I think it fixes first two points from #864, but I don't understand the third issue. Could you visualise it?


dubslow:
This only fixes the first point I think. The second two are separate issues about the same line separating the contact info at the top and the main message window.

agilob:
The only difference I can see is different font size. I think I really need some visualisation for that.

TheLastProject:
@agilob
453acfae-7597-11e4-8d85-cafe9c101be1
I've coloured the lines red and blue and drew them a bit further than they really are, to show to difference they have, they are not on the same "height", which is what I meant with not aligning.

dubslow:
Is setting the color necessary? Won't that mess with some themes?

zetok:
@dubslow colour is there to show what needs to be done, not to be used

agilob:
LOL. Volume and microphone buttons have pictures 22x18 what also scales to 25x20. Layout height is limited to 40px so adding a spacing between those two buttons was the problem. Fixed by removing fixed size of buttons, scaling bottom button to 25x18, adding spacing between buttons and aligning them to top/bottom.

The difference in height between lines on my screen was 2px, nice catch @TheLastProject

qtox1
qtox3

PR #943, opened 2014-12-17
Make quote colour more readable

zetok:
screenshot3 `new | old`

It's not according to mock-up colour scheme, but it looks much better and, what is more important, text is more readable.

### DIFF
--- a/ui/chatArea/innerStyle.css
+++ b/ui/chatArea/innerStyle.css
@@ -34,7 +34,7 @@ div.date_me {
 }
 
 span.quote {
-  color: #6bc260;
+  color: #279419;
 }
 
 div.green {


suhr:
We need some wizards of colour.
cast @ItsDuke

ItsDuke:
@zetok while I think _looks much better_ is subjective, I do agree that the one on the left is more legible. I'm curious, how much and how often will quoted text be used in a typical conversation? I'm not trying to build a case, just legitimately curious.

suhr:
> I'm curious, how much and how often will quoted text be used in a typical conversation?

Well, it is actually used pretty often, but usually as few lines not as big chunks of text.

agilob:
It doesn't matter how often it will be used, when it's more readable/looks better.

zetok:
@ItsDuke I don't have a metrics for how often, but for me it was often enough to make me patch qTox for myself and ask people in groupchat what they think about the difference. Since opinions that I got were positive, I've made PR with a change.

Yep, I just realized some time after making PR that _looks much better_ is in fact how I feel about colour based on how easy it is to read text :P

It's rarely used with larger pieces of text, screenshot above has a lot of text to better show difference between colours.

ItsDuke:
@suhr I figured as much, just wanted to be sure

@agilob ayy lmao

@zetok I feel ya. Do you have the specific color so I can fiddle with the mockups to see how it looks?

zetok:
@ItsDuke it's `#279419`. I don't think that it's the best one, just something that looked readable, was green and didn't look too gloomy. If you think that it should be changed to some other colour that will be in green range and will be readable, please say so :)

PR #947, opened 2014-12-18
Autorun on OS startup

ghost:
Adds autorun on OS startup behaviour when "Start in tray" option is checked. Supported platforms are windows and xfreedesktop-compatible unixes. It works with multiple profiles although after creating/switching profile checkbox in settings is not updated. I am unsure how to go about it. Maybe it is another problem of not updating settings UI after switching profile?

OSX support is missing - someone please contribute. I added warning about missing implementation [here](https://github.com/novist/qTox/compare/tux3:master...novist:autorun?expand=1#diff-a095c041ed4fe02895e04ee7e2ddff6fR473). When it is implemented entire #ifdef / #endif block should be removed. With these changes behaviour on OSX does not change.

I hijacked "Start in tray" option because in code its called like get**Autostart**InTray(). This implied intention to use that option for autorun functionality in the future and so i did.

Windows implementation used std::wstring / QString::toStdWString(). While most likely it is not a problem im mentioning it here in case it is required to work with Qt compiled without stl support. Non-stl solution would be rather ugly so i opted in for tiny bit stl.

### DIFF
--- a/src/platform/autorun_win.cpp
+++ b/src/platform/autorun_win.cpp
@@ -1,0 +1,82 @@
+/*
+    Copyright (C) 2014 by Project Tox <https://tox.im>
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    This program is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+
+    See the COPYING file for more details.
+*/
+
+#include <QApplication>
+#ifdef Q_OS_WIN32
+#include "src/platform/autorun.h"
+#include "src/misc/settings.h"
+#include <Windows.h>
+#include <string>
+
+namespace Platform
+{
+    inline std::wstring currentCommandLine()
+    {
+        return ("\"" + QApplication::applicationFilePath().replace('/', '\\') + "\" -P \"" +
+                Settings::getInstance().getCurrentProfile() + "\"").toStdWString();
+    }
+
+    inline std::wstring currentRegistryKeyName()
+    {
+        return (QString("qTox - ") + Settings::getInstance().getCurrentProfile()).toStdWString();
+    }
+}
+
+bool Platform::setAutorun(bool on)
+{
+    HKEY key = 0;
+    if (RegOpenKeyEx(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run",
+                    0, KEY_ALL_ACCESS, &key) != ERROR_SUCCESS)
+        return false;
+
+    bool result = false;
+    std::wstring keyName = currentRegistryKeyName();
+
+    if (on)
+    {
+        std::wstring path = currentCommandLine();
+        result = RegSetValueEx(key, keyName.c_str(), 0, REG_SZ, (PBYTE)path.c_str(),
+                               path.length() * sizeof(wchar_t)) == ERROR_SUCCESS;
+    }
+    else
+        result = RegDeleteValue(key, keyName.c_str()) == ERROR_SUCCESS;
+
+    RegCloseKey(key);
+    return result;
+}
+
+bool Platform::getAutorun()
+{
+    HKEY key = 0;
+    if (RegOpenKeyEx(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run",
+                    0, KEY_ALL_ACCESS, &key) != ERROR_SUCCESS)
+        return false;
+
+    std::wstring keyName = currentRegistryKeyName();;
+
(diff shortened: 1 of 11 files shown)


src/platform/autorun_win.cpp
```diff
@@ -0,0 +1,82 @@
...
+    inline std::wstring currentRegistryKeyName()
+    {
+        return (QString("qTox - ") + Settings::getInstance().getCurrentProfile()).toStdWString();
+    }
+}
+
+bool Platform::setAutorun(bool on)
+{
+    HKEY key = 0;
+    if(RegOpenKeyEx(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Run",
```
zetok: According to qTox style, there should be space between `if` and `()`; `if()` → `if ()`
Could you please change that and other occurrences in code?

dubslow:
I should think that starting qTox in tray and auto starting on boot are unrelated options, regardless of what the original author might have thought.

ghost:
I would not agree. Think about it. What is desired when application runs on startup? That application runs in the background until we need it. What do we want when we start application manually? We want to interact with application immediately. Start in tray option would become obsolete because usually noone runs an application manually only to do nothing with it. I personally use "Start in Tray" option only with autorun feature (currently added manually to OS startup programs). Other tox instances have that option unchecked because i run them at will when i need them.

Zer0-One:
"start in tray" and "run on boot" are unrelated options, and should have separate checkboxes.

zetok:
@novist I have to disagree. I would prefer starting qTox in tray **not** when system starts if I knew that qTox will actually start. Currently I don't have this option on, because I don't feel like qTox is stable enough to be able to start properly without any sort of problem, and I need a visual confirmation that it actually has started.

Also, based on what has been said, and what little I could understand from code, you don't seem to consider scenario where user sets their os/de to start qTox with system, rather than checking box in qTox (no, _requiring_ user to do something through qTox when it can be done in different way is not a good thing)

ProMcTagonist:
What Zetok said.

ghost:
Well if you guys insist i can make another option but i really feel it would be a bloat ;)

agilob:
I like when application starts on boot in tray and this feature should be supported in IM. I don't think it would be a bloat.

ProMcTagonist:
As long as start in tray and start at boot are different checkboxes.

dubslow:
Still need OS X implementation :P

stqism:
@dubslow That's fairly easy to implement, going to need to install Qbloat first though.

stqism:
@novist Fix your merge conflict, I cba to write both a launchagent and figure out how a bunch of Qt stuff works.

ghost:
@stqism any update on this?
@dubslow i still think better having 66% implemented instead of 0%. Maybe this could be merged while osx support could be contributed via new PR? At the moment on mac checkbox is disabled so its graceful-enough failure. Or it can be hidden if you prefer.

ProMcTagonist:
Start in Tray and Autostart are not the same.

ghost:
That setting was split long ago, just like you wanted ;)

ProMcTagonist:
<3

stqism:
@novist Totally forgot

stqism:
Yeah, I remember, I put that on hold because your pr didn't build with core and you didn't mention that it got fixed.

ghost:
Dont new commits generate notification for people watching PR? I thought everyone gets notified auto. Hmm..

stqism:
Nope, I just watch the repo and don't remember getting notified.

zetok:
@novist nope, pushing new commits to PR doesn't generate notification

ghost:
@stqism did you start any work based on this PR? Asking because i would fix merge conflicts rebasing changes on to upstream for nicer history. Otherwise if you are already working on stuff i would do merge commit.

dubslow:
I am generally not a fan of rebasing

ghost:
Clean > spaghetti history. One just have to know when to rebase and when to merge ;)

ghost:
Solved conflicts. Since OSX support does not seem to be coming any time soon i suggest this to be merged. There simply is no reason to have windows/linux users waiting for OSX support. I know for a fact that it is major deficiency for non-expert windows users. Not everyone knows of startup folder in start menu. Worse yet - it is kind of gone from windows 8. I had to dig that up deep in %APPDATA% and toss qTox shortcut in there just so my family member would be online when he is online instead of being online whenever he remembers to start qTox. So if all good merge pl0x.

Also maybe we want this option to be enabled first time qTox is installed and run? I think its usually what majority would want.

stqism:
Feel free, I need to finish my updater work, hook in to the notify api, and create a launchd service before it'll ever be start-able on boot. 

Thing to note though, we need to make it a part of the UI with some sort of getting started somehow. With Linux we can't just shove a check box in to the deb install.

ghost:
If you mean autorun by default after install - well i was thinking it being automatic. Check if its first launch, if it is set autorun=true. As for more controlled approach - well windows installer could have autorun checkbox but i dont think it it would work. Problem i see here is using profile name in launch command line. Its not exactly known at install time. Besides its problematic to know if it is first install or not. If app is being updated and autostart was disabled by user we do not want to override this behaviour.

PR #949, opened 2014-12-18
Fixed linked libs for win32

### LINKED ISSUE #843: Windows compilation instructions incomplete
Instructions for windows compilation end (linux instructions begin) immediately after explaining dependency compilation without explaining how to actually compile qTox.

Furthermore, there may be more dependencies or a mistake in the instructions for the dependencies. After running qmake and make with qtox.pro, the project failed to find the following libraries while linking: jpeg, tiff, png, jasper, IlmImf, Half.

qtox.pro file also needs to be modified to update the opencv version number from 248 to 249: "-lopencv_core249 -lopencv_highgui249 -lopencv_imgproc249"

### LINKED ISSUE #948: win32 build config screwed?
I went through pains of building qTox on windows and i discovered that somehow it fails to build with current config.

I had to move **-liphlpapi** after -lws2_32 because of linker errors.
I had to move **-lsodium** after -ltoxdns because of linker errors.
qTox links to bunch of unnecessary libraries. I removed **-ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf** and finally it built correctly.

All these changes were done inside win32 { } case. Now i wonder how win32 builds are compiled by buildbot if default config is broken? I did this with mingw 4.9, nothing fancy.

So should i make PR with these fixes..?

ghost:
Closes #948
Closes #843

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -74,9 +74,9 @@ contains(JENKINS,YES) {
 # Rules for Windows, Mac OSX, and Linux
 win32 {
     RC_FILE = windows/qtox.rc
-    LIBS += -liphlpapi -L$$PWD/libs/lib -lsodium -ltoxav -ltoxcore -ltoxencryptsave -ltoxdns -lvpx -lpthread
-    LIBS += -L$$PWD/libs/lib -lopencv_core248 -lopencv_highgui248 -lopencv_imgproc248 -lOpenAL32 -lopus
-    LIBS += -lopengl32 -lole32 -loleaut32 -luuid -lvfw32 -ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf -lws2_32 -lz
+    LIBS += -L$$PWD/libs/lib -ltoxav -ltoxcore -ltoxencryptsave -ltoxdns -lsodium -lvpx -lpthread
+    LIBS += -L$$PWD/libs/lib -lopencv_core249 -lopencv_highgui249 -lopencv_imgproc249 -lOpenAL32 -lopus
+    LIBS += -lopengl32 -lole32 -loleaut32 -luuid -lvfw32 -lws2_32 -liphlpapi -lz
 
     contains(DEFINES, QTOX_FILTER_AUDIO) {
         contains(STATICPKG, YES) {
--- a/windows/bootstrap.sh
+++ b/windows/bootstrap.sh
@@ -1,0 +1,109 @@
+#!/bin/sh
+QTOX_DIR=`pwd`/..
+
+if [ -d $QTOX_DIR/libs ]; then
+    echo "Remove ./libs and redownload/recompile dependencies?"
+    read -p "m/a/N (missing/all/no): " yn
+    yn=$(echo $yn | tr "[:upper:]" "[:lower:]")
+    if [ "$yn" == "a" ]; then
+        rm -rf $QTOX_DIR/libs
+    elif [ "$yn" == "n" ]; then
+        exit -1
+    elif [ "$yn" != "m" ]; then
+        exit -1
+    fi
+fi
+
+mkdir -p $QTOX_DIR/libs
+cd $QTOX_DIR/libs
+
+
+## toxcore
+if [ ! -f "libtoxcore-win32-i686.zip" ]; then
+   wget --no-check-certificate http://jenkins.libtoxcore.so/job/libtoxcore-win32-i686/lastSuccessfulBuild/artifact/libtoxcore-win32-i686.zip
+   rm -rf include/tox
+fi
+
+if [ ! -d "include/tox" ]; then
+   $QTOX_DIR/tools/unzip -o libtoxcore-win32-i686.zip -d ./
+fi
+
+## filter_audio
+if [ ! -f "bin/libfilteraudio.dll" ]; then
+   $QTOX_DIR/install_libfilteraudio.sh $QTOX_DIR/libs/filter_audio $QTOX_DIR/libs
+fi
+
+
+## OpenAL
+if [ ! -f "openal-soft-1.16.0.tar.bz2" ]; then
+   wget http://kcat.strangesoft.net/openal-releases/openal-soft-1.16.0.tar.bz2
+   rm -rf openal-soft-1.16.0
+fi
+
+if [ ! -d "openal-soft-1.16.0" ]; then
+   tar -xvf openal-soft-1.16.0.tar.bz2
+   rm bin/OpenAL32.dll
+fi
+
+if [ ! -f "bin/OpenAL32.dll" ]; then
+   pushd openal-soft-1.16.0/build
+   cmake -G "MSYS Makefiles" -DQT_QMAKE_EXECUTABLE=NOTFOUND -DCMAKE_BUILD_TYPE=Release -DALSOFT_REQUIRE_DSOUND=NO -DCMAKE_INSTALL_PREFIX=$QTOX_DIR/libs ..
+   make
+   make install
+   popd
+fi
+
+## opencv
+if [ ! -f "opencv-2.4.9.tar.gz" ]; then
+   wget --no-check-certificate https://github.com/Itseez/opencv/archive/2.4.9.tar.gz -O opencv-2.4.9.tar.gz
+   rm -rf opencv-2.4.9
+fi
+
+if [ ! -d "opencv-2.4.9" ]; then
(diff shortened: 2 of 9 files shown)


tux3:
This PR and commit includes quite a lot of changes that are unrelated to the title and commit message. 
Please try to avoid that, now I have to try to carefully make sense of every unexplained change you made.

ghost:
A lot? Lets see...
- Fixed warning on windows [here](https://github.com/novist/qTox/commit/4c4b0f023009b688848972a07f79717712426b13#diff-dc7eafdfcdfc16aed754a99d1d3d7e8fL24)
- opencv2 includes modified in order to make it actually compile on windows because of how opencv installs itself. Even if stuff is disabled and not included in package main header still includes them. Its good mojo to include only whats needed anyway.
- install_libfilteraudio.sh fixed to work on windows. That should be covered by main msg i think.
- Removed unnecessarily linked libs because there are no such libs when building on win32. Those deps are statically linked into other dependencies and thats why it also failed to build on win32.

Yeah i indeed sneaked in that win32 warning fix. If it bad to discard very unimportant stuff as unimportant then guilty ;)

dubslow:
I don't like those modifications to install_libfilteraudio.sh; not providing an install location, I.e. using the default, shouldn't change the file extension... id rather see a dedicated option/argument to change the extension, or maybe modify windows/bootstrap.sh to rename the output

qtox.pro
```diff
@@ -74,9 +74,9 @@ contains(JENKINS,YES) {
 # Rules for Windows, Mac OSX, and Linux
 win32 {
     RC_FILE = windows/qtox.rc
-    LIBS += -liphlpapi -L$$PWD/libs/lib -lsodium -ltoxav -ltoxcore -ltoxencryptsave -ltoxdns -lvpx -lpthread
-    LIBS += -L$$PWD/libs/lib -lopencv_core248 -lopencv_highgui248 -lopencv_imgproc248 -lOpenAL32 -lopus
-    LIBS += -lopengl32 -lole32 -loleaut32 -luuid -lvfw32 -ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf -lws2_32 -lz
+    LIBS += -L$$PWD/libs/lib -ltoxav -ltoxcore -ltoxencryptsave -ltoxdns -lsodium -lvpx -lpthread
+    LIBS += -L$$PWD/libs/lib -lopencv_core249 -lopencv_highgui249 -lopencv_imgproc249 -lOpenAL32 -lopus
```
tux3, line 78: This line will break the automated windows builds, that's not much of a problem, but it needs to be addressed before merging.

tux3:
The fix to src/platform/timer_win.cpp should probably go in a separate commit, I'd gladly merge that asap.

Same for the #include changes in src/video/camera*.h, they have nothing to do with linked libs, but it's an improvement :)

Thanks.

qtox.pro
ghost, line 78: How about passing a variable QTOX_CROSSCOMPILING to qmake on auto-build machine? if that is present old set of libs could be used otherwise set in my commit. Would that work?

EDIT:
Could you elaborate on how it breaks? I wonder if its only missing -ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf that i removed or is it lib order too? Because as you can see i moved around some libs. I think it would probably be best if we add additional libs that are used on buildserver instead of having all 3 lines copied twice and modified. Basically what im asking if this would work?

```
LIBS += -L$$PWD/libs/lib -ltoxav -ltoxcore -ltoxencryptsave -ltoxdns -lsodium -lvpx -lpthread
LIBS += -L$$PWD/libs/lib -lopencv_core248 -lopencv_highgui248 -lopencv_imgproc248 -lOpenAL32 -lopus
LIBS += -lopengl32 -lole32 -loleaut32 -luuid -lvfw32 -ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf -lws2_32 -liphlpapi -lz
```

Also buildserver version should probably keep opencv 248 version libs? Or can they be updated?

ghost:
@tux3 okie ill split them up ;)

@dubslow why do you prefer having install_libfilteraudio.sh broken on windows while it could be working?

EDIT: I see your point @dubslow. Will fix it ^_^

qtox.pro
tux3, line 78: It's not a problem with the cross-compiler, just that many Linux distributions are still packaging OpenCV 2.4.8. An option named "QTOX_CROSSCOMPILING" would be misleading. "USE_OPENCV_248" maybe, but I don't think it's needed in the first place.
It just means that I'll have to build OpenCV 2.4.9 DLLs for the Jenkins server, and that's not a problem at all.

ghost, line 78: Okay, please re-check my edited post (timing heh). I also want to know if linked libs i removed are needed at all. I think checking dependencies i did not see them loaded at all. So its unnecessary bloat if they arent used.

ghost:
@tux3 if there is any special handling for buildserver needed let me know what it is. Though i guess it would be perfect if buildserver could be accommodated to this build config so we dont have to use different one.

qtox.pro
tux3, line 78: I'm pretty sure that " -ljpeg -ltiff -lpng -ljasper -lIlmImf -lHalf" are not actually needed, they are historical remains from when I tried to build OpenCV statically. And I should be able to update the build server to OpenCV 249

I'll make some tests and see what I can remove and what I can update.

tux3:
@novist I'm currently building OpenCV 2.4.9 on the build server, as soon as that's done I'll make some builds and test them.

tux3:
Alright, the 32bit OpenCV 2.4.9 seems to build fine, but the -ws2_32 in the LIBS needs to come back or the build fails.

ghost:
Actually its [there](https://github.com/novist/qTox/commit/46cc81b4825b2d5b66cb0a0e76f8ec24d9ee7fde#diff-6a074c693ab3fcda2a48f60a4ef75bd5R79). Could you tell me where it should go to make build happy? I would test it on win32 too.

tux3:
@novist Oh my bad, I had only seen the first two added lines, not the last unchanged one...
Alright then, the build server should be ready, this part is ok.

@dubslow  are you happy with the changes to the bootstrap scripts ? If yes, then let's merge it now, otherwise I can cherry-pick the 3 others now and keep the PR open for the bootstrap changes.

windows/bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+    fi
+fi
+
+mkdir -p $QTOX_DIR/libs
+cd $QTOX_DIR/libs
+
+
+## toxcore
+if [ ! -f "libtoxcore-win32-i686.zip" ]; then
+   wget --no-check-certificate http://jenkins.libtoxcore.so/job/libtoxcore-win32-i686/lastSuccessfulBuild/artifact/libtoxcore-win32-i686.zip
```
dubslow, line 23: Not checking certificates is something that should be done with care, https is still useful, despite the flaws in the certificate system. (I assume it's related to running a Unix utility on Windows that SSL/TLS becomes Difficult™?)

ghost, line 23: Indeed but i added this so getting stuff actually works. Weird thing is that browser seems to recognize certificate just fine. In case of wget - should it not use windows certificate store? This is error that it spews anyway:

```
C:\Users\Administrator\Desktop\qTox>wget https://jenkins.libtoxcore.so/job/libtoxcore-win32-i686/lastSuccessfulBuild/art
ifact/libtoxcore-win32-i686.zip
--2015-01-20 10:40:27--  https://jenkins.libtoxcore.so/job/libtoxcore-win32-i686/lastSuccessfulBuild/artifact/libtoxcore
-win32-i686.zip
Resolving jenkins.libtoxcore.so... 104.219.184.40
Connecting to jenkins.libtoxcore.so|104.219.184.40|:443... connected.
ERROR: cannot verify jenkins.libtoxcore.so's certificate, issued by `/C=IL/O=StartCom Ltd./OU=Secure Digital Certificate
Signing/CN=StartCom Class 1 Primary Intermediate Server CA':
Unable to locally verify the issuer's authority.
To connect to jenkins.libtoxcore.so insecurely, use `--no-check-certificate'.
```

Probably most portable thing to do would be bundling root certificates in repo and using _--ca-directory_ flag passing the path. But its rather ugly. If you have better ideas im all ears.

dubslow, line 23: That's basically what I expected. For now, I guess we'll just "oh well" this.

PR #956, opened 2014-12-22
Update INSTALL.md

zetok:


dubslow:
Err... ./bootstrap.sh runs ./install_filter_audio?

zetok:
It does?
huh, I somehow managed to miss that

Issue #960, opened 2014-12-24
Video preview is not affected when settings are changed by mouse scroll


kushagrasurana:
Hi
The video preview present in video settings does not update when the slider is moved using mouse scroll.
It changes when slider is moved using mouse press.

kushagrasurana:
writing valueChanged(int) slot for all the sliders solved for me. ( copied from slider moved slot )
eg. for Hue Slider, i implemented valueChanged slot -:

void AVForm::on_HueSlider_valueChanged(int value)
{
    Camera::getInstance()->setProp(Camera::HUE, value / 100.0);
}

dubslow:
The sliders are stubs... I don't know if that's how they were intended to be implemented - were it so easy, why wasn't it done in the first place?

At any rate, feel free to submit a PR with the implemented slots :)

zetok:
since #962 was merged, close?

PR #962, opened 2014-12-25
issue 960

kushagrasurana:
Video setting sliders works on mouse scroll.

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -201,3 +201,23 @@ void AVForm::onFilterAudioToggled(bool filterAudio)
 {
     Settings::getInstance().setFilterAudio(filterAudio);
 }
+
+void AVForm::on_HueSlider_valueChanged(int value)
+{
+    Camera::getInstance()->setProp(Camera::HUE, value / 100.0);
+}
+
+void AVForm::on_BrightnessSlider_valueChanged(int value)
+{
+    Camera::getInstance()->setProp(Camera::BRIGHTNESS, value / 100.0);
+}
+
+void AVForm::on_SaturationSlider_valueChanged(int value)
+{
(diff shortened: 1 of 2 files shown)


Brottweiler:
I'd imagine this would be intended.

kushagrasurana:
The slider works now on mouse scroll, after my fix.

Brottweiler:
They currently work for me with mouse scroll without your fix.

zetok:
@Brottweiler so you don't experience any symptoms of #960 ?

EDIT: I for one experience those issues…

zetok:
And this fix works for me.

Brottweiler:
@zetok Oh, maybe it's a different issue. I can manipulate the slider using my mouse wheel, but I don't know what happens when i save the settings.

dubslow:
@krepa098 

If I don't see any response to this by the time I get proper access to my own computers again (i.e. not just ssh, probably no more than two weeks ish?), I'll see about testing and merging this and 920 @lumirayz

krepa098:
Seems to work fine.
@kushagrasurana please remove the _sliderMoved slots as they are now redundant.

agilob:
+1, I have the same problem as described by @kushagrasurana 
scrolling on sliders doesn't change camera settings until clicked on sliders.

dubslow:
@krepa098 @kushagrasurana This is probably something really stupid, but how does this actually help? As far as I can tell, the new slots aren't connected to any signals (just like the old, non-functioning slots)...?

krepa098:
They are connected (http://qt-project.org/doc/qt-4.8/designer-using-a-ui-file.html#automatic-connections). And the old slots had been connected too.

Anyway, changing theses properties produces garbage results on Windows and I think we should disable them (on Windows, they work fine on linux and OSX(?)).

tux3:
Rebased and merged in 64bd3b7b47661bbdbbacca11a3aa190f72e41954

PR #965, opened 2014-12-28
Proxy type selection

### LINKED ISSUE #963: qtox doesn't compile anymore due to refactoring of proxy code in toxcore
qtox won't compile with the git-version of toxcore, due to refactoring of their proxy code, e.g. the variable proxy_enabled does not exist anymore, see https://github.com/irungentoo/toxcore/commit/e9bf38499e42eec2b01ef5024f0b7092539bc11c .

apprb:
also fixes #963

### DIFF
--- a/src/widget/form/settings/generalsettings.ui
+++ b/src/widget/form/settings/generalsettings.ui
@@ -513,7 +513,7 @@
               </property>
               <item>
                <property name="text">
-                <string>Direct Connect</string>
+                <string>None</string>
                </property>
               </item>
               <item>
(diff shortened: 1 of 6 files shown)


src/widget/form/settings/generalsettings.ui
```diff
@@ -492,11 +495,40 @@
...
+             <widget class="QComboBox" name="proxyType">
+              <property name="sizePolicy">
+               <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+                <horstretch>0</horstretch>
+                <verstretch>0</verstretch>
+               </sizepolicy>
+              </property>
+              <item>
+               <property name="text">
+                <string>Direct Connect</string>
```
zetok: Wouldn't it be better to just call it `None` instead of `Direct Connect`, which is a bit strange?

Issue #940, opened 2014-12-16
[Groupchat] Message failed to send


TheLastProject:
Original message:

```
[blocks B      ] <dev-qt/qtmultimedia-4.8.6:4 ("<dev-qt/qtmultimedia-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtscript-4.8.6:4 ("<dev-qt/qtscript-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtwebkit-4.8.6:4 ("<dev-qt/qtwebkit-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtxmlpatterns-4.8.6:4 ("<dev-qt/qtxmlpatterns-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtdbus-4.8.6:4 ("<dev-qt/qtdbus-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qttest-4.8.6:4 ("<dev-qt/qttest-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/designer-4.8.6:4 ("<dev-qt/designer-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtdeclarative-4.8.6:4 ("<dev-qt/qtdeclarative-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtsvg-4.8.6:4 ("<dev-qt/qtsvg-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qt3support-4.8.6:4 ("<dev-qt/qt3support-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtopengl-4.8.6:4 ("<dev-qt/qtopengl-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[blocks B      ] <dev-qt/qtcore-4.8.6:4 ("<dev-qt/qtcore-4.8.6:4" is blocking dev-qt/qtchooser-0_p20140613)
[...]

PR #966, opened 2014-12-28
messages splitting fix

apprb:
As was mentioned in #940 messages in group chats should be split by `MAX_GROUP_MESSAGE_DATA_LEN` and `MAX_GROUP_MESSAGE_DATA_LEN` < `TOX_MAX_MESSAGE_LENGTH`.
Too hard to calculate exact length, so I just hardcoded 1024.


dubslow:
I would advise you to figure out what the actual number is and submit a PR to core adding that to tox.h

Issue #740, opened 2014-11-13
Do not move offline contacts with unread messages to the bottom of the list


TheLastProject:
If a contact sends you a message and then goes offline before you read it, they are sorted to the bottom of the contact list, making it hard to see who sent a message.

Personal suggestion: add contacts that just went offline on the top of the "offline contacts" list, instead of bottom. It would be even better if they would not be pushed down by other contacts going offline as long as there are still unread messaging from that user.

TheLastProject:
I'm not sure if I'd truly consider this fixed. Sure, it's an improvement, but if someone sends a message and then goes offline, they're sorted to the top of offline users, still forcing you to scroll sometimes.

EDIT: I just realized I got exactly what I asked for. Maybe I should've asked a smarter question ;)

PR #972, opened 2015-01-02
Fixes issue #740, now contacts with new events are not moved to the bottom when going offline

Nokta-strigo:
It works, but perhaps that's not the best way to do it. May be it's better to create two layouts for status==offline: one for contacts having new events and the other - for contacts without new events?
And why layouts was of type QLayout, not QVBoxLayout?


tux3:
Merged after a rebase in 96fd00d8b6f924d8ae99486f61431b6202955543

PR #973, opened 2015-01-02
Alphabetize groupchat user list

### LINKED ISSUE #701: Sort names in groupchat list in alphabetical order
So that finding whether a particular user is/isn't in the chat doesn't require scanning the whole list.

Zer0-One:
fixes #701 , and cleans up the tab-completion code a little. Because I sort the peer-list beforehand, there's no more need to store the completion-list as a map, since we iterate over the sorted peer-list from beginning to end. This, in turn, obviates the need for "SortableString".

@dubslow are you still wondering why QObject is necessary here? :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -145,7 +145,9 @@ GroupWidget *Group::getGroupWidget()
 
 QStringList Group::getPeerList() const
 {
-    return peers.values();
+    QStringList peerNames(peers.values());
+    peerNames.sort(Qt::CaseInsensitive);
+    return peerNames;
 }
 
 void Group::setEventFlag(int f)


dubslow:
Yes. Shouldn't QString include QObject, since the former are the latter?

What about when someone gets around to implementing sorting by last activity, as I mention in comments?

Why is the comment block deleted? Perhaps it is opaque, but better improve it than delete it...

Zer0-One:
If you'd like, I can undo everything except for the changes to getPeerList() (imo, we should leave out the clutter and just file a feature request). I haven't looked into it yet, but if there's a way to find out whose activity was more recent, I can just go ahead and implement sorting by last activity.

dubslow:
I think that's best for now. Implementing sort by activity separately would also be wonderful.

Zer0-One:
done

Zer0-One:
btw, for future reference, QString actually isn't a QObject

PR #979, opened 2015-01-04
Added typing notification support and enabled by default

ghost:
Adds a text label between text log and input box in normal ChatForm. Setting enabled by default since it helps people to be polite and not to interrupt each other. Label font size is small. Overall its much like in pidgin IM. Typing flag is enabled when input box contains text. It is turned off when input box is cleared.

### DIFF
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -189,7 +189,7 @@ void Settings::load()
     s.endGroup();
 
     s.beginGroup("Privacy");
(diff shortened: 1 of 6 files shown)


tux3:
Hm, it works but it's not very clear what that "Typing notification" checkbox does.

If you uncheck it, you'll still see typing notifications, it's just that others won't see yours.
Maybe we could change it to "Send typing notifications" ?

Merging regardless.

PR #981, opened 2015-01-04
Some wording improvements

zetok:
not tested. No idea whether tooltip thing works, or even if I did it right..


dubslow:
The tool tip is nearly correct, it uses the contents of the string tag, rather than an attribute like extracomment.

PR #985, opened 2015-01-06
Encryption

dubslow:
Unfortunately I still don't know what's causing main() to exit on @zetok's computer, but this is otherwise feature complete, and no other bugs are known.

I however would like a lot more people to attempt to use this additional code, so I'm making the PR for visibility. In particular, @TheLastProject, can you confirm whether or not your Gentoo installation crashes when running this branch?

tux3 I merged it myself since you had at least one issue with the rebase that you fixed, and frankly the merge was quite simple (approximately 30 lines of conflicting code over the 4 conflicting files), rather less work I think than double checking your rebase (and I've become more of a fan of keeping history intact recently).

At some point I want to tweak the password meter to put more emphasis on longer passwords rather than special characters, caps, numbers etc, but it doesn't affect the functionality of the PR.

Edit: Actually, zetok reported the same issue on _master_ currently as well, though with the checkboxes disabled it doesn't really matter much. So it's not an issue with any additional code from this PR, and it may well be worth merging right now.


zetok:
Didn't seem to segfault after little of testing on newest master and with Qt 5.4.0.

On the other hand, there is a problem with this PR deleting whole history if one ~~tries to~~ looks into ways to unencrypt it -.-

Edit: 
Steps to reproduce:
1. put something in history
2. check in `encrypt history`, use main password
3. click to uncheck
4. click no
5. check again `encrypt history`
6. aaannnnd it's gone.  ._.

dubslow:
@zetok did you actually read what clicking no means? Hint: that's the intended behavior.

TheLastProject:
@dubslow I must be honest, I still haven't tried the encryption

zetok:
@dubslow oh, you're right. Sorry, I was a bit tired when I was "testing" that.

Anyway, here's a real problem:
orly enabled
↑ showed when I tried to close qTox after clicking 'yes'

Same message appeared after launching qTox again, and to my surprise `Encrypt Tox data file` was unchecked.

qTox currently silently unencrypts users' data file..

Edit: and why would you **delete or unencrypt** logs? I.e., have you considered an idea that someone could have clicked on checkbox accidentally and doesn't actually want their history to be removed/unencrypted?

Edit2: No, currently `No` doesn't work as intended behaviour. `No` means "no", not some `delete all my data`.

dubslow:
@zetok that's entirely with the data file encryption, correct? Not with the chat logs? or both?

Edit: Wait that's dumb, has to be history to ask the question.

So if you hit yes, then next thing you do when trying to close qTox is seeing that message? I guess I should update that to say which password it's talking about... but I guess it means history in this case.

And then the data encryption was unchecked after unchecking history encryption and restarting?

1) So you're suggesting a third cancel button and 2) rewording No to something more dangerousy?

zetok:
> And then the data encryption was unchecked after unchecking history encryption and restarting?

Yep

> I guess I should update that to say which password it's talking about... but I guess it means history in this case.

There's only 1 password used after all, same for history and data file…

> 1) So you're suggesting a third cancel button and 2) rewording No to something more dangerousy?

`Yes` → `Yes, remove encryption`
`No` → `No, remove data` ← This option should have another confirmation window with yes/no warning that _all_ history will be permanently deleted.
`Cancel` ← nothing is being changed

EDIT: or maybe without `Yes` and `No`, just `Remove encryption`, `Remove history` and `Cancel`?

ghost:
I still think the right way to do history encryption is by making [qt plugin](https://github.com/sijk/qt5-sqlcipher) for [sqlcipher](https://github.com/sqlcipher/sqlcipher). Encrypting only some rows is terrible hack. Besides why have only some stuff encrypted when entire sqlite database file could be encrypted. And bonus would be that there is probably less than 20 lines of code required for [encrypting/decrypting](https://www.zetetic.net/sqlcipher/sqlcipher-api/#sqlcipher_export) sqlite database password. Only one side-effect - sqlcipher uses openssl or libtomcrypt. Thats really the only downside.

apprb:
Probably you right, some time ago I tried https://github.com/devbean/QtCipherSqlitePlugin and it caused very minor changing of the current code and it can be easily compiled along with main project.
But on another hand if toxcore offers encryption capability why we shouldn't use it? 

PS:

```
$ ldd qtox | grep ssl
libssl.so.1.0.0 => /usr/lib/libssl.so.1.0.0 (0x00007fad99e5b000)
```

(current master branch)

PS2:
If it is required for somebody I can prepare working example with https://github.com/devbean/QtCipherSqlitePlugin.

dubslow:
I would prefer not to use any other encryption than that offered by core,

ghost:
Indeed. sqlcipher has sort of crypto toolkit api. [openssl](https://github.com/sqlcipher/sqlcipher/blob/909658402abf454e8c34ac4a6b7269fdce198ebd/src/crypto_openssl.c) and [libtomcrypt](https://github.com/sqlcipher/sqlcipher/blob/master/src/crypto_libtomcrypt.c) plugins here. But can core provide all these?

```
int sqlcipher_openssl_setup(sqlcipher_provider *p) {
    p->activate = sqlcipher_openssl_activate;
    p->deactivate = sqlcipher_openssl_deactivate;
    p->get_provider_name = sqlcipher_openssl_get_provider_name;
    p->random = sqlcipher_openssl_random;
    p->hmac = sqlcipher_openssl_hmac;
    p->kdf = sqlcipher_openssl_kdf;
    p->cipher = sqlcipher_openssl_cipher;
    p->set_cipher = sqlcipher_openssl_set_cipher;
    p->get_cipher = sqlcipher_openssl_get_cipher;
    p->get_key_sz = sqlcipher_openssl_get_key_sz;
    p->get_iv_sz = sqlcipher_openssl_get_iv_sz;
    p->get_block_sz = sqlcipher_openssl_get_block_sz;
    p->get_hmac_sz = sqlcipher_openssl_get_hmac_sz;
    p->ctx_copy = sqlcipher_openssl_ctx_copy;
    p->ctx_cmp = sqlcipher_openssl_ctx_cmp;
    p->ctx_init = sqlcipher_openssl_ctx_init;
    p->ctx_free = sqlcipher_openssl_ctx_free;
    p->add_random = sqlcipher_openssl_add_random;
    return SQLITE_OK;
}
```

Some might be optional like iv and maybe kdf function, but not hmac. Though for hmac maybe [this](http://qt-project.org/wiki/HMAC-SHA1) can be used or something. Idk if its valid hmac code though. Or maybe it could be implemented with libsodium?

dubslow:
Core just wraps NaCl/libsodium.

I'd still rather merge this once I fix the bug zetok reported though.

dubslow:
@zetok ^

Any other issues?

zetok:
@dubslow there's still not way to cancel removing encryption from logs.
Also, those buttons should have confirmation dialog, since they mess with user data
delete/unencrypt

In case where one sets encryption where there is no history (setting for it is checked, just there is no history), there's shown warning box, and apparently history is being disabled after restarting qTox. Why?
why disable history

# 

Also, if I may say so, depending on openssl to provide encryption in those times isn't a really good thing; after all there is a reason why toxcore doesn't use it…

zetok:
**Segfault**: https://gist.github.com/zetok/b5b84b165ed43bda91e6

Steps to reproduce:
1. $ qtox -P $encrypted_profile
2. press enter when asked for password
3. press `Cancel` when asked to select profile
4. segfault

# 

**History loss** ← needs history to be encrypted with profile password
1. $ qtox -P $encrypted_profile
2. uncheck `Encrypt Tox data file`
3. restart qTox
4. press enter when asked for password
5. enable history
6. you have no history

# 

**Confirm removing encryption** ← if user sets encryption, they should be shielded from accidental removal of encryption. Thus unchecking `Encrypt Tox data file` needs a confirmation dialog

# 

I've "tested" encryption, but I have a feeling that testing wasn't long or deep enough.
When issues above will get fixed, encryption will probably be ~ready for merging, provided that there won't be anything else lurking in there (I think that looking at how db and data encryption are working together is worthy looking at, considering that there were already several issues with it)…

dubslow:
All three of those things are fixed.

Since 826d60c8767ac1c there is literally no way for interaction troubles between the two passwords, that commit's purpose was to make them utterly independent since there were so many issues with the previous way of linking them. (The second issue in your most recent post didn't in fact require "use other password".)

srkunze:
I built it myself to see if it works the way, I like it.

1) your general qtox make-it-yourself docu seems a bit outdated

2) the encryption feature works for me +1 big thanks to dubslow! :)

3) Why on earth do I need to have two passwords one for the logs and one for the tox data?

4) Encrypting and so on requires a restart of the client as my complete history was empty. But after encrypting and restarting, everything seems fine. Should be noted somewhere in order to not scare people away. :)

dubslow:
It's _very nearly_ done, but there is still one issue I can't really troubleshoot for a couple of days. It should also be easy to use the same password :) thanks for the other feedback too.

srkunze:
Guess, what I do. ;) But honestly, could we not have just on password for both and be done with it?

Correct me if I am wrong, but managing two passwords (beside being user-unfriendly) would cause me head-aches during implementation such as:

1) what happens if one of the passwords the user enters is wrong? Will he be able to chat without unlocking the history or the other way round? Neither would make sense to me but people might want to do that.

2) people will confuse the one password with the other for sure. So, why the hassle with dealing two passwords (also code-wise)? It seems to violate KISS.

Don't get me wrong, I want to have that branch upstream finally. But the dual-password is quite confusing (at least to me). Maybe, because it is a completely new concept to me but on the other had I question the necessity for it.

dubslow:
It's pretty damn trivial to use the same password for both. When setting the password for one you can always choose to use the other password (if it's set), and when the code needs a password it always tries the other password before asking, so it should generally be seamless to use one password for both.

Code wise there were some issues (see the commits), but after changing/simplifying how it's done I see no reason to remove working code. And particularly paranoid users (of which there are certainly some, and probably some with good reason) will appreciate the ability.

Hopefully I'll be able to fix the last remaining bug tomorrow, but no guarantees of course.

srkunze:
Maybe, there is a misunderstanding here:

> when the code needs a password it always tries the other password before asking

This in fact is not true. When I start the client, I need to type in the same password twice (one for account data, one for history).

dubslow:
Really? That's a bug. Can you explain what exactly you do to produce that?

srkunze:
I have one client, two profiles, hence four passwords where the passwords for the same profile remains the same (same for data and for history). So, I actually have only two passwords.

1) Open the client.
2) The client asks for the password for profile A.
3) I type it in and press Enter (errors or not is irrelevant here; after providing the correct password, 4) will occur)
4) The client asks for the password for the history of profile A. << that is the bug, I suppose

Same goes for profile B.

dubslow:
@srkunze I am unable to reproduce. When I click "use other password" and restart, it doesn't ask for the history password. When I use a different password, it rightfully asks for the different password. In either case the history correctly loads.

Note that simply typing the same password in doesn't suffice, because a new randomly generated salt is used each time, causing the key derived from the password to differ. The "Use other password" copies the key and salt, rather than deriving a new key with a different salt.

srkunze:
> Note that simply typing the same password in doesn't suffice, because a new randomly generated salt is used each time, causing the key derived from the password to differ.

Really? -.-

dubslow:
Yes really. That's very basic cryptographic best practice. Not doing so would be gross incompetence. That's why the `Use other password` button is provided.

srkunze:
I see. Now, it works the way, you intended it to work.

BUT:

I am sorry but that is a usability fail.

srkunze:
Something else:

When decrypting your profile data, you need to click the button "Set Password". I think that a button label like "Decrypt" or "Decrypt Profile" or "Unlock" is better to understand.

"Set Password" sounds more like changing the password. But you actually want to unlock your account.

dubslow:
@zetok @tux3 I fixed the last issue I know about, one final test please? @srkunze ?

srkunze:
@dubslow Will do in 5h.

srkunze:
Well, looks good.

Are there any tests to test that new functionality?

dubslow:
qTox doesn't have any automated testing....

srkunze:
Okay. :(
Let's merge it.

PR #1008, opened 2015-01-12
fix videosurface (missing buffer flip) and some glitches in av settings

derMart:
Hello,

this PR addresses three issues:
1. Opengl buffers are not flipped during paintGL. At least on Windows this prevents video from working. Enabling setAutoBufferSwap solves the problem.
2. Audio devices using special characters are not displayed correctly and cannot be opened.
According: http://openal.org/pipermail/openal/2014-August/000202.html at least on Windows the strings are UTF8 rather than any local encoding. I'm not sure about other platforms, so I just applied the change on windows.
3. Video is not correctly initialized in the av settings form (resolution not set). Furthermore the 0x0 resolution is removed and audio initialization code tries to avoid opening audio devices twice.

cheers
Martin

### DIFF
--- a/src/video/cameraworker.cpp
+++ b/src/video/cameraworker.cpp
@@ -112,7 +112,7 @@ void CameraWorker::_probeResolutions()
 
             //qDebug() << "PROBING:" << res << " got " << w << h;
 
-            if (!resolutions.contains(QSize(w,h)))
+            if (w>0 && h>0 && !resolutions.contains(QSize(w,h)))
                 resolutions.append(QSize(w,h));
         }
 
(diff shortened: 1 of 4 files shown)


dubslow:
Excellent stuff, neat and tidy separate commits to boot :)

derMart:
thank you very much :)

derMart:
Video in settings screen hangs after waking from tray icon. Seems the showEvent in AFForm was missing -> I was so free to add a commit therefore too ;)

dubslow:
@suhr At least the sound works now. I'll close the old related issues, and you can make a new one regarding the clipped name.

PR #1024, opened 2015-01-19
Context menu style fix for chat title

saneki:
This doesn't make the context menu "useful" (due to CroppingLabel hiding the QLineEdit when focus is lost), but makes it look better and consistent with the other context menus.


tux3:
I'm not sure what context menu you're talking about. This changes means that the black/white style now only applies to QLabels and QLineEdits in the headers above the main content instead of applying to the whole header. 
But this doesn't result in any visible change, at least for me.

Can you show two before/after screenshots of what this changes for you ?

PR #1030, opened 2015-01-21
Notifications + settings UI cleanup

ghost:
Name mentions in group chats now work only if name is mentioned as full word and not somewhere in middle of other word. This gets rid of false-positive mentions. 

Added option for group chats to always notify (like normal chats) (default off). Otherwise its hard to keep an eye on chat when noone is mentioning you. In the future this ideally should be per-chat setting.

"Focus window on new message" option was split into two: show window (default on) and focus window (default off). Most of the time on new message we want window to be visible highlighted in taskbar for example but we do not want window to steal focus on every new message.

General settings form received lots of love. Now it looks much more neat. Stuff was moved around and lined up so it both fits good within context of other settings and saves space.
settings

The only thing i am bit unsure of is portable setting - it seems out of place. However keeping it below translations creates lots of empty space on the right. I suggest moving portable setting to advanced tab altogether because not so many people are going to use it. Also use of this setting may cause some things happen that normal user would not expect and not wish to have. So maybe it would be right to classify it as advanced option?

While i was at it i cleaned up `onFriendMessageReceived`, `onGroupMessageReceived` and `newMessageAlert` to get rid of duplicate code which by the way was confusing to make sense of. Now it is more straightforward.

I tried to test all cases i could think of. Tested it on windows too. Additional testing would be lovely though.

### DIFF
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -136,7 +136,9 @@ void Settings::load()
         currentProfile = s.value("currentProfile", "").toString();
         autoAwayTime = s.value("autoAwayTime", 10).toInt();
         checkUpdates = s.value("checkUpdates", false).toBool();
+        showWindow = s.value("showWindow", true).toBool();
         showInFront = s.value("showInFront", false).toBool();
+        groupAlwaysNotify = s.value("groupAlwaysNotify", false).toBool();
         fauxOfflineMessaging = s.value("fauxOfflineMessaging", true).toBool();
         autoSaveEnabled = s.value("autoSaveEnabled", false).toBool();
         globalAutoAcceptDir = s.value("globalAutoAcceptDir",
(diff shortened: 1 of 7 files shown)


suhr:
- “Make Tox portable” looks related to translation, but it is not.
- Consider “Translation” → “Language”
- “Save files in” → “Save in”
- Move “Emoticon size” after “Smiley pack”
- Maybe increase space between “Autoaccept files” and “Save in”?

Zer0-One:
"On new message" looks a little weird. Perhaps group it a little more tightly with the checkboxes it belongs to? Maybe a horizontal line/bar underneath that first row?

ArchangeGabriel:
Yes, I think that “Make Tox portable” should be moved to advanced indeed.

PR #1032, opened 2015-01-21
Fixed translation errors.

Lednerb:
Added / Fixed Strings in the German translation.

### DIFF
--- a/translations/de.ts
+++ b/translations/de.ts
@@ -614,8 +614,8 @@ Alias:</translation>
     </message>
     <message>
         <location filename="../src/widget/form/settings/generalsettings.ui" line="196"/>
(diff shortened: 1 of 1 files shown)


Lednerb:
How can i translate the NO-Button?

no-button

Should be nearby the tox1-protocol prompt.

tux3:
You can't change the text on the button if that's what you're asking, not without some code changes anyway. What would you put instead ?

Lednerb:
Yes, this was what i wanted to ask.
I wanted to change "No" to the German  "Nein".

Edit: Can be merged, so at least the text is translated.

tux3:
Note: There is a way to translate message box buttons using Qt's official translation files as well as our own.

PR #1042, opened 2015-01-25
URLs and emoticons in nicknames/status mesages in chat/friend list

TheLastProject:
Windows Live Messager. Those were the days, right? So yes, this pull request adds support for URLs and smileys in nicknames/status messages in chats/friend list.

However, IT IS NOT COMPLETE. Please do NOT consider merging this yet. I'm making this pull request in the hope of getting some support to fix the last issues:
- [ ] Messages ending on a link do not get ellipsis

snapshot142
- [ ] Images get broken if there is not enough space for them

snapshot142
- [ ] Disabling/enabling emoticons doesn't instantly hide/show them in the friend list
- [ ] The code is ugly :(

Other discussion-worthy questions:
- Do we want to allow URLs in nicknames?
- This breaks editing group chat titles by clicking on them which are completely URL because the URL gets opened. Is this to be considered a regression? Should this be solved? If so, how? (Please note that you can still change the title by right-clicking it in the friend list and selecting "Set Title...")

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -858,6 +858,9 @@ QSplitter:handle{
                  <property name="text">
                   <string>Your name</string>
                  </property>
+                 <property name="openExternalLinks">
+                  <bool>false</bool>
+                 </property>
                 </widget>
                </item>
                <item>
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -940,6 +940,9 @@ QSplitter:handle{
                  <property name="text">
                   <string>Your status</string>
                  </property>
+                 <property name="highlightURLs">
+                  <bool>true</bool>
+                 </property>
                 </widget>
                </item>
               </layout>
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -858,9 +858,6 @@ QSplitter:handle{
                  <property name="text">
                   <string>Your name</string>
                  </property>
-                 <property name="openExternalLinks">
-                  <bool>false</bool>
-                 </property>
                 </widget>
                </item>
                <item>
@@ -942,9 +939,6 @@ QSplitter:handle{
                  </property>
                  <property name="text">
                   <string>Your status</string>
-                 </property>
-                 <property name="openExternalLinks">
-                  <bool>false</bool>
                  </property>
                 </widget>
                </item>
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -940,6 +940,9 @@ QSplitter:handle{
                  <property name="text">
                   <string>Your status</string>
                  </property>
+                 <property name="openExternalLinks">
+                  <bool>false</bool>
+                 </property>
                 </widget>
                </item>
               </layout>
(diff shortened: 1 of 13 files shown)


zetok:
> Do we want to allow URLs in nicknames?

No. Status is all users need. No need to overcomplicate it.

> Images get broken if there is not enough space for them

Images?

> This breaks editing group chat titles by clicking on them which are completely URL because the URL gets opened. Should this be solved?

Yes, it is a regression, and a serious one at that. Don't break user experience.

> If so, how?

Well, don't break it.
Solution would be to provide copying/opening/whatever by menu that would appear when r-clicked, while l-click would still lead to editing title.

# 

On a screenshot from groupchat I've seen that it also applies to status message of user, if that's still the case, please change it, since there's no need for it. Not only that, but moreover it complicates UX, since now user is forced to do some strange things to just edit status message, where before it was easy and simple.

TheLastProject:
@zetok Codewise, it is less complicated to also support nicknames (in fact, this is currently supported).

Images in the sense of emoticons.

Your solution to group chat titles is probably the easiest solution, as it'd only mean disabling setOpenExternalLink for group chat titles.

Regarding your own status message, yes, links and emoticons get highlighted and shown, but clicking your status message will never open a link, but always let you edit it.

zetok:
> Codewise, it is less complicated to also support nicknames (in fact, this is currently supported).

I may be ~easier, but it doesn't mean that it'll be better UX-wise. When I click on someone in list to start conversation, I don't carefully aim to click _just on avatar_ to start conversation, in case where someone thinks that setting link as nick and status is a good idea. With compact contact list that would leave only **tiny** amount of space to click on, i.e., avatar and status orb. Thus nick shouldn't be a clickable link.
There's also reason for not even "showing" that it's a link in any special way…

> Regarding your own status message, yes, links and emoticons get highlighted and shown, but clicking your status message will never open a link, but always let you edit it.

That's nice. Unfortunately there's still problem of links looking just simply bad in qTox's GUI outside of places specially designated for them. I for one would hate how link would be highlighted in my own status message, destroying that nice colour composition it currently has. Thus I think that not highlighting it as a link would be nice, when it can't even be opened by l-clicking.

tux3:
I think we could do that like most terminals/irc clients do: just underline links (maybe only when the mouse is over them, maybe always) and do nothing particular when you click on them. 
But when you CTRL+click on them they open in a browser, and if you right click the context menu lets you open them too.

zetok:
It just looks bad.
noew

On fist screenshot I can't even really see where that link points to… (Actually I can when I try, but I know that there are people with worse sight…)

Compared to
gud

ProMcTagonist:
Tux3's CTRL+click solution takes care of the most issues without breaking the default experience, both for group title editing, and clicking on friends to chat. It's also a convention from Microsoft Word.

And there's no reason to restrict where users can put links and where they can't. Are you going to block names from containing https://? It's arbitrary and confusing.

TheLastProject:
@zetok You're right, thanks

@ProMcTagonist It's not to restrict, names can contain links just fine, they just won't be clickable or highlighted like links.

dubslow:
It may be conducive to development to create a bunch of small tasks, and submit a separate PR for each.

TheLastProject:
Closing until I can figure out the last few issues and get this merge-ready. Help welcome.

PR #1048, opened 2015-01-26
Arrow up to paste previous message

agilob:
If message box is empty, it will paste previous message on arrow up
if message box is not empty, it will copy current(2) text and paste previous(1) message, to paste previous message(2) press arrow down, press arrow down twice to clear mesage box, only previous message(1) is available to paste now.


tux3:
It looks like you're clearing the text edit anytime someone presses down, that sounds wrong :/

Maybe only restore the current message if there is one (that implies that we just pressed up not long ago) and ignore the event otherwise.

Or another idea, maybe just do it "like bash does it". Have a history, any modification is kept, up and down circle through history, and you can't use up and down to navigate the text box at all. Maybe use shift+up and shift+down.

But clearing the text box everytime someone pressed down is not so good.

PR #1052, opened 2015-01-26
Window focusing fixes

ghost:
Currently if window is minimized it is still considered visible and is hidden. That requires user to click tray icon two times in order to bring up window. This patch implements window showing on tray icon click also when window is minimized therefore requiring always one tray icon click to bring window up if it is not visible and active window. This extends to window activation behaviour when application is executed again (via shortcut or pinned on taskbar).

In addition other two expected behaviours with tray icon implemented:
Double-clicking tray icon shows window because on windows double-clicking tray icons is quite common. Does not affect current default single-click toggling in any way.
Middle-clicking tray icon hides window. Browsers associated middle-click with (tab) closing action therefore using middle mouse for hiding is intuitive behaviour even if might not be that obvious at first.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -56,8 +56,7 @@ void toxActivateEventHandler(const QByteArray& data)
 {
     if (data != "$activate")
         return;
-    Widget::getInstance()->show();
-    Widget::getInstance()->activateWindow();
+    Widget::getInstance()->forceShow();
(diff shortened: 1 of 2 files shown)


dubslow:
Merged by cherry pick to avoid extra commits

dubslow:
Er, nvm, unfair, and I like history

PR #1061, opened 2015-01-28
fix greentext, take 2

### LINKED ISSUE #1059: Quoting is Broken
If > is the first character, everything on that line should be green.

I often have cause to quote someone or something who used characters that the latest regex arbitrarily deems unquotable.

suhr:
Fixes https://github.com/tux3/qTox/issues/1059.


suhr:
@dubslow r?

PR #1071, opened 2015-01-30
Updated German translation

Lednerb:
Removed vanished entries, updated all unfinished entries.

Found 1 translation bug:
Translation of the **Theme color:** - dropdown menue does not work correctly.
Colors are translated in German yet.


dubslow:
Oops, rebase please

tux3:
And this is why you're not supposed to rewrite history after pushing it!
This was a nightmare to rebase, so I cherry-picked and tried my best at fixing conflicts

Cherry-picked in f43f09e, merged in c1f41cd

dubslow:
I believe it's possible to rebase this in a way that squashes the correct part of the previous commit (the one that accidentally updated all the *.ts files), so that no conflict resolution is necessary, but I was lacking for time and I had also deleted the relevant commit. Whoops.

I still think that in the case it was a net gain. It was very recent history with little written past it, and the video-perf commit was basically unviewable with Javascript enabled (the GitHub highlighting script, specifically). Definitely not a tool to be used lightly though.

tux3:
> I believe it's possible to rebase this in a way that squashes the correct part of the previous commit (the one that accidentally updated all the *.ts files), so that no conflict resolution is necessary

It really isn't if someone already committed on top of the old history. The way git works, every commit links to its parent. If you kill its parent, then git has to go back to the most recent common ancestor, so all the commits until that common ancestor (those you squashed) with will be part of the merge.

> I still think that in the case it was a net gain. It was very recent history with little written past it, and the video-perf commit was basically unviewable with Javascript enabled

Yeah, I fucked up. But turns out people still had the time to commit on top of the old history, so it was a bit late to rewrite it.

Issue #1068, opened 2015-01-30
Enter Key Not Sending Message


sierramountain:
Why doesn't the enter key send when messaging, it only seems to return? I'm using the 32bit window version.  Does anyone know how to fix this issue, not sure if its my computer or the program. Thanks!

dubslow:
On GNU-Linux, my numpad enter key doesn't work, but the primary enter key on the keyboard does.

sierramountain:
Good point, however for me neither the numpad or the primary enter key can 'send', both return

srkunze:
Is the "primary enter key" not the RETURN key?

I remember them to have different purposes like: return -> new line, enter -> submit form

sierramountain:
Yes it is has two functions 'return' and 'submit' generally speaking but finding that it will only 'return' when using it in Tox

PR #1076, opened 2015-01-31
Fix keypad enter not sending message

worfox:
Issue #1068

This fixes the issue when you press the keypad enter the message was not sent because keypad enter sends a modifier, and if any modifier was present it would not send the message.

### DIFF
--- a/src/widget/tool/chattextedit.cpp
+++ b/src/widget/tool/chattextedit.cpp
@@ -27,7 +27,7 @@ ChatTextEdit::ChatTextEdit(QWidget *parent) :
 void ChatTextEdit::keyPressEvent(QKeyEvent * event)
 {    
     int key = event->key();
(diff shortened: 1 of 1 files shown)


dubslow:
I wonder how many one character commits are in qTox's history? At least 2 in the last few days lol

Also: I never would have found that, so much appreciated :) (dunno if tux3 would have of course)

sierramountain:
Thank you for resolving this issue so quickly. What do I need to do to resolve this on my version on my computer?  Thanks!

dubslow:
@sierramountain what operating system? Either your package manager or the built-in updater should have this already.

sierramountain:
Windows 7, unfortunately, my updater doesn't update automatically.  Looking to unistall tox and reinstall it completely

Issue #1074, opened 2015-01-31
Make microphone slider work


soredake:
Because I hear some of my friends is pretty quiet, and I have to do qtox louder through windows mixer. Thanks.

ghost:
As he says. Audio sliders affect nothing and you need to change volume in windows audio mixer. It's not a problem tet-a-tet, but group voice chat with 3 more people is kinda hell.

Windows 7@64-bit

fenque:
When will it be implemented? I hear your friends quietly because of this.

tux3:
Implemented by 4a40269698d2e8a8b492916e43916e59f95a9f31

PR #1082, opened 2015-02-01
Add support for playback audio slider setting

worfox:
This adds the support to change the sliders for the playback audio in the settings. There might be a better way to do this. This does work however, from what I've tested.
#1074


tux3:
It doesn't seem to support changing audio mid-call, and doesn't work on e.g. the ringtone, but it's definitely an improvement.

PR #1087, opened 2015-02-02
Tray new message notification

### LINKED ISSUE #547: Tray icon should animate while there are unread incoming messages.
Pleeeeease. :) For linux.

ghost:
Closes #547.
Adds tray icon notification on new messages: demo
Uses 1s timer (same one which is checking for idle user). Figured it can serve well as being general purpose timer for tasks like these. Btw timer of that gif is also 1s. If it displays otherwise your system is confused somehow ;)

Another change is use of svg images for tray icons. Problem is evident when resizing taskbar (say on linux) - tray icons do not scale too well. Now not only scaling problems are gone but code is cleaned up too. We no longer need `_x2.png` bloated because Qt scales svg just fine. It probably improves look of tray icon on HDPI monitors as well. Tested on windows too - works just fine. I believe svg icon plugin is already shipped with install anyway.

Oh and i renamed taskbar_**idle**.svg to taskbar_**away**.svg so status in app is same as part of icon name. Now we can avoid weird if/else madness and have cleaner code.


zetok:
uh, sorry to comment now, when you've already done it, but to make it more friendly for people with sight problems, I think that white space in lock also should change its colour to orange when it "pings" about new message

ghost:
With white space orange it would look ugly. It would so play out of tune with rest icons. As for people with sight problems i think it is out of scope of this change. Reasoning is that majority of people do not run high contrast desktop theme because they do not have sight problems and it looks ugly. Therefore i think right here and now coherent theme and good visual design should be prioritized.

However i do not mean visually impaired people should be neglected. qTox should most certainly take them into account. But best way IMHO is to create a dedicated theme. Then it would be work done right instead of slapping bits here and there and claiming good support while it would be of little help to those people because simply visual icon does not matter if rest of application is not well-suited for them.

Hey @ItsDuke since you are the artist here maybe you feel like designing a (coherent) theme for visually impaired?

ItsDuke:
@novist when you say "theme", you mean try icon, correct?

> majority of people do not run high contrast desktop theme because they do not have sight problems and it looks ugly

I'm not really arguing that the majority of our users _do_ have visual impairments, but you seem to think you have hard factual data. Also, since when does high-contrast or colorblind-mode look ugly?

src/widget/widget.cpp
```diff
@@ -1103,20 +1106,20 @@ bool Widget::isFriendWidgetCurActiveWidget(Friend* f)
...
                 activeChatroomWidget->resetEventFlags();
                 activeChatroomWidget->updateStatusLight();
             }
-        case QEvent::MouseButtonPress:
-        case QEvent::MouseButtonRelease:
-        case QEvent::Wheel:
-        case QEvent::KeyPress:
-        case QEvent::KeyRelease:
-            if (autoAwayActive)
-                onUserAwayCheck();  // Just so we get back from away faster when interacting with app
```
dubslow, line 1119: Why remove this?

src/widget/widget.cpp
```diff
@@ -297,19 +299,17 @@ void Widget::setTranslation()
...
-        pic = ":img/taskbar/" + color + "/taskbar_idle_2x.png";
-    else if (status == "busy")
-        pic = ":img/taskbar/" + color + "/taskbar_busy_2x.png";
+    QString status;
+    if (eventIcon)
+        status = "event";
     else
-        pic = ":img/taskbar/" + color + "/taskbar_offline_2x.png";
+    {
+		status = ui->statusButton->property("status").toString();
```
dubslow: spacing issue lol

src/widget/widget.cpp
```diff
@@ -265,7 +267,7 @@ void Widget::init()
     connect(setStatusAway, SIGNAL(triggered()), this, SLOT(setStatusAway()));
     connect(setStatusBusy, SIGNAL(triggered()), this, SLOT(setStatusBusy()));
     connect(addFriendForm, SIGNAL(friendRequested(QString, QString)), this, SIGNAL(friendRequested(QString, QString)));
-    connect(idleTimer, &QTimer::timeout, this, &Widget::onUserAwayCheck);
+    connect(timer, &QTimer::timeout, this, &Widget::onTick);
```
dubslow: Wouldn't it be better just to have on function for each thing using the timer, and then have one connect for each function?

As opposed to having a singular onTick() that does a billion things. (It's only two small things now, but if this is meant to be a general purpose timer then onTick wouldn't scale well.

src/widget/widget.cpp
ghost, line 1119: Its pretty much useless noise. It is triggered only when user get back from away and does any of those actions in app window. Basically starts typing to last selected contact. Window has to be in focus and right contact has to be open. Chance is real slim of that happening. Most of the time user will either start moving mouse and it will trigger away reset via timer. So i figured for the sake of cleaner code we can get rid of this without loosing anything.

src/widget/widget.cpp
ghost: Indeed, turns out some tabs crept in. Got all of them cleaned.

src/widget/widget.cpp
ghost: Good point, adjusted as you said ;)

ghost:
@ItsDuke actually i mean full theme for tox. Including tray icons and all. You know something like high contrast themes windows has in ease of access options:
vis_impair

dubslow:
With this branch, I don't have any tray icon whatsoever. Maybe my system can't handle .svg's...?

ghost:
What system is that? Basically what you need is qt having access to svg and ico icon plugins. Should be shipped basically with any qt version.

dubslow:
I have Mint Debian, Qt 5.2 from the debian testing repos (though from last summer, I can't really update packages that I snipe from testing very easily). How would I go about checking the Qt install?

ghost:
You might want seeing if you have svg icon plugin and svg image plugin. On my system they are at `/usr/lib/x86_64-linux-gnu/qt5/plugins/iconengines` `/usr/lib/x86_64-linux-gnu/qt5/plugins/imageformats`. You might want to try creating qt.conf next to compiled executable and add paths to conf manually. Like:

```
[Paths]
Plugins = /some/path
```

More docs: http://doc.qt.io/qt-5/qt-conf.html
Someone had similar problem [here](http://qt-project.org/forums/viewthread/3641)

dubslow:
Okay after much futsing around with temporarily adding the Debian testing repository to my apt sources, I've managed (with much manual download and installation) to install the Qt svg stuff, and this now works. I will be merging shortly.

@novist The busy tray icon looks like it has an extra little red to the right side of the image. I had previously imported some stuff from the Graphics repository, but noticed that some of the svgs were similarly not-quite-right. Would it be possible for you to fix it? Additionally, the busy icon in the UI, next to your name and in the drop menus for changing status still have the old busy icon that is the same shape as the away icon.

I will still merge for now as its an improvement, but someone please fix these issues.

ghost:
@dubslow could you provide a screenshot? I think i know what you are talking about because i have similar problem (KDE desktop). However it happens only when app starts and icon is fixed as soon as status changes. From then on its alright. Happened with png icons too. I tried looking for possible workaround for that but failed. I wonder if its same issue or not..

As for busy icons - ill replace them in all other spots and make PR.

dubslow:
screenshot from 2015-02-05 11 06 09

If you [look at it](https://github.com/tux3/qTox/blob/b7dd4bf8293fc5374f92c5af7fd48d055202f7d1/img/taskbar/dark/taskbar_busy.svg) in GitHub, it appears the top inside of the circle has issues as well. I believe some of the other svgs also have similar issues. @ItsDuke

ghost:
Oh its different problem then.

ItsDuke:
@novist obviously this is a WIP idea, but I'd imagine a high-contrast version of tox would look something like this:

highcontrast

@dubslow that's really odd. I went in and looked, and it seemed like only the offline, busy, and invisible svgs had those anomalies. Apparently when I saved them, they had a few too many anchors right next to each other, so I just went in and simplified all of them to have fewer anchors, and updated them all in my repo. Let me know if it gives you any further trouble.

ghost:
High contrast theme looks pretty good actually.

PR #1093, opened 2015-02-03
Audio code rearrangement

apprb:
1. qTox's audio related stuff moved to separate folder in order to match video subsystem.
2. `#ifdef` magic into one place has been moved into one place, which made code prettier.


dubslow:
I honestly prefer the old way, it's easier to see that it may or may not be active. With all the defines in one place, instead everywhere else looks like there's always audio filtering.

dubslow:
I mean, just because I think something doesn't mean it's right... surely you had a reason?

Issue #1027, opened 2015-01-21
Emoji selector is too wide with large emoji lists


ProMcTagonist:
![stretched emoji selector](https://cloud.githubusercontent.com/assets/4358079/5832022/e90d43b0-a0f2-11e4-9921-1ccb215ba34e.png)

Reproduce:
- Download and install http://kde-look.org/content/show.php/Google+Hangout+Emojis+Complete?content=162161
- Restart qTox
- Select GoogleEmojiComplete as the Smiley Pack in Settings
- Open a conversation and click the smiley button

dubslow:
@apprb @krepa098

apprb:
I really want to help, but unfortunatelly till the end of this month (or even more) I'm very busy with my work.

krepa098:
![text4615-9-1-6](https://cloud.githubusercontent.com/assets/4587853/6087771/c7204c92-ae4f-11e4-8805-509c8801863e.png)

suhr:
cast @ItsDuke

ProMcTagonist:
@krepa098 has the right idea. The popup should be remodeled after Facebook's, in preparation for stickers.
![popup](https://cloud.githubusercontent.com/assets/4358079/6088836/91e9a19e-ae0f-11e4-8d6c-9ebb7353e285.jpg)

redsteakraw:
This was fixed when the emoticon widget was switched to an 8x8 grid.  @dubslow

ProMcTagonist:
No, it wasn't. The current implementation is fundamentally flawed. It will be a problem again with larger lists of emoji and TwitterEmojiSVG is still misaligned because of the long row of pager dots. It needs this redesign.

A note, the tab bar should be vertical and on the side, because it's awkward and unpleasant to navigate Facebook's horizontal one.

redsteakraw:
@ProMcTagonist  this was a bug not a feature request as far as bugs are concerned it was fixed, there isn't a huge list for large Emoji sets as far as what your feature request should be filed separately and this should be closed.

ProMcTagonist:
The core problem isn't fixed, 
![emoji paginator too wide](https://cloud.githubusercontent.com/assets/4358079/6374488/150e5fd2-bcc6-11e4-8131-d5b1bb8269e3.png)
and will get worse again whenever increasingly longer lists of emoticons are used. This is because the problem is inherent to the current pagination model. 

[...]

PR #1105, opened 2015-02-05
Added Twitter emoji SVGs with copyright notice and my emoticons.xml file

redsteakraw:
Added Twitter emoji SVGs with copyright notice and my emoticons.xml file this brings color emoji support to qtox it also will bring color emoji to previous emoticon strings like :-) It is a very large emoticon set compared to the previous two.


zetok:
Well, I think that it should be set to be default emoticon pack, due to it supporting most (all?) of them(?)

redsteakraw:
Twitter doesn't have complete flags, nor the middle finger emoji and a lot of the Unicode 7 emoji but that can be added as they update their set.  It is a lot more complete than the current ones included and it exposes a few bugs (such as [issue#1027](https://github.com/tux3/qTox/issues/1027) ) that should be fixed before it is default.   However having it included at the very least makes it easier to test.

dubslow:
Could you put all the smileys in a separate subfolder, so that the documentation is easy to find? Additionally, can you estimate how large the combined smileys are, so that we can consider embedding them into the binary?

redsteakraw:
Size wise the SVGs add 2.2 MB 16x16 is 338KB 32x32 458 KB 72x72 713KB,  I went with the SVGs because it can be scaled to any size without dupication even larger ones for HiDPI displays.  However if png files are prefered I could easily swap to any size my emoticons.xml file works for all of them without modification.  I can do a subset, however it will lack the color emoji support for the greater emoji standard, given Mobile clients will use them it  will leave the clients with reduced experience.  Ideally I would like to [Organize them in to seperate pages as suggested by the upcoming unicode standard](http://www.unicode.org/Public/emoji/1.0/emoji-ordering.html) but emoticons seem to be all or nothing you can't lable or have seprate groupings other than the placement in the emoticons.xml file.

zetok:
IMHO size isn't that important (especially since it's next to ~nothing) when it comes to benefit in quality that SVGs offer.

dubslow:
People complain all the time about how qTox is 30-50 MB vs uTox's less than 10 (?) I believe, so yes, size does matter. If no one cares though, then @redsteakraw should add a commit that would properly embed these in the binaries.

krepa098:
Resources are compressed by default (http://qt-project.org/doc/qt-4.8/resources.html#compression) so its probably less than 2.2MB :).

redsteakraw:
Dubslow so you want the pngs as well?  What changes do I need to make before this can be merged?

dubslow:
No no, just the svg's is fine. Take a look at the `qrc.res` file, that should get you started on how to embed these into the binary. Or maybe @krepa098 you have better instructions?

krepa098:
Hmm I don't see a point in merging this now. The emoticons widget simply cannot handle this amount of items. :disappointed: 

> Ideally I would like to Organize them in to seperate pages as suggested by the upcoming unicode standard 

That would be nice to have!

> but emoticons seem to be all or nothing you can't lable or have seprate groupings other than the placement in the emoticons.xml file.

Yes, shouldn't be too hard to add though.

> Or maybe @krepa098 you have better instructions?

We can have multiple qrc files afaik. Maybe we can put each smiley pack in its own qrc?

redsteakraw:
Well I added the files to res.qrc you can merge this if you want if the emoticon input / system is improved I am more than willing to update or reorder the emoticon.xml or switch the order.

dubslow:
Okay I see it's added to the res.qrc. However, it's still not showing up in the GUI. @krepa098 I glanced at the code, but it seems that all that's necessary is to add it to res.qrc, because the smiley code already searches for embedded smileys... any ideas?

krepa098:
File extensions are missing:
`<emoticon file="1f31f">` should be `<emoticon file="1f31f.svg">`

redsteakraw:
The emoticon files doesn't need file extensions, as it is working fine now without it on my system

redsteakraw:
I found the problem, the emoticon.xml file was not listed in the res.qrc file

redsteakraw:
My latest commit sets the emoticon input box to 8x8 grid for emojis which should [fix issue #1027](https://github.com/tux3/qTox/issues/1027)

dubslow:
Hmm, now I don't see any of the embedded packs in the GUI list.

Edit: You dropped the `smiley/` prefix from the new .qrc, but the paths are relative to the top dir regardless of the .qrc's location. I'll make my own commit with the correct changes.

PR #1111, opened 2015-02-06
simplemake: Added svg-Support 

Lednerb:
Avoids errors by compiling qTox.


dubslow:
Can you also find out what the yum/pacman packages are called?

Lednerb:
I think i've found them.
But could not test these packages.
`apt-get` command works fine.

PR #1112, opened 2015-02-06
Fixed busy and invisible taskbar icons

ghost:
Updated other status icons with latest stuff from @ItsDuke's repo.
Removed duplicate status icons from ui/statusButton. That icon setting via css is fancy and all but person who figured having two copies of same icons was a good idea deserves to loose a finger :p Because for some reason using icons from other resource path in css no longer worked i had to add manual icon setting for button. Still way better than two copies of icons.

@ItsDuke you might want to grab these updated svg icons and update your repo.


ItsDuke:
@novist I must be not reading correctly - what is the difference between the SVGs I'm currently hosting and these?

ghost:
Your busy and invisible SVGs have glitches. Uneven curves and all. Although i found it strange that PNGs you exported (i suspect) have no said glitches. My guess would be maybe software you are using has a bug and it displays your SVGs right while inkscape and Qt interprets them with glitches. So i fixed those glitches with inkscape and thats the difference ;)

PR #1113, opened 2015-02-06
Added German translation for new Encryption-Dialogs and -forms

Lednerb:
Still some problems with dialogs, whenever there is an `\n` inside.

Example:

dialog

### DIFF
--- a/translations/de.ts
+++ b/translations/de.ts
@@ -1265,12 +1265,6 @@ Soll die alte Historiedatei gelöscht werden?</translation>
         <source>Local file encryption</source>
         <translation>Lokale Dateien verschlüsseln</translation>
     </message>
-
-
-
-
-
-
     <message>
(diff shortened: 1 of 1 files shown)


dubslow:
Could you specify which source lines have a `\n` inside them? I can easily move those outside the `tr()` call.

Lednerb:
I just copy and paste the locations here:

```
<location filename="../src/widget/form/settings/privacyform.cpp" line="116"/>
<location filename="../src/widget/form/settings/privacyform.cpp" line="146"/>
```

Can you pls generate the translation files with these great "vanished" and "unfinished" tags?
Would be less painful to translate ;-)

dubslow:
It appears the Italian translation works fine...

Edit: https://github.com/tux3/qTox/commit/883d9cda4ab1659b8594135f998e4a232509b017?diff=unified#diff-307f5ff0d91dcf4b13295b8c45511315R1350

Just make a new line, rather than trying to use the escape sequence.

PR #1124, opened 2015-02-08
Attempted to make Greentext better.

Dirius77:
If I understand how the strings here work. This should make it so that only text that contains at least 1 'word' or 'diget' character, will be greentexed. allowing for things like >_< to be completely fine, while >What are you talking about   will be greentexted, as well as anything more obscure such as > _Im_not_sure_why_youd_greentext_this     as well.

Please tell me if I've misunderstood how the regex's work, I find them a bit confusing, and I will happily fix this if I've done it wrong.

### DIFF
--- a/src/widget/tool/chatactions/messageaction.cpp
+++ b/src/widget/tool/chatactions/messageaction.cpp
@@ -63,7 +63,7 @@ QString MessageAction::getMessage(QString div)
     message_ = "";
     for (QString& s : messageLines)
     {
-        if (QRegExp("^&gt; .*(\\w{1,}|\\d{1,}).*").exactMatch(s))
+        if (QRegExp("^&gt;.*(\\w{1,}|\\d{1,}).*").exactMatch(s))
             message_ += "<span class=quote>" + s + "</span><br/>";
         else
             message_ += s + "<br/>";
--- a/src/widget/tool/chatactions/messageaction.cpp
+++ b/src/widget/tool/chatactions/messageaction.cpp
@@ -63,7 +63,7 @@ QString MessageAction::getMessage(QString div)
     message_ = "";
     for (QString& s : messageLines)
     {
-        if (QRegExp("^&gt;( |[[]|&gt;|[^_\\d\\W]).*").exactMatch(s))
+        if (QRegExp("^&gt; .*(\\w{1,}|\\d{1,}).*").exactMatch(s))
             message_ += "<span class=quote>" + s + "</span><br/>";
         else
             message_ += s + "<br/>";


dubslow:
`+` is a synonym for `{1,}`, and also there's a space before the first character, meaning `>this` wouldn't be greentexted.

Other than the required space, what do you think @zetok @suhr?

suhr:
`\w` captures `_` (and we want to avoid it) and digits, and why `(\w+|\d+)` instead of `[\w\d]+`? I believe this regexp can be replace by `^&gt;.*[^_\\W].*`.

Actually, something like `>_< ‹some text of frustration›` may cause problems. But this is pretty rare kind of messages.

Dirius77:
I just wrote out the regex the way that made the most sense to me, and I find it easier to read as if(string.containsWord || string.containsDiget) then { } . I'm not sure how exactly I'd read the bracket part. But I can switch that out if you want.

Also, with the emote and text, I think at some point we have to draw the line for nitpicking, and I find that people very rarely type out the emote before the text. I'd be more likely to type:
`Augh this doesn't work >_<`
than I would:
`>_< Augh this doesn't work`

zetok:
> \w captures _ (and we want to avoid it) (…)

do we?

Why break consistent behaviour just because of `>_<` / `>_>` / `>.<` / `>.>` / other variation?

Why removing green colour from those is so important, to the point of crippling everything?

suhr:
@zetok, please. It could be just replaced by `^&gt; .*` for greater consistency and no user confusion. Just because 4chan'ers won't be happy it is not done yet.

ProMcTagonist:
It has nothing to do with imageboards. You're beating a dead scarecrow. It's about quoting not acting inconsistent and confusing.

suhr:
> It's about quoting not acting inconsistent and confusing.

What exactly is inconsistent (or even confusing) in using `>` with space as greentext syntax instead of raw `>`?

ProMcTagonist:
> What exactly is inconsistent (or even confusing) in using > with space as greentext syntax instead of raw >?

**Inconistent**: it's only necessary sometimes within qTox. It's inconsistent with every other implementation ever. It alienates existing users as well as the conceptual Average User.

**Confusing**: Requiring a space as a part of quote syntax. Why? Users will have to figure out this behavior on their own. It looks weird. It's just odd. 

And no, a tutorial is not the answer to this broken behavior.

Dirius77:
Why not make greentext highlighting an option within the settings and make
it any > ?
I dont really understand the config area of this code well enough to
implement it yet, but I can look into it.

suhr:
> and make it any >

You mean, allow user to choose the greentext syntax?

Dirius77:
I mean, whenever there is a > you greentext it, and just give them the
option to turn it off. Perhaps have the ability to set the regex under the
advanced tab.

tux3:
This PR doesn't seem to have been making much progress in the last two weeks, it also conflicts with master. Closing for now.

PR #1133, opened 2015-02-09
Fixed emoji size settings to change emoji size added Middle finger emoji

redsteakraw:
...le Finger emoji


dubslow:
I'm waiting on this until new chatform is merged, that has a fix for the resize issue

dubslow:
The rescaling should be fixed now, `reset --hard` to the first commit here and force push.

PR #1134, opened 2015-02-10
Avatar Choosing Dialog

CosmicToast:
Explicitly filtering out images for the avatar selection dialog.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -344,7 +344,10 @@ QString Widget::getUsername()
 
 void Widget::onAvatarClicked()
 {
-    QString filename = QFileDialog::getOpenFileName(this, tr("Choose a profile picture"), QDir::homePath());
+    QString filename = QFileDialog::getOpenFileName(this,
+        tr("Choose a profile picture"),
+        QDir::homePath(),
+        tr("Images (*.png *.jpg)"));//TODO: cover all types of images QPixmap can use reliably
     if (filename.isEmpty())
         return;
     QFile file(filename);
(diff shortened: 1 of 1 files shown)


dubslow:
I believe this was discussed, and we didn't want to artificially limit what users could choose?

Although your list is explicitly what QPixmap can handle... @tux3 ?

zetok:
Currently it's already impossible to add as image something that isn't an image - there's an error shown that something couldn't be loaded as image. With this PR though, if you have an image that doesn't end in one of formats, you'd have to rename it to one of things that qTox would _require_. IMHO that's not an improvement.

CosmicToast:
A friend of mine (while testing) had his Qt freak out and only be able to select executables.
My theory was that this is because no filters are explicitly up (that and/or memory corruption). Specifying a set of supported images (and making it as extensive as possible) would be idea.
Ideally, the filter should be generated from [QImageReader::supportedImageFormats](http://doc.qt.io/qt-5/qimagereader.html#supportedImageFormats), which I'm planning to implement eventually.

dubslow:
I think I'll punt on merging this until someone implements QImageReader, as you suggested you might do.

tux3:
I think the list should also include TIFF and WEBP since Qt support those.

CosmicToast:
Added dynamic generation of the list.
Maybe make it static? Not sure how much the code would actually benefit from that.
Another TODO: move the code into a QString getSupportedFormatsFilter().
Note: while this should work, I'm currently not in a position to test it, if you could add support for travis CI it'd greatly increase (at least my, but hopefully others') productivity.

PR #1155, opened 2015-02-13
Fix #941

### LINKED ISSUE #941: Enable building with shared libraries
Building with shared libraries would make using different tox clients in parallel much easier.
If the dependencies are alreadiy installed it would also reduce the build process to:
qmake
make

sudden6:
Build with system libraries by default.


dubslow:
As discussed in the issue, this isn't necessary

PR #1156, opened 2015-02-14
Audio code rearrangement

apprb:
Almost the same description as for previous one, but this time without any additional classes.


tux3:
It's not clear that this PR is an improvement.

The calls to filter audio look like they are being made, but they actually don't do anything, this is a bit misleading. When QTOX_FILTER_AUDIO is not defined, there's no external indication that some functions don't actually do what their name say they do.

Additionally, this change leaves empty useless code in the binary and wastes some memory ( <code>AudioFilterer*Core::filterer[TOXAV_MAX_CALLS]</code>) instead of compiling it out.

Closing for now.

PR #1157, opened 2015-02-14
Rename 'Start automatically' → 'Autostart'

zetok:


dubslow:
Er... isn't the first one more clear?

zetok:
~~no~~ I don't think so, since "Autostart" seems to be widely used in windows ecosystem, thus (most?) of users may already be familiar with it.

EDIT: also, `Start automatically` sounds weird to me

PR #1188, opened 2015-02-16
Created TwitterEmojiSimple doesn't replace ascii smiles

redsteakraw:
Due to demand / complaints I have made an alternative emoticon theme that shares all the same assets with TwitterEmojiSVG but has a different emoticons.xml file.  It doesn't replace ascii smiles like :-)   ;-)   :-(  and is a more simplified emoticons.xml file.  This was created in response to complaints asking that the ascii emoticons aren't replaced.  Also Theme emoji theme tweaks.


zetok:
Oh! That's awesome. I feel like I finally will be able to have emoticons turned on.

PR #1194, opened 2015-02-16
Simplified Chinese translation

broncotc:
This is the zh-CN translation I created today.


dubslow:
Hmm it looks like you didn't actually run Qt Linguist, since you copied the French file? It's slightly out of date, a lot of file line numbers have changed...

broncotc:
I manully translated it from the <source>s in the french file. When I put it into Linguist, all the ticks are green.

broncotc:
All the translation is finished. lupdate is executed. QT 4.8.6 is used.

dubslow:
I'll merge this, but in the future try to avoid updating all language files, only update the one(s) you're working on.

dubslow:
You also accidentally added a spurious file "null"

broncotc:
Ok, get it~

PR #1205, opened 2015-02-17
[INSTALL.md] Same build dependencies for Debian and Ubuntu

zetok:


zetok:
Or would it be better to just make it as single entry `Debian / Ubuntu / etc.`?

dubslow:
Single entry is better. I suggested it when you first reorganized, but you didn't like it then.

PR #1221, opened 2015-02-18
Hide is typing status when friend disconnects

TheLastProject:
Currently, when a contact goes offline, the typing notification remains:
typingafteroffline

This patch fixes that (`src/widget/widget.cpp`, line 687 in this patch)

It should also resend typing notification when a contact comes back online, so that the typing notification message comes back after a reconnect if the contact is still typing (obviously, other clients need to implement this as well, so this would currently only be available on qTox-to-qTox discussions).

The support for this last feature, however, has not been tested as I could not get a second qTox instance to run on TCP-only to play with disconnecting while typing a message. I do, however, see no real reason why this wouldn't work, but would still recommend it to be tested.

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -901,6 +901,11 @@ QString ChatForm::secondsToDHMS(quint32 duration)
     return cD + res.sprintf("%dd%02dh %02dm %02ds", days, hours, minutes, seconds);
 }
 
+bool ChatForm::getIsTyping()
+{
+    return isTyping;
+}
+
 void ChatForm::setFriendTyping(bool isTyping)
 {
     chatWidget->setTypingNotificationVisible(isTyping);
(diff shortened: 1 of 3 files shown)


dubslow:
Um, when the contact comes online, it sends whether or not we are typing, to them. Is that what you intended?

TheLastProject:
I probably explained it badly, but yes, that part was also intended, for the following to work:
1. Alice starts typing
2. Bob gets typing notification
3. Alice and Bob lose a connection
4. Bob's client removes Alice's typing notification, because Alice is no longer online
5. Alice and Bob connect again
6. As Alice is still typing, her client tells Bob she is still typing
7. Bob's client shows a typing notification shown again

Unless I'm mistaking, resending it upon reconnecting is required for it to work this way. Then again, thinking of it, maybe the typing notification could also simply be hidden and unhidden on going offline and online, as clients would probably queue their "is typing" status messages and tell us if they're no longer typing?

PR #1228, opened 2015-02-19
Updated windows/bootstrap.sh to build filter_audio using Makefile

ghost:
Depends on https://github.com/irungentoo/filter_audio/pull/10


ghost:
@tux3 maybe consider temporary fork of filter_audio and pointing windows script to that. Since noone seems to have any interest in merging required PR any time soon so until that happens windows users will have bootstrapping errors that will have to be solved manually somehow.

PR #1231, opened 2015-02-19
Allow user to open all transferred files, but warn for executable files

### LINKED ISSUE #1174: Open transfered files
<SylvieLorxu> You can open any extension through transferred files
<SylvieLorxu> But only some images through a file transfer in a chat

TheLastProject:
This allows users to open any file sent to them in in-chat file transfers, and in the file transfer history.

When the user tries to open an executable file, they are warned, with the response defaulting to No for security reasons:
executablewarning

Note: I wanted to use the warning icon, but then I couldn't use Yes/No :(

This fixes #1174.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -502,7 +502,7 @@ bool Widget::confirmExecutableOpen(const QFileInfo file)
 {
     if (file.isExecutable())
     {
-        if(!GUI::askQuestion(tr("Executable file", "popup title"), tr("You have asked qTox to open an executable file. Executable files can potentially damage your computer. Are you sure want to open this file?", "popup text"), false, true))
+        if(!GUI::askQuestion(tr("Executable file", "popup title"), tr("You have asked qTox to open an executable file. Executable files can potentially damage your computer. Are you sure want to open this file?", "popup text"), false, false))
         {
             return false;
         }
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -502,7 +502,7 @@ bool Widget::confirmExecutableOpen(const QFileInfo file)
 {
     if (file.isExecutable())
     {
-        if(!GUI::askQuestion(tr("Executable file", "popup title"), tr("You have asked qTox to open an executable file. Executable files can potentially damage your computer. Are you sure want to open this file?", "popup text"), false, false))
+        if(!GUI::askQuestion(tr("Executable file", "popup title"), tr("You have asked qTox to open an executable file. Executable files can potentially damage your computer. Are you sure want to open this file?", "popup text")))
         {
             return false;
         }
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -59,6 +59,8 @@
 #include <QByteArray>
 #include <QImageReader>
 #include <QList>
+#include <QDesktopServices>
+#include <QProcess>
 #include <tox/tox.h>
 
 #ifdef Q_OS_ANDROID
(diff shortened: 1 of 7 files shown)


dubslow:
What's wrong with Ok/Cancel buttons for this question?

I wonder if the confirmExecutable function should go in the GUI class... or maybe Nexus? Hmm...

TheLastProject:
"Do you want to do X?" is clearly a Yes/No question in my view.

dubslow:
But it's not `Do you want to do X?`, it's a `Are you sure you want to do X?`

And Ok/Cancel is still fine for either IMO :P

TheLastProject:
I don't know, "OK/Cancel" feels wrong to me, because if I'd be asked this question in real-life I'd answer either Yes or No. Is there no way to have the warning symbol but still Yes/No as buttons?

dubslow:
You can directly call QMessageBox yourself (see the code for askQuestion())

TheLastProject:
I instead opted for slightly refactoring askQuestion to allow you to explicitly choose yes/no instead of ok/cancel, because this seemed cleaner to me. I also checked all current dialogs using askQuestion and decided that all but one are better suited as a yes/no question, so I made that the default.

I hope my changes are acceptable. Otherwise, feel free to cherry pick this. Up to 5b6e2d5 for the Yes/No question, up to 5b6e2d5 excluding 718aad2 for the Ok/Cancel warning, or the whole thing for a Yes/No warning. You work a lot more on qTox than me so I'm going to leave the final judgement to you, although I personally feel a Yes/No warning is best for this.

yesnowarning

src/widget/widget.cpp
```diff
@@ -498,6 +498,19 @@ void Widget::onTransferClicked()
     activeChatroomWidget = nullptr;
 }
 
+bool Widget::confirmExecutableOpen(const QFileInfo file)
+{
+    if (file.isExecutable())
```
tux3: You can't use QFileInfo::isExecutable on windows and more generally on filesystems that don't have permissions, because it always returns true.
And on Linux/Mac file transfer doesn't preserve permissions, so there's no way to detect if a received file is executable, it would always be false (since default perms are non-executable).

Maybe check if the filename finishes with common executable/dangerous extensions.
Here's a list from google : .exe, .com, .pif, .msi, .msp, .src, .app, .dmg, .hta, .cpl, .msc, .jar, .py, .bat, .vb, .vbs, .vbe, .js, .jse, .ws, .wsf, .wsc, .wsh, .ps1, .ps1xml, .ps2, .ps2xml, .psc1, .psc2, .msh, .msh1, .msh2, .mshxml, .msh1xml, .msh2xml, .scf, .lnk, .reg

TheLastProject: According to [StackOverflow](http://stackoverflow.com/a/15083959) it should return true on .exe, .com and .bat. On Linux and the likes, it indeed will always return false, but I'm pretty sure they won't ever be executed as programs there anyway? At least, the last time I sent myself a .sh file it simply got opened in my text editor.

I will, however, replace it with the extension list anyway, because I can imagine the odd situation where WINE may execute a .exe despite it not being marked as executable (don't know if it actually does that, though) or something similar. It's probably the safer way to go about it.

tux3: @TheLastProject that stackoverflow post seems to be wrong.

If you look at the Qt5 source code (I also looked at 4.8 and 4.2) it seems to only be checking permissions. The doc also implies that it just check permissions, it never mentions filenames.

snapshot5

tux3:
We also need to change the permissions before we're able to run a file. We need to add u+x if the user agrees to run it.

PR #1243, opened 2015-02-20
IPC rework and #840 fix

### LINKED ISSUE #840: minimize to tray doesn't disable opening new instance
- On Windows "Pin to Taskbar" the open instance of qTox
- then minimize qTox to Tray 
- then I click on the Icon in my starter Taskbar (not the tray icon) then there is started a new instance of qTox.

Instead the minimized instance should maximize.

(tested on Windows 7 64bit and Windows 8.1 64 bit)

ghost:
To fix #840 (yes, its not fixed even though its closed) i had to introduce new functionality to IPC and since it was a mess i rewrote it.
#### Simpler design

Uses structs and arrays instead of magic offsets and memory pages. Now it is much easier to read/modify/maintain the code.
#### Named events

Now only handler of received event is called as opposed to calling all handlers and having them check data object to decide if they want to handle it or not. Now data object can hold more meaningful data.
#### Additional functionality

Now we can check if our event was handled and accepted. Main instance can handle event but in case handler returns `false` it will not count as accepted. Also we can send event to instance that runs specific profile. Last two features are important for fixing #840.
#### #840 fix

Now new instance launched with `-p` flag will send `activate` event to instance that runs profile specified with `-p`. In case no instance handles such event then new instance with specified profile will be launched. If no `-p` flag is specified event is sent to main instance that is already running. If it is not running app starts normally.

Note that i have changed global memory key to `qtox-IPC_PROTOCOL_VERSION`. This is done because on unix systems shared memory may not be always released. In such case updating qtox to new version would crash because size of shared memory changes. Having unique key solves this problem.

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -81,6 +81,15 @@ IPC::~IPC()
     }
 }
 
+IPC& IPC::getInstance()
+{
+#ifdef Q_OS_ANDROID
+    Q_ASSERT(0 && "IPC can not be used on android");
+#endif
+    static IPC instance;
+    return instance;
+}
+
 time_t IPC::postEvent(const QString &name, const QByteArray& data/*=QByteArray()*/, uint32_t dest/*=0*/)
 {
     QByteArray binName = name.toUtf8();
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -16,37 +16,39 @@
 
 
 #include "src/ipc.h"
+#include "src/misc/settings.h"
 #include <QDebug>
 #include <QCoreApplication>
+#include <unistd.h>
 
-IPC::IPC() :
-    globalMemory{"qtox"}
+
+IPC::IPC()
+    : globalMemory{"qtox-" IPC_PROTOCOL_VERSION}
 {
     qRegisterMetaType<IPCEventHandler>("IPCEventHandler");
 
-    ownerTimer.setInterval(EVENT_TIMER_MS);
-    ownerTimer.setSingleShot(true);
-    connect(&ownerTimer, &QTimer::timeout, this, &IPC::processEvents);
+    timer.setInterval(EVENT_TIMER_MS);
+    timer.setSingleShot(true);
+    connect(&timer, &QTimer::timeout, this, &IPC::processEvents);
 
     // The first started instance gets to manage the shared memory by taking ownership
-    // Every time it processes events it updates the global shared timestamp
+    // Every time it processes events it updates the global shared timestamp "lastProcessed"
     // If the timestamp isn't updated, that's a timeout and someone else can take ownership
     // This is a safety measure, in case one of the clients crashes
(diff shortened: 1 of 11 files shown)


src/ipc.cpp
```diff
@@ -16,37 +16,39 @@
...
+#include "src/misc/settings.h"
 #include <QDebug>
 #include <QCoreApplication>
+#include <unistd.h>
 
-IPC::IPC() :
-    globalMemory{"qtox"}
+
+IPC::IPC()
+    : globalMemory{"qtox-" GIT_VERSION}
```
tux3: Is there any reason you would give a separate shared memory to each version ?

That seems bad, having multiple clients running on the same profile could cause inconsistencies, and after an update the new client obviously won't be able to talk to previous, which could be confusing.

If you just want to prevent clients with different IPC protocols from messing with each other, I'd suggest using a new fixed name for the global memory like "qtox2" or "qtox-#1243".

dubslow:
Are you suggesting that we support multiple instances at the same time? That's liable to cause a whole host of inefficiencies and bugs relating to config file management... I'd really rather not go down that route...

ghost:
Well.. For what it is wroth qTox already can run with two profiles and it brings its problems. This patch makes things go nicer. While there still can be problems there certainly will be less of them. Although i do not see many issues with this. We simply should allow writing of global settings by global instance (which handles IPC) and still allow every profile write it's own settings. That way risk to crash and burn should be minimal.

src/ipc.cpp
ghost: I do not believe this is an issue. Simply because you cant update a running app on windows and unix users should know better to restart alpha software after update. Anyhow ill make said change.

zetok:
> That's liable to cause a whole host of inefficiencies and bugs relating to config file management... I'd really rather not go down that route...

qTox already is on that path since long time ago…

# 

> We simply should allow writing of global settings by global instance (which handles IPC) and still allow every profile write it's own settings. That way risk to crash and burn should be minimal.

+1

ghost:
Something went south, i would really appreciate another set of eyes to help figuring this out. Worse yet it worked yesterday but somehow today code is broken and i dont know what to believe.

Problem:
`currentProfileId` is set [here](https://github.com/tux3/qTox/pull/1243/files#diff-a095c041ed4fe02895e04ee7e2ddff6fR766). However pointer pointing to that variable somehow ends up few levels deep in QtCore right on [this](https://github.com/novist/qTox/blob/autorun-fix/src/main.cpp#L98) line. Down-below QFileInfo gets called and it overwrites `currentProfileId`.

I cant spot the problem so far, any help would be greatly appreciated.

ghost:
Okay i sorted it out. Turns out Settings was freed and reallocated. Problem is corrected and patch is ready for testing.

As you can see i added restriction to write main settings only for instance that owns IPC. But i think for the perfect solution we also need something like a lock file for each profile in temp dir. You know so two instances cant run same profile at the same time.

PR #1244, opened 2015-02-21
Delayed tray icon creation

ghost:
This change solves issues/crashes where tray is not immediately available when qTox is started. It is common on linux desktops. While for example lxqt has option to delay application autostart until panel (and thus tray) is available other desktops (like KDE) do not. Adding checks around use of `icon` object was not enough because application may start a little bit earlier than panel is available. For that reason tray icon creation is delayed to timer event and tried multiple times with delay of one second. Usually after few tries icon creation succeeds and signal is disconnected.

In case tray is not available qTox window is shown. This creates a side effect where starting qTox before tray is available will make application window briefly appear and when tray is available window will be hidden or remain visible as per settings. Window has to be visible if tray is unavailable because otherwise users may end up with qTox running and no way to access it. If application is started and tray is available no window flashing happens and settings are restored as usual.

Without this patch if qTox started before tray is available window is shown and changing tray icon visibility option crashes application. Thats probably a known issue.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -112,46 +112,6 @@ void Widget::init()
     statusBusy->setIcon(QIcon(":img/status/dot_busy.png"));
     connect(statusBusy, SIGNAL(triggered()), this, SLOT(setStatusBusy()));
 
-    if (QSystemTrayIcon::isSystemTrayAvailable())
-    {
-        icon = new SystemTrayIcon;
-        updateTrayIcon();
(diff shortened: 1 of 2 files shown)


zetok:
I have a request. If you are writing explanation, can it be put into commit message instead?

ghost:
If noone objects long commit messages then sure

tux3:
I love long commit messages, that's what they are for.

dubslow:
@tux3 Any reason to not merge this?

PR #1248, opened 2015-02-21
Fix filter_audio build on linux

strk:
and more error handling


dubslow:
This file isnt even used anymore, pending a particular PR from novist. (I can help more on Monday)

Issue #1265, opened 2015-02-23
missing tray support when using Status Notifier


zetok:
on fdb86ca7e924493a025e5a425701f81a2f67ed0b
When compiled with support for Status Notifier, qTox is missing tray icon.

```
SystemTrayIcon: Using Status Notifier backend
```

When compiled without, it's working as it is supposed to.

Happens under i3.

Also a regression that hits users of binaries.

likwrk:
With the latest binary on Ubuntu 14.04 lxde or simply openbox with tint2 tray icon disappeared

tux3:
Could both of you post the result of "echo $XDG_CURRENT_DESKTOP" ?

likwrk:
I have nothing in that variable

tux3:
@likwrk then please try again with the very latest version, it should work like it did before.

agilob:
I have very similar problem on KDE4.

One commit completely broke support for tray icons in KDE4. Application starts hidden in tray, but no icon is visible. qTox is running in background without any possibility to show window or anything. I had to kill it by PID.

Removing lines: 40-63 from https://github.com/tux3/qTox/blob/master/src/widget/systemtrayicon.cpp#L40 fixed the problem here.

```
    #ifdef ENABLE_SYSTRAY_STATUSNOTIFIER_BACKEND
    else if (desktop == "kde")
    ...
    ...
        status_notifier_register(statusNotifier);
    }
    #endif
```

tux3:
@agilob that's very strange, I'm on KDE 4.12.2 right now and it works just fine. It also works on the official KDE 5 live cd.
Can you tell what exact version of KDE4 you use ?
Can you also post the content of $DESKTOP_SESSION, $KDE_SESSION_VERSION and $KDE_FULL_SESSION ?

zetok:
`$ echo $XDG_CURRENT_DESKTOP` returns nothing for me.

``` bash
$ echo $DESKTOP_SESSION
i3
```

likwrk:
@tux3 thanks. Latest binary version works know

agilob:
kdease 14.12.2-1
qt5-base 5.4.0-3

```
➜  qTox git:(traykde4) ✗ echo  $DESKTOP_SESSION
default
➜  qTox git:(traykde4) ✗ echo $KDE_SESSION_VERSION
4
➜  qTox git:(traykde4) ✗ echo $KDE_FULL_SESSION 
true
```

tux3:
[...]

PR #1268, opened 2015-02-23
fix tray icon support for KDE4

agilob:
It fixes my problem with tray icon on KDE4, reported in [comment here](https://github.com/tux3/qTox/issues/1265).


tux3:
I can't merge it as is since StatusNotifier is the only one that works on KDE5.

agilob:
So there will be no support for tray icon in KDE4?

tux3:
@agilob of course there'll be support for tray icon on KDE4, I'm trying to understand why it doesn't work for you. This patch is just not the solution. Not only does it break the system tray icon on KDE5, but it doesn't even remove the feature cleanly.

Issue #988, opened 2015-01-06
Groupchat peer list layout is badly broken


suhr:
Picrelated.

![](http://a.pomf.se/hycsab.png)

zetok:
duplicate of #919?

Edit: it's not really broken, if one doesn't count layout. Someone just set their nick to something long.

agilob:
@zetok no it's a duplicate of #752 which I broke some time ago ;)

It's fixed [here](https://github.com/tux3/qTox/pull/989)

Issue #919, opened 2014-12-10
qTox can't handle displaying more peers in groupchat


zetok:
![screenshot](https://cloud.githubusercontent.com/assets/3148759/5384020/1b5b4476-80ae-11e4-8c5d-a0331751d3af.png)
↑ doesn't even display them

Should be fixed when #677 and #701 will be implemented

tox-user:
Still relevant or can be closed?

zetok:
Does it work with 100+ people in chat?

PR #1275, opened 2015-02-24
closes again tux3/qtox #988

agilob:
I think that's it.

@zetok @dubslow could you test it?

It looks good here.
kkk
kkk1


dubslow:
Hmm, with long lists it significantly lengthens the vertical space taken... still an improvement I guess. (I haven't tested it)

zetok:
improvement?
Seems to improve situation with smaller groups, I have yet to see how it'll behave with >40 people, where there is #919

Issue #1295, opened 2015-02-27
Putting HTML in names?


JFreegman:
Who thought this was a good idea? https://imgur.com/9rTdN0r

agilob:
It's standard QLabel behavior to parse HTML. Every Qt application does it.
I think parsing HTML is a feature, not bug.

 You can also put JS, QML and CSS in there.

srkunze:
Lol? IMHO, a name is a atomic string.

JFreegman:
@agilob

> Every Qt application does it.

And if all Tox clients were written in Qt you would have a valid argument.

TheLastProject:
This has been fixed in #1299.

srkunze:
Can be closed, right?

Issue #1224, opened 2015-02-18
[Feature request] ability to change colour of group title / topic


griffondor:
users would like to put the colors in groups of securities

griffondor:
I have a feature request : users want to put the topic in the group of color tox software when creating a lounge titles can not mettres color and fat also
it is possible to have a function like this?

dubslow:
Essayez encore en français svp, je ne peux pas te comprendre

griffondor:
seras t'il possible que lorsqu'on crée un salon dans qtox de mettre les titres en couleurs ? au lieu de noir et blanc

ArchangeGabriel:
@griffondor want to be able to modify the “markup” of group titles, by changing color or using bold.

agilob:
You can do that already, just use HTML:

```
<strong>bold</strong>
<u>underlined</u>
<strong><u>bold underline</u></strong>
```

At least works here, both in status and group description:
![strong](https://cloud.githubusercontent.com/assets/2544251/6400725/eb0a1bf8-bdf2-11e4-8ab3-652517683b9d.png)

---

![strongu](https://cloud.githubusercontent.com/assets/2544251/6400726/eb0c704c-bdf2-11e4-8087-69fb62357b41.png)

---

![red](https://cloud.githubusercontent.com/assets/2544251/6400793/6587b0ca-bdf3-11e4-9c79-d4ffac3e416d.png)

I think HTML should be striped from window title.

griffondor:
the group titles are always in black is annoying, you can not put color?

agilob:
![red1](https://cloud.githubusercontent.com/assets/2544251/6401473/a3189648-bdf8-11e4-809c-7c2185983fe0.png)

griffondor:
it's run by cons if I do not want to underscore in the title, I should be how?

griffondor:
it works for the titles to the color in the group ;)

agilob:
`<font color="red">sdasda</font>`

You can change "red" to any other color name.

griffondor:
Is it be possible to put in émiticones group titles ?

sudden6:
Colors will not be supported by all clients, thus closing. Emoticons are supported now.

PR #1299, opened 2015-02-27
qlabels will not support richtext, plaintext only

### LINKED ISSUE #1298: [Security] Potentially exploitable titles and names
exploiting hurr durr

http://qt-project.org/doc/qt-4.8/qdeclarativejavascript.html#inline-javascript

Please fix asap, since there probably are exploits <128 characters…

thanks to @agilob for providing link

EDIT: related to  #1295

agilob:
removes support for HTML, QML, JS and CSS in username/status labels

this commit closes #1298(bug) #1295(bug) and #1224(feature request)

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -57,7 +57,8 @@ ChatForm::ChatForm(Friend* chatFriend)
     statusMessageLabel->setObjectName("statusLabel");
     statusMessageLabel->setFont(Style::getFont(Style::Medium));
     statusMessageLabel->setMinimumHeight(Style::getFont(Style::Medium).pixelSize());
-
+    statusMessageLabel->setTextFormat(Qt::PlainText);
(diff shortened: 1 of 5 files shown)


redsteakraw:
The html tags have brought joy to many qTox users.  I think this is better set this to a checkbox in the settings to disable richtext.

agilob:
I find this as a feature, but too many users on our groupchat disagreed with me :<

srkunze:
It might be a feature. But the name itself and its styling should be handled, transferred and stored separately as they concerns different concepts.

This way, clients can choose not to get the styling of the name and display text only names.

JFreegman:
This would be a great qTox-only feature to have, but it shouldn't break compatibility with other clients. Like @srkunze says, it should all be handled separately from the name. Maybe you could use a custom packet to transfer the HTML data.

Issue #1288, opened 2015-02-26
Too wide whitespaces in chat


agilob:
Would it be possible to eliminate them or make smaller?

![whitespaces](https://cloud.githubusercontent.com/assets/2544251/6400404/8947ce12-bdf0-11e4-9bd8-cd84fcaad420.png)

Related to #795 (should be closed?) and #867

TheLastProject:
I actually feel the spacing is comfortable, it makes qTox feel calm. Although I'm not going to knock a possible change before trying it.

dcharles525:
White space is important in modern design. I'd maybe make them slightly smaller, but I agree with @TheLastProject that it makes qTox feel pretty good.

agilob:
Due to those whitespaces qTox window is too big to fit on 1/2 width of my screen, I can't make the window smaller, I can't shrink those whitespaces.

If you could provide new setting to remove those whitespaces, it would be great.

Just to compare, Gajim (which I find the most comfortable IM), full chat window vs qTox chat area only:
![gajm](https://cloud.githubusercontent.com/assets/2544251/6404352/e63f92d0-be0e-11e4-84e6-ff74529b8781.png)

zetok:
Whitespaces are needed, but..

> If you could provide new setting to remove those whitespaces, it would be great.

I think that plenty of people would feel that compact mode would be great.

dcharles525:
@zetok I agreed that a setting would be optimal, but you don't see people complaining about Google Hangouts and Skype because they're not compact enough. I'm not saying copy them, just saying that they are successful and no one talks about the white space, look on this very web page, you see a lot of white space, but you don't see people making issues about it. Just a thought.

As for the setting, there is a plethora of settings in qTox already, how much more do people need? Tox as a whole is suppose to be "plug and play" easy, with all these options flooding the settings panel, I fear the basic users will be scared away by the site of it. As for all of the "techy" guys who are testing qTox right now, we don't care about the extra setting.

zetok:
Looks like this issue can be closed after #2013 gets merged.

jessica181920:
left side whitespace isn't huge anymore/padding seems appropriate

Issue #383, opened 2014-10-05
[Feature request] Multiple chat windows


zetok:
User should be able to use many chat windows at once. "Detach" and reattach chat windows to main qTox window.

Idea:

There's main window, which has contact list and beside it chat window. Chat window can be "cloned" into separate window. Both windows would display same content of chat, "cloned" window would show only content of chat with one peer selected at time of cloning. Main window would still allow to freely switch between chats.

When there will be support for minimizing to tray, user should have an option to either minimize just main window, or main window and all "cloned" windows to tray.

msteinbeck:
+1

iav:
And a way to set something like "remember separated status and widow size and position" on a per-contact basis - to not drag all chats by sides on every new program start

Chiitoo:
Teegrins!

Been meaning to suggest something similar for a long time!  Just got registered, finally, so I'll add my thoughts here, as it's almost the same thing (but not quite; let me know if it should be a separate issue).

Not only would I like to be able to have the main-window (contact list) separate from the chat-windows, I would also like to have the option for a tabbed chat-window.  I've used KMess since 2010, which I can get to pretty much just as I want it, and have gotten quite used to it indeed!

I do not want the main-window (that has the contact list) to have a chat-window at all.

Does this seem like a possible thing to do, or should it be left for another project?  (I have the want to start creating this vision myself, but it will probably take at least ages, if not longer, to get to anything useful!)

Thanks!

dubslow:
It's probably possible, but at the moment there are so many features to add and bugs to fix that it's not particularly a priority.

tux3:
Qt being Qt, this should actually be fairly easy to do.
We'd just need to modify GenericChatForm to not add itself to the main layout. (Note that we can't have two cloned widgets displaying the same thing as described in the OP with this method.)

Proof of concept, with just two lines of code changed:
![pic12](https://cloud.githubusercontent.com/assets/5155966/4756185/880f1c2c-5acb-11e4-8ba6-3fab9d19eda6.jpg)

Although, @dubslow is right that this isn't exactly a priority at the moment.

Chiitoo:
I figured it should be relatively simple (Qt being Qt), and was already poking around a bit the other day (well actually I was poking at Qt GUI by nurupo; perhaps I'll take a whack at qTox next!).

[...]

PR #1302, opened 2015-02-27
Allow making main window smaller

### LINKED ISSUE #867: Minimum window size is too large
I can't fit qTox into one corner/quarter of my laptop screen...

agilob:
by hiding list of contacts when width of the windows is equal minimum size, set minimum size to 540

somehow fixes #867 @dubslow ?
qtoxmin2

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -337,7 +337,16 @@ void Widget::resizeEvent(QResizeEvent *event)
 {
     Q_UNUSED(event);
     saveWindowGeometry();
-
+    
+    if(this->width() <= 775) {
+         // width that doesn't break file transfer layout is 535
+        this->setMinimumWidth(540);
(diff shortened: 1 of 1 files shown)


dcharles525:
could also help out with #1288!

agilob:
I had this in mind ;)

zetok:
> by hiding list of contacts

No.
IMHO it should be done properly, i.e. #383

tux3:
I'm not convinced this is great from a usability PoV. I can already see the confused users wondering where their contact list is gone, or making it disappear by accident. It isn't really a common thing.

Zer0-One:
I agree with @zetok in that this should be implemented as described in #383 . As-is, this might just end up confusing people, and might be annoying for people trying to resize the window to fit a particular desktop layout (could be wrong about this, haven't tested this for myself).

PR #1311, opened 2015-02-28
Add image preview on mouseover

### LINKED ISSUE #980: Bigger image preview when hovering over thumbnail
When sending an image to a friend, it shows a small thumbnail. It'd be nice if a bigger preview could be displayed when hovering your mouse over the image, to hopefully allow looking at text in an image without opening the image itself. Fixing #573, however, would probably be a better improvement. Regardless of that, I'd like to post this suggestion/feature request anyway.

TheLastProject:
On placing your mouse onto the image preview in a file transfer, a tooltip will pop up, showing the image. If the image is too large, it is scaled down to make sure it never takes more than 50% of the screen.
qtox_preview

fixes #980

### DIFF
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -26,6 +26,7 @@
 #include <QFile>
 #include <QMessageBox>
 #include <QDesktopServices>
+#include <QDesktopWidget>
 #include <QPainter>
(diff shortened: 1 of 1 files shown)


dubslow:
Fuck uTox, this is amazing

agilob:
that gave me an idea :)

PR #1323, opened 2015-03-02
Enable high DPI support

stqism:
This basically makes things using pixels scale up like they do on a Mac, so on high resolution displays (muh 5k) things are actually readable and you can tell what you're looking at.

In some cases it reduces CPU usage because pixmaps on high DPI displays are horribly inefficient.

Also, _slight_ chance this isn't supported by Linux, but I can't test it because Linux on the desktop is broken ootb.

Depends on Qt 5.4+ so it's behind a define, just enable HIGH_DPI.

Sorry about the commit spam, used the web UI to do it quickly because I don't have my pgp key on hand but actually verified it worked in production™

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -471,6 +471,7 @@ void Widget::onStatusSet(Status status)
 void Widget::setWindowTitle(const QString& title)
 {
     QString tmp = title;
+    /// <[^>]*> Regexp to remove HTML tags, in case someone used them in title
     QMainWindow::setWindowTitle("qTox - " + tmp.remove(QRegExp("<[^>]*>")));
 }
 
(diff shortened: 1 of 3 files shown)


stqism:
Note that it'll look exactly the same for 99.9% of the people using it, so don't actually be afraid to enable it.

zetok:
could you squash those commits?

src/widget/widget.cpp
```diff
@@ -470,7 +470,9 @@ void Widget::onStatusSet(Status status)
 
 void Widget::setWindowTitle(const QString& title)
 {
-    QMainWindow::setWindowTitle("qTox - " + title);
+    QString tmp = title;
+    /// <[^>]*> Regexp to remove HTML tags, in case someone used them in title
+    QMainWindow::setWindowTitle("qTox - " + tmp.remove(QRegExp("<[^>]*>")));
```
tux3, line 475: Please be careful not to send commits unrelated to the PR's title and description, ever. I'll just cherry-pick the last commit this time.

PR #1328, opened 2015-03-03
Move the identity settings tab into its own dedicated profile page (fixes #1012)

### LINKED ISSUE #1012: There's no way to quickly get own Tox ID
It should be accessible in easy, fast way, **max** 2 clicks.
Currently to copy Tox ID one has to go to settings, click on tab and copy ID. **4 clicks if one knows where to look**. IMHO that should be changed. I've tried to think about that problem, but haven't come with best way that would provide global access in easy and user-friendly way (user must know that it's about getting ID)

Some ideas:
1. It could be done like in Venom, i.e., after clicking on green status button beside being able to change status one could copy Tox ID. Handicap is that there are still 2 clicks required, and user is not in any way informed/pointed to that way of copying ID.

\2. There could be added some button under green status button (no idea where else could it be located, since practically no other place is accessible globally all the time). This button would clearly indicate that it would copy Tox ID. Handicap is that there really isn't much of free space beside green button, and it would be kinda outside of mockup looks. On the plus side, there's only 1 click, and (provided that button will be self-explanatory [button design + tooltip]) it would be user-friendly.
# 

Additionally, in user-input fields in chats, when r-clicking there could be added field to menu `Paste my Tox ID`

Any other ideas?
@ItsDuke ?

ovidiusabou:
Make the page directly accessible from the main screen by clicking the avatar or the username. Changing the avatar or the username are now possible only from the profile page.
#1012 for the discussion

This is what it would look like
qtox-profile

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -111,7 +111,7 @@ ProfileForm::ProfileForm(QWidget *parent) :
 ProfileForm::~ProfileForm()
 {
     delete bodyUI;
-    delete head;
+    head->deleteLater();
 }
 
 void ProfileForm::show(Ui::MainWindow &ui)
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -36,6 +36,16 @@
 #include <QBuffer>
 
 
+void ProfileForm::refreshProfiles()
+{
+    bodyUI->profiles->clear();
+    for (QString profile : Settings::getInstance().searchProfiles())
+        bodyUI->profiles->addItem(profile);
+    QString current = Settings::getInstance().getCurrentProfile();
+    if (current != "")
+        bodyUI->profiles->setCurrentText(current);
+}
+
 ProfileForm::ProfileForm(QWidget *parent) :
     QWidget(parent)
 {
@@ -317,4 +327,10 @@ void ProfileForm::enableSwitching()
         bodyUI->loadButton->setEnabled(true);
         bodyUI->newButton->setEnabled(true);
     }
+}
+
+void ProfileForm::showEvent(QShowEvent *event)
+{
+    refreshProfiles();
+    QWidget::showEvent(event);
 }
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -15,9 +15,11 @@
 */
 
 #include "src/core.h"
+#include "src/nexus.h"
 #include "ui_identitysettings.h"
-#include "identityform.h"
+#include "profileform.h"
 #include "src/widget/form/settingswidget.h"
+#include "src/widget/maskablepixmapwidget.h"
 #include "src/misc/settings.h"
 #include "src/widget/croppinglabel.h"
 #include "src/widget/widget.h"
@@ -26,13 +28,16 @@
 #include "src/misc/style.h"
 #include <QLabel>
 #include <QLineEdit>
+#include <QGroupBox>
 #include <QApplication>
 #include <QClipboard>
 #include <QInputDialog>
 #include <QFileDialog>
+#include <QBuffer>
 
-IdentityForm::IdentityForm() :
-    GenericForm(tr("Identity"), QPixmap(":/img/settings/identity.png"))
+
+ProfileForm::ProfileForm(QWidget *parent) :
+    QWidget(parent)
 {
     bodyUI = new Ui::IdentitySettings;
     bodyUI->setupUi(this);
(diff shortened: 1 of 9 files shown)


zetok:
Introduces regression, it's no longer possible to change profile

zetok:
Other profile-related stuff (except for creating new Tox ID) also seems to be broken

TheLastProject:
I haven't tested it, but I'm a bit unsure about this. I mean, yes, it makes it easier to find your Tox ID and so, but I quite like being able to edit my nick and status with a single click...

ovidiusabou:
@zetok can you please take another look? sorry for being so careless. It takes some time until the profile list is populated, but this seems to be related to the settings save/load mechanism rather than the profile form.

@TheLastProject I could try to do it differently if you have a better idea. I completely agree that changing nickname and avatar is more involved now, but I still think it's worth doing it (I wouldn't have bothered making the change of course)

ProMcTagonist:
This breaks one-click editing of user, status message, and avatar, and still requires multiple clicks. A simpler, better solution would be to just make the Identity tab the first settings tab. It's not less "intuitive" than this, because neither make any instinctual sense. They'll learn after the first click either way.

dubslow:
> just make the Identity tab the first settings tab

That's an excellent idea, if you guys vote to reject this design change.

I think I'm still in favor of this PR though.

zetok:
@TheLastProject 

> but I quite like being able to edit my nick and status with a single click...

When was the last time when you have changed your nick?

@ProMcTagonist 

> (…) breaks one-click editing of (…) status message (…)

What do you mean?

> A simpler, better solution would be to just make the Identity tab the first settings tab. It's not less "intuitive" than this, because neither make any instinctual sense. They'll learn after the first click either way.

Nope.

Is your average $someone_who_cannot_into_computers randomly exploring capabilities of software?
no.
Would they click on settings button?
No.
Is clicking on your profile name/avatar widely recognised as something that will lead to opening your profile?
Yes. (all the websites, other software as well)

This leads to question how possibly users could learn, provided that they don't click on settings?
(please keep in mind that settings aren't there so that people have to go into them, but rather qTox should provide sane defaults, so that people don't have to go into settings)

Answer is: users won't learn when profile stuff is in settings. _Or_ they will learn, after fighting with an unintuitive software.

On the other hand, users **will have to** click on at least their name (to change it from default user name), which would open profile (consistency with muh web), and there they could learn about most important things in qTox, like Tox ID and multiple profiles.

Whereas if profile was kept in settings, users would have to go on exploring on their own to discover that.

# 

@ovidiusabou works nicely now, except for 1 minor thing. When switching profile, qTox "resets" itself to "Add friends" page, instead of keeping "Profile" open. I think that it should keep "Profile" page open?

@dubslow um, if it's not clear yet, I support idea behind this PR (i.e. if code and behaviour is good enough, IMHO it should be merged).

Zer0-One:
+1 for this PR from me as well

Edit: You probably didn't need to bother changing every instance of "Identity" to "Profile"

tux3:
> just make the Identity tab the first settings tab

I also agree with that, regardless of this. But I'd like more opinions on this PR, it's a pretty big, visible change.

ovidiusabou:
@zetok I fixed the section switch issue, the profile page remains open when switching the profile

ProMcTagonist:
@zetok I've been convinced, and I +1 this if the implementation won't feel weird. There should be some effect, because right now, it's just some items on a black background with no indication of unification or clickability.

zetok:
@ovidiusabou :3

https://gist.github.com/zetok/fa7130214563c9c380df

Also, after profile is created, new profile is not listed in profile list.

ovidiusabou:
@zetok The app doesn't crash anymore when closing it. I think the profile list bug is unrelated to this PR because I can reproduce it with a downloaded build. I will take a look though...

ovidiusabou:
The list is now refreshed, but only after the new id gets online, that's when it becomes accessible in the core and the name of the profile seems to be defined by core, so it's not smart to preemptively add it to the list, but I'm very new to this so maybe somebody else has a better idea

zetok:
Well, there is still problem with new profile not showing up in list upon creation, but that's a problem with master qTox, not this PR…

@tux3 I don't think that you'll get more input on that before introducing functionality to people.
PR is clearly an improvement, so the only obstacle is initially people who were used to the old way of getting to Profile page.. and even then it's a tiny one.

So if code is good enough, merge pls?

src/widget/form/profileform.cpp
```diff
@@ -46,42 +70,59 @@ IdentityForm::IdentityForm() :
     
     bodyUI->toxGroup->layout()->addWidget(toxId);
 
+    profilePicture = new MaskablePixmapWidget(this, QSize(64, 64), ":/img/avatar_mask.png");
+    profilePicture->setPixmap(QPixmap(":/img/contact_dark.png"));
+    profilePicture->setClickable(true);
+    connect(profilePicture, SIGNAL(clicked()), this, SLOT(onAvatarClicked()));
+    QHBoxLayout *publicGrouplayout = qobject_cast<QHBoxLayout*>(bodyUI->publicGroup->layout());
+    publicGrouplayout->insertWidget(0, profilePicture);
```
dubslow, line 78: You should modify the .ui file rather than doing this manually.

dubslow:
Any reason why you removed the inheritance from GenericForm? That changes very little, other than specifying the image that goes in the header (which should still be kept IMO).

Edit: Meh, the image in the header was handled from the settingswidget.cpp class, not the genericform class as I misrecalled.

src/widget/form/profileform.cpp
dubslow, line 78: I take this back. Although feasible, by modifying the MaskablePixmapWidget to not require constructor arguments, the main profile picture is also handled this way. Oh well...

PR #1331, opened 2015-03-04
Ask user to move to Applications folder on launch

stqism:
Works perfectly, I think it looks neat enough.

screen shot 2015-03-03 at 7 26 08 pm

(Also, I want to drop the dmg for a self-extracting zip because it's nicer to use and takes less work and this is the first step)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -132,7 +132,7 @@ int main(int argc, char *argv[])
         AskInstall.setIcon(QMessageBox::Question);
         AskInstall.setWindowModality(Qt::ApplicationModal);
         AskInstall.setText("Move to Applications folder?");
-        AskInstall.setInformativeText("I can move myself to the Applications folder, keeping your downloads folder less cluttered\r\n");
+        AskInstall.setInformativeText("I can move myself to the Applications folder, keeping your downloads folder less cluttered.\r\n");
         AskInstall.setStandardButtons(QMessageBox::Yes|QMessageBox::No);
         AskInstall.setDefaultButton(QMessageBox::Yes);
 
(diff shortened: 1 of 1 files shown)


dubslow:
The linewrapping looks funny, also period please

dubslow:
Did you consider using GUI::askQuestion?

As for your earlier comments about issues with qApp->exit(), it may not be possible to do magic qApp stuff from main... just a thought

src/main.cpp
```diff
@@ -29,11 +29,16 @@
...
 #include <QMutexLocker>
+#include <QProcess>
 
 #include <sodium.h>
 
 #include "toxme.h"
 
+#include <unistd.h>
+
+#define EXIT_UPDATE_MACX 218 //We track our state using unique exit codes when debugging
```
zetok, line 40: Where are those codes listed?

stqism:
@dubslow AskQuestion isn't real. Also, Qt is pretty dumb with formatting and totally ignores resizing hints so this is by far the best it's going to be. Added the period btw.

Also also also, fixed those issues I was having.

src/main.cpp
```diff
@@ -119,6 +124,44 @@ int main(int argc, char *argv[])
...
+            QProcess *sudoprocess = new QProcess;
+            QProcess *qtoxprocess = new QProcess;
+
+            QString bindir = qApp->applicationDirPath();
+            QString appdir = bindir;
+            appdir.chop(15);
+            QString sudo = bindir + "/qtox_sudo rsync -avzh --remove-source-file " + appdir + " /Applications/qtox.app";
+            QString qtox = "open /Applications/qtox.app";
+
+            if (fork() != 0) { //uTox grade cheap hack
```
tux3: This comment is not really needed.

src/main.cpp
```diff
@@ -119,6 +124,44 @@ int main(int argc, char *argv[])
...
+
+        QMessageBox AskInstall;
+        AskInstall.setIcon(QMessageBox::Question);
+        AskInstall.setWindowModality(Qt::ApplicationModal);
+        AskInstall.setText("Move to Applications folder?");
+        AskInstall.setInformativeText("I can move myself to the Applications folder, keeping your downloads folder less cluttered.\r\n");
+        AskInstall.setStandardButtons(QMessageBox::Yes|QMessageBox::No);
+        AskInstall.setDefaultButton(QMessageBox::Yes);
+
+        int AskInstallAttempt = AskInstall.exec(); //Check if user sucks
```
tux3: Let's not start insulting our users in the source code, shall we :) ?

stqism:
@tux3 Changed.

src/main.cpp
```diff
@@ -119,6 +124,44 @@ int main(int argc, char *argv[])
...
+        int AskInstallAttempt = AskInstall.exec(); //Actually ask the user
+
+        if (AskInstallAttempt == QMessageBox::Yes) {
+            qDebug() << "Installing";
+            QProcess *sudoprocess = new QProcess;
+            QProcess *qtoxprocess = new QProcess;
+
+            QString bindir = qApp->applicationDirPath();
+            QString appdir = bindir;
+            appdir.chop(15);
```
tux3, line 148: This is really hackish, but I guess it'll have to do for now.

PR #1332, opened 2015-03-04
Fix quoting; everything after '>' is a quote

zetok:


ghost:
+1

stqism:
This is a regression.

suhr:
@stqism Indeed it is.

ghost:
No it's not.

tux3:
This PR is pointless if I'm going to get another PR two days from now to revert it. Let's try to find an actual solution instead, like adding an option in the settings, or maybe making a plugin system so you can have all sorts of crazy formatting only yourself can see.

zetok:
@tux3 This is an actual solution. At this point, qTox not only is inconsistent with its own behaviour, but also with what other Tox clients do.

tux3:
@zetok Consistency by itself is not enough. Removing quoting support entirely would be just as if not more consistent than considering everything after '>' a quote.
Obviously, not everyone agrees that everything after '>' should be a quote (emoticons are an issue in particular), but many people still want quoting support. I would rather not just ignore either half of the problem, but instead try to find a solution that addresses the issue in its entirety.
A simple option in the settings might be a solution, this is something we could discuss.

optimumtact:
Stop letting the core team push you around and grow a spine.

tux3:
@optimumtact would you care to expand on that ?

zetok:
> Removing quoting support entirely would be just as if not more consistent than considering everything after '>' a quote.

Nope, it would be inconsistent with other Tox clients.

> try to find a solution that addresses the issue in its entirety.

Issue here is that people disagree. So.. how are you going to solve it?

ProMcTagonist:
+1

mindcombatant:
actual solution would be to have an option in the settings finally, no one will complain once there's

<> quote greentexting ON
<> quote greentexting OFF
<> quote greentexting CUSTOM [type in your own regex]

or something similar in the settings

tsdko:
With ON as the default, obviously.

dcharles525:
Awesome, more settings to bloat up qTox. No one uses greentext, Tox is being written to be simple and easy for the common person, not to have a million settings.

4DA:
Option would be fine, just hide it under "Advanced settings".

ayyyy:
Keep greentext, dcharles525 can %*#$ off with his generalizations.

dubslow:
Language please :)

zetok:
@dubslow just saying; "language" wouldn't be needed if problem was solved..

TheLastProject:
Honestly, I think greentext is pretty okay, but if we can't agree on what should be green and what not, let's just get rid of greentexting completely. Nobody seems to be happy with any solution, and not using greentexting at all is the most consistent and most professional looking choice.

zetok:
@TheLastProject 

> Nobody seems to be happy with any solution

Not "nobody". Also, you're not being specific when it comes to what you mean by "solution"…

ProMcTagonist:
> Nobody seems to be happy with any solution

What? Everyone is happy with some solution.

> not using greentexting at all is the most consistent and most professional looking choice.

A completely unrelated and baseless statement, pushing your own personal preference, sprinkled with random feel-good words.

kushagrasurana:
How about quoting only from context menu. User selects some text, right-click then click on "Quote".
We can have shortcut for the action.

ProMcTagonist:
That sounds terrible to use and also very far removed from what we're talking about here.

dcharles525:
@ayyyy Didn't mean to piss in your coffee mate! I guess make it a setting in advanced settings so everyone is happy?

stal888:
Tox will make greentexting popular (again).

This could be our quirky killer feature, but only if all clients support greentext.

optimumtact:
Yes agreed, all clients support by default, on by default
lame ass weeb emoticons can go jump

JFreegman:
> Not supporting greentext

Into the trash it goes

Zer0-One:
@JFreegman 
trash

stal888:
@Zer0-One 
no

ayyyyyyyyyyy:
Could you please merge PR?

zetok:
still broken

↑ qTox users without this patch still have to suffer broken quoting

Edit: it still applies cleanly against master qTox..

ProMcTagonist:
Please patch this bug

zetok:
Since it doesn't apply cleanly on master any longer, I've created new PR: #1732

PR #1340, opened 2015-03-05
Resize images. Fixes #1204

### LINKED ISSUE #1204: New message circle is misaligned
When you have one or more unread messages from a contact, their status circle gets pushed slightly to the left due to the "new message" circle around it.
misalign

TheLastProject:
I'm not quite happy with dot_groupchat_*, as they were 17x17px, which makes it impossible to size nicely like the others, but this is still better than the other pull request I'd think. I can't test currently, but I don't think 18x18px will break it if 16x16px worked.


TheLastProject:
@urras Did you even test the previous one? Because I opened qTox and I instantly see that the online indicator next to my own nick in the top left is insanely small :/

Will fix that up today if nobody beats me to it...

urras:
@TheLastProject I briefly checked over https://github.com/TheLastProject/qTox/commit/2e44de36a4d96107cbb43e086c9a5b5931eeaee9, and everything looked fine to me. Maybe I'm just having some trouble seeing...

Issue #1119, opened 2015-02-07
  Ending a call while call window is in OSX fullscreen mode results in an unclosable fullscreen desktop


mckeesh:
The setup:
Machines tested on were a late 2013 iMac and a mid 2013 Macbook air, both running OS X Yosemite 10.10.2.
The Tox version does not seem available anywhere, but both Tox applications were downloaded tonight (2/6/15)

To reproduce:
1. Make a call between two OS X machines (might work with one of them being a Windows machine). 
2. Make the call window full screen on one of the OS X machines. 
3. End the call before taking the call window out of full screen mode.

The result:
The resulting screen is completely black, and the escape key will not close it. The only way that I have been able to close it is by quitting qTox completely.

zetok:
Since 03329d81714dcb526ab837e80d953dce4ccada8d is in place, does that mean that this issue is fixed?

Additionally, qTox doesn't have separate window for video calling, so..

zetok:
No response for over a month, assuming it was fixed, thus closing.

mckeesh:
My apologies, the notification for this somehow slipped by. I made a pull request here:
https://github.com/tux3/qTox/pull/1343

But it being my first open source contribution, I was having trouble getting my commits squashed. It looks like the relevant change was cherrypicked, so it seems like it's probably all good. Sorry for the confusion.

zetok:
> My apologies, the notification for this somehow slipped by.

Well, sometimes github has problems with delivering email notifications (aka doesn't deliver them).

PR #1343, opened 2015-03-06
Fix for OS X full screen bug on call end (issue #1119)

mckeesh:
It isn't beautiful, but it is a workaround for an issue that seems to already be known by Qt.


zetok:
If it's a workaround, would you mind clearly stating that in comments for relevant code, along with link to the known issue?

Also, could you squash those commits?

dubslow:
Squash those commits, and see if you can also remove that spurious merge from the history, and then force push

dubslow:
I cherry picked the relevant commit (though forgot to remove the unneeded include).

PR #1344, opened 2015-03-06
Removes the flickering "Busy" notification when resized

### LINKED ISSUE #1213: Chat form flickers and says "Busy" during any resizing
Very ugly and distracting.

kushagrasurana:
fixes #1213 .

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -456,11 +456,22 @@ void ChatLog::startResizeWorker()
             workerAnchorLine = visibleLines.first();
     }
 
+    // switch to busy scene displaying the busy notification if there is a lot
+    // of text to be resized
+    int txt = 0;
+    for(ChatLine::Ptr line : lines)
+    {
+        if(txt>500000)
+            break;
+        for(ChatLineContent* content : line->content)
+            txt += content->getText().size();
+    }
+    if(txt>500000)
+        setScene(busyScene);
+
     workerLastIndex = 0;
     workerTimer->start();
 
-    // switch to busy scene displaying the busy notification
-    setScene(busyScene);
     verticalScrollBar()->hide();
 }
 
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -123,7 +123,7 @@ ChatMessage::Ptr ChatMessage::createBusyNotification()
     ChatMessage::Ptr msg = ChatMessage::Ptr(new ChatMessage);
 
     // TODO: Bigger font
-    msg->addColumn(new Text(QObject::tr("Busy..."), Style::getFont(Style::ExtraBig), false, ""), ColumnFormat(1.0, ColumnFormat::VariableSize, ColumnFormat::Center));
+    msg->addColumn(new Text(QObject::tr("Resizing"), Style::getFont(Style::ExtraBig), false, ""), ColumnFormat(1.0, ColumnFormat::VariableSize, ColumnFormat::Center));
 
     return msg;
 }


zetok:
I am against that.

If something, there should be added a timer, so that instead of flickering, `Busy` would be displayed after `1s` since resizing operation begun.

Also, `Busy` should be renamed to `Resizing`.

zetok:
Also, please squash/ammend your commit when you don't have real changes, or it can be put in 1 commit.

kushagrasurana:
nice idea. i will add the timer and squash my commits to one.

kushagrasurana:
I implemented the timer, but it was not good.
 When number of lines in chatlog is high (more than 2000), resizing makes the UI unresponsive and the "Resizing" notification takes more than 1s to appear.

I have another solution, we can show "resizing" notification (without any wait) if chat lines are more than 1000, and wait for resizing to finish .. because resizing takes no time if chat lines are less than 1000

zetok:
Whatever works the best. I am simply against removing `Resizing`

ProMcTagonist:
What is wrong with the chatform that makes it unable to resize smoothly, without hiding behind something? Hiding the window contents on resize is Windows 95 stuff.

kushagrasurana:
Chatform will resize smoothly now without any flickering notification. The <code>Resizing</code> notification will appear when resizing a long chat.

src/chatlog/chatlog.cpp
```diff
@@ -456,11 +456,18 @@ void ChatLog::startResizeWorker()
             workerAnchorLine = visibleLines.first();
     }
 
+    // switch to busy scene displaying the busy notification if there is a lot
+    // of text to be resized
+    int txt = 0;
+    for(ChatLine::Ptr line : lines)
```
dubslow, line 462: Why not short circuit these at 500,000 or whatever limit? If that chat is indeed very large, then just counting it could take a while.

ProMcTagonist:
The chat should not ever need to cover itself up in order to resize. It should only be displaying as much as it can handle, with the majority of the history stuffed under a scrollback.

src/chatlog/chatlog.cpp
kushagrasurana, line 462: I had thought of it, and so i calculated time taken for counting chatlog of many sizes.

```
    long long t =  QDateTime::currentMSecsSinceEpoch();
    int txt = 0;
    for(ChatLine::Ptr line : lines)
        for(ChatLineContent* content : line->content)
            txt += content->getText().size();
    qDebug()<<(QDateTime::currentMSecsSinceEpoch()-t)<<" txt="<<txt;
    if(txt>500000)
        setScene(busyScene);
```

and the results were very good. When resizing took 5 seconds, the time taken for counting was 0 milliseconds.
This was the qDebug message i got from the above code.
<code>0  txt= 2378625</code>
I never got time taken more than 0 milliseconds.

dubslow, line 462: Interesting, that honestly makes me suspicious if the only number you were ever able to see is 0. What are your system's specs? 

I'd still feel safer quitting the loop once it hits the limit. There's no reason not to, and then at least it's bounded no matter the chat size.

dubslow:
@ProMcTagonist You are of course correct, but in the meanwhile this PR is a slight improvement.

src/chatlog/chatlog.cpp
kushagrasurana, line 462: Yeah quitting the loop is right, saves a lot of computations(if chat is very long). So i have amended the commit.
Also i experimented with this code

```
    long long t = QDateTime::currentMSecsSinceEpoch();
    for (int i=0; i<500000; i++);
    qDebug()<<QDateTime::currentMSecsSinceEpoch()-t;
```

Sadly i got 1 or 2 ms for <code>i < 500000</code>. And 15 to 17 ms for <code>i < 500000</code>. 

Specs - i7 4702MQ, 8 gb ram and my qTox runs on ubuntu 14.04.

PR #1351, opened 2015-03-07
Qrcode as HTML tooltip in Identity tab

### LINKED ISSUE #663: Show QR code of tox id
Helps with adding friend on mobile, it seems no know desktop client support that

agilob:
I had to use external lib `qrencode` for this, I added `-lqrencode` in qtox.pro, but I'm not sure if I did it right, please check it.
I used a snippet from SO, modified it a bit, so it doesn't show a qwidget but converts qimage to base64 html, png format, that image is shown as tooltip :)

The tooltip is probably too big, 350x350 is hardcoded size.

closes https://github.com/tux3/qTox/issues/663
My toxid: 7119F8F3B55FCA07F78CC78003F9F9FCE70B2C0AEF3649F3A84C87EA56BE927F3A9119A7992E
The same is decoded by barcodescanner on Android.

qrcodetox

Thanks @TheLastProject for [the idea](https://github.com/tux3/qTox/pull/1311) :)


TheLastProject:
I wonder if it'd be better to mark it like a Tox URI, so, make it start with `tox:`, followed by the Tox ID?

dubslow:
Wouldn't it be more useful to save it as a file? The whole point is that you can share the qr code, not use it yourself.

@TheLastProject We can do this for the image preview: scale the QImage, and then write it to the html tooltip as base64.

zetok:
@agilob please update also `INSTALL.md`. Updating just qtox.pro isn't enough.
Also, what @TheLastProject said, `tox:TOX_ID` should be used instead of just `TOX_ID`

@dubslow 

> Wouldn't it be more useful to save it as a file?

Why putting it in terms of `or`, rather than saying that beside toolTip there could be a button (or by r-click) that would allow to save it to file?

agilob:
> Wouldn't it be more useful to save it as a file? The whole point is that you can share the qr code, not use it yourself.

and you would send a .png file instead toxID string?

dubslow:
Or you could import the .png into whatever imaging program you use, maybe to put on posters or something. I guess there's no reason to remove the tooltip, but I still don't see the reason to add it in the first place.

ProMcTagonist:
It's completely useless as a tooltip. It should be permanently displayed near the ID in the Identity tab with a little explanation. Thanks for finally adding QR codes.

zetok:
@agilob dependencies are still not updated in INSTALL.md

INSTALL.md
```diff
@@ -15,7 +15,7 @@
 | OpenCV       | >= 2.4.9    | core, highgui, imgproc                            |
 | OpenAL Soft  | >= 1.16.0   |                                                   |
 | filter_audio | most recent |                                                   |
-
+| qrcode       | most recent |                                                   |
```
zetok: https://pypi.python.org/pypi/qrcode ?

Also, why would you put a requirement on most recent? Does your code use some functionality that is only available in most recent version?

agilob:
Any ideas where the button to save QR should be and how look like?

ProMcTagonist:
Near the permanently-displayed QR code.

agilob:
> permanently-displayed

it's useless. how often do you want to scan your own QRcode? It's a waste of space.

ProMcTagonist:
Space that's being used for what? How are they supposed to discover hovering over their ID to get a QR code on the screen?

NoxNoctium:
For a mostly on desktops used client, IMO space is definitely not the limiting factor.
image
If the smallest possible window size is used, there's already a vertical scrollbar, I fail to see how a permanently shown QR-code would break anything here.
image
And with actually useable mobile clients I'd probably go for the QR-Code in most cases anyways.

agilob:
Do you think it looks better now?
formats4

ProMcTagonist:
Yes, but it doesn't need to be quite that big. It could take up half the width of the Profiles widget above it, with a little explanation and save button next to it.

Zer0-One:
If it was half the width, it should be centered.

ProMcTagonist:
I was picturing something like this, @Zer0-One 
untitled

NoxNoctium:
I'd suggest to display the QR-code in the same tab as the Tox ID but otherwise it looks great.

ProMcTagonist:
@markus1540 +1 for it being in the same box

agilob:
What would copy button do?
formats5

dubslow:
Can you not copy images to the clipboard now on modern OSs?

agilob:
@dubslow  do you mean copying bitmap?

dubslow:
After running 

```
sudo apt-get install qrencode
[sudo] password for bill: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
The following packages were automatically installed and are no longer required:
  lib32z1 libmpg123-0:i386 libusb-0.1-4:i386 libxcomposite1:i386 libxcursor1:i386 libxslt1.1:i386 lsb-core lsb-security pax
Use 'apt-get autoremove' to remove them.
The following extra packages will be installed:
  libqrencode3
The following NEW packages will be installed:
  libqrencode3 qrencode
0 upgraded, 2 newly installed, 0 to remove and 5 not upgraded.
Need to get 68.9 kB of archives.
After this operation, 185 kB of additional disk space will be used.
Do you want to continue? [Y/n] 
Get:1 http://mirror.metrocast.net/linuxmint-debian/latest/ testing/main libqrencode3 amd64 3.4.2-1 [37.7 kB]
Get:2 http://mirror.metrocast.net/linuxmint-debian/latest/ testing/main qrencode amd64 3.4.2-1 [31.2 kB]
Fetched 68.9 kB in 0s (191 kB/s)                         
Selecting previously unselected package libqrencode3:amd64.
(Reading database ... 333534 files and directories currently installed.)
Unpacking libqrencode3:amd64 (from .../libqrencode3_3.4.2-1_amd64.deb) ...
Selecting previously unselected package qrencode.
Unpacking qrencode (from .../qrencode_3.4.2-1_amd64.deb) ...
Processing triggers for man-db ...
Setting up libqrencode3:amd64 (3.4.2-1) ...
Setting up qrencode (3.4.2-1) ...
Processing triggers for libc-bin ...
```

I still get this error: 

```
src/misc/qrwidget.cpp:6:26: fatal error: qrcodewindow.h: No such file or directory
 #include "qrcodewindow.h"
                          ^
compilation terminated.
make: *** [qrwidget.o] Error 1
```

Despite the apt-get command, the file doesn't exist on my system. Nor does `qrencode.h`. And by the way, system libraries should be included with <>, while "" is for local project files.

Looks like on Debian based systems you need `libqrencode-dev` as well as `qrencode`.

dubslow:
After installing the dev package, I still don't have a qrcodewindow.h on my system, but I do have qrencode.h and including that was sufficient for compilation.

Can you comment one way or the other about qrcodewindow.h?

zetok:
With media-gfx/qrencode-3.4.3 installed:

```
src/misc/qrwidget.cpp:6:26: fatal error: qrcodewindow.h: No such file or directory
 #include "qrcodewindow.h"
                          ^
compilation terminated.
```

agilob:
It's a compilation error, local file was missing.
"#include "qrcodewindow.h" it's actually never used. It's from months ago when I tried implementing QRcode for the first time and QRcode was in separate qwidget.

dubslow:
@agilob What about the format? Shouldn't it be QR_MODE_AN, rather than the 8-bit mode? (It was never the 8 bit data, it was always the ToxID in _text_ form.)

dubslow:
No, it shouldn't be, any mode other than _8 or _KANJI is an error according to docs (that's somewhat confusing)

ProMcTagonist:
The "This QR code contains..." string is giving me a horizontal scrollbar. Could it be made to wrap instead?

agilob:
@ProMcTagonist just noticed the same problem, in my repo on qrcode branch it doesn't happen, in tux3/qtox I have horizontal scrollbar too.

agilob:
@ProMcTagonist do you still have this problem? After recent PRs I don't.

ProMcTagonist:
Nope, wraps now, no more horizontal scrollbar :)

agilob:
Probably that PR with disabling QComboboxes changed something. I don't know.

ProMcTagonist:
Weird, because I still get a horizontal scrollbar in the General settings form at the smallest window size

dubslow:
I made a one line change that set the QR text to wrap. I didn't make any changes to the combo boxes, other than merging the related PRs.

PR #1359, opened 2015-03-09
Screenshot file transfer

TheSpiritXIII:
Menu was added under the "attach" button that has an option to create a screenshot of the desktop and send that. The screenshot is created as a temporary file and sent.

To accomdate for the new menu, I increased the size of the "attach" button in order to display an arrow. I completely redid the smiley icon too.


src/widget/form/chatform.cpp
```diff
@@ -879,6 +889,52 @@ void ChatForm::loadHistory(QDateTime since, bool processUndelivered)
...
+    }
+
+    if (result == QDialog::Accepted)
+    {
+        QScreen *screen = QGuiApplication::primaryScreen();
+        if (screen)
+        {
+            // Have to wait for the rubber band to disappear.
+            // Using flush, sendPostedEvents and processEvents does not work.
+            QTimer::singleShot(100, this, [=]()
```
dubslow: This is quite the hack.... I'm not too comfortable with this, it's magical and literally just praying and hoping that it works.

src/widget/form/chatform.cpp
```diff
@@ -879,6 +889,52 @@ void ChatForm::loadHistory(QDateTime since, bool processUndelivered)
...
+
+                if (!file.open())
+                {
+                    QMessageBox::warning(this, tr("File not read"), tr("qTox wasn't able to save the screenshot"));
+                    return;
+                }
+                file.setAutoRemove(false);
+                // QPixmap is copy-on-write. We have to copy to crop anyway.
+                (screen->grabWindow(0).copy(region)).save(&file, "PNG");
+                if (file.isSequential())
```
dubslow: Uh... why would your temporary file ever be sequential? We're not sanitizing user input here, I don't see the purpose

src/widget/tool/screenshotdialog.h
```diff
@@ -0,0 +1,51 @@
+/*
+    Copyright (C) 2014 by Project Tox <https://tox.im>
```
zetok: >living in the past

src/widget/form/chatform.cpp
dubslow: I don't see any such timers in the Qt example you linked me to, only one that delays the screenshot by a user-controlled value.

zetok:
broken ?

zetok:
black future
↑ why there's a black something when I want to take a screenshot?

Edit: almost forgot: after testing, I don't like that drop-down menu.

Edit2: Actually, that black box is the only "area" where screenshot can be taken…

src/widget/form/chatform.cpp
TheLastProject: I can't say I like the idea of timeouts but, well, [even KSnapShot seems to do it](https://projects.kde.org/projects/kde/kdegraphics/ksnapshot/repository/revisions/master/entry/regiongrabber.cpp#L44), assuming I'm not misunderstanding their code. Still, worth some research to figure out if there isn't a better way, because there really should be.

ProMcTagonist:
The new buttons look too high in @zetok's screenshot. 

Is the [attach file][dropdown for others] arrangement concrete?

TheSpiritXIII:
This is how it looks like on my end:
Screenshot

I'm investigating all issues. I believe that @zetok's issue with the black screen is because his window manager does not support compositing as described in [Qt's documentation](http://doc.qt.io/qt-5/qwidget.html#creating-translucent-windows). I will look for alternatives.

tux3:
This does not rebase cleanly on master, and seems to have some more fundamental issues. I'm closing this for now, feel free to send an updated version.

PR #1363, opened 2015-03-11
Allow writing messages in a lonely groupchat

### LINKED ISSUE #1246: Don't actually send message when alone in a groupchat
Because toxcore doesn't like it and thus it confuses users (why isn't my message sent?). Just add them to the log and pretend we sent them. We're alone, anyways.

Flakebi:
This allows users to write messages if they are alone in a groupchat, before they were rejected because there is noone connected. This should fix #1246.

### DIFF
--- a/src/widget/form/groupchatform.cpp
+++ b/src/widget/form/groupchatform.cpp
@@ -101,11 +101,9 @@ void GroupChatForm::onSendTriggered()
     msgEdit->setLastMessage(msg);
     msgEdit->clear();
 
-    bool isAction = msg.startsWith("/me ");
-
     if (group->getPeersCount() != 1)
     {
-        if (isAction)
+        if (msg.startsWith("/me "))
         {
             msg = msg.right(msg.length() - 4);
             emit sendAction(group->getGroupId(), msg);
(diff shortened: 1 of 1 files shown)


src/widget/form/groupchatform.cpp
```diff
@@ -101,12 +101,22 @@ void GroupChatForm::onSendTriggered()
...
+            emit sendMessage(group->getGroupId(), msg);
+    }
+    else
     {
-        msg = msg.right(msg.length() - 4);
-        emit sendAction(group->getGroupId(), msg);
-    } else {
-        emit sendMessage(group->getGroupId(), msg);
+        QDateTime timestamp = QDateTime::currentDateTime();
+        addSelfMessage(msg, isAction, timestamp, true);
```
dubslow: Insignificant style question: why not just single line this?

src/widget/form/groupchatform.cpp
```diff
@@ -101,12 +101,22 @@ void GroupChatForm::onSendTriggered()
     msgEdit->setLastMessage(msg);
     msgEdit->clear();
 
-    if (msg.startsWith("/me "))
+    bool isAction = msg.startsWith("/me ");
+
+    if (group->getPeersCount() != 1)
+    {
+        if (isAction)
```
dubslow: Also why not just put the test here instead of storing it?

Flakebi:
Thanks for your suggestions, I updated my code.
I stored `isAction` in a separate variable because it's used two times, I just thought it was cleaner.

dubslow:
Ah, I didn't notice its secondary use, whoops. At least it's still only executed once.

PR #1366, opened 2015-03-11
Support for theming qTox tray and window icons

ghost:
This patch adds ability to override qTox window and tray icons with ones from user's desktop theme. Some people prefer keeping tray icons consistent to achieve best looks.

Following theme icons used:
- qtox: general window icon which is also used in taskbar
- qtox-online, qtox-offline, qtox-busy, qtox-away, qtox-invisible, qtox-event: tray icons

If theme icon is not available then default built-in icon will be used.
## 

I noticed window (and thus taskbar) icon is static, default tox icon. I would like to propose/discuss making window/taskbar icon identical to tray icon. Reason is that some people may choose to not use tray icon, but keep application always in taskbar. Something that recent windows versions endorse. However currently such choice results in degraded usability because user can not see his own status. This is mostly a problem when user wishes to know if he is away, but does not want to break auto-away status by interacting with computer. If such change is acceptable please drop a note here and i will implement it. It would also remove need of "qtox" theme icon for window/taskbar leaving us with remaining six status icons.

### DIFF
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -331,7 +331,7 @@ void SystemTrayIcon::setVisible(bool newState)
     }
 }
 
-void SystemTrayIcon::setIcon(QIcon &&icon)
+void SystemTrayIcon::setIcon(QIcon &icon)
 {
     if (false);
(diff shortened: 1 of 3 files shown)


dubslow:
I see no reason not to.

Issue #1330, opened 2015-03-04
Visual glitch when renaming group from right-click


ProMcTagonist:
![group rename glitch](https://cloud.githubusercontent.com/assets/4358079/6477544/9036e158-c1d9-11e4-901c-1b5bf30811a4.png)

Reproduce:
1. Start or join group
2. Deselect the group
3. Right click > Set title...
4. Rename group

It disappears when the mouse moves, or when it gets moused over again if the cursor was moved away while the rename dialog was open.

kushagrasurana:
Unable to reproduce on my ubuntu. is it fixed?

ProMcTagonist:
Still happening to me on a fresh git pull.

agilob:
Same here, but it looks a bit differently
![qtoxgroupchat](https://cloud.githubusercontent.com/assets/2544251/6546151/f565a798-c59f-11e4-884c-eb8815c203e9.png)

zetok:
@agilob it looks differently because your contact list isn't big enough to show a scrollbar, am I right?

dubslow:
Also unable to reproduce

zetok:
@dubslow right. well, lets see differences

@ProMcTagonist @kushagrasurana @agilob @dubslow mind filling relevant part of https://github.com/tux3/qTox/wiki/Useful-bug-reports#template ?

qTox: 2ea3d73756b5e846dac461f9762eaca12dad8134

```
Info:
OS: Gentoo Linux x86_64
Kernel: 3.18.7
Xorg: 1.16.4
nouveau: 1.0.11
Qt: 5.4.1
DE/WM: i3

all compiled with
GCC: gcc version 4.8.3 (Gentoo Hardened 4.8.3 p1.1, pie-0.5.9)

Hardware: GPU nvidia GTX 460

Reproducible: Always
```

ProMcTagonist:
```
Info:
OS: Debian Jessie GNU/Linux x86_64
Kernel: 3.16.0-4
Xorg: 1:7.7+7
nouveau: 1:1.0.11-1
Qt: 5.3.2
DE/WM: Openbox 3.5.2-8

gcc: 4:4.9.1-5
gcc-4.8: 4.8.3-13
gcc-4.9: 4.9.1-19

Hardware: NVIDIA Corporation G84GLM [Quadro FX 570M] (rev a1)

Reproducible: Always
```

agilob:
```
Info:
OS: Arch
Kernel: 3.18.6-1
[...]

PR #1367, opened 2015-03-11
Probably closes tux3/qtox#1330

agilob:


zetok:
fixes problem for me

aside from that, shouldn't there be put comment along code that it fixes this particular problem?

agilob:
https://github.com/tux3/qTox/issues/1330#issuecomment-78243033

Moving mouse pointer over groupwidget results in CSS effect mouse-over(?). Changing group title repaints only changed element - title, the rest of the widget stays in the same CSS as it was on mouse over. Repaining whole widget fixes style problem.

PR #1371, opened 2015-03-11
Alphabetized Friends List

Zer0-One:
Adds functionality for sorting the friends list, merges Online/Away/Busy layouts into a single layout, and adds a configuration option that lets the user choose whether groupchats will be placed at the top of the friends list, or below the online contacts.

alphabetized
checkbox


dubslow:
Would you be willing to split this into smaller commits?

Like one commit for merging layouts, one commit for the conf option, etc

TheLastProject:
I'm really not a fan of putting even more options in the options screen, it's already way crowded. I think that putting group chats on top is more logical anyway, and should just be the new (and only) default, as it better separates group chats from "normal" chats, making it much easier to find back what you're looking for.

EDIT: On that note, could groupchats maybe be sorted alphabetically too, while you're at it?

EDIT 2: Sorry if I sound too negative, I think this is way overdue and am quite glad you wrote this. I always have so much trouble finding back contacts...

Zer0-One:
@TheLastProject I thought about sorting groupchats by alphabet as well, but it really doesn't make sense to me, since topics probably change a lot more frequently than names. I think we should wait until we have a design in the UI for sorting by activity, which I will implement at some point if no-one else wants to do it (i.e "if i have to").

And don't worry, I know you mean well :smile:

Edit: As far as making the groupchat layout a bit more distinguished, I'm going to try changing the status lights to blue or possibly adding a horizontal separator line. I'll add that in a separate PR that people can comment in.

ProMcTagonist:
If there was a separation between Contacts and Recent (conversations) then groupchats would be only in the conversations tab. Just a thought regarding organization.

And I agree that there are too many options already and groupchats should just be on top with no option

zetok:
Doesn't apply cleanly any longer on master

zetok:
Doesn't apply cleanly any longer on master again

Zer0-One:
@dubslow looks good?

PR #1380, opened 2015-03-13
Add debug statement about audio filtering and echo cancellation

PKEv:


dubslow:
I was not able to get this to compile, something about the strange header, which is why I didn't include it.

tux3:
This should probably be moved to Audio::getInstance() in audio.cpp, where it's more in context and the required headers are already included.

PR #1384, opened 2015-03-14
Update generalform.cpp

dbuday:
Added croatian (Hrvatski)


tux3:
Please don't send three different pull requests for a single logical change (adding a translation). Instead, make three commits in a branch and send a PR for that branch.

Also, it seems that you're trying to add a translation to qTox, but you never actually uploaded the translation files (.ts), those are required.

PR #1386, opened 2015-03-14
Croatian (hrvatski) translation

dbuday:
I have just finished translating qTox to croatian.
Sorry for three pointless pull requests - i just removed them.
Here is the whole job - necessary changes to three files + language file translations/hr.ts

I hope this time I did everything correctly.

### DIFF
--- a/res.qrc
+++ b/res.qrc
@@ -53,6 +53,7 @@
         <file>translations/es.qm</file>
         <file>translations/fi.qm</file>
         <file>translations/fr.qm</file>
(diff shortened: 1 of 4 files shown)


dbuday:
Very quick indeed.  I`ve updated qTox from repo - it`s already there. Thanks
Now I see that I let out a few letters. Only few - barely visible. I 
will check every word, make changes and improvements and update 
translation in few days to conclude it.

Keep up the good work.

14.03.2015 u 17:00, tux3 je napisao/la:

PR #1391, opened 2015-03-14
Align file bubbles better

### LINKED ISSUE #1178: File bubbles are misaligned
misaligned file bubble

They start too low down, making the selection padding uneven.

Flakebi:
File bubbles were misaligned with the username. This patch changes the alignment of a ChatLineContentProxy so it looks better aligned. I added a margin of 5 to the bounding rect to add a bit space between the bubble and the selection rectangle.
This should fix #1178.


Flakebi:
I choose 7.0 because it looks good on my system. I hope it's not dpi dependent but I don't know much about it.

PR #1397, opened 2015-03-15
fix  #1394

### LINKED ISSUE #1394: segfault on starting
qtox segfaults after the following console output:

```
IPC: Our global ID is  17187678335339865315
IPC: Attaching to the global shared memory
Settings: Loading from  "/home/sudden/.config/tox/qtox.ini"
Using embeded bootstrap nodes list
built on:  00:14:32 Mar 15 2015 ( 1426374838 )
commit:  ac9b327fe7eb6c05c9ac353128de65931c2dc9fb 

IPC: postEvent  "activate" to 0
IPC: Previous owner timed out, taking ownership 6477497898772817312 -> 17187678335339865315
Nexus: Starting up
Core: loading Tox from "/home/sudden/.config/tox/sudden.tox"
Audio:  "Trying to open output "
Audio: Opening audio output  ""
Audio:  "Trying to open input "
Audio: Opening audio input  ""
QFSFileEngine::open: No file name specified
```

Backtrace:

```
(gdb) bt
#0  0x00007ffff20a5b0f in QString::operator=(QString const&) () at /usr/lib64/libQt5Core.so.5
#1  0x00007ffff36b265c in QWidget::setToolTip(QString const&) () at /usr/lib64/libQt5Widgets.so.5
#2  0x0000000000437825 in GeneralForm::reloadSmiles() ()
#3  0x000000000043835d in GeneralForm::GeneralForm(SettingsWidget*) ()
#4  0x0000000000436256 in SettingsWidget::SettingsWidget(QWidget*) ()
#5  0x0000000000457e15 in Widget::init() ()
#6  0x00000000004d3964 in Nexus::start() ()
#7  0x00000000004309fc in main ()
```

sudden6:
Probably more a hack, but it fixes the problem.

### DIFF
--- a/src/widget/form/settings/generalform.cpp
+++ b/src/widget/form/settings/generalform.cpp
@@ -27,6 +27,7 @@
 #include <QTime>
 #include <QFileDialog>
 #include <QStandardPaths>
+#include <QDebug>
 
 #include "src/autoupdate.h"
 
@@ -313,8 +314,11 @@ void GeneralForm::onReconnectClicked()
 void GeneralForm::reloadSmiles()
 {
     QList<QStringList> emoticons = SmileyPack::getInstance().getEmoticons();
-    if(!emoticons.isEmpty())
+    if (emoticons.isEmpty())
+    // sometimes there are no emoticons available, don't crash in this case
     {
+        qDebug() << "Settings GeneralForm: No emoticons found";
+    } else {
     QStringList smiles;
     smiles << ":)" << ";)" << ":p" << ":O" << ":["; //just in case...
 
--- a/src/widget/form/settings/generalform.cpp
+++ b/src/widget/form/settings/generalform.cpp
@@ -313,6 +313,8 @@ void GeneralForm::onReconnectClicked()
 void GeneralForm::reloadSmiles()
 {
     QList<QStringList> emoticons = SmileyPack::getInstance().getEmoticons();
+    if(!emoticons.isEmpty())
+    {
     QStringList smiles;
     smiles << ":)" << ";)" << ":p" << ":O" << ":["; //just in case...
 
@@ -331,6 +333,7 @@ void GeneralForm::reloadSmiles()
     bodyUI->smile3->setToolTip(smiles[2]);
     bodyUI->smile4->setToolTip(smiles[3]);
     bodyUI->smile5->setToolTip(smiles[4]);
+    }
 }
 
 void GeneralForm::onCheckUpdateChanged()


zetok:
Could you add comment to your code?

sudden6:
I don't really know what I should comment, this change catches a case where there are no emoticons found. This case only happened once and I'm now unable to reproduce it :/

agilob:
Did it happen after you upgraded some packages?

sudden6:
packages may be upgraded, but mainly I was upgrading to the newest qTox git

zetok:
@sudden6 dunno, something like `// sometimes there are no emoticons available, which would lead to crash if not caught` after `if(!emoticons.isEmpty())` ?

Actually, if you're going to work on hacks like that, I think that it would be good to reform it a bit; i.e. instead of just 

``` C
if(!emoticons.isEmpty())
{code}
```

why not add some debug statement in `if else`, i.e.

``` C
if (emoticons.isEmpty())
{
qDebug()<<
} else if (!emoticons.isEmpty())
{code}
```

unless something else in `void GeneralForm::reloadSmiles()` logs when emoticons are empty?
_(just a suggestion, I'm not saying that it should be done that way)_

# 

Aside from that, your coding style is a bit off, i.e. you should put a space between `if` and `(`, i.e. instead of `if(!emoticons.isEmpty())` it should be `if (!emoticons.isEmpty())`

Edit: oh, and sorry for making your PR harder; I would have done it myself if I knew how to code

sudden6:
I added some improvements to the previous commit as @zetok suggested.

> Edit: oh, and sorry for making your PR harder; I would have done it myself if I knew how to code

No problem, I'm here to learn a better programming style, so improvements to my code are always welcome ;)
And I don't actually know how to C++ too, I just did trial and error^^

src/widget/form/settings/generalform.cpp
```diff
@@ -313,6 +314,11 @@ void GeneralForm::onReconnectClicked()
 void GeneralForm::reloadSmiles()
 {
     QList<QStringList> emoticons = SmileyPack::getInstance().getEmoticons();
+    if (emoticons.isEmpty())
+    // sometimes there are no emoticons available, don't crash in this case
+    {
+        qDebug() << "Settings GeneralForm: No emoticons found";
+    } else {
```
dubslow, line 321: If we're gonna be picky, this isn't Allman style which qTox largely adheres to, and also the rest needs to be indented.

It may be better (from a style standpoint) to just return after the qDebug, and not put the rest of the function inside the conditional.

dubslow:
Also for something like this, better to amend your previous commit than make a new one :) You can force push amended commits all you like, until such time as I merge it.

dubslow:
Squashed and merged in https://github.com/tux3/qTox/commit/42fe29e83df19c2faadeb52d1e06f1ef799ba04a

Note that the merge commit modifies your code as I suggested.

Also @zetok in this case a comment isn't really necessary at all, it's pretty darn obvious exactly what, how and why the `if` statement is doing.

PR #1398, opened 2015-03-15
Translation system menu

PKEv:


dubslow:
Could you elaborate on what this does?

Do you mean this translates strings in the Qt source that qTox doesn't have access to? If so, can you use the translation selected by the user in GeneralForm, instead of just the system locale?

If not, then please explain.

PKEv:
This allows you to translate the language of dialogue and system menu qTox, such as copy, paste, the buttons in dialog boxes, etc.
Regarding the transfer of the selected user, it seems to me it is not so important because the user is using the system, but I think as it is possible to take into account

agilob:
> This allows you to translate the language of dialogue and system menu qTox, such as copy, paste, the buttons in dialog boxes, etc.

we had at least one bug report about it. I think system language is better than qTox lang in this case.

dubslow:
> I think system language is better than qTox lang in this case.

Really? Why?

agilob:
We had an issue about hat, someone had Windows completely in Russian, but QFileDialog was in English. Can't find it now ;/

dubslow:
That's what this is supposed to fix, and I don't see how it's related to user-selected language vs system locale.

PR #1406, opened 2015-03-16
ver.2 System menu translation

PKEv:
Does not work with all languages, but takes into account the user's settings.
In the future, you need to fix an array of "locales" in the file "generalform.cpp" according to the names of available locales in QT.
I did not find this information.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -59,6 +59,7 @@
 #include <QList>
 #include <QDesktopServices>
 #include <QProcess>
+#include <qlibraryinfo.h>
 #include <tox/tox.h>
 
 #ifdef Q_OS_ANDROID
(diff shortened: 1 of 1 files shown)


agilob:
You could just commit to previous PR ;_;

dubslow:
Can't you force push this commit to the branch that your first PR is? And also can you expand on why this works but the previous one doesn't? And please the relevant information in the commit message as well as in the PR comments.

PKEv:
I have some problems with "smartgit" or with hands ...
let's continue here:
Previous also worked, but only took the information from the locale of the system and only at startup.
This works right when you change the language.
Note, that in the current assembly generally works very strange change the interface language - sometimes there is no entry in the configuration file.

PKEv:
Correct, please merge.

It looks terribly:

dubslow:
Can you clean up this PR? It looks like only the last commit is the relevant one.

PKEv:
good?

PKEv:
this PR Not logged in 86e9ee92 tux3:master

PR #1418, opened 2015-03-17
Added chatroom status to window title

rohilsurana:
I thought that the status of active chat room in the window title would be nice. The status strings displayed are same as that in the Status class and when there's a new message (New Message) is displayed instead of status.
screenshot


agilob:
1. Add space before `(STATUS)`
2. Strings must be translatable (commented)
3. Optimize that huge `if` (commented)
4. I was in chat A, got message from B, switched to chat with B, title was still "(New message)" at the end, should be "(Online)" after switching.

rohilsurana:
How about creating a seperate function for the if statements?

agilob:
It's up to you, I just don't like that huge else if block, you don't need half of the comparisons there.
All string that you show in GUI must be in tr() so they can be translated.

rohilsurana:
have a look at the recent commit 5229312e2bac98b3fc141d1a78497e0edeed6acf

rohilsurana:
Sorry my bad... :) Corrected the Null check and the indentation  and also the fixed the problem where New Mesaage was still showing up. 335cd655e5a6636ad0311d78992c7b771fcc3dd0

dubslow:
Can you rebase on master please? You seem to have some extra stuff in your commit.

tux3:
Rebased and merged in 1feea9e6fab8f2533cd7ef47f24445f037ed93d2

PR #1429, opened 2015-03-19
IPC improvements

ghost:
- Removed waitUntilProcessed() because waitUntilAccepted() fits the job. They were nearly identical too so decreased code duplication
- Global events are set as processed only by instance that accepts them. Solves issue where global event would be consumed by first instance that saw it even if that instance ignored that event
- Fixed bug where running qtox instance would not properly exit after sending window activation event that was accepted by already running instance

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -150,7 +150,7 @@ void IPC::registerEventHandler(const QString &name, IPCEventHandler handler)
     eventHandlers[name] = handler;
 }
 
-bool IPC::isEventProcessed(time_t time)
+bool IPC::isEventAccepted(time_t time)
 {
     bool result = false;
     if (globalMemory.lock())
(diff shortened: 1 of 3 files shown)


src/ipc.cpp
```diff
@@ -291,9 +262,19 @@ void IPC::processEvents()
             auto it = eventHandlers.find(name);
             if (it != eventHandlers.end())
             {
-                evt->accepted = runEventHandler(it.value(), evt->data);
                 qDebug() << "IPC: Processing event: " << name << ":" << evt->posted << "=" << evt->accepted;
```
dubslow, line 265: Isn't evt->accepted now uninitialized in this line?

ghost, line 265: When new event struct is allocated its zeroed [here](https://github.com/novist/qTox/blob/ipc-improvements/src/ipc.cpp#L116) so its ok, its still initialized to false.

PR #1431, opened 2015-03-19
Add some error handling in attempt to fix #1240

### LINKED ISSUE #1240: qtox audio/video configuration menu causes a crash
As I wrote in the subject, when you try to configure the audio and video from the specific menu, qTox dies after a few seconds spent in the menu ("Audio/Video" menu).

run on Fedora 21 with XFCE

sudden6:
This should prevent the application from crashing due to missing/not working webcam, but there is still no way to tell there is no webcam attached.


dubslow:
You've tested it?

Issue #1413, opened 2015-03-17
Clicking tray icon flashes strangely


TheLastProject:
Previously, clicking the tray icon make qTox minimize to tray, and clicking it again made qTox return to the size it was. However, now it acts quite strange.

I always run qTox maximized, and when I click the tray icon, the qTox window minimizes (to the task bar, not to the tray). When I click qTox from the minimized task bar entry, it'll appear in a non-maximized windowed mode, instead of my previous maximized mode. When I click qTox from the tray icon, it'll appear back from the task bar, then disappear and flash back into view, in both cases ignoring the fact that it was maximized before.

KDE4 here.

TheLastProject:
Oops, duplicate of #1411. Well, almost. It is a similar bug, but not exactly the same.

zetok:
Yeah, also happens for me.

Info:
OS: Hardened Gentoo Linux x86_64
DE/WM: i3
Qt: 5.4.1
qTox: de423571eb16bab230f8de5e277c3b0b20315e00

Reproducible: Almost always

ProMcTagonist:
When my qTox is maximised and I click the tray icon, then bring it back by clicking again, it comes back unmaximized.

Debian Jessie x86_64
Openbox/Tint2
Qt 5.3.2
qTox de423571eb16bab230f8de5e277c3b0b20315e00
Reproducible always.

agilob:
related https://github.com/tux3/qTox/issues/1542

zetok:
@TheLastProject is it still reproducible?

agilob:
yes, it will be fixed with #1542

agilob:
#2365

sudden6:
This seems to have been fixed somewhere until https://github.com/qTox/qTox/commit/bdbf2b461bd2fde5c2cc00fca364f76e5b448356

Please open a new issue if it still happens.

PR #1433, opened 2015-03-19
Correct for Maximized window

PKEv:
Correct #1413
Added handling of the state Maximized

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -84,6 +84,7 @@ Widget::Widget(QWidget *parent)
       eventFlag(false),
       eventIcon(false)
 {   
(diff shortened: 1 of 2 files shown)


src/widget/widget.cpp
```diff
@@ -84,8 +84,10 @@ Widget::Widget(QWidget *parent)
       eventFlag(false),
       eventIcon(false)
 {   
+    installEventFilter(this);
     translator = new QTranslator;
     setTranslation();
+    WasMaximized = false;
```
dubslow: This is C++11, you can initialize this from the declaration. Also the first w should be lower case.

PKEv:
Correct

Issue #1417, opened 2015-03-17
Problem with build


PKEv:
problem with lib qrencode build in windows

ProMcTagonist:
@stqism is this fixed now? after what you said in #1408? Or was that only for the Linux binaries?

stqism:
@ProMcTagonist https://github.com/tux3/qTox/issues/1408#issuecomment-82548509

PKEv:
Could not find how to collect qrencode under windows

PKEv:
Maybe add https://code.google.com/p/qrencode-win32/source/checkout "qrencode-win32" fork in reporizitary tux3 for create (correct) script "bootstrap.sh" (build automation).
Especially the news: http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html

PR #1438, opened 2015-03-19
manual how to build in Windows with qrencode

PKEv:
decided to write..
can anyone suggest another solution?
#1417


dubslow:
@tux3 runs the Windows build server, and he will shortly get around to updating the usual Windows downloads with the new QR code. When he does that he can also presumably comment on this.

tux3:
Yes, those instructions sound about right.

PR #1446, opened 2015-03-20
Some style fixes

zetok:
"if(" → "if ("
"for(" → "for ("
"while(" → "while ("

I also would like to ask whether https://github.com/tux3/qTox/wiki/Coding is correct, and if not, then what is correct version, since qTox's source is inconsistent as hell…

When there will be an agreement to what coding style is, could you please _enforce_ it?

### DIFF
--- a/src/chatlog/chatline.cpp
+++ b/src/chatlog/chatline.cpp
@@ -27,9 +27,9 @@ ChatLine::ChatLine()
 
 ChatLine::~ChatLine()
 {
-    for(ChatLineContent* c : content)
+    for (ChatLineContent* c : content)
     {
-        if(c->scene())
+        if (c->scene())
             c->scene()->removeItem(c);
 
         delete c;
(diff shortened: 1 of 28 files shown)


dubslow:
No that's not correct, qTox uses Allman style generally. I usually enforce it, but I didn't bother with the chatform thing since, well... chatform.

Other than that I _usually_ massage my merge commit to silently fix the formatting in PRs, but sometimes something slips through the cracks...

```
if ()
{
    thing;
}
else
{
    otherThing;
}
```

```
QObject* asterisksGoWithTheType;
uint8_t* array = new uint8_t[count];

QObject notToMentionThatWeUseCamelCase;
// *camelCase, not CamelCase
```

agilob:
Merging this will make other PRs conflict.

dubslow:
Indeed. Now that I think about it, it's probably best to wait with this until after new_api hits. I'm largely throwing core.cpp out the window, for example.

agilob:
You can use astyle -A1 at any time if you want to convert code to allman style.

PR #1453, opened 2015-03-22
Added option to disable notification sounds.

Jookia:
This is a simple option I find needed.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -745,17 +745,20 @@ void Widget::newMessageAlert(GenericChatroomWidget* chat)
             setWindowState(Qt::WindowActive);
     }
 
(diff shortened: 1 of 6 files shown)


src/widget/widget.cpp
```diff
@@ -745,17 +745,20 @@ void Widget::newMessageAlert(GenericChatroomWidget* chat)
             setWindowState(Qt::WindowActive);
     }
 
-    static QFile sndFile(":audio/notification.pcm");
-    static QByteArray sndData;
-
-    if (sndData.isEmpty())
+    if(Settings::getInstance().getNotifySound())
```
dubslow: See the wiki code style guidelines (space between `if` and `(`)

Jookia: Oops, fixed.

PR #1449, opened 2015-03-21
fixes #1448, CTRL+C and CTRL+A works now

kushagrasurana:
Dont know why QShortcut is not working.

Issue #1019, opened 2015-01-17
List of keyboard shortcuts to implement


agilob:
- [ ] Screenshot  #1678 // CTRL + ALT + S
- [ ] ~~Copy my ID (only in Identity tab)~~
- [ ] Start/End audio/video call
- [ ] Create group chat // CTRL + N
- [ ] Find in chat  // CTRL + F
- [ ] Enable/disable logging // CTRL + SHIFT + P (shift my privacy setting)
- [ ] Leave group chat // CTRL + W
- [x] Paste previous message // arrow up(?)
- [x] Clear chat window  // clear and CTRL + SHIFT + L
- [x] Chat with contact above/below // CTRL + Pg_Up/Down

Anything more?

List of duplicates:
https://github.com/tux3/qTox/issues/1043 https://github.com/tux3/qTox/issues/997

zetok:
Why limiting ID copying to Identity tab?

agilob:
Depends what keyboard shourtcut we choose, What about Ctrl+C in Identity tab only? Overriding ctrl+c in other tabs is a bad idea.

zetok:
overriding ctrl + c anywhere is a bad idea

TheLastProject:
My suggestions:
- Clear chat window -> Don't, people will get confused when they accidentally tap it
- Chat with contact above/below -> Ctrl/Alt + Up/Down arrow
- Find in chat -> Ctrl+F seems sanest, also most common
- Enable/disable logging -> This seems too uncommon to even be used
- Leave group chat -> Escape (but maybe with a warning)

Regarding the ID: Seeing how it is something you need to be able to share easily, it may be worth thinking of putting it somewhere where it is easier to find.

agilob:
> Clear chat window -> Don't, people will get confused when they accidentally tap it

Ctrl+L, it's difficult to press both accidentally, of course there should be a manual of shortcuts and possibly, a way to change them in settings.

AlexDaniel:
Clear chat window -> /clear
Leave group chat -> ctrl+w

zetok:
@AlexDaniel using `/clear` is a bad idea, considering that if something, it should be easy to use (commands aren't, they're not even easy to use by english-speaking people, while people who don' t speak english will have it even harder.
Not to mention that we're talking about keybidings here..

@TheLastProject I personally don't like keybinding to ESC, `ctrl + w` sounds much saner

[...]

Issue #997, opened 2015-01-08
acceleration key support


xuhdev:
Keyboard shortcut to switch to the next contact, previous contact, previous message, etc.

AlexDaniel:
Yes, please!

agilob:
@xuhdev What do you mean by 'previous message'?

xuhdev:
@agilob The previous sent message. For example, I sent you a message but want to correct something, or need to type something similar again, I can just quickly call back the previous sent message and modify it. Similar to calling back the previous command in bash

agilob:
arrow up = pastle previous message?

xuhdev:
@agilob Similar to that, I would suggest that only when the cursor is at the first line.

agilob:
or message box is empty.

xuhdev:
correct

agilob:
Partly done.
https://github.com/tux3/qTox/pull/1048

xuhdev:
@agilob Great! Thanks! We can also add a keyboard shortcut to move between contacts on the left list. Maybe `Alt-N` `Alt-P` or `Shift-Down` `Shift-Up`?

srkunze:
From [Switching and Re-ordering Chats like Browsers do](https://github.com/tux3/qTox/issues/1043#issuecomment-71395165)

ctrl + page-up/down <<< switch to next tab
ctrl + shift + page-up/down <<< move current tab down or up

chat = tab

zetok:
When there'll be new keyboard shortcuts, please edit https://github.com/tux3/qTox/wiki/Keyboard-shortcuts accordingly (it's only for shortcuts that qTox already supports)

zetok:
Closing, since mentioned functionality is in qTox - i.e. switching next/previous contact + pasting previous message.

As for switching between multiple messages, there already exists #310.

PR #1454, opened 2015-03-22
Add ability to cycle through the contacts list.

Jookia:
I've added the ability to cycle through the contacts list with Ctrl+Tab and Ctrl+Shift+Tab. See commits for discussion of the implementation. I haven't tested this on friends lists with just one friend but my calculator says it should work.

The last commit may conflict with #1449 if they fix the shortcuts for the chat system.
This should work with #1371 as it walks the widgets rather than using internal structure.
Improves #997 and #1019 by adding these keybinds.

### DIFF
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -169,22 +169,10 @@ GenericChatForm::GenericChatForm(QWidget *parent)
     connect(emoteButton, &QPushButton::clicked, this, &GenericChatForm::onEmoteButtonClicked);
     connect(chatWidget, &ChatLog::customContextMenuRequested, this, &GenericChatForm::onChatContextMenuRequested);
 
-    new QShortcut(Qt::CTRL + Qt::Key_PageUp, this, SLOT(previousContact()));
-    new QShortcut(Qt::CTRL + Qt::Key_PageDown, this, SLOT(nextContact()));
     new QShortcut(Qt::CTRL + Qt::SHIFT + Qt::Key_L, this, SLOT(clearChatArea()));
 
     chatWidget->setStyleSheet(Style::getStylesheet(":/ui/chatArea/chatArea.css"));
     headWidget->setStyleSheet(Style::getStylesheet(":/ui/chatArea/chatHead.css"));
-}
-
-void GenericChatForm::previousContact()
-{
-    parent->previousContact();
-}
-
-void GenericChatForm::nextContact()
-{
-    parent->nextContact();
 }
 
 bool GenericChatForm::isEmpty()
--- a/src/widget/form/genericchatform.h
+++ b/src/widget/form/genericchatform.h
@@ -76,8 +76,6 @@ protected slots:
     void clearChatArea(bool);
     void clearChatArea();
     void onSelectAllClicked();
-    void previousContact();
-    void nextContact();
 
 protected:
     QString resolveToxID(const ToxID &id);
--- a/src/widget/friendlistwidget.cpp
+++ b/src/widget/friendlistwidget.cpp
@@ -61,6 +61,32 @@ QVBoxLayout* FriendListWidget::getFriendLayout(Status s)
     return layouts[static_cast<int>(Status::Online)];
 }
 
+QList<GenericChatroomWidget*> FriendListWidget::getAllFriends()
+{
+    QList<GenericChatroomWidget*> friends;
+
+    for (int i = 0; i < mainLayout->count(); ++i)
+    {
+        QLayout* subLayout = mainLayout->itemAt(i)->layout();
+
+        if(!subLayout)
+            continue;
+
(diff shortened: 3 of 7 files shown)


agilob:
Ctrl+Tab and Ctrl+Shift+Tab are global/system shortcuts, please set it on CTRL+PG UP/DOWN.

Jookia:
I don't understand. Most applications use the same shortcuts, they don't seem global or system shortcuts?

agilob:
Both used keyboard shortcuts exist in KDE, this commit would override them.
Ctrl+PG up/down already do the same thing in firefox, gajim, piding, clementine, netbeans, ms/libre office...

Jookia:
I see, what do you mean exactly override? Does it steal it from the window manager somehow? Shouldn't it be deciding what its global keys do before applications? This commit also supports Ctrl+Page Up and Ctrl+Page Down since it uses QKeySequence::NextChild and QKeySequence::PreviousChild.

zetok:
@agilob they exist in KDE? from what I remember when I was using KDE I was able to use `CTRL`+`TAB` / `CTRL`+`SHIFT` + `TAB` to switch between tabs in firefox without any problem… (yes, it would seem that TAB does in ff ~same thing `Ctrl+PG up/down` does

zetok:
@Jookia if both ways are supported, then it's the same way ff does it, so IMHO it's as it should be..

Jookia:
@zetok Searching the web seems it may be assigned to switching desktop lists sometimes? Either way if it's overriding global hot keys it's a bug.

zetok:
@Jookia it's not a bug if ff uses it.

Jookia:
@zetok I imagine if you have a key set to globally do something qTox shouldn't affect it or override them, qTox shouldn't get the keys in the first place.

zetok:
@Jookia how do you detect then that there's "global" keybinding, so that qTox could be prevented from grabbing it if it's in use already?

Jookia:
Ideally it shouldn't be getting the keybinding in the first place, so I'm guessing I'm not quite understanding what agilob means, unless there's a bug in KDE.

agilob:
ok, i will provide a way to set such parameters anyway. I'm still saying ctrl+pg up/down is much obvious as more applications use it.

Jookia:
Ah I see. Checking again, I was mistaken- This code does not support Ctrl+Page Up or Ctrl+Page Down, I must've had it working at some point during development and later removed it. I'll have to add this in before any merging happens. Apologies!

Jookia:
I've updated the pull request to use shortcuts and support both key bindings explicitly. This should now actually work since #1457 is merged. In future if there's an option to bind keys, the shortcut parts should be changed as well as other shortcuts across the application.

TheLastProject:
@agilob Err, Ctrl+(Shift+)Tab seems way more common for switching between tabs, because of Firefox and Chrome and the likes doing it, which is basically what you're doing in qTox.

@Jookia supporting both sounds sane, thanks :)

zetok:
@Jookia just a nitpick;
How did you manage to not include info about keybindings in commit messages? Like, they're awesome and stuff, but main point is MIA..

Jookia:
The idea was that it used the existing key bindings, though I should explicitly note that. I'll fix that.

Jookia:
Done. It now mentions the key bindings at the bottom of the last commit. I figure people who want to know more would read the actual commit message.

zetok:
on top of f55fd40ec66e43625b4f206c08c9b511b6934d97

```
* Applying user patches from /etc/portage/patches//net-im/qtox ...
 *   1332-zetok-quoting-fix.patch ...                                                                                                        [ ok ]
 *   1371-zero-one-alphabetized-friend-list.patch ...                                                                                        [ ok ]
 *   1454-Jookia-keybindings-for-cycling-through-contact-list.patch ...

 * Failed Patch: 1454-Jookia-keybindings-for-cycling-through-contact-list.patch !
```

Jookia:
It works fine for me - I've rebased it anyway to make merging easier, perhaps try now? Are you sure you've collected all of the patches and not just the last?

zetok:
Um, it fails to apply on top of other patch, I guess it would work on clean master.

And yep, `wget https://github.com/tux3/qTox/pull/1454.diff -O diff.patch` gets all patches

Jookia:
Oh, it could be that alphabetized friend list patch that's doing things. I'll have a look to see if I can get it to merge cleanly now instead of later.

Jookia:
I'm can't seem to find a compromise to make them merge happily automatically as both branches modify the same areas of code, meaning you have to move some lines around. Try this branch for a patch that includes the friends list code: https://github.com/Jookia/qTox/tree/cyclecontacts_friend_list

tux3:
Rebased on master and merged in 821b01a90b38313d5f1d2ba41308ccdccfaf7f98

PR #1460, opened 2015-03-23
autocomplete will work for : also,fixes #1377

### LINKED ISSUE #1377: Groupchat: :[tab] will autocomplete for all nicks, not only for those starting with `:`

kushagrasurana:
added <code>:-</code> in regex.
Works for me

### DIFF
--- a/src/widget/form/tabcompleter.cpp
+++ b/src/widget/form/tabcompleter.cpp
@@ -49,7 +49,7 @@ void TabCompleter::buildCompletionList()
     nextCompletion = completionMap.begin();
 
     // split the string on the given RE (not chars, nums or braces/brackets) and take the last section
-    QString tabAbbrev = msgEdit->toPlainText().left(msgEdit->textCursor().position()).section(QRegExp("[^\\w\\d-_\\[\\]{}|`^.\\\\]"), -1, -1);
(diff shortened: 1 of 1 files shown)


agilob:
The same bug with `;`
Will autocomplete work with nicks like `:3` with that commit?

kushagrasurana:
so autocomplete should work with ; also ?
yes it will autocomplete nicks like <code>:3</code>

agilob:
http://webmup.com/HmDJQ/vid.webm

dubslow:
The base problem is that each special character needs to be specially added to the regex... what to support can be difficult.
@kushagrasurana What's with the extra `-` character in your commit?

kushagrasurana:
@dubslow without that extra <code>-</code>, autocomplete wont work for <code>-</code>.
At preset, we are whitelisting all characters which should work with autocomplete.
We can do it other way, that is blacklisting all those characters which we dont want to include in autocomplete like <code>whitespace</code>.

dubslow:
That might indeed be the better option....

kushagrasurana:
Can you tell me all the blacklisted characters ( like whitespace )

tux3:
Finally a PR that merges cleanly against master \o/ !

It could maybe be done better, but this is an improvement for now.

agilob:
well only one line changed...

tux3:
@agilob After the horrible things I had to go through for the other 15 PRs, this is paradise.

New rule: From now on all PRs must be 1 line long!

Issue #773, opened 2014-11-16
Screenshot button implementation


zetok:
Thanks to people in groupchat I got idea how to implement it UI-wise, here's quick mock-up made in gimp:
![mock-up](https://cloud.githubusercontent.com/assets/3148759/5061395/046380ca-6d8b-11e4-955c-aa86e4967c05.png)

Also, some explanation (sorry, I'm too lazy to write it again):
![explanation](https://cloud.githubusercontent.com/assets/3148759/5061397/1a074b6e-6d8b-11e4-9588-5839dd0ce872.png)

That "list" is supposed to be extensible in the future.

Hope it helps.

tux3:
A screenshot button ? But, why ?

It's the OS's job to handles that.

zetok:
@tux3 people seem to **really** want screenshot button in Tox client, and saying that it's OS's job doesn't help with that...

ghost:
I remember that uTox has a screenshot button, just in place of qTox's attachment button.

tux3:
I don't get it.

What's the usecase ? 
Why would it be better than hitting Print Screen ?
What's wrong with print screen ?

The GUI's gonna get cluttered and confusing if we had tons of buttons.

zetok:
@tux3 to put it straight: use case is people who cannot into computers, and because of it they need screenshot in their qTox, because otherwise they won't be able to make a simple screenshot and send it.

Also, screenshot button isn't supposed to work like Print Screen, but offer a way to select only a part of screen…

TheLastProject:
@tux3 This is exactly my opinion. To me, it makes absolutely no sense to have a screenshot button. The argument I've been given is that it's "easier", but I really don't feel like it is really an improvement. You can already either press PrintScreen and then paste it, or save the screenshot to a file and send the file through Tox. It feels quite redundant to reimplement a standard OS feature.

tux3:
@zetok

If someone can't into computers, a simple GUI that does only what you'd expect will probably help more than tons of options imo.

> Also, screenshot button isn't supposed to work like Print Screen, but offer a way to select only a part of screen…

I would use a screenshot tool for that, not an IM client, don't you think ?
I don't think one of the main features of a Tox client should be a tool to take screenshots of a region of the screen. Not enough to justify a button anyway. A keyboard shortcut, maybe.

dubslow:
[...]

PR #1462, opened 2015-03-24
Fancy  screenshot grabber

Papierkorb:
Took TheSpiritXIII's work and rewrote the grabber to use QGraphicsView instead of QRubberBand, allowing to have fancy features like better visual feedback and being able to resize and move the selection further.


zetok:
```
 * Applying user patches from /etc/portage/patches//net-im/qtox ...
 *   1332-zetok-quoting-fix.patch ...                                                                                                        [ ok ]
 *   1371-zero-one-alphabetized-friend-list.patch ...                                                                                        [ ok ]
 *   1462-Papierkorb-screenshot-button.patch ...                                                                                             [ ok ]
 * Done with patching
```

>tfw no screenshot

How one is supposed to take that screenshot? → Could you please use idea that makes it clear, since obviously currently it's **not** intuitive, and you're _hiding_ functionality that should be somewhat exposed to user.

Colour selection is _bad_. E.g. nicely done colour in ksnapshot: 
ksnapshot nice

And qTox:
qTox bad

Also, I don't think that including commits with binary blobs (.png) that aren't used anyway is good…

Beside that, it would be nice if you could put https://github.com/tux3/qTox/wiki/Coding#coding-style into practice, since currently PR has mixed style..

zetok:
Oh, and I forgot to mention; selection performance is bad.

Papierkorb:
> How one is supposed to take that screenshot?

Granted, I focused on implementing the feature, and less on exposing it nicely. The current context-menu is a sub-par solution. I was thinking about replacing the current "attach" button functionality with that it simply shows a popout menu, where the user can choose "Attach file" and "Attach screenshot". Not sure if this is ideal, but certainly better I think. And considering that most users probably drag-drop a file into the window to send it, it may not even introduce a single extra click for most users?

The PNG is an left over of the PR I based this upon. Will throw it out when we decided how to expose this feature, that is, either do what I propose above, or have the "dropout" arrow next to the attach button. (I'm fine with doing it, but as a user, I don't like it when the UI is cramped full with those downwards arrows)

> Colour selection is bad.

What would be a better choice? Some lighter blue color like KSnapshot?

> selection performance is bad.

Really? Could you elaborate what's lagging or "feels bad"? Does it lag when you create the selection? Or move it, or resize it using the handles? Could you provide some HW details? Works fine on my notebook with i5 520M CPU/GPU.

> Beside that, it would be nice if you could put https://github.com/tux3/qTox/wiki/Coding#coding-style into practice, since currently PR has mixed style..

Will have a look, thanks.

ProMcTagonist:
@Papierkorb What does a "popout" menu look like?

Papierkorb:
Note: Just noticed I forgot to swap out the license header in my IDE. Sorry. Will also fix this right away.

Papierkorb:
@ProMcTagonist  A popout menu would look like this (It's effectively a context-menu appearing when you click the button):
menu_right
or this
menu_bottom

zetok:
> Or move it, or resize it using the handles?

Yep.
Flamegraph: https://a.pomf.se/ppkamb.svg

ksnapshot's selection works just fine, but it uses Qt 4, so..

```
Kernel: 3.18.7-gentoo
Xorg: 1.16.4
nouveau: 1.0.11
CPU: AMD Phenom II 550 BE @ 3.6 GHz
GPU: nvidia gtx 460
```

# 

> A popout menu

No.

# 

You probably would want to implement button according to #773, since it seems to be superior to what you propose. (if you want for me to elaborate on that, just ask (though don't expect quick answer))

Papierkorb:
@zetok Okay, it spends most of its time for drawing the QPixmap (which is the screen capture). Currently, most of the screen will be redrawn on each change. I'll look into it to make it draw less, and then I can just hope that Qt will do some trick and only draw the pixmap partially.

The button. So, I hover (or click?) over the attach button, and a list of buttons (like a toolbar) expands to the left, showing buttons for e.g. screen capture?

zetok:
> The button. So, I hover (or click?) over the attach button, and a list of buttons (like a toolbar) expands to the left, showing buttons for e.g. screen capture?

Yep, hover. Note that just clicking on attach button remains a default action, i.e. it sends file.

ProMcTagonist:
I like #773 but they should be shaped like this

```
/__|__|__|__/
```

instead of how they're shown in the first post in #773 which is

```
|__/|__/|__/|__/
```

Can't fire up GIMP right now, hope that makes sense.

zetok:
If I understood correctly, `|__/|__/|__/|__/` refers to corners, right?
Um, you know, #773 was just a quick copypaste made in gimp, so…  `;)`

What you propose seems good to me :3

ProMcTagonist:
Yes, exactly.

Cool.

Papierkorb:
Like this?
screenshot_slider

ProMcTagonist:
Flip the monitor one vertically.

Papierkorb:
Alternative versions: 

First:
screenshot_slider_edge

Second: (My favorite)
screenshot_slider_up

ProMcTagonist:
Second :+1:

ghost:
@Papierkorb Can you make it slide out on mouse hover?

Papierkorb:
@mannol That's the plan :)

Progress: Created the button SVG based on my GIMP mockup.

ghost:
@Papierkorb Oh right, tl;dr.

Papierkorb:
This took way longer than needed. But it works for the most part. Thoughts?

ProMcTagonist:
Does it introduce any new dependencies?

zetok:
@ProMcTagonist ?

zetok:
https://a.pomf.se/quwzvf.webm

It's awesome, though, can screenshot button be made to not hide after moving cursor 'attach → screenshot → attach' ?

zetok:
Also, when making screenshot: 
qtox-screenshot-ztvu2649a

Papierkorb:
Mh, should the flyout collapse first and then take a screenshot (Delays it by ~300ms), or should the screenshot taken while it's out, or should the flyout just disappear?

zetok:
I don't know, but halfway through isn't really good

src/widget/form/chatform.cpp
```diff
@@ -879,6 +884,37 @@ void ChatForm::loadHistory(QDateTime since, bool processUndelivered)
     chatWidget->verticalScrollBar()->setValue(savedSliderPos);
 }
 
+void ChatForm::onScreenshotClicked()
+{
+    ScreenshotGrabber *screenshotGrabber = new ScreenshotGrabber (this);
```
zetok: space ()

Papierkorb: Will be fixed in the now upcoming commit

Papierkorb:
Ok, the flyout should not be visible at all anymore on screenshots, and doing the button <-> flyout dance should not collapse it either. If everythings fine, I'd take look at the performance stuff now.

@ProMcTagonist No additional dependencies. Pure Qt.

zetok:
> This adds a delay of ca. 1/3

Adding a delay is kinda worse than just a "glitch" of halfway hidden button

Papierkorb:
Well, does it feel bad? Before that commit, there was already a delay which was a left over of the context-menu stuff. Sure we can change it, just want to make sure that this isn't a placebo where something feels worse than before without any real change in timing towards the user.

zetok:
> Well, does it feel bad?

Yes.

Papierkorb:
Could you check again the screen grabber performance @zetok ? Threw out a forced redraw, but I think we'll need to split the capture into tiles and display those instead ..

zetok:
When resizing selection, performance is ~on par (AFAICT with naked eye) with ksnapshot's performance - it's awesomely smooth.

There is still bad performance when dragging selection around.
Flamegraph: https://a.pomf.se/quuihw.svg

Also, there's a graphical glitch when selecting area: https://a.pomf.se/qjjizc.webm

src/widget/tool/screenshotgrabber.cpp
```diff
@@ -0,0 +1,183 @@
...
+void ScreenshotGrabber::useNothingSelectedTooltip()
+{
+    this->helperTooltip->setHtml(tr("Click and drag to select a region to select it. Press <b>ESC</b> to cancel.",
+                                    "Help text shown when no region has been selected yet"));
+    adjustTooltipPosition();
+}
+
+void ScreenshotGrabber::useRegionSelectedTooltip()
+{
+    this->helperTooltip->setHtml(tr("Press <b>Return</b> to accept, or move and resize the selected region.",
```
TheLastProject: I don't quite like this text. Return is more commonly known as "Enter", "accept" is vague, there is no text about cancelling and the text implies you can't make a new selection altogether. What about changing it to this:

`Press <b>Enter</b> to send a screenshot of the selected region, <b>Escape</b> to cancel, or select a new region.`

src/widget/tool/screenshotgrabber.cpp
```diff
@@ -0,0 +1,183 @@
...
+    useNothingSelectedTooltip();
+    
+    connect(this->chooserRect, &ScreenGrabberChooserRectItem::doubleClicked, this, &ScreenshotGrabber::acceptRegion);
+    connect(this->chooserRect, &ScreenGrabberChooserRectItem::regionChosen, this, &ScreenshotGrabber::useRegionSelectedTooltip);
+    connect(this->chooserRect, &ScreenGrabberChooserRectItem::regionChosen, this->overlay, &ScreenGrabberOverlayItem::setChosenRect);
+}
+
+void ScreenshotGrabber::useNothingSelectedTooltip()
+{
+    this->helperTooltip->setHtml(tr("Click and drag to select a region to select it. Press <b>ESC</b> to cancel.",
```
TheLastProject: "select a region to select it" is very double. What about simply:

`Click and drag to select a region. Press <b>Escape</b> to cancel.`

Papierkorb: Agreed. I changed the texts to yours with little changes. Ty

Papierkorb:
So, the drawing artifacts should be gone now. Also fixed a off by one pixel error. It's actually surprising, that moving the selection is more costly than resizing, would've expected it to be the other way around...

zetok:
So, since PR didn't change in last few days, and last time I've downloaded patches is from `03-28`, there are still bugs with it.

groupchat screenshot - gl
↑ never-hidding screenshot button in groupchat

I'm not sure how to reproduce whole combination reliably, since it's a bug on top of another bug.

Well, first bug is that it **never hides**, steps to reproduce (haven't tested them a lot):
1. Move cursor between `attach` and `screenshot` buttons, and at some point move cursor outside of qTox window in a way that will leave `screenshot` button visible
2. Minimize qTox to tray by clicking on a tray icon
3. Restore qTox from tray by clicking on tray icon
4. You've just gained never-hiding `screenshot` button! :3 (it would hide after clicking on it, _but who would want to click on it anyway_, right?)

So, the second bug where it **changes its place**.. hmm, I've just been using qTox for some time, and it would seem that for some reason button first moved down, to appear above later on. Not quite sure what it causes, since I've changed chat windows, switched between chat windows and groupchat, opened settings panel..

Both of those bugs are reproducible.

Papierkorb:
Knock knock, would be great if someone would provide some feedback. Thanks

zetok:
I still can reproduce never-hiding `screenshot` button. A little bit harder, but.. :3

To reproduce, environment should allow you to close/quit given application using keyboard shortcut, or in any other way that doesn't involve doing anything with mouse.

Steps to reproduce:
1. Set `Close to tray` in settings
2. In contact window move cursor to show `screenshot` button
3. Use keyboard shortcut to close qTox window, which closes qTox to tray
4. Particular contact window now has never-hiding `screenshot` button

Other contacts (unless done separately for them) don't have never-hiding `screenshot` button, and groupchats don't have that problem.

If I may ask you, could your code have screenshot button for groupchats, but be "commented out", so that when groupchats will finally support sending / sharing files, all that would be needed is "uncommenting" that code to make screenshot button work in groupchats?

Papierkorb:
Nice catch. I'll just bind it to the hideEvent() of the widget, so if it's hidden, the flyout will be hidden. That should get rid of it once and for all :)

> If I may ask you, could your code have screenshot button for groupchats, but be "commented out", so that when groupchats will finally support sending / sharing files, all that would be needed is "uncommenting" that code to make screenshot button work in groupchats?

That will just work™: The flyout is enabled if the attach file button is enabled. There's a small method in ChatForm which opens the screen grabber, but it just needs to be moved to the GenericChatForm, just like the method used to select and send a file.

I thought about moving that to GenericChatForm, but for that I'd also need to move the "sendFile" signal (No idea why that's a signal) from the ChatForm and the GroupChatForm into the generic class. That'd be a nice thing, but yeah..

zetok:
I think that I just found a way to reproduce flying screenshot button. 

Steps to reproduce on 181d748c5b5c46a455906acc224515512aaad700:
1. Make it never-hiding
2. Resize qTox's window in i3wm  (not sure if it will "work" in other DE / WM, since i3wm is the only one under which I've tried)
↓
flying screenshot button

Papierkorb:
@zetok Fixed :3

Papierkorb:
Closing this PR, please see https://github.com/tux3/qTox/pull/1510 instead.

PR #1467, opened 2015-03-25
Blue status lights for groupchat widgets

Zer0-One:
The images have been sitting in /img for a while. I'm not quite sure why they haven't been used up to this point, but here you go.

### DIFF
--- a/src/widget/groupwidget.cpp
+++ b/src/widget/groupwidget.cpp
@@ -36,7 +36,7 @@ GroupWidget::GroupWidget(int GroupId, QString Name)
     : groupId{GroupId}
 {
     avatar->setPixmap(Style::scaleSvgImage(":img/group.svg", avatar->width(), avatar->height()), Qt::transparent);
-    statusPic.setPixmap(QPixmap(":img/status/dot_online.svg"));
+    statusPic.setPixmap(QPixmap(":img/status/dot_groupchat.svg"));
     statusPic.setMargin(3);
     nameLabel->setText(Name);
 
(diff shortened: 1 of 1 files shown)


ProMcTagonist:
They were used in the past, declared nonsensical, and disabled. Don't know why they're still in the resources.

Zer0-One:
@ProMcTagonist in lieu of another method of distinguishing regular friend widgets from groupchat widgets (horizontal line, etc..), I think that blue status lights work alright.

ProMcTagonist:
They're distinguished by #1371. Additionally, blue status orbs do nothing but add confusion. We had these discussions before and that's why they were removed.

Zer0-One:
I disagree. A cursory glance at the friends list doesn't make it obvious enough where the friends end and where the groupchats begin.

ProMcTagonist:
Groups have the cluster icon. Friends will nearly always have custom avatars, and even when they don't, the single body is different from the cluster icon. Groups are cleanly grouped at the top now.

And making the dots blue is arbitrary and weird. It doesn't mean anything. It's confusing. 

Furthermore, they don't need to be strongly differentiated further. With the grouping from #1371 you know where the groups are and where the friends are, and you scan for either titles, names, or new messages.

zetok:
Why not just remove indicator for groups, since it doesn't mean anything anyway?

Solves both problem of them not being marked, and strange green orb that with contacts means online

ProMcTagonist:
Then how would new group messages be indicated?

zetok:
@ProMcTagonist oh, I forgot about this aspect.. Thanks for pointing that out. Sadly I can't think of a way that would allow to handle this, so just disregard my idea.

tux3:
I'm not convinced that we want the blue orb back, so closing this for now.

Issue #1402, opened 2015-03-15
duplicate of the sender's name


PKEv:
Add the sender's name in the chat window if the interval between messages more than 2 minutes.
It is difficult to understand the message history especially if the sender does not uses a combination of shift + enter

dubslow:
You mean in groupchats...?

PKEv:
No, everywhere.
when communication is not only a dialogue but just reset the links, or one of the users do not otvechat long time (departed).
allows you to separate groups of messages sent to each other

tux3:
Implemented by #1470

PR #1470, opened 2015-03-25
#1402 duplicate of the sender's name

PKEv:
if interval more 5 minuts -> repeated the name of the sender
not tested for group chats!!!

### DIFF
--- a/src/chatlog/chatlog.h
+++ b/src/chatlog/chatlog.h
@@ -58,6 +58,8 @@ public:
 
     ChatLine::Ptr getTypingNotification() const;
     QVector<ChatLine::Ptr> getLines();
+    // repetition interval sender name (sec)
+    const uint repNameAfter = 5*60;
 
 signals:
     void selectionChanged();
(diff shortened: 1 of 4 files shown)


ProMcTagonist:
I think it would be cleaner just to make names handle like section headers in iOS alphabetized lists.

That would make sure you can always see who sent each block, no matter what (which this wouldn't even do) and without extra clutter.

PKEv:
Did not understand your idea.
this PR is in part easy viewing the history of messages if the messages are sent only unilaterally. And when the interval between messages is longer than 5 minutes then duplicated the name of the sender because it assumes that the subject is likely to have changed. Now the story sometimes looks weird and you have to figure out what's what be treated.

src/chatlog/chatlog.h
```diff
@@ -142,6 +144,8 @@ private slots:
     QMargins margins = QMargins(10,10,10,10);
     qreal lineSpacing = 5.0f;
 
+
+
```
tux3, line 148: Why the gratuitous whitespace :) ?

PR #1492, opened 2015-03-29
Suppresses warning

kushagrasurana:
Did not removed the argument as it might be usefull in some later pull requests.


tux3:
In doubt it's better to remove crap. If it turns out it's needed in another branch, we can add it back in the merge resolution.
I actually removed that argument completely while working on something else a while ago.

PR #1496, opened 2015-03-30
Fixes action bugs

kushagrasurana:
Bugs which this pr will fix.
1) action received gets modified on restart ( action is not saved properly in history )
2) sender's name is written twice if action is sent using offline messaging
3) /me is written when action is sent in groupchat having one peer
4) action sent is not being saved as action in last message history.


tux3:
Merged in https://github.com/tux3/qTox/commit/2954ee3c958575f6f88fe00beda73a47f600f066

It will reach master on Monday if all goes well.

PR #1500, opened 2015-04-01
Added tool menu active button indicator.

rohilsurana:
Now the button whose window is active has light background color and also the same button is disabled to remove hiding and repainting of ui each time it is pressed.


tux3:
Rebased on master and merged in 8d91997ae5308f783e8c5222713c66f22916bd24

Issue #1501, opened 2015-04-02
qTox does not compile due to missing header


quatrox:
src/toxme.cpp is missing #include <QtDebug>, hence it fails building.

Platform: Gentoo Linux
Qt version: 5.4

Adding the missing #include makes it build fine

quatrox:
Sorry for the mess.  I submitted a pull request instead

PR #1502, opened 2015-04-02
Make qTox build again

quatrox:
Without this #include, qWarning() cannot be streamed to (in Qt 5.4)


quatrox:
This fixes issue #1501

PR #1509, opened 2015-04-07
Added add friend on window on first start

rohilsurana:
When qTox starts it shows the 'Add friend' widget but it was not reflected in the title.


tux3:
Merged in https://github.com/tux3/qTox/commit/1c51cd788ba37df0c5c26b0e64d62afb4f15d832

PR #1518, opened 2015-04-10
Fixed cbGroupchatPosition overlay

mpxc:
Introduced by 27f58c9ca6922b322892a89d9287b130d3f2419e

### DIFF
--- a/src/widget/form/settings/generalsettings.ui
+++ b/src/widget/form/settings/generalsettings.ui
@@ -391,7 +391,7 @@ will be sent to them when they appear online to you.</string>
             </property>
            </widget>
           </item>
-          <item row="2" column="1">
+          <item row="2" column="2">
            <widget class="QCheckBox" name="cbGroupchatPosition">
             <property name="toolTip">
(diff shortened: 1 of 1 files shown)


tux3:
This renders strangely :

selection_027

I don't think that was the desired effect, closing for now.

mpxc:
Well in fact it was, because as it is now "groupchats" covers "faux offline" checkbox, making it invisible :-) Since it was my first PR here, I opted to just move it one column space to the right, and not waste space. I am assuming that the undesired effect was making it look like it was thrown too far to the right, but that's because of the other checkbox relations that need tuning-up as well. As a matter of a fact this is not the last UI change that I've made, few other things were sticking out like a sore thumb while using qTox... :-)

tux3:
> "as it is now "groupchats" covers "faux offline" checkbox, making it invisible"

Oh, that's a new bug, then.
Could you open a new issue with a screenshot and some info on what system you're running? 
I'm unable to reproduce at the moment.

mpxc:
Here it is: https://github.com/tux3/qTox/issues/1578
Looked a bit closer this time and you are right, it ends up in the spacer with this change. Best to put it below all of the others?

PR #1523, opened 2015-04-12
German translation updated. [de.ts]

AWeinb:
Hey,
i updated the german translation. A merge should be possible without problems.

-AxP


tux3:
Hi, there seems to be a lot of unrelated change along with your translation update.

The diff doesn't match the PR description. Please don't do that, ever.

I'll assume you only meant to send      da08459, I'll cherry pick it in https://github.com/tux3/qTox/commits/back_in_the_game and it'll reach master on Monday if all goes well.

AWeinb:
Ah, sorry, my mistake. Did not think that this would happen. I remember it for the next time.

tux3:
Okay, thanks :)

PR #1552, opened 2015-04-28
Fix for unwanted auto-away status removal

ghost:
At least windows version has a problem where auto-away is unset without user interaction. If user is gone for a prolonged period of time other people will see lots of "user is online now / user is away now" messages. OS incorrectly reports user activity. Proposed solution to this is to remove auto-away only on continous user interaction with PC. Simple counter counts how many timer interactions in a row can detect last input time occouring during last second. Two seconds of continous interaction mean user is back and auto-away is removed. As a side-effect this would protect against accidental auto-away removal when for example person bumps the table and mouse moves briefly.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1060,11 +1060,17 @@ void Widget::onUserAwayCheck()
             qDebug() << "Widget: auto away activated at" << QTime::currentTime().toString();
             emit statusSet(Status::Away);
             autoAwayActive = true;
+            continousActivity = 0;
         }
     }
-    else if (ui->statusButton->property("status").toString() == "away")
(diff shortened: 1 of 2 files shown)


tux3:
The problem of toxcore spamming user status callbacks isn't limited to auto-away (online/offline spam happens often too) and not a bug with the auto-away. Unfortunately, this PR fixes the wrong problem.

So, NAK and closing in anticipation of a better solution, either a fix in toxcore (unlikely to happen in a timely manner) or a filter in qTox's status callback.

ghost:
Oh, so this observed on other platforms than windows too? Because i noticed it only on windowses, figured workaround needed..

PR #1588, opened 2015-05-07
Fixing clang build on OSX

tucnak:
std::abs() is function from <complex>, current HEAD doesn't build on OSX with latest clang because of missing include. This commit fixes the build.

### DIFF
--- a/src/widget/tool/screengrabberchooserrectitem.cpp
+++ b/src/widget/tool/screengrabberchooserrectitem.cpp
@@ -16,6 +16,8 @@
 
 #include "screengrabberchooserrectitem.h"
 
+#include <complex>
+
 #include <QGraphicsSceneMouseEvent>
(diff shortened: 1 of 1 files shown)


tux3:
I think it'd be more appropriate to use the std::abs overload for integers in `<cstdlib>` rather than the full blown generic complex-number one in `<complex>`, since we just use it on a couple of ints.

Does it work with `<cstdlib>`, and if so can you update the PR?

tucnak:
It does work with with `<cstdlib>` indeed. Just committed it straight away. Am I right that github doesn't support squash PR merges, so I need to do some rebasing on it?

tux3:
Github should support whatever Git supports, if you do it on the command line. Although you might have to force push in some cases.

Thanks for the PR, merging it right now.

Issue #1594, opened 2015-05-09
OSX: A damn huge hall for improvements


tucnak:
Alright, so I spent some time for satisfying all dependencies (since default osx packaging approach is broken bullshit) and finally got a debug build. So far, inb4 even actual chatting, but just using the app, I noticed a few points that make qtox an absolutely unusable product for end user:
- Tray icon just isn't working. Sometimes it appears, sometimes it doesn't, but the actual icon is not being drawn. Tbh, I haven't seen it there (in the tray), ever. Status icons in tray menu are misrendered (what you get when you scale a 16x16 icon up to 32x32 representation). After pressing the tray icon, first, application window raises and after 2s (with enabled osx animations) tray menu triggers. I suspect creepy `SystemTrayIcon` you guys have there. By the way, why would you want to write manual support for all the backends, such as Gtk/Unity/KDE when your functionality doesn't go beyond helloworld QSystemTrayIcon API?
- Sometimes after leaving app focus, you can't raise it with clicking the dock icon (can't really reproduce everytime.
- Global application style is Fusion, looks completely creepy in context of OSX

I understand that there haven't been any sort of serious development for OSX and would like to help you out, but I'd need some hints. Can you please comment on the points listed above?

tux3:
First, thank you for trying things out.

> Tray icon just isn't working. Sometimes it appears, sometimes it doesn't, but the actual icon is not being drawn.

What is the tray icon backend you're getting, what does the qtox.log/stdout output say?
Unfortunately, I don't have a Mac available anymore so I can't do much testing on OS X.

> After pressing the tray icon, first, application window raises and after 2s (with enabled osx animations) tray menu triggers.

This makes it sound like you're getting the Unity systray backend, which is a horrible hack to work around some platform restrictions. If so, this'll need to be changed.

> By the way, why would you want to write manual support for all the backends, such as Gtk/Unity/KDE when your functionality doesn't go beyond helloworld QSystemTrayIcon API?

Because it turns out that QSystemTrayIcon doesn't actually work on several platforms. Including Unity (Ubuntu) and KDE5. We do have a QSystemTrayIcon backend to our system tray code, and we use it <a href="https://github.com/tux3/qTox/wiki/Tray-icon">where it works</a>. As Qt versions improve, we start using it more.

> Sometimes after leaving app focus, you can't raise it with clicking the dock icon (can't really reproduce everytime.

[...]

PR #1610, opened 2015-05-10
Fixing tray icon visibility issue on OS X

tucnak:
Some quirking since Qt doesn't render SVG tray icons on OS X. It won't leave a significant footprint on general performance and we can leave it as it is till the proper fix from the Qt's side.

Related issue: #1594

### DIFF
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -4,6 +4,8 @@
 #include <QMenu>
 #include <QFile>
 #include <QDebug>
+#include <QPainter>
+#include <QBitmap>
 #include "src/misc/settings.h"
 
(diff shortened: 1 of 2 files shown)


tux3:
Does it work better/more reliably than the normal icon? Does it work with both the light/dark icon?

tucnak:
Seems like it needs some more testing and improvements. Gimme a while!

tucnak:
> Does it work better/more reliably than the normal icon?

As long as working icon is better / more reliable then the not working one, yep, it does.

> Does it work with both the light/dark icon?

In fact light icon makes absolutely no sense on OS X while running light theme, since all white pixels are automatically translated to transparent ones in the actual tray bar. In case if you enable dark osx theme and put on "light icon" switch in qTox settings, yippy, it works as expected.

Issue #768, opened 2014-11-15
Add labels with corresponding values to sliders in Audio/Video Settings


cotti:
Sometimes fine-tuning (specially in case of video settings) is desirable, and a small label to the right of each slider reflecting its current value would do the job just fine.

PR #1621, opened 2015-05-11
closes tux3/qtox#768

agilob:
added values to slides in av settings, converted slot to qt5 style
audio sliders are disabled, they don't change anything anyway


tux3:
Thanks, I should have done that long ago!

Although if we're going to give a significance to `Audio::outputVolume` now, it's important that the volume be applied consistently (or someone somewhere is going to expect a low volume and get blasted).

Could you ammend your commit with an `alSourcef(alMainSource, AL_GAIN, outputVolume);` in `Audio::playMono16Sound` so nobody gets surprised/deafened by the notification/call ringing sound?

agilob:
sorry for the second commit, didn't think about that change :P

PR #1624, opened 2015-05-11
QFormLayout for Public Information

saneki:
Current:
qtox-old-crop

New:
qtox-new-crop

### DIFF
--- a/src/widget/form/profileform.ui
+++ b/src/widget/form/profileform.ui
@@ -39,8 +39,8 @@
        <rect>
         <x>0</x>
         <y>0</y>
-        <width>422</width>
-        <height>639</height>
+        <width>565</width>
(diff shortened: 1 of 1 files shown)


tux3:
I have no objections, but I'd like more opinions on the design change because I have no tastes.

agilob:
If you could make avatar be the same size as the form it would be great! Just fit it to height of the form.

TheLastProject:
When aligned correctly, it's surely fine. I'd like the colon to be removed though. So not "Name:" but simply "Name".

saneki:
Update (vertically aligned, avatar size is the same):
qtox-new-2-crop

@TheLastProject I added colons after the fieldnames for consistency, as they're used in the options forms.

Issue #968, opened 2014-12-29
Tray icon looks bad with Qt 5.4.0


zetok:
![qt5.4.0 a shit](https://cloud.githubusercontent.com/assets/3148759/5572411/e4850a80-8f95-11e4-9401-35b01bc490c3.png)

Was fine with Qt 5.3.2

agilob:
Can't reproduce with 5.4.0-1

zetok:
Have you tried with i3wm?

suhr:
Try using Qt 5.5.0.
PS: actually had this bug before with lesser qt version. May be qt bug.

agilob:
With KDE 4.14

stqism:
@suhr Isn't it a tad early for 5.5.0?

apprb:
![png](https://cloud.githubusercontent.com/assets/6405341/5585440/6c36a79a-90d6-11e4-8952-1c7e9628766d.png)
Qt 5.4.0-1 under awesome: bug is reproducible.

ArchangeGabriel:
No problem here with QT 5.4 under Plasma 5.1.2 with KF 5.5.

nbvcxz:
"bad" you probably mean white background of the icon. Confirmed - reproduced with Arch Linux 64 lxpanel, qt 5.4.0

nurupo:
Sounds related to [QTBUG-35832](https://bugreports.qt.io/browse/QTBUG-35832), which should have been [partially fixed in Qt 5.4.0](https://qt.gitorious.org/qt/qtbase/source/34ce66cd89ea1c618d8f63dd2d9b95aed0a81b11:dist/changes-5.4.0#L435), but it seem to have broken things for @zetok instead of fixing them : )

There is also another tray-related issue that is open [QTBUG-31762](https://bugreports.qt.io/browse/QTBUG-31762). They all are DE-related.

> Apparently the latest version of Ubuntu Gnome has also broken everything as well (a much lower priority for us).

So just get patient and wait till it's fixed in some future version.

tux3:
Is this still a problem with the new systray code and an up to date Qt version?

agilob:
@apprb could you check what's the different between other icons with transparent background and qtox icon?

Find and icon under /usr/share/application/...  and check its type with `file /path/to/a/file.extenstion`

**SOME** qtox icons are RGBA (colors + transparency), but most of the icons are just RGB (just colors, transparent is white).

```
➜ file img/icons/qtox-*          
[...]

PR #1628, opened 2015-05-11
converted images to RGBA

agilob:
@zetok @apprb @suhr  please try this branch if https://github.com/tux3/qTox/issues/968 is fixed.

Don't merge yet.


apprb:
png
Still the same.

PR #1499, opened 2015-04-01
Style sheet and some minor changes

kushagrasurana:
Corrected tab ordering in general settings
Added visual change of disabled settings
Stylesheet changes
faux offline messaging is visible now

PR #397, opened 2014-10-07
css styles for mainHead and mainContent in widget.cpp

agilob:
fixes problems with ugly UI on non-standard system theme (KDE example)
before: https://i.imgur.com/XJ0mbpi.png
after: https://imgur.com/a/UbgNz

PR #1629, opened 2015-05-11
Fixed settings tabbar style

saneki:
Current:
qtox-tab-old-crop

New:
qtox-tab-new-crop

It should look better for other styles as well, not just Fusion.

### DIFF
--- a/ui/settings/mainContent.css
+++ b/ui/settings/mainContent.css
@@ -71,11 +71,6 @@ QTabWidget
     background-color: white;
 }
 
-QTabBar
-{
-    background-color: white;
-}
-
 QScrollArea
 {
     background-color: white;


agilob:
[NOOOOOO](http://www.nooooooooooooooo.com/)



Please revert this commit.

tux3:
But it looks better!

agilob:
It does not.

tux3:
Does it not?

EDIT: Oh, I missed your edits.

saneki:
@agilob How did you make your tabs appear dark like that?

agilob:
Screenshot above ^

tux3:
@agilob Whatever you did to make half the theme dark and the other half light seems to be broken.
What did you do?

saneki:
Setting `QTabBar::tab { background-color: white }` might fix that..

agilob:
It's enough to have [photophobia](https://en.wikipedia.org/wiki/Photophobia) and use dark theme **everywhere**, otherwise, my eyes are bleeding after several minutes.

@tux3 I use [this](https://github.com/tux3/qTox/pull/1499#issuecomment-94318122) everyday, it's not 100% ready, it will be soon.

tux3:
@agilob Then maybe we should find a way to all of qTox comply with the theme?

agilob:
There are two ways:
1. remove all .css files, style will be 100% inherited from system theme
2. implement dark theme

I'm working on second solution, but when people change css or widget.cpp or main.cpp or setting tab they cause conflicts in my branch and it takes more time ;,,;

tux3:
I like what you've done with <a href="https://github.com/tux3/qTox/pull/1499#issuecomment-94318122">this</a>, but how does this PR break it?

Wouldn't removing more CSS help, so that we automatically fallback on the theme?

agilob:
@tux3 nope. This one doesn't, but see the screenshot above. It makes it even worse, [see my first pull request](https://github.com/tux3/qTox/pull/397) where I fixed that problem.

@saneki nope. This is 3rd try where someone reverses my changes, it makes me cry (literally, this contrast make my cry).

tux3:
@agilob .... but why is the settings form white on your screenshots here if it's dark [there](https://github.com/tux3/qTox/pull/1499#issuecomment-94318122) ?

I don't get it. Why don't you use your dark theme here? Why don't you set all of it to dark?

I'd love to help but I'm not sure what the problem is and how to solve it.

saneki:
@agilob What about `QTabBar::tab-bar { background-color: white }`? If that doesn't work, I'm not sure what else to suggest..

agilob:
When I'm working on qtox I use code from tux3/master, when I start my laptop, qtox from my private branch is started where I have 4/5 of dark theme implemented. I do it to avoid conflicts in the code, as a lot of code in widget.cpp, widget.ui, settings and chat had to be reimplemented. The screenshot is from tux3/master branch.

tux3:
So we should work on getting your dark theme ready and mainlined then, shouldn't we?
I don't think it should be on by default, but I wouldn't mind having it as an option either if possible.

agilob:
Everyone with dark system theme will see something similar to what I see now.

agilob:
fixed in https://github.com/tux3/qTox/pull/1630

agilob:
btw. The only difference between those screenshots is the line under tabs? I can't see anything more.

tux3:
Yes, the only difference for me is the border and the small shadow near it. It's not pixel-perfect.

I suppose it's better to revert this to not break the dark theme, but it'd be nice to have a real solution eventually.

tux3:
Reverted by c8ff943daeb40f5ac2a677038cacaa31e067cc26

agilob:
Thanks, I'll try to fix it soon and commit with my dark themes.

PR #1630, opened 2015-05-11
Fix in white theme for me

agilob:
fix for ugly theme, without reverting 567b0d4

before:
before

after:

### DIFF
--- a/ui/settings/mainContent.css
+++ b/ui/settings/mainContent.css
@@ -4,6 +4,12 @@ QWidget
     font: @big;
 }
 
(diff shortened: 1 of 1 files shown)


agilob:
You probably should test it before merging.

tux3:
It exactly reverts to the old behavior for me, so it's as good as reverting #1629 but with a couple more lines.

tux3:
Superseded by reverting #1629 in c8ff943daeb40f5ac2a677038cacaa31e067cc26

PR #1631, opened 2015-05-11
Another fix for settings tabbar style

saneki:
So I set a dark system theme and replicated the issue @agilob was having with pull request #1629. While Qt's CSS is a bit weird/unexpected in how it displays, setting a white QPalette to the QTabBar seemed to work for all styles except GTK+ with a dark system theme.


tux3:
@agilob can you confirm whether this works for you?

agilob:
Solution with css would work for all current and future tabs. Colors are ok in general settings, but there is still 'File transfer' tab, where tabs are dark. :+1: And this will be (probably) removed because it conflicts with dark themes (overrides CSS from other themes).

agilob:
In my PR 'dark themes' it looks like this
qtox-light

tux3:
Ok, then let's close this for now.

PR #1637, opened 2015-05-13
make /ME work

ProMcTagonist:
Getting excited and talking in all-caps shouldn't break actions


tux3:
It's probably better/faster/shorter to just pass `Qt::CaseInsensitive` as the second argument of `QString::startsWith` if you want to do that.

ProMcTagonist:
Thanks :)

PR #1654, opened 2015-05-15
removed file NURUPOWILLSUEME

msteinbeck:
I think we don't need this file anymore :)


tux3:
You need to update/remove the COPYING too, it still refers to NURUPOWILLSUEME.

...at least we'll still have RMS's face hidden at the end.

msteinbeck:
done.

tux3:
Oh no, where is stallman!

PR #1663, opened 2015-05-16
Style for comobobox in friendlist panel

agilob:
It looks like qlineedit which is on left side of the combobox.


tux3:
It looks great, except that "Friends" and "Groups" don't fit in the combobox anymore.

ss024

agilob:
oh

SkyzohKey:
We got this in Micro$hit Windows 8.1
screen

agilob:
qtox
That's better, it should be 50/50 now.

Combobox will have arrow-down there when dark theme branch is merged, the branch has nice arrow down icon.

@FishTFM is that screenshot from my branch? oO
Wow your font rendering is so ugly :<

Issue #1671, opened 2015-05-17
After language change the user should be notified to restart qTox


ghost:
Simple issue: If you change the language it isn't applied immediately. Instead of this you have to restart qTox.
But the "usability issue" which "occurs" here is that no restart message is displayed. The best message would be one where the user just has to click on "Yes" and qTox will restart automatically.

PR #1672, opened 2015-05-17
closes tux3/qtox#1671 restarting after changing lang

agilob:


src/widget/form/settings/generalform.cpp
```diff
@@ -210,6 +211,16 @@ void GeneralForm::onTranslationUpdated()
...
     Widget::getInstance()->setTranslation();
+    Settings::getInstance().save();
+    QMessageBox::StandardButton reply;
+      reply = QMessageBox::question(0,
+                                    tr("Do you want to restart now?"),
+                                    tr("You must qTox restart to change the language"),
+                                    QMessageBox::Yes|QMessageBox::No);
+      if (reply == QMessageBox::Yes) {
+          QProcess::startDetached(QApplication::applicationFilePath());
+          exit(12);
```
tux3, line 222: Why the nonzero exit status?

agilob, line 222: It doesn't restart with any other exit code, just closes and doesn't start again.

tux3, line 222: ...what?

That's obviously not right.

agilob, line 222: Could you try it? On my desktop any other code doesn't restart qtox

tux3, line 222: Yes, it works with any return code here.

The problem is not the return code. It's probably just the new instance racing with the current one. If it starts before we exit, it'll just maximize the current running instance (which is about to quit).
So we need to release the IPC's hold on the shared memory before starting the new process.

zetok:
Wouldn't it be better to make language switch on runtime?

tux3:
@zetok That's the ideal solution.

TheLastProject:
@agilob "You must restart qTox" is correct grammar, your current "You must qTox restart" is wrong.

PR #1675, opened 2015-05-17
Pulled out ToxID from corestructs.h

msteinbeck:
- Added separate file "toxid.h".
- Replaced the struct with a class.
- Renamed ToxID to ToxId. This might be pedantic but currently the source contains both: ID and Id. I think Id is more "natural" because otherwise variables like "id" should be named "ID" or "toxId" should be named "toxID". Moreover Id was used at first (for example getSelfId etc.).
- Renamed isMine to isActiveProfile because a user can have various profiles with different ids. The name isMine is not very precise here, isActiveProfile on the other hand is.

<del>I would like to remove the function isActiveProfile from ToxId, because this class just represents a tox id. It doesn't know anything about users or profiles. I think we should add another class named Profile which contains a ToxId. By getting the current active profile we can check if an id is equals to the id of the current active profile. </del>

Keep things simple.


tux3:
Please try to avoid pushing unrelated commit in your PR branches. I have to check every single one and that's annoying.

msteinbeck:
Usually I like staying on master, but will avoid it in future.

tux3:
It's fine to track master, as long as you do it in a way that doesn't pollute the PR.

PR #1684, opened 2015-05-19
Better error message for image resizing

zetok:
Now user will know what is the acceptable file size limit, and if
they don't feel satisfied, they can always report a bug.

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -213,14 +213,14 @@ void ProfileForm::onAvatarClicked()
     file.open(QIODevice::ReadOnly);
     if (!file.isOpen())
     {
-        GUI::showError(tr("Error"), tr("Unable to open this file"));
+        GUI::showError(tr("Error"), tr("Unable to open this file."));
         return;
     }
 
     QPixmap pic;
     if (!pic.loadFromData(file.readAll()))
     {
-        GUI::showError(tr("Error"), tr("Unable to read this image"));
+        GUI::showError(tr("Error"), tr("Unable to read this image."));
         return;
     }
 
@@ -242,7 +242,8 @@ void ProfileForm::onAvatarClicked()
     // If this happens, you're really doing it on purpose.
     if (bytes.size() > 65535)
     {
-        QMessageBox::critical(this, tr("Error"), tr("This image is too big"));
+        QMessageBox::critical(this, tr("Error"),
+            tr("The supplied image is too large.\nPlease use an image that is less than 64 KiB in size."));
         return;
     }
 


src/widget/form/profileform.cpp
```diff
@@ -242,7 +242,8 @@ void ProfileForm::onAvatarClicked()
     // If this happens, you're really doing it on purpose.
     if (bytes.size() > 65535)
     {
-        QMessageBox::critical(this, tr("Error"), tr("This image is too big"));
+        QMessageBox::critical(this, tr("Error"),
+            tr("This image is too big.\nPlease supply image smaller than 64 KiB, or try with different one.\nIf you think that your image should work, please report a bug with it attached."));
```
tux3: Some english mistakes as far as I can tell:
Please supply image -> Please supply an image
try with different one -> try with a different one

I'm not sure if "If you think that your image should work, please report a bug with it attached." should be in the error message in this case.
This error will probably never happen, so no need to handheld the user here. I'm not convinced that it's technically possible at all to craft a file that would trigger this error. It's probably impossible. If it happens, then you really tried very hard to make it happen and you can find the bugtracker by yourself :)

zetok:
Thanks, fixed mistakes.

I'm too not sure about last sentence, but.. I have no knowledge whether there won't be some legit image that could trigger this behaviour, and because of that worst-case scenario should IMHO be assumed - i.e. there **_will**_ be people who will run into that issue. Thus for a really rare edge case user should be encouraged to report a bug, along with attaching an image, so that edge-case problem could be fixed, rather than letting user think that "_qTox is broken, since it can't even use my picture_".

tux3:
> I have no knowledge whether there won't be some legit image that could trigger this behaviour

There will not. No legit image could possibly trigger this.

I took the most non-legit image I could, a colorful JPG photo that I tried to make it as bloated and poorly encoded and bloated as I could, with extra pointless metadata and all compression off, and I ended up with this:

ss039

It's 4.2kB and I really tried to make it big. The limit is 64kB, 15 times bigger.

If it happens, this isn't a legit image or a mistake, this is someone with way too much free time.
I don't know if it's possible, maybe it is, but it will be someone who knows exactly what they are doing.

zetok:
If there is consensus that it's not needed, then I'll remove it from PR, ok?

(from my own testing: I wasn't able to trigger this condition, but I haven't tried with many files since it's bothersome to set avatar again and again)

Zer0-One:
IMO you can trim this down to: "The supplied image is too large. Please use an image less than 64KB in size."

zetok:
updated ~according to @Zer0-One's suggestion

PR #1722, opened 2015-05-24
Update profileform.cpp

TheLastProject:
This fixes several issues caused by a stupid zetok fix.
- Doesn't tell the user to pick an image under 64 KiB when bigger images, quoting tux3, "work 100% of the time", thus, doesn't mislead the user
- Doesn't bother the user with unnecessary technical details

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -243,7 +243,7 @@ void ProfileForm::onAvatarClicked()
     if (bytes.size() > 65535)
     {
         QMessageBox::critical(this, tr("Error"),
-            tr("The supplied image is too large.\nPlease use an image that is less than 64 KiB in size."));
+            tr("The supplied image is too large.\nPlease use another image."));
         return;
     }
 


tux3:
And most importantly, this removes a whopping 28 bytes of bloat. uTox is finished.

zetok:
> This fixes several issues caused by a stupid zetok fix.

Status: `merged almost instantly`

@tux3 that makes me wonder what kind of contribution is encouraged…

tux3:
@zetok More like merged after 15 minutes of bikeshedding in -ontopic. Sorry that the commit message says your fix was stupid, I didn't write it.

Now I can see how we might not want to mislead people into thinking that they _have_ to give an image smaller than 64kB, so I suppose this commit is a tiny improvement.

But if we could all stop spending so much time arguing over trivial issues, like the proper way of writing that error message that will _never_ appear, that'd be wonderful. I'm okay with small changes and incremental improvements, but this isn't Wikipedia.

Issue #1173, opened 2015-02-15
[Regression] Broken quoting ( greentext )


zetok:

griffondor:
lol

ProMcTagonist:
Zetok, myself, and many others have demonstrated again and again the problems with the current implementation's odd, arbitrary, and confusing criteria for what should and should not be quotestyled.
It's a mess. 

Any line that begins with > or ＞ should be green. Simple.

suhr:
> Any line that begins with > or ＞ should be green. Simple.

Things like `>_<` must not be green. Super simple.

ProMcTagonist:
Why?

suhr:
Because it is not a quote. It is a text form of an emoticon.

ProMcTagonist:
Understood and agreed. But until a suitable way is found to accomplish that **without** breaking hundreds of normal ways quotes are used daily for the sake of a couple emoticons that average users will never see...

It's a matter of priorities.

suhr:
Actually, https://github.com/tux3/qTox/pull/1124 is a compromise. The string that doesn't have spaces, digits or letters is hardly a quote.

> couple emoticons that average users will never see

Wrong, this kind of emoticons is pretty popular.

ProMcTagonist:
> The string that doesn't have spaces, digits or letters is hardly a quote.

That's pretty arbitrary of you.

> Wrong, this kind of emoticons is pretty popular.

The "Average User" who's been thrown around as an excuse has graphical emoticons enabled and will therefore never see any text emoticons affected by quote styling.

suhr:
> That's pretty arbitrary of you.

Probably so, yet I'd prefer to find a compromise instead of saying NOTABUG.

> The "Average User" who's been thrown around as an excuse has graphical emoticons enabled and will therefore never see any text emoticons affected by quote styling.

Well, the main problem with the “Average User” is that he doesn't exist: there's different groups of users with the different preferences. And the group that uses `>_>`-like emoticons in text form is not that small to completely ignore it.

zetok:
![broken quotes](https://cloud.githubusercontent.com/assets/3148759/6227390/ecce2e38-b696-11e4-87d6-25328c357e7b.png)

@suhr

> And the group that uses `>_>`-like emoticons in text form is not that small to completely ignore it.

No, but it's a lot smaller than group affected by qTox failing at having consistent behaviour.
Consistent behaviour is much better than trying to provide "perfect" solution, and creating bugs.

[...]

Issue #1099, opened 2015-02-04
[Regression] Broken quoting ( greentext )


zetok:
![broken greentexting](https://cloud.githubusercontent.com/assets/3148759/6039184/143612d2-ac5c-11e4-8ab3-3219eddf51d2.png)

Please fix.

DanTheBritish:
#945 Isn't this intended behaviour?
Also related: #1059

suhr:
Not a bug. It is not supposed to greentext everything.
Add a space after `>` to ensure that message will be greentexted.

mindcombatant:
all of this discussion and disagreement about what should and shouldn't be greentexted makes me think it should be a per-user setting...

perhaps 3 radio boxes for greentext:
*OFF (default)
*ON
*CUSTOM [box for inputting your own regex] <link or hover-tooltip to description of how to format>

zetok:
@LapFox it's not an intended behaviour, since `>_some_underlined_text_should_be_green_` is not being greentexted

suhr:
But `> _some_underlined_text_should_be_green_` is greentexted perfectly.

ProMcTagonist:
Basically everything I said in #1059 still ands. 

Occasional green text smilies are not a problem; disallowing quoting of a large amount of things is.

On top of that, by default users have graphical emoticons displaying instead of text emoticons, and those aren't affected at all by quoting. So quoting has been broken for a lot of users, for no benefit at all to our average user, who happily sees cute yellow smiley faces no matter what. 

I agree that users should be able to choose if quoting happens in their client or not, but it should be on by default. Bare > walls are confusing, and other formatting interferes with the plain text. Coloring them is a perfect, quiet, innocent solution that works.

@suhr 

> But > _some_underlined_text_should_be_green_ is greentexted perfectly.

That's nice, but irrelevant, unless The Messenger For Everyone's new suggested workflow is scanning every quote and manually editing every instance of deviation from its rules on what should and shouldn't be quoted.

suhr:
> That's nice, but irrelevant

It is not. Consider `> text` as the true quoting syntax and `>text` just a syntactic sugar for 4chan'ers.

DanTheBritish:
E-mails for example use `> text` instead of `>text` and has been the standard in clients for a long time.

Therefore, `> _my_quote` is correctly green'ed, but even if it isn't green'ed, it's not like they're impossible to read. Other e-mail and IM clients don't colour at all.

[...]

PR #1732, opened 2015-05-25
Fix quoting; everything after '>' is a quote

### LINKED ISSUE #1059: Quoting is Broken

zetok:
Plenty of people voiced their opinion that everything after '>' should be a
quote. Trying to conform to anything other brings only chaos and makes
situation worse, since it is impossible to catch every exception.

There are already plenty of cases where regexp simply can't work with unicode.

In the future, qTox should have easy to use r-click options for 'Paste text as
 a quote' and 'Copy as a quote', which makes it a must to have quote detection
as simple as possible, to avoid situation where user wants to paste something
as a quote, and suddenly it doesn't seem to work.

Fixes #1059 #1099 #1173
Continuation of #1332 

If you want examples of things that can't be done with regexp, just read related issues / previous PR.

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -211,7 +211,7 @@ QString ChatMessage::detectQuotes(const QString& str, MessageType type)
         // don't quote first line in action message. This makes co-existence of
         // quotes and action messages possible, since only first line can cause
         // problems in case where there is quote in it used.
-        if (QRegExp("^(&gt;|＞)( |[[]|&gt;|[^_\\d\\W]).*").exactMatch(messageLines[i])) {
+        if (QRegExp("^(&gt;|＞).*").exactMatch(messageLines[i])) {
             if (i > 0 || type != ACTION)
                 quotedText += "<span class=quote>" + messageLines[i] + "</span>";
             else


SkyzohKey:
+1

tux3:
I personally don't mind either way, but what about smileys like >_< ?

zetok:
Personally, I like `>.<` to be green, it simply looks weird without colour

SkyzohKey:
Maybe should u make a strawpoll ? x)
EDIT: http://www.strawpoll.me/4453608/

suhr:
:-1:

Diadlo:
+

optimumtact:
:+1:

ayyyyyyyyyyy:
I feel it should be that everything on a _new line_ after a ">" ought to be green texted.

stal888:
:+1:

ProMcTagonist:
:+1:

SkyzohKey:
The strawpoll is self-explaining i guess. :3

suhr:
> In the future, qTox should have easy to use r-click options for 'Paste text as
> a quote' and 'Copy as a quote', which makes it a must to have quote detection
> as simple as possible, to avoid situation where user wants to paste something
> as a quote, and suddenly it doesn't seem to work.

I.e `>` + space. This _perfectly_ solves the problem and doesn't make an regression with things like `>_>`

zetok:
@suhr on the other hand, languages like Taiwanese don't use spaces at all.. Are you going to tell Taiwanese people that qTox is a special snowflake and simply requires spaces?

Your proposal doesn't solve problem in cases where people don't use spaces because it's faster to quote something using `>` alone.

suhr:
> http://upload.wikimedia.org/wikipedia/commons/e/ef/Standard_Taiwanese_Keyboard_Layout.svg

Actually, I can find the long blank key on that keyboard.

tux3:
The strawpoll (inb4 rigged) is at 16Y/3N, so democracy wins =]

PR #1733, opened 2015-05-25
use toxcore error codes

suhr:
Subj. I hope I haven't messed it up.


suhr:
Oh no I probable have.

tux3:
What?

suhr:
No, I haven't. But qTox segfaults on invalid proxy.

tux3:
Yes, the emit failedToStart(); seems to be broken, it probably needs to be made blocking.

Issue #1696, opened 2015-05-20
qTox saves screenshots in installation dir


ghost:
Tox saves the screenshots (at least on Windows) in it's installation dir. At least on Windows this is not nice.
At least if we set the default installation dir to %programfiles%\qTox as indicated in #1607 this also causes an error message as the %programfiles% dir is only writeable with admin rights.

So instead of using the dir of the qTux exe it should use the temporary dir or the data dir (%appdata%\Tox on Windows).
Or like it Skype does - in the dir the user choose for "autoaccepting" files.

agilob:
related: https://github.com/tux3/qTox/issues/1648 

question: as temporary solution, should it be saved in the same dir as toxprofile or ask OS for `Pictures` directory like it works on android?

zetok:
> as temporary solution, should it be saved in the same dir as toxprofile

yep, in
`~/.config/tox/screenshots`

PR #1739, opened 2015-05-25
Create dir for screenshots #1696

PKEv:
But the question remains - when to remove them?
correct #1696


tux3:
Not saving them in the first place would work just as well.

ProMcTagonist:
This is a complicated issue because it's already weird that file transfers aren't visible in the log after a restart. They really should be, they're a part of the conversation. And that means the corresponding files should exist as well, if I haven't moved them, including screenshots.

Issue #1731, opened 2015-05-25
Just lost all my profiles


agilob:
after update today I restarted qtox, my profile was removed. name of the profile was stored, but after each restart qtox creates new toxid. My contacts were also removed.
I can _load_ `agilob` profile but it doesn't change anything, after restart new toxid is created, previous is removed. name of the profile stays the same. It worked in the morning, stopped working in the afternoon after that https://github.com/tux3/qTox/commit/7c8334bca90bf77ac745c484d1485b9e30e5de70 commit

tux3:
I reverted the commit while we investigate.

tux3:
I can't reproduce yet. 

What precisely are the exact steps that lead to losing you profile?

agilob:
1. Update qtox
2. update toxcore
3. restart qtox
4. -100% profit

what I have now in .config/tox is:

```
➜  tox  ls -1
7119F8F3B55FCA07F78CC78003F9F9FCE70B2C0AEF3649F3A84C87EA56BE927F.txt
agilob.ini
agilob.lock
agilob.qtox_history
agilob.tox
agilob.tox.DI1692
agilob.tox.dN1666
agilob.tox.xG2338
avatars
data
qTox User.ini
qTox User.tox
qtox.ini
qtox.log
settings.ini
tox_save
```

none of the files have anything that would be useful. None of .ini files have my list of +20 contacts.
I can't load profiles because somehow there are removed before loading - something prevents profiles with duplicated names.

Files 

```
agilob.tox.DI1692
agilob.tox.dN1666
agilob.tox.xG2338
```

have 0 bytes in total -.-' How fucked am i?

tux3:
Note that the friends are stored in the .tox files, not in the .ini files.

I still can't reproduce. Please make a copy of your ~/.config/tox/ folder asap in the meantime, it should be possible to recover everything.

agilob:
`agilob.tox` is empty too.

tux3:
@agilob can you post the qtox.log, entirely?

Do you have any custom patch to toxcore or qTox?

agilob:
No. I update toxcore from `./bootstrap.sh`. It looks like a problem with toxcore, after removing folder `.config/tox` qtox creates new profiles in `/home/agilob` instead in .config/

qtox.log has only debug messages, nothing that would help -.-'

tux3:
@agilob Do you have "Make qTox portable" enabled in the advanced settings?

agilob:
no

[...]

PR #1751, opened 2015-05-29
core: adjust to change in tox_new() API

intelfx:
`tox_new()` has been changed in irungentoo/toxcore@8e80ced "Move savedata to options struct."

Probably related to #1731.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -179,9 +179,15 @@ void Core::make_tox(QByteArray savedata)
             toxOptions.proxy_port = proxyPort;
         }
     }
+    
+    if (!savedata.isEmpty()) {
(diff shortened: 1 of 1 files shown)


intelfx:
Ah, damn, I didn't see that this was already done and subsequently reverted.

intelfx:
Actually, I've encountered the same failure as in #1731 (irungentoo/toxcore@8a32b3a, tux3/qTox@a841224).

Then I suspected an ABI mismatch (which is correct, these versions do mismatch), attempted to rebuild everything (irungentoo/toxcore@29c37bf, tux3/qTox@6fff75b), which failed due to API mismatch, fixed it by myself, rebuilt again and now it all works. This is just FYI.

tux3:
@intelfx Thanks. I think I'll go ahead with this fix and see what happens.

PR #1759, opened 2015-05-31
Change max avatar size to 256kB

zetok:
Otherwise avatars look bad in clients that display them on bigger surfaces,
e.g. in Antox.

Patch needed for time being, in future there might be a switch to webp, which should make it possible to have avatars that are looking ~well even with really small size.

For details, look at related IRC chat: https://gist.github.com/zetok/6204b08ddda5a27b60be

Related PRs in other clients:
- https://github.com/Tox/toxic/pull/338

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -224,23 +224,29 @@ void ProfileForm::onAvatarClicked()
         return;
     }
 
-    // Limit the avatar size to 64kB
-    // We do a first rescale to 256x256 in case the image was huge, then keep tryng from here
+    // Limit the avatar size to 256kB
+    // We do a first rescale to 512x512 in case the image was huge, then keep tryng from here
+    // TODO:
+    //  * check whether 512x512 is really the best starting point
+    //  * better resizing algorithm needed, so that image will (if needed) be
+    //    resized to $max_avatar_size - 1
     QByteArray bytes{picToPng(pic)};
-    if (bytes.size() > 65535)
+    if (bytes.size() > 262143)
     {
-        pic = pic.scaled(256,256, Qt::KeepAspectRatio, Qt::SmoothTransformation);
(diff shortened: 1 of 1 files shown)


tux3:
I'm not fundamentally against increasing the size to 256k, but I think that not using PNG for everything would be a great step forward. 

I have to agree with irungentoo that VPX still images are probably the way to go.
Even just using JPEG would do wonders on some avatars, it's still a surprisingly relevant format.

Is there an example of an avatar where both VPX and PNG can't reach a decent quality in 64kB?

zetok:
> I have to agree with irungentoo that VPX still images are probably the way to go.

I agree with that too. Though, until using VPX for avatars won't be a standard, it'll be problematic. Thus, until standard will say that other format is to be used, this PR would serve as workaround.

Feel free to close it if there'll be standard saying that VPX is to be used instead of deprecated PNG.

> Is there an example of an avatar where both VPX and PNG can't reach a decent quality in 64kB?

I don't think so - even full hd images with vp8 looked quite decent, and even better with vp9 while maintaining size  of 64 KB. That being said, qTox resizing algorithm would have to change to fully utilize capabilities of vpx, since it would have to resize to size..

tux3:
> I agree with that too. Though, until using VPX for avatars won't be a standard, it'll be problematic. Thus, until standard will say that other format is to be used, this PR would serve as workaround.

True, although both VPX and larger avatars would break compatibility with other clients (uTox), so if we're going to break we might as well do it once and do it well.

> I don't think so - even full hd images with vp8 looked quite decent, and even better with vp9 while maintaining size of 64 KB. That being said, qTox resizing algorithm would have to change to fully utilize capabilities of vpx, since it would have to resize to size..

Good to know. The resizing algorithm could definitely be improved.

tux3:
Closing for now, since there's no consensus with other clients.

PR #1769, opened 2015-06-03
Correct redraw after filtering

### LINKED ISSUE #1757: visual bug when filtering
1. put any character in field "Search contacts" to filter at least one contact
2. change the value of the "combo box" so that the result was an empty list
3. on the screen will remain white rectangle

default

PKEv:
fix #1757
It looks a bit strange, but a redraw on another call fails. Moreover, order calling functions radically changes the behavior of the widget

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -889,8 +889,7 @@ void Widget::removeFriend(Friend* f, bool fake)
     if (ui->mainHead->layout()->isEmpty())
         onAddClicked();
 
-    contactListWidget->hide();
-    contactListWidget->show();
+    contactListWidget->reDraw();
 }
 
 void Widget::removeFriend(int friendId)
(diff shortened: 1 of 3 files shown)


src/widget/widget.cpp
```diff
@@ -1447,6 +1447,7 @@ void Widget::searchContacts()
 
     contactListWidget->hide();
     contactListWidget->show();
+    contactListWidget->resize(QSize());
```
tux3: Add a comment to explain why this hack is here, maybe?

PKEv: structure:
contactListWidget->hide();
contactListWidget->show();

It is found in the file widget.cpp three times, can make a separate function Update ()?

tux3: Yes. Don't call it update though, that's already taken. It needs a more descriptive name, and a comment so we don't forget why it's needed.

PR #1813, opened 2015-06-06
Improved readme ! 3:]

SkyzohKey:


ghost:
Even more offtopic: is funny how much energy is wasted in the uTox/qTox fight. The phrase should actually  read "Fuck Skype"

tux3:
NAK.

zetok:
@kehugter it's not a fight. What @FishTFM "proposed" here is IMHO simply disgusting.. and that's it, since we don't want me saying something ad hominem under @FishTFM address.

# 

Anyway.. there is no real fight between µTox and qTox. What is there, is competition - it allows both clients to be better, i.e. having something they can really compete against in their own little environment is good, since it encourages to keep both clients "better than the other one". At the same time, no feature parity allows for both clients to see whether feature that other client has is popular, and whether implementation in other client of it is the best that could be done - e.g. difference between the way screenshotting is done in µTox and qTox - µTox was first to implement that feature, and even though people liked it, it wasn't done well UI-wise. Being popular proved that it's needed in qTox, and a bit of time allowed to implement it in a much cleaner UI-wise way.

Also, scopes of clients differ - µTox aims to be lightweight client, while qTox has features.

ghost:
Thanks for the insight, @zetok.

Issue #1837, opened 2015-06-08
qTox switches tabs when receiving a new message


towlie:
This gets annoying when writing to other contacts. And could lead to privacy problems.
(Windows)
Settings:
![messagenew](https://cloud.githubusercontent.com/assets/6986141/8033138/609c396c-0ddc-11e5-9817-19f303979b13.jpg)

towlie:
This is caused by the show window option.
Show window shouldn't focus tabs.
 I could imagine that you want to read some messages later. And if you received a lot messages from other contacts then you will not know what you did read already if qTox behaves like this.

PR #1838, opened 2015-06-08
Fix tux3/qTox#1837

Ansa89:


Ansa89:
No more need for this.

Issue #1747, opened 2015-05-27
Notify if notification is outside of screen


TheLastProject:
With increasingly long friend lists, and friends sorted on name only, it's easy to sometimes miss notifications of friends because they fall out of the screen. qTox should display it if there are unseen notifications caused by contacts falling out of the screen.

Slack's "more unread" screenshot as reference:
![url](https://cloud.githubusercontent.com/assets/1885159/7844771/276f3856-04b3-11e5-813b-ff7c1ef13ef6.png)

tux3:
Or we could sort friends with notifications first temporarily, so the most important/urgent conversations are what you see first. This way you wouldn't have to go search through your contact list for that one friend with a notification. Thoughts?

TheSpiritXIII:
I'd rather have what tux3 mentioned. It saves more time than having to scroll all over the place.

Zer0-One:
I still think we should have a separate "tab" where friends are sorted by activity.

zetok:
@TheLastProject how is the way in which mockup handles this worse than what you're proposing?

@tux3 @TheSpiritXIII ↑ ?

TheLastProject:
Randomly resorting the sorted friend list is a horrible hack. @Zer0-One says what we need too, but the alphabetical list should remain alphabetical.

@zetok The mock-up doesn't even show any case of new message notifications being invisible because the contact list is too long and the contact falls out of the screen.

zetok:
@TheLastProject so? What you're describing is `default` view, and it should behave like that.

On the other hand, there also should be `Recent` view… Can you say what is worse about mock-up approach, compared to your proposal?

TheLastProject:
Again, I can't compare the mock-up because it does not deal with this issue at all... Do you even understand the problem? The problem is that my friend list is so long that when someone whose nick starts with z messages me, I often don't notice because they, and therefore their new message ring, are not visible in my friend list until I scroll down

On 28 May 2015 10:32:12 CEST, zetok notifications@github.com wrote:

> @TheLastProject so? What you're describing is `default` view, and it
> should behave like that.
> 
> On the other hand, there also should be `Recent` view… Can you say
> what is worse about mock-up approach, compared to your proposal?
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/tux3/qTox/issues/1747#issuecomment-106231700

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

zetok:
[...]

PR #1843, opened 2015-06-09
Notification edge marker

TheSpiritXIII:
This pull request adds notification edge markers as those described by #1747

It works by checking whether the chatroom that received a notification is on the screen or not, and if not, it adds an edge marker accordingly. Even if we have a recent list, a notification edge marker is still useful, such as when the recent list is long. Edge markers are removed when each chatroom is on the screen. Alternatively, users may click on them and the scroll area will move to the closest chatroom towards their chosen direction (and is not removed until all chatrooms have been seen).

Note, that this does not affect group chats unless alerts for group chats are enabled.

### DIFF
--- a/src/widget/notificationedgewidget.cpp
+++ b/src/widget/notificationedgewidget.cpp
@@ -32,7 +32,7 @@ NotificationEdgeWidget::NotificationEdgeWidget(Position position, QWidget *paren
     QHBoxLayout* layout = new QHBoxLayout(this);
     layout->addStretch();
 
-    QLabel* textLabel = new QLabel(tr("Unread message(s)"), this);
+    textLabel = new QLabel(this);
     textLabel->setMinimumHeight(textLabel->sizeHint().height()); // Prevent cut-off text.
     layout->addWidget(textLabel);
 
@@ -47,6 +47,11 @@ NotificationEdgeWidget::NotificationEdgeWidget(Position position, QWidget *paren
     layout->addStretch();
 
     setCursor(Qt::PointingHandCursor);
+}
+
+void NotificationEdgeWidget::updateNotificationCount(int count)
+{
+    textLabel->setText(tr("Unread message(s)", "", count));
 }
 
 void NotificationEdgeWidget::mouseReleaseEvent(QMouseEvent *event)
--- a/src/widget/notificationedgewidget.cpp
+++ b/src/widget/notificationedgewidget.cpp
@@ -1,0 +1,56 @@
+/*
+    Copyright © 2015 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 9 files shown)


zetok:
newest qTox fails to compile with this patch applied, using gcc-4.9.2

src/widget/notificationedgewidget.cpp
```diff
@@ -0,0 +1,54 @@
...
+    QLabel* textLabel = new QLabel(tr("Unread message(s)"), this);
+    textLabel->setMinimumHeight(textLabel->sizeHint().height()); // Prevent cut-off text.
+    layout->addWidget(textLabel);
+
+    QLabel* arrowLabel = new QLabel(this);
+    if (position == Top)
+        arrowLabel->setPixmap(QPixmap("://ui/chatArea/scrollBarUpArrow.svg"));
+    else
+        arrowLabel->setPixmap(QPixmap("://ui/chatArea/scrollBarDownArrow.svg"));
+    layout->addWidget(arrowLabel);
```
zetok, line 46: Just some nitpicking on coding style:
after 

```
if asdf()
    something;
else
    other_thing;
```

there should be an empty newline before any other code that isn't a part of conditional code.

src/widget/notificationscrollarea.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+void NotificationScrollArea::resizeEvent(QResizeEvent *event)
+{
+    if (topEdge != nullptr)
+        recalculateTopEdge();
+    if (bottomEdge != nullptr)
+        recalculateBottomEdge();
+    AdjustingScrollArea::resizeEvent(event);
```
zetok, line 114: same here

src/widget/notificationscrollarea.cpp
```diff
@@ -0,0 +1,201 @@
...
+}
+
+NotificationScrollArea::Visibility NotificationScrollArea::widgetVisible(QWidget *widget) const
+{
+    int y = widget->mapTo(viewport(), QPoint()).y();
+    if (y < 0)
+        return Above;
+    else if (y + widget->height() > viewport()->height())
+        return Below;
+    return Visible;
```
zetok, line 196: and here

ui/notificationEdge/notificationEdge.css~
```diff
@@ -0,0 +1,5 @@
+NotificationEdgeWidget
+{
+    background-color: #6bc260;
+    color: white;
+}
```
tux3: Please don't commit editor backup files!

zetok:
more contact pls
It would be nice if whole contact was shown, or bigger part of it (leave 1px not shown if you have to)

Also, notification stays on the bottom even if groupchats are deleted, and whole contact is shown:
y u no dissapear

TheSpiritXIII:
I have fixed both aforementioned bugs. Please confirm that this is now bug-free. :)

zetok:
pls new msg
↑ message from contact `:3`

ghost:
Idea: Instead of showing 'Unread message(s)', would it be possible to show 'message' if there's only one, and 'messages' if there's more? I don't think it would really make a difference, but it could be a nice touch.

ProMcTagonist:
@Tirifto +1

TheSpiritXIII:
@Tirifto It is doable but a translation file will have to be created for English, as the "Handling Plurals" section [here](http://doc.qt.io/qt-5/i18n-source-translation.html) suggests. I will add this capability.

zetok:
@tux3 merge pls? Or is there something that should be done?

tux3:
@zetok I need to find some time to test it and I'll merge.

PR #1874, opened 2015-06-15
Add friend name in removal dialog(fix #1860)

### LINKED ISSUE #1860: [Usability improvement] include name of to-be-deleted friend in pop-up confirmation
It's really bad when one wants to make sure that they're deleting right contact, and "confirmation" pop-up doesn't even include name of contact that is going to be deleted…

Every time I delete contact, I'm really worried that I might have misclicked.

Please fix.

PKEv:


Zer0-One:
Could you make this say "Are you sure you want to remove <name> from your contact list?"

PKEv:
In this dialog, two problems solved:
- Confirmation of account deletion
- Confirm the deletion of the history of correspondence

Are you sure you should ask two questions to the user in a dialog box?

PKEv:
@zetok : What do you think about this?

ProMcTagonist:
"Delete history?" checkbox in the dialog that remembers what it's set to?

Issue #1704, opened 2015-05-21
autoaccept in settings doesn't work


Doom032:
What I did:

1) Check the "Autoaccept and save files" checkbox and select an existant folder(/home/thomas/Downloads)
2) Send a few inline images from my µTox account to my qTox account

What I expected:

It automatically accepts the images and saves them in Downloads

What actually happened:

I see the "Accept to receive this file" prompt a few times, and the file isn't accepted and downloaded automatically.

zetok:
I also am affected by that bug on qTox a841224683116e75852d6d2ebc9a9f1dc70ec957.

agilob:
I use it everyday from Antox to qtox and it always works.

Inline images? We don't support that, I didn't know it's a thing

zetok:
@agilob there isn't such thing as inline images, but some µTox people call screenshots that..

Doom032:
@agilob For clarity, this is about the global Autoaccept setting, not the per-user setting, which does work for me. Not sure if you tested that one.

tux3:
Fixed by #1890

Issue #775, opened 2014-11-16
Auto-accept and corresponding folder not being remembered.


raisingw:
Using the GUI to apply auto-accept and the folder is forgotten after closing and reopening. Manually specifying these settings in settings.ini does work and they show on next load. Ubuntu 14.10.

dubslow:
Do you mean the settings page button, or right clicking on contacts? I'm going to guess the former.

raisingw:
Settings page, yes. Contacts right click is remembered.

dubslow:
Err... this patch (move the global settings after the per-friend loop) seems to fix the issue.

```
git diff
diff --git a/src/misc/settings.cpp b/src/misc/settings.cpp
index 0770c42..012fa82 100644
--- a/src/misc/settings.cpp
+++ b/src/misc/settings.cpp
@@ -317,10 +317,10 @@ void Settings::save(QString path)
     s.endGroup();

     s.beginGroup("AutoAccept");
-        s.setValue("autoSaveEnabled", autoSaveEnabled);  
-        s.setValue("globalAutoAcceptDir", globalAutoAcceptDir);
         for (auto& id : autoAccept.keys())
             s.setValue(id, autoAccept.value(id));
+        s.setValue("autoSaveEnabled", autoSaveEnabled);
+        s.setValue("globalAutoAcceptDir", globalAutoAcceptDir);
```

I have no idea what the problem is or why this fix works. I did check that at least one friend's individual folder is also correctly written with this patch.

@tux3 Merge this manually I guess if this doesn't bother you

apprb:
problem is in these lines:

```
for (auto& id : autoAccept.keys())
    s.setValue(id, autoAccept.value(id));
```

and

```
for (auto& key : s.childKeys())
   autoAccept[key] = s.value(key).toString();
```

because `autoSaveEnabled` and `globalAutoAcceptDir` params are also stored in `autoAccept`.

dubslow:
Ah, when I added the AutoAccept group I meant for it to stay separate specifically because of that issue lol.

As long as we put any further items after the for loop in the writing case, I'm inclined to just use the patch above.

Otherwise we can condition the loop body, or maybe move some of these settings to a different group.

zetok:
[...]

Issue #755, opened 2014-11-15
Windows crash on startup


dubslow:
http://pastebin.com/FJnNBkQA

bricky149:
No crashes on Windows here and the pastebin's been removed.

dubslow:
I guess I'll close as insufficient data, or possibly fixed by the user and thus spurious.

PR #1890, opened 2015-06-20
Codecleanup

agilob:
fixes issues with global autoaccept:
if per-contact is specified then checks if global autoaccept is enabled, is so, saves to global autoaccept dir.
opens new file chooser in home dir instead in current location of qtox (/home/agilob/projects/qtox/)

removes unused attributes from settings.h

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -190,10 +190,7 @@ void Settings::loadGlobal()
     s.endGroup();
 
     s.beginGroup("GUI");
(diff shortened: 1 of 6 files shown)


agilob:
closes #1704 and #775 

edit.  oh well it doesn't actually solve #755 instead new file chooser is opened in home directory. when user cancelled file chooser we don't get it's last location so it can't be saved.

tux3:
I'm not sure what this has to do with #755?

Issue #1668, opened 2015-05-17
About window?


ghost:
A simple "issue" but I would really like to see an about window (as a part of the existing one-window-style).
So maybe the advanced settings are something like this - there is at least a license.
However I can't see a version number or something like this anywhere... :expressionless:

ghost:
+1
Is a crucial thing to have. Quite often do I need to make sure that my less tech-savvy contacts are using an updated version of qTox, but there's no way to get that info.

Commit hash as "version number" maybe?

ghost:
> Commit hash as "version number" maybe?

Yes I think this would be good as there is actually no real version number of qTox.

Zer0-One:
@blobgo I'll put this together sometime soon. Expect a PR in the next week or so

kiug:
Maybe is good idea to include information about status of version (alpha/beta/stable) in "about window".

zetok:
@kiug yep, good idea :3

ghost:
Besides the proposed idea, what do you think about a chat command that outputs the same information? So, instead of giving the instructions on where to find it, one could simply say "copy and send me this please /qtox-info".

zetok:
@kehugter 

> Besides the proposed idea, what do you think about a chat command that outputs the same information?

no.

Zer0-One:
@kehugter nah, it's not really necessary.

zetok:
![now I know](https://cloud.githubusercontent.com/assets/3148759/8267031/797ddaca-1747-11e5-81a4-5f0fd80804c5.png)

would be good, if not for 1 thing;

why are you linking to some other repo, when qTox wiki has section too for bug reporting?

PR #1895, opened 2015-06-21
Added Proper About Window

Zer0-One:
This PR adds a proper "About" tab to the settings menu, closing #1668 

about
advanced
The license text has been removed from the Advanced tab, and moved to the About tab

### DIFF
--- a/src/widget/form/settings/aboutform.cpp
+++ b/src/widget/form/settings/aboutform.cpp
@@ -1,0 +1,46 @@
+/*
+    Copyright © 2014-2015 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


src/widget/form/settings/aboutform.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+AboutForm::~AboutForm()
+{
+    Translator::unregister(this);
+    delete bodyUI;
+}
+
+void AboutForm::retranslateUi()
+{
+    bodyUI->retranslateUi(this);
```
tux3, line 45: This will restore the commit hash label to what you set in Qt Designer when the language is changed, you need to copy the `bodyUI->label_4->setText(bodyUI->label_4->text().replace("GIT_VERSION", QString(GIT_VERSION)));` line here.

Zer0-One:
@tux3 whoops! fixed it.

PR #1907, opened 2015-06-25
translate "aboute page" in Russian

PKEv:


suhr:
Remove obsolete items, please (`-no-obsolete` option of lupdate).

ghost:
@suhr should all translations use the `-no-obsolete` option? If yes, the command in the wiki should be updated

https://github.com/tux3/qTox/wiki/Translating#if-youre-just-updating

suhr:
It's better to use `-no-obsolete` after translation, but not before.

ghost:
I don't understand when I should use it. My workflow is

`pull` -> `lupdate` -> `linguist` -> `push`

should I `lupdate -no-obsolete` without other arguments after `linguist`?

suhr:
pull → `lupdate` → `linguist` → `lupdate -no-obsolete qtox.pro -ts ‹path to ts file›` → push
This allows you to see (and use) obsolete translations in qt linguist, but get rid of them when everything is translated.

PR #1908, opened 2015-06-25
Windows bootstrap ffmpeg

ghost:
This includes patch for ffmpeg because vanilla source does not build. Due to some reason`windows.h` has to be included before `unistd.h` to make compiler happy.

Also i am sneaking in commit that re-aligns bootstrap.sh with 4 spaces. No idea why i used 3 spaces before. Oh well, thats fixed now.


tux3:
I'm surprised ffmpeg wouldn't build, have you tried reporting it upstream?

Issue #1861, opened 2015-06-13
lupdate "Qualifying with unknown namespace/class"


ghost:
Started happening with commits from about 3-4 days ago.

`lupdate -pro qtox.pro -ts translations/es.ts` outputs

```
/home/path/qTox/src/widget/form/settings/advancedform.cpp:39: Qualifying with unknown namespace/class ::AdvancedForm
/home/path/qTox/src/widget/form/settings/avform.cpp:142: Qualifying with unknown namespace/class ::AVForm
/home/path/qTox/src/widget/form/settings/generalform.cpp:98: Qualifying with unknown namespace/class ::GeneralForm
/home/path/qTox/src/widget/form/settings/generalform.cpp:305: Qualifying with unknown namespace/class ::GeneralForm
/home/path/qTox/src/widget/form/settings/generalform.cpp:361: Qualifying with unknown namespace/class ::GeneralForm
/home/path/qTox/src/widget/friendwidget.cpp:60: Qualifying with unknown namespace/class ::FriendWidget
/home/path/qTox/src/widget/friendwidget.cpp:175: Qualifying with unknown namespace/class ::FriendWidget
/home/path/qTox/src/widget/friendwidget.cpp:266: Qualifying with unknown namespace/class ::FriendWidget
/home/path/qTox/src/widget/tool/chattextedit.cpp:70: Qualifying with unknown namespace/class ::ChatTextEdit
/home/path/qTox/src/widget/widget.cpp:216: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:377: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:388: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:419: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:433: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:444: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:512: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:521: Qualifying with unknown namespace/class ::Widget
/home/path/qTox/src/widget/widget.cpp:629: Qualifying with unknown namespace/class ::Widget
[...]

PR #1914, opened 2015-06-26
translate to russian "contacts groups"

PKEv:


PKEv:
after load i see that names of tab not translated...
but in "ru.ts" all was translated... 
What could be the problem?
default

tux3:
They aren't translated properly. I opened the ru.ts and none of them are validated.

ss013

Did you forget to click the validate translation button (green check mark) ?

PKEv:
correct, but this don't help... 
default
perhaps this is due to #1861

tux3:
That's interesting, since it still works in every other language.
Have you tried removing the "final" specifiers in the "class XXX final : public GenericForm" lines?

PKEv:
yes, removing the "final" specifiers  is help

Issue #1871, opened 2015-06-15
[Regression] No way to delete whole history


zetok:
I don't remember how exactly it was doable before, but I remember that it was possible to remove history at least through: `encrypt history` →  `remove password` → `delete history`. 

Even if it wasn't intuitive, one could have removed history from within qTox's interface.

History shouldn't be deleted just because someone unticks "Keep history".
There should be a separate button for it, with a confirmation dialog to make sure that user didn't press it accidentally.

Perhaps located in `Advanced` settings?

linux-modder:
if its that important back it up  non locally  imo.

zetok:
> if its that important back it up non locally imo.

@linux-modder what that has to do with this issue?

kangasbros:
"History shouldn't be deleted just because someone unticks "Keep history".
There should be a separate button for it, with a confirmation dialog to make sure that user didn't press it accidentally."

I disagree. It most definitely shouldn't keep history if the keep history box is unchecked. Mainstream users will be very nervous, if the app keeps history even when the box is unchecked. I at least freaked out when I saw that the app (qtox) actually keeps logs even though I had the box unchecked.

zetok:
@kangasbros
I see.
There's other side of the coin though - users might accidentally click, unchecking `Keep history` box.

To prevent accidental removal of history, there should appear a confirmation window, with a question whether user wants to remove all history so far, or only stop saving new history.

Would that ↑ be a good middle ground for removing history, while protecting user from accidental removal of history and allowing flexibility if user wants to keep old history, but doesn't want new history?

kangasbros:
Sounds like a good solution to me. 

On larger scale, I wouldn't see losing history as that critical for many users who use this kind of software.  It seems that most privacy-oriented chat software is quite keen on keeping the logs and storing them by default. At least I know large enough userbase who on default would not want to have the record. So, not having the history by default would nicely differentiate from the other solutions.

linux-modder:
+1 I could  see a value to the dialog

ProMcTagonist:
+1 for confirmation upon unchecking keep history

PKEv:
closed?

zetok:
yep

PR #1928, opened 2015-06-28
Add way to delete whole history #1871

PKEv:


PKEv:
look like this:
default

ProMcTagonist:
Change it to "Do you want to permanently delete all chat history?"

PKEv:
correct
default

Issue #1948, opened 2015-07-01
crash with sorting "by activity" 


PKEv:
Today found
When you select the sorting "by activity", qTox falls

win7 x32

PKEv:
Most likely it is due to damage to the ini file..

ghost:
Can reproduce using tux3's updated apt package (x64) - segmentation fault. Can this be because none of my contacts have been active since I updated to qTox having support to sort by activity?

TheSpiritXIII:
The reason why is interesting. It's because the last 7 days is not this month, so "This Month" does not get created, making the system get the wrong index for the "Unknown" activity (it goes 1 too far).

PR #1949, opened 2015-07-01
fix sorting "by activity" #1948

PKEv:
fix spaces


TheSpiritXIII:
Can you not use C style cast and use a C++ cast instead? static_cast might do.

PKEv:
ok

PR #1952, opened 2015-07-01
fix pointer's style

PKEv:


tux3:
Dynamic casts? Don't you mean static_cast?

PKEv:
sorry...

tux3:
Alright, no problem. Now that's still not going to compile because the syntax is wrong. You forgot parentheses.

And I'd almost argue that safe C++ style casts look a bit strange in the middle of all that faux-OO `void*` heavy C style code riddled with macros, but why not.

PKEv:
you're right

PR #1953, opened 2015-07-01
OSX main menu bar

TheSpiritXIII:
Adds a global menu bar for OS X instead of having an empty one. Includes common OS X shortcuts/menus, including preferences, about and quit, as well as a few window-based shortcuts.

This does not include some actions such as cut/copy/paste because they're a little bit harder to implement for OS X. It would require signals on each widget which is tedious work and will likely be a different PR.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -258,37 +258,32 @@ void Widget::init()
     new QShortcut(Qt::CTRL + Qt::Key_PageDown, this, SLOT(nextContact()));
 
 #ifdef Q_OS_MAC
-    mainMenu = new QMenuBar(0);
-    mainMenu->setNativeMenuBar(true);
+    QAction* windowMenu = Nexus::getInstance().windowMenu->menuAction();
+    QAction* fileMenu = Nexus::getInstance().globalMenuBar->insertMenu(windowMenu, new QMenu(tr("File"), this));
 
-    QMenu* fileMenu = new QMenu(tr("File"));
-    mainMenu->addAction(fileMenu->menuAction());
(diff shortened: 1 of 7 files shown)


tux3:
Your PR and commit message don't match what you changed. Don't do that.

It looks like you tried to fix the INSTALL.md, that's good, but put that change in a separate and properly named commit. I'm sure that was a mistake, but be careful please. Don't just hide unrelated changes in.

TheSpiritXIII:
The change you're talking about is what I told you about before where you merged my OS X install instructions too early. zetok told me to fix the readme format for OS X in the IRC channel to match the others. I fixed it and squashed it but I was a few minutes too late and it could no longer be merged properly.

tux3:
@zetok stop telling people to squash things, it messes with the history.

TheSpiritXIII:
Removed the INSTALL.md changes.

TheSpiritXIII:
Don't merge yet. Just realized that the menu bar doesn't play nicely with the login screen (since it's a global menu bar).

EDIT: It's working now.

src/widget/widget.cpp
```diff
@@ -342,6 +415,11 @@ Widget::~Widget()
     delete trayMenu;
     delete ui;
     instance = nullptr;
+
+#ifdef Q_OS_MAC
+    //Nexus::getInstance().globalMenuBar->clear();
+    //Nexus::getInstance().globalMenuBar->addMenu(Nexus::getInstance().windowMenu);
+#endif
 }
```
tux3, line 420: Can be removed?

PR #1960, opened 2015-07-02
README.md: update client image

zetok:


tux3:
I don't know, I liked the old image!

Opinions?

ProMcTagonist:
The new messages alert is good to show off, but the existing screens have better conversations.

agilob:
A few things changed in the UI, I think we should update it, but with better conversation.

tux3:
@zetok could you send a new client image with the old conversation?

zetok:
@tux3 no. While I _could_ do it, I won't.

Parts that I dislike about old conversation:
- "centralizing"
- Sean
- avatar (although that one doesn't really matter)

As for better conversation - anyone. feel free to make a screenshot of such that I could use to put in an image.. or make you own screenshot `:)`

Unless you want me to do ↓ with past conversation? Then I guess I could use it..
a) replace `Sean` with `Lee`
b) replace `Dubslow` with `Zetok`
c) replace "centralising" with something other, perhaps `Who let the bots out?` (better status message is needed for screenshot)

suhr:
> replace Sean with Lee

Stop advertising your bot.

> replace Dubslow with Zetok

Explain.

zetok:
> Stop advertising your bot.

Sure `:D`
But what else should be used?

> > replace Dubslow with Zetok
> 
> Explain.

That replacement is needed, since your own nick in conversations is **bold**

PR #1938, opened 2015-06-29
Multi-window - Separate Window for chats

TheSpiritXIII:
This pull request adds the functionality to detach chats from the main window. You can either detach manually by right clicking on friends or set a setting that detaches all windows by default.

Detached windows still have a friends list! You can chat with multiple friends on a single detached window. Of course, you can also detach friends from detached windows into their own window if you'd like. You're able to drag friends between windows, too!

PR #1961, opened 2015-07-03
Better contact removal dialog

ghost:
This dialog solves all confusion that arose from current verification method and is accident-proof. Current method asks if user also wants to remove chat history. However confused user may not carefully read message and click "no" because he does not want to remve contact. However in this case contact is in fact removed but history is preserved. It is also open to possiblity that key smashing deletes contact by accident. This is very inconvenient because tox ids are long and hard to memorize. If someone removes contact by accident then he would need to find id of that contact in order to contact him/her. Sometimes this may be difficult.

New contact removal confirmation dialog prompts use to enter "YES" in the text boxin order to enable "OK" button so removal can be accepted. It also has checkbox for history removal. Dialog will work with translations. Russian user would need to enter "Да" in order to accept removal.

rmc

Please test if dialog placement is alright. On my machine its weird, not centered on parent widget. Any advice how to solve this would be greatly appreciated. So far i tried centering it with `move()` but no luck no matter what i do.

### DIFF
--- a/src/widget/tool/removefrienddialog.cpp
+++ b/src/widget/tool/removefrienddialog.cpp
@@ -1,0 +1,31 @@
+#include "removefrienddialog.h"
+#include <QPushButton>
+
+
+RemoveFriendDialog::RemoveFriendDialog(QWidget *parent, const Friend *f)
+    : QDialog(parent)
+{
+    setWindowFlags(windowFlags() & ~Qt::WindowContextHelpButtonHint);
+    setAttribute(Qt::WA_QuitOnClose, false);
+    ui.setupUi(this);
+    ui.label->setText(ui.label->text().replace("&lt;name&gt;", f->getDisplayedName()));
+    auto removeButton = ui.buttonBox->button(QDialogButtonBox::Ok);
+    removeButton->setEnabled(false);
+    removeButton->setText(tr("Remove"));
+    connect(ui.yes, &QLineEdit::textChanged, this, &RemoveFriendDialog::onTextChanged);
+    connect(ui.buttonBox, &QDialogButtonBox::accepted, this, &RemoveFriendDialog::onAccepted);
+    connect(ui.buttonBox, &QDialogButtonBox::rejected, this, &RemoveFriendDialog::close);
+    setFocus();
+}
+
+void RemoveFriendDialog::onAccepted()
+{
+    _accepted = true;
+    close();
+}
+
+void RemoveFriendDialog::onTextChanged(QString text)
+{
+    ui.buttonBox->button(QDialogButtonBox::Ok)->setEnabled(text == ui.yes->placeholderText());
+}
+
--- a/qtox.pro
+++ b/qtox.pro
@@ -34,7 +34,8 @@ FORMS    += \
     src/widget/form/settings/advancedsettings.ui \
     src/widget/form/settings/avsettings.ui \
     src/widget/form/settings/generalsettings.ui \
-    src/widget/form/settings/privacysettings.ui
+    src/widget/form/settings/privacysettings.ui \
+    src/widget/form/removefrienddialog.ui
 
 CONFIG   += c++11
 
@@ -493,7 +494,8 @@ SOURCES += \
     src/widget/genericchatitemwidget.cpp \
     src/widget/friendlistlayout.cpp \
     src/widget/genericchatitemlayout.cpp \
-    src/widget/categorywidget.cpp
+    src/widget/categorywidget.cpp \
+    src/widget/tool/removefrienddialog.cpp
 
 HEADERS += \
     src/audio/audio.h \
@@ -536,4 +538,5 @@ HEADERS += \
     src/widget/genericchatitemwidget.h \
     src/widget/friendlistlayout.h \
     src/widget/genericchatitemlayout.h \
-    src/widget/categorywidget.h
+    src/widget/categorywidget.h \
+    src/widget/tool/removefrienddialog.h
--- a/src/widget/form/removefrienddialog.ui
+++ b/src/widget/form/removefrienddialog.ui
@@ -1,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ui version="4.0">
+ <class>RemoveFriendDialog</class>
+ <widget class="QDialog" name="RemoveFriendDialog">
+  <property name="geometry">
+   <rect>
+    <x>0</x>
+    <y>0</y>
+    <width>300</width>
+    <height>180</height>
+   </rect>
+  </property>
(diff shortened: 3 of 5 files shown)


DanTheBritish:
The red is hard for me to read on that background, possible for it to be 1 or two shades darker? (Opinions needed)

ghost:
REMOVE or *remove* would probably be better, no extra colours would be necessary.

DanTheBritish:
What about just regular bold, rather than changing the styling of the letters?

ghost:
The contact name is already in bold, otherwise it looks strange.

suhr:
Too much protecting. Also, red `remove` is unreadable and takes too much attention.
But “Remove history” as a checkbox is a good idea.

agilob:
What's the difference between OK and YES?
I think is should be two buttons Remove | Cancel

ProMcTagonist:
@agilob YES is a text box.

PKEv:
I think flag "also remove chat history" must be setup by default

PKEv:
Typing "yes" is not quite intuitive, can add a label specifying what to do...

ghost:
For me "remove" is not hard to read at all. Besides "that background" is my desktop theme, dont think it would be everywhere like that ;) This dialog follows Qt theme that is set.

Regarding textbox - i would preffer less labels with explanations but i probably should add tooltips/help text (used with question mark button near maximize button). Besides its just like web2.0. Never seen textbox with gray text saying "e-mail" which disappears when you start typing in email?

suhr:
That red color (and any emphasis of the “remove” word) is not useful at all. Better to have “Remove” instead of “OK” as the confirm button.
Also, people who runs bots won't like “YES” typing. Can this be optional?

ghost:
Ok i guess it makes sense.

Err what bots? since when qTox is for bots?

suhr:
> since when qTox is for bots?

Since time people run bots and add them temporary as contacts in qTox.

ghost:
I dont get it..

zetok:
> For me "remove" is not hard to read at all

Why do you consider your own case as the only valid one?

suhr:
> I dont get it..
1. Run bot
2. Add it to your qTox
3. Use the bot
4. Halt the bot
5. Remove the bot from your qTox

ghost:
And you are saying this is done so often that this safeguard makes it VERY inconvenient to use qTox?

suhr:
I won't say “VERY”, but are you sure the safeguard is THAT useful?

ghost:
We already have some people who managed to do some keyboard-smash-contact-removal by accident. Current confirmation is certainly too lacking. This is more of a problem with tox because we can not memorize IDs of our friends and if one is accidentally removed then its cumbersome to re-add him/her. If you have better ideas how to make keyboard-smash-proof contact removal confirmation im all ears.

Regarding bots.. I am not sure its worth to add yet another option for 0.001% of total qTox users (if we are lucky). I think everyone would benefit more from universally good design rather than option.

One thing i can think of is confirmation slider, on one end being "No", on the other "Yes". But i have not seen such thing anywhere and i am not sure it would intuitive for everyone.

ProMcTagonist:
There's another issue open right now for having a list of recently removed friends you can easily re-add. Something to keep in mind when considering this removal protection at the cost of convenience.

ghost:
I would argue that such list is overengineering. If we remove contact means we do not want it. Would it not make more sense to prevent accidents instead of adding yet another rarely used feature? Besides how often do we actually remove contacts to call typing in 3 letter word inconvenient? I personally have not removed contact at least in a year (counting jabber too mind you). People close to me have not done it in months. To call it inconvenient i think we need better arguments.

@tux3 @dubslow care to state your opinions?

ghost:
Renamed contact -> friend to match rest of tox code (filenames, class names). Removed bold red "remove" and changed button from "OK" to "Remove".

rmc

zetok:
I applied this patch on top of patch from #1938, and as result:

```
In file included from src/widget/widget.cpp:51:0:
src/widget/tool/removefrienddialog.h:6:35: fatal error: ui_removefrienddialog.h: No such file or directory
 #include "ui_removefrienddialog.h"
                                   ^
compilation terminated.
Makefile:2503: recipe for target 'widget.o' failed
```

Same happens without #1938 applied.

ghost:
I did evil thing of overwriting this branch so not to trash commit history with these renames that should not be there in the first place. I bet if you delete that branch and check out again it would work. (because it works for me and i double-checked that no files are missing from commit)

P.S. please report how dialog placement works on your system. On mine its funky and im not sure why. But that surely needs solving.

zetok:
> I bet if you delete that branch and check out again it would work. (because it works for me and i double-checked that no files are missing from commit)

nope, I didn't checkout anything.

What I did, is simple `wget` patch from this PR.

And no, after getting new patch, it still doesn't work on top of master.

ghost:
Oh sorry, my bad. I had qTox.pro customized and i did not add stage it for commit because of those custom changes, completely forgetting about new aditions required for this PR. Now it should be good.

zetok:
@r-ku your commit message has gone.. MIA?

ghost:
Thank you for spotting this! Must be some ctrl-v accident. Fixed now.

suhr:
@tux3 @dubslow Merge it?

suhr:
cast @agilob

src/widget/tool/removefrienddialog.cpp
```diff
@@ -0,0 +1,30 @@
...
+{
+    setWindowFlags(windowFlags() | Qt::Tool);
+    ui.setupUi(this);
+    ui.label->setText(ui.label->text().replace("&lt;name&gt;", f->getDisplayedName()));
+    auto removeButton = ui.buttonBox->button(QDialogButtonBox::Ok);
+    removeButton->setEnabled(false);
+    removeButton->setText(tr("Remove"));
+    connect(ui.yes, SIGNAL(textChanged(QString)), this, SLOT(onTextChanged(QString)));
+    connect(ui.buttonBox, SIGNAL(accepted()), this, SLOT(onAccepted()));
+    connect(ui.buttonBox, SIGNAL(rejected()), this, SLOT(close()));
```
agilob: Could you change it to new Qt5 SIGNAL/SLOT syntax?
http://wiki.qt.io/New_Signal_Slot_Syntax

agilob:
Oh how nice! I left a comment about old singal/slot syntax, I'll merge it when it's fixed!

ghost:
Did as asked. Also figured out that weird dialog placement, now its ok.

antis81:
@agilob Think you want to merge this. Looks good to me.

PR #1988, opened 2015-07-10
Add missing ffmpeg libs and set correct depency order of them

abbat:
Add missing ffmpeg libs and set correct depency order of them (for static build). This fixes link errors like:

```
ld: warning: libavfilter.so.5, needed by /usr/lib/tox/libavdevice.so, not found
ld: warning: libswresample.so.1, needed by /usr/lib/tox/libavcodec.so, not found
```

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -165,12 +165,12 @@ win32 {
             contains(STATICPKG, YES) {
                 target.path = /usr/bin
                 INSTALLS += target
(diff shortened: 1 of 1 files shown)


tux3:
I haven't tried static ffmpeg builds in a while, but for dynamic builds this is not right, we don't use swresample and postproc, they shouldn't be linked with.

PR #1989, opened 2015-07-10
Add profile file name in profile page

### LINKED ISSUE #1951: No way to know profile's name
as the user name does not always coincide with the name of the profile that it would be convenient to the profile name displayed somewhere in the profile settings
+
Maybe add the name of the user in the window title - useful when you have multiple copies running qTox?

PKEv:
Add profile file name in profile page and Added indication of the profile folder on the profile page
fix #1951
default

(profile folder is clickable - opens the profile folder)

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -118,6 +118,12 @@ ProfileForm::ProfileForm(QWidget *parent) :
     Translator::registerHandler(std::bind(&ProfileForm::retranslateUi, this), this);
 }
 
+void ProfileForm::prFileLabelUpdate()
+{
+    Nexus& nexus = Nexus::getInstance();
+    bodyUI->prFileLabel->setText(tr("The name of the current file Profile: ") + nexus.getProfile()->getName() + ".tox");
+}
+
 ProfileForm::~ProfileForm()
 {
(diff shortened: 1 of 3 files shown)


ProMcTagonist:
Please change the text strings to

> Current profile:

and

> Current profile location:

tux3:
I second changing the text, I can't merge this as is.

PKEv:
I fixed strings.

src/widget/form/profileform.cpp
```diff
@@ -118,6 +118,12 @@ ProfileForm::ProfileForm(QWidget *parent) :
     Translator::registerHandler(std::bind(&ProfileForm::retranslateUi, this), this);
 }
 
+void ProfileForm::prFileLabelUpdate()
+{
+    Nexus& nexus = Nexus::getInstance();
+    bodyUI->prFileLabel->setText(tr("Current profile: ") + nexus.getProfile()->getName() + ".tox");
```
agilob: I would say there should be no spacebar after `profile:`, this adds confusion for traslators, some people will not translate with with spacebar, so it will look ugly. Please add spacebar by source code.

PKEv:
fix

PR #1994, opened 2015-07-12
Update .gitattributes

jscott0:
Fix GitHub being spelled github


tux3:
Why not, I suppose :)

PR #2005, opened 2015-07-15
Q_OS_LINUX -> Q_OS_UNIX: *nixes should behave similarly to Linux

yurivict:


tux3:
For v4l2, we only mean to compile it on Linux, Q_OS_UNIX would also include it for Mac OS

yurivict:
Will resubmit

PR #2010, opened 2015-07-16
Q_OS_LINUX -> Q_OS_UNIX: *nixes should behave similarly to Linux

yurivict:
Resubmitting (see https://github.com/tux3/qTox/pull/2005)

New logic is Q_OS_UNIX && !MACOS

Reason: FreeBSD and other BSDs should be included.

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -72,7 +72,7 @@ ChatLog::ChatLog(QWidget* parent)
     connect(copyAction, &QAction::triggered, this, [this]() { copySelectedText(); });
     addAction(copyAction);
 
-#ifdef Q_OS_LINUX
+#ifdef Q_OS_UNIX
     // Ctrl+Insert shortcut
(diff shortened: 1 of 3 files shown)


yurivict:
> V4L stands for "video for linux", right? Can you ensure this works for other platforms? Which are the target platforms?

At least on BSD V4L is supported through installing webcamd + some other support packages: libv4l and v4l_compat.

But ok, if this is the debate-causing issue I will just convert this into the FreeBSD port patch.

antis81:
@yurivict Sorry, I missed your last message as I was not subscribed automatically (why??) - now I am. I'm not starting a debate and I don't have deep knowledge on that topic. Just asking to make a bit more clear, what this is used for in the in-file comment. Do you have the implementation ready? Could you then maybe link to PR from here? Otherwise I don't see any reason to close this PR. It actually contains only good changes.

**EDIT** All good changes - except for the duplicate shortcut ... :). Can you please split commits?

PR #2024, opened 2015-07-21
Update INSTALL.md

wngr:
Added missing dependencies to build toxcore on fedora.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -163,7 +163,7 @@ sudo apt-get install libtool autotools-dev automake checkinstall check libopus-d
 
 Fedora:
 ```bash
-sudo dnf install libtool autoconf automake check check-devel libsodium-devel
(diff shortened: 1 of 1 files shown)


linux-modder:
possibly  re-word to indicate that more like @irungentoo's  comment on  /usr/local/   in  .../toxcore/.../INSTALL.md

linux-modder:
@zetok, might wanna show how that is bullshit especially when it builds JUST fine as written in me gist.

zetok:
@linux-modder https://github.com/tux3/qTox/blob/master/INSTALL.md#dependencies ← here you go, read this section in context of my previous comment.

linux-modder:
@ze

linux-modder:
@zetok  do you even use  Fedora, and only part that is even remotely relevant is the qt comment.... which IS valid

PR #2030, opened 2015-07-22
Translation: Spanish update

ghost:
Obsolete strings removed as discussed in https://github.com/tux3/qTox/pull/1907#issuecomment-116011549


suhr:
cast @zetok

PR #2042, opened 2015-07-29
Grammar correction for Norwegian

quatrox:
The Norwegian translation is suffering from English grammar rules.
This commit converts it to Norwegian grammar rules.


suhr:
cast @zetok

PR #2045, opened 2015-07-30
Fix friend list when day changes

### LINKED ISSUE #2043: "Activity" groups don't disappear when there are no members of it left
Description:
`Activity` "groups" are shown when there are members of it, but aren't hidden again when there are no members of it left - e.g. when when one talks to contact, it appears in `Today` "group", but after >24h when contact is no longer a member of this group, it still is shown. Additionally, it shows `1/1` members.

How to reproduce:
1. Message contact to make it appear in `Today` "group"
2. Wait >24h with no activity, so that contact could disappear from this group
3. Group still exists

Current behaviour:
"Group" doesn't disappear when there are no members of it left.

Expected behaviour:
"Group" should disappear when there are no members of it left.

Additional info:
muh members

Most likely other than `Today` "groups" are affected too.

@TheSpiritXIII

TheSpiritXIII:
Fixes #2043 but untested as of now because of the ridiculous steps to reproduce.

### DIFF
--- a/src/widget/friendlistwidget.cpp
+++ b/src/widget/friendlistwidget.cpp
@@ -31,6 +31,7 @@
 #include <QMimeData>
 #include <QDragEnterEvent>
 #include <QDragLeaveEvent>
+#include <QTimer>
 #include <cassert>
 
 enum Time : int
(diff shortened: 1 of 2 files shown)


agilob:
Thanks for the fix!

Issue #2048, opened 2015-08-01
Falling back to _tox.name.tld lookups doesn't work.


alexbakker:
As the title states, toxdns lookups after falling back to doing a _tox.whatever.tld lookup fail.

To reproduce this I removed toxme.se from the pinned servers list, recompiled and tried to lookup impy@toxme.se. After trying it a couple of times I got the following results: It prints to the console: `src/net/toxdns.cpp:195 : Warning: Server  toxme.se  returned no record, assuming the Tox ID doesn't exist` or it tries to fall back to tox1, which is not what we want.

My guess is that toxme.se doesn't return a result because it was unable to decrypt the request.

PR #2051, opened 2015-08-03
Fixed issue #2048

alexbakker:


alexbakker:
First of all, if 'pubkey' is 'NOT empty' it means that the server has responded to our request.

Now, if a _tox.toxme.se (for example) lookup request is done, the public key is sent in the response as a hex encoded string. What qTox was assuming however, was that the public key is sent in raw bytes, not encoded in any way. This caused actual tox3 lookup requests to that dns server to fail for obvious reasons.

That's how this fix works.

TheSpiritXIII:
@agilob do you approve this?

PR #2052, opened 2015-08-03
Disables maximize button on login window

agilob:


antis81:
:+1:

agilob:
Since we can now, please someone take a look at source code and merge if you don't see anything wrong. Test it on your machine if possible!

PR #2053, opened 2015-08-03
remove inherited style in rename profile window, closes #1674

### LINKED ISSUE #1674: Inconsistent style of popup windows
The "popup windows" which are doing only some small stuff (mostly just an edit/text box) are differently styled.
E.g. the "rename profile" window looks like a HTML-designed one while all password input boxes use system styles.

I think this should be generalized to one style.

agilob:


src/widget/form/profileform.cpp
```diff
@@ -247,10 +247,14 @@ void ProfileForm::onRenameClicked()
     Nexus& nexus = Nexus::getInstance();
     QString cur = nexus.getProfile()->getName();
     QString title = tr("Rename \"%1\"", "renaming a profile").arg(cur);
+
     do
     {
-        QString name = QInputDialog::getText(this, title, title+":");
-        if (name.isEmpty()) break;
+        QString name = QInputDialog::getText(0, title, title + ":");
```
antis81, line 253: Please revert this line. A dialog always requires a parent widget.

agilob:
@antis81 if you know any other way to fix this annoying issue, please make PR.

src/widget/form/profileform.cpp
antis81, line 253: @agilob If I understand issue topic right, you ~~don't~~ want to inherit parent stile right? What you need is set the flag Qt::WA_WindowPropagation. Just change to this:

``` c++
QString name = QInputDialog::getText(this, title, title + ":", QLineEditNormal, QString(), 0, Qt::WA_WindowPropagation);
// also check this window's flags, please
```

Might be Qt bug - let me check

antis81, line 253: Sorry, my mistake. It's a widget attribute, not a window flag. Hm ... only way I see is to implement it right (not use the convenience method):

``` c++
forever {
    QInputDialog dialog(this);
    // probably this one:
    dialog.setAttribute(Qt::WA_WindowPropagation, true);

    dialog.setWindowTitle(title);
    dialog.setLabelText(title + ":");
    int ret = dialog.exec();

    QString name;
    if (ret) {
        name = dialog.textValue();
    }

    // do the checks
}
```

Does this help?

agilob, line 253: Yes, some widgets (like frames mostly) inherit style from parent,  some inherit form OS (buttons and file browsers), we want to use OS style in such dialogs.

antis81, line 253: Then change the `setAttribute` line to false. It should use OS style then.

**EDIT** No Qt bug then. The attribute `Qt::WA_WindowPropagation` says: "Use style of parent window." (which is the default). If you remove that attribute, system style is used.

agilob, line 253: Nope, it didn't help
snapshot8

agilob, line 253: nope... 
snapshot9

antis81, line 253: Can you please also try disabling `Qt::WA_StyledBackground` and/or enable `Qt::WA_NativeWindow`?

agilob, line 253: None of these works. The only way I found a few months ago was to remove parent.

snapshot10

`dialog.setAttribute(Qt::WA_SetStyle, false);` doesnt work as well.

antis81, line 253: Removing the parent is no fix and raises another issue - dialog is mispositioned.

This issue is a bit less trivial, than it initially seems. It comes down to `QStyle` used in QTox (see file src/widget/style.h). This is set somewhere as application style and cannot be removed easily. I'm searching a way to tell the QTox style: "Don't apply to this widget." Will take some time though, sorry :(

agilob, line 253: > Removing the parent is no fix and raises another issue - dialog is mispositioned.

Never had this problem.

antis81, line 253: It's not a problem, it's a fact. You will get to the problem when using more than one monitor. You pretty much always want to give a dialog a parent, so it can be centred nicely to it - and not the screen (or worse inbetween two or more screens).

agilob, line 253: > You will get to the problem when using more than one monitor. You pretty much always want to give a dialog a parent

New dialog is always on parent window, even if parent is set 0. Never had this problem you described.

antis81, line 253: How do you think, a widget is centred on it's parent, when there is none?. See for yourself:
bildschirmfoto von 2015-08-06 09 27 45

Again, this is no solution for the actual problem here.

agilob, line 253: Again... as said twice before, I it is above parent (even if parent is null) my computer. I can't reproduce your issue...

antis81:
@agilob Sorry, if my reaction was somewhat rough. This is clearly not against you! It's about learning to solve problems right. I also had to look up lots of information about that topic and now I found a real solution.

Like I tried to write above, the solution to this problem is fixing the style and teach it to not style certain widgets. As we're using qss style sheets, this comes down to defining a css rule similar to this one:

``` css
.QDialog /* maybe a not(QMessageBox), not(QInputDialog), etc. */ {
/*
     QTox style rules here ...
*/
}
```

@tux3 @TheSpiritXIII It is better and easier to handle to implement a QTox application style by inheriting QStyle (QProxyStyle) instead of using using style sheets. @agilob Would you like to do that? You can get a template from [style.hpp](https://github.com/macgitver/libHeaven/blob/development/libBlueSky/Style.hpp) and [style.cpp](https://github.com/macgitver/libHeaven/blob/development/libBlueSky/Style.cpp) files in [libHeaven project](https://github.com/macgitver/libHeaven) (maintained by myself and a friend - @scunz FYI).

Code would look something like that:

``` c++
// Pseudo code

class QToxAppStyle : QStyle {
public:

  /**
   * If this function returns false, QTox uses the system style
   */
  static styledByQTox(QWidget* w) {
    if (w) {
      // example code:
      if (w->isWindow() && w->window()->windowFlags().testFlag(Qt::Dialog)) {
         return false;
      }
    }

     return true;
  } 
protected:
  // virtual overrides from QStyle. One example taken from own project:
      void Style::drawComplexControl(ComplexControl control, const QStyleOptionComplex* option, QPainter* painter, const QWidget* widget) const
    {
        if (!styledByQTox(widget)) {
            return QProxyStyle::drawComplexControl( control, option, painter, widget );
        }

        if( control == CC_ToolButton )
        {
            if( const QStyleOptionToolButton* buttOpt =
                    qstyleoption_cast< const QStyleOptionToolButton* >( option ) )
            {
                // this is just an example!!

                if (widget->parentWidget()->property("qtoxToolButton").toBool()) {
                    QStyleOptionToolButton opt = *buttOpt;
                    opt.palette.setColor( QPalette::ButtonText, opt.palette.color( QPalette::BrightText ) );
                    QProxyStyle::drawComplexControl( control, &opt, painter, widget );
                }
            }
        }
    }
```

Sorry, this is a little work. But it has to be done, if we want to gain full control about our styles here.

agilob:
Thanks, I'll take a look at it.

agilob:
Thanks, I'll take a look at it.

agilob:
I'll rewrite this over the weekend so closing for now.

antis81:
@agilob Any news on this one?

PR #2057, opened 2015-08-05
Changed toxcore urls to point to new buildserver in windows/bootstrap.sh

ghost:
So bootstrap script works again.


suhr:
cast @zetok

Zer0-One:
@r-ku why do you use --no-check-certificate?

zetok:
@Zer0-One IIRC that was the "recommended default" for the old jenkins - if that's not needed / something with new jenkins, feel free to open PR

Issue #1818, opened 2015-06-06
[Feature request] Add an option to set a default download directory without autoaccept


towlie:
So you just accept a filetransfer and the rest is done automatically.

TheLastProject:
While we're add it, I think it'd be good to have a default download directory, something like Downloads/Tox, be enabled by default already. You almost always want to save things in the same place anyway, so I feel it is a much nicer default.

linux-modder:
Vote to close on grounds this already exists  with  autoaccept files from
'john doe'   via right click on contact

Corey W Sheldon
Freelance IT Consultant, Multi-Discipline Tutor
(p) 310.909.7672
G+: https://www.plus.google.com/+CoreySheldon
LinkedIn: https://www.linkedin.com/profile/view?id=70127804
Github: https://www.github.com/linux-modder
Facebook: https://www.facebook.com/corey.sheldon
Several Communities on Stack Exchange https://www.stackexchange.com

http://www.facebook.com/1stclassmobileshine

Tutoring in person or via any of the following platforms:
HackHands https://www.hackhands.com
Wizpert https://www.wizperts.com
FieldNation https://www.fieldnation.com
AirPair https://www.airpair.com
Truelancer http://www.truelancer.com

## {PayPal,Google Wallet/Play store, Apple Pay}

pub  3072D/718BF597
http://pgp.mit.edu/pks/lookup?op=get&search=0xE958C5D6718BF597 2014-12-08
      Key fingerprint = 2930 99EB 083D D332 0752 88C4 E958 C5D6 718B F597

## uid     Corey Sheldon (Fedora Key) sheldon.corey@gmail.com

TheLastProject:
You know, @linux-modder, it would be good if you would actually read the title of the bug report to see that it was about being able to set a default download directory WITHOUT autoaccepting files. What would be even greater would be if your signature would not be over 10 times the length of your message.

ProMcTagonist:
I would like a default download directory of Downloads/Tox/%friendname

zetok:
@ProMcTagonist what should be done in case where friend name is composed entirely of characters that can't be used as a file (dir) name?

ProMcTagonist:
@zetok Stripping would be unavoidable. I prefer that to having all the files lumped together in one directory, and the Skype users I've talked to about it agree. Sticking all downloaded files in one place is actually a big complaint. Just an anecdote at the moment, we can do a survey if we need to.

ProMcTagonist:
[...]

PR #2062, opened 2015-08-07
Save file transfer accept default directory

TheSpiritXIII:
@Chiitoo requested on IRC that he would like the last used file transfer directory to be saved instead of having the directory always show the home directory. This small commit adds that functionality.

### DIFF
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -24,6 +24,7 @@
 #include "src/core/core.h"
 #include "src/widget/style.h"
 #include "src/widget/widget.h"
+#include "src/persistence/settings.h"
 
 #include <QMouseEvent>
 #include <QFileDialog>
@@ -143,6 +144,8 @@ void FileTransferWidget::acceptTransfer(const QString &filepath)
 
     //everything ok!
     Core::getInstance()->acceptFileRecvRequest(fileInfo.friendId, fileInfo.fileNum, filepath);
+
+    Settings::getInstance().setSaveFileDirectory(QDir(filepath).path());
 }
 
 void FileTransferWidget::setBackgroundColor(const QColor &c, bool whiteFont)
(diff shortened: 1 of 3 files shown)


agilob:
Doesn't it conflict now with global save dir? Will the new dir be reset after qtox restarts?

snapshot12

TheSpiritXIII:
@agilob The global save directory is used when the "Autoaccept files" check box is checked while mine only applies for manually saved files. Therefore, no, this does not conflict.

agilob:
No. Default directory to save files is global, should work for autoaccepted and manually accepted files.

agilob:
Well it looks like there is a bug with manually accepting files.

TheSpiritXIII:
I'm not sure that changing this to work with the default directory edit in the settings will produce desired behavior. After saving, users expect that the next time they save will be in the same location. With your suggestion, it will also be changing the location in the settings, which might confuse users.

agilob:
I think the assumption was to have one default directory for autoaccepted and manually accepted files, I dont mind your solution, but personally I would find it confusing to have two different saving locations, from which only one can be changed in setttings. 

You need to add path validator on start, check if path exists, if not set to home, users might delete folder when qtox is not running.

Chiitoo:
I feel the location for auto-accepted files indeed should be separate.  Perhaps the 'default location' could be renamed to make it clearly be about that?

That, or add a separate option for default /and/ auto-accept (yeah, I like options... I know others don't... much).

I'll try the patch soon!

towlie:
@TheSpiritXIII Does this fix #1818 ?

agilob:
@towlie No, it's already implemented, but apparently default download location for manual accept was removed during some refactoring. That's what I'm talking about, this PR is duplicating (bugged) feature.

Chiitoo:
Finally got around to applying this.  (I made the changes by hand on top of the multi-window changes because I can't tox without those!)

Haven't noticed any issues so far, though I haven't come up with a lot of things to try.

I did try what happens if I delete directories in-between saves, and it looks like if two(?) or more of the directories are deleted (for example a path 'asd1/asd2/asd3'), it seems to revert to the original location, which is nice, though for one directory deleted only and it seems to go to the one below that.

I wonder if the 'origin' is a bit weird being the binary directory on Windows for example (the place where qtox.exe lives).  I don't mind it, but I think a friend (who is a Windows user; I'm mainly on Linux, just testing on Windows now) suggested 'My Documents' I believe.

I think on Linux the 'origin' is “$HOME”?

Chiitoo:
Oh, I noticed something that might be unwanted: sending a file and saving a file seem to share the 'save slot'.  That is, after sending a file, saving the next file will start at the directory where the file was saved.  This at least on Linux... forgot to try on Windows while I still had it running.

Will try more later.

agilob:
I think this can be closed as auto-accept per friend and global are fixed now, for me, this functionality would introduce only confusions.

PR #2080, opened 2015-08-16
Fix compiling qTox against static Qt 5.5

zetok:
http://stackoverflow.com/questions/12802536/c-multiple-definitions-of-operator

PR thanks to @towlie


suhr:
cast @antis81

antis81:
Please move `inline` implementations to header.

PR #2083, opened 2015-08-17
Possible Fix for Tray Icon Detection

gitgud-software:
Call it a hunch, but XDG_SESSION_DESKTOP and XDG_CURRENT_DESKTOP aren't as interchangeable as they may have seemed. Notably, the XDG_CURRENT_DESKTOP environment variable in Ubuntu GNOME 15.04 (systemd woo) is set like this:

XDG_SESSION_DESKTOP=gnome

which is a variable qtox is expecting from XDG_CURRENT_DESKTOP. On the other hand:

XDG_CURRENT_DESKTOP=GNOME

The casing is different, obviously. Two solutions emerge: change the casing in those if statements, or change the environment variable we're checking. I've done the former.

Even if this solution doesn't work at all, my guess as to why the tray icons for all this time is this environment variable setup. Also to be noted is that XDG_SESSION_DESKTOP takes one item, whereas XDG_CURRENT_DESKTOP takes a colon-seperated list.

Some stuff about XDG_SESSION_DESKTOP is over here:

http://www.freedesktop.org/software/systemd/man/pam_systemd.html

### DIFF
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -28,7 +28,7 @@
 
 SystemTrayIcon::SystemTrayIcon()
 {
-    QString desktop = getenv("XDG_CURRENT_DESKTOP");
+    QString desktop = getenv("XDG_SESSION_DESKTOP");
     if (desktop.isEmpty())
         desktop = getenv("DESKTOP_SESSION");
     desktop = desktop.toLower();
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -28,7 +28,7 @@
 
 SystemTrayIcon::SystemTrayIcon()
 {
-    QString desktop = getenv("XDG_SESSION_DESKTOP");
+    QString desktop = getenv("XDG_CURRENT_DESKTOP");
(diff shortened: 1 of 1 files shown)


agilob:
I'll test it with KDE5 soon (before Tuesday) and merge if that works. I found a problem with tray icon and action menu from tray icon on KDE5, not sure if that's related to tray backend or styling...

gitgud-software:
I'm pretty certain that KDE isn't detected with the first XDG desktop check.

Right now I need to launch the app with env XDG_CURRENT_DESKTOP=GNOME before the qtox command. Setting it up for startup still causes problems though.

I think it's a problem with systems at this point.

agilob:
qtox doesn't start all all for me. it should start in tray, but it doesn't show any icon or window at all.

> [12:25:42.767] src/widget/systemtrayicon.cpp:126 : Warning: Detected a KDE5 session, but we don't have Status Notifier support. Disabling the systray icon

Now I don't have it in tray or as standard window. can't merge it.

agilob:
That will be fixed by my awaiting PR. Thanks for your contribution anyway!

PR #2092, opened 2015-08-21
Friend request and group invite buttons

TheSpiritXIII:
This PR replaces the current dialogs for friend requests and group invites and adds a button that brings you to a screen that lets you see current requests.

Friend requests are saved, so you can accept or reject a friend any time you want. It is now available under the "Add friend" button.

Group invites are only saved until you close qTox. This can change in the future, e.g. when Toxcore gets new group API merged and groups are persistent. Invites are now viewable under the "Add group" button. To compensate that the "Add group" button now takes you to a new screen instead of adding a new group immediately, I've added new context menus. You can right click on an empty portion of the friends list to create a new group or you can right click on a friend and immediately create and invite them to a group simultaneously.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -1062,19 +1062,24 @@ void Core::groupInviteFriend(uint32_t friendId, int groupId)
     tox_invite_friend(tox, friendId, groupId);
 }
 
-void Core::createGroup(uint8_t type)
+int Core::createGroup(uint8_t type)
 {
     if (type == TOX_GROUPCHAT_TYPE_TEXT)
     {
-        emit emptyGroupCreated(tox_add_groupchat(tox));
+        int group = tox_add_groupchat(tox);
+        emit emptyGroupCreated(group);
+        return group;
     }
     else if (type == TOX_GROUPCHAT_TYPE_AV)
     {
-        emit emptyGroupCreated(toxav_add_av_groupchat(tox, &Audio::playGroupAudioQueued, this));
+        int group = toxav_add_av_groupchat(tox, &Audio::playGroupAudioQueued, this);
+        emit emptyGroupCreated(group);
+        return group;
     }
     else
     {
         qWarning() << "createGroup: Unknown type "<<type;
+        return -1;
     }
 }
 
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -301,6 +301,21 @@ void Settings::loadPersonnal(Profile* profile)
         ps.endArray();
     ps.endGroup();
 
+    ps.beginGroup("Requests");
+        unreadFriendRequests = ps.value("unread", 0).toUInt();
+        size = ps.beginReadArray("Request");
+        friendLst.reserve(size);
+        for (int i = 0; i < size; i ++)
+        {
+            ps.setArrayIndex(i);
+            QPair<QString, QString> request;
+            request.first = ps.value("addr").toString();
+            request.second = ps.value("message").toString();
+            friendRequests.push_back(request);
+        }
+        ps.endArray();
+    ps.endGroup();
+
     ps.beginGroup("General");
         compactLayout = ps.value("compactLayout", false).toBool();
     ps.endGroup();
@@ -467,7 +482,22 @@ void Settings::savePersonal(QString profileName, QString password)
             if (getEnableLogging())
                 ps.setValue("activity", frnd.activity);
 
-            index++;
+            ++index;
+        }
+        ps.endArray();
+    ps.endGroup();
+
+    ps.beginGroup("Requests");
+        ps.setValue("unread", unreadFriendRequests);
+        ps.beginWriteArray("Request", friendRequests.size());
+        index = 0;
+        for (auto& request : friendRequests)
+        {
+            ps.setArrayIndex(index);
+            ps.setValue("addr", request.first);
+            ps.setValue("message", request.second);
+
+            ++index;
         }
         ps.endArray();
     ps.endGroup();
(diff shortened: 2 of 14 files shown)


agilob:
I think this can be merged, could you take a look @antis81? I don't want to be the only person who reads the code :<

src/core/core.cpp
```diff
@@ -1062,19 +1062,24 @@ void Core::groupInviteFriend(uint32_t friendId, int groupId)
...
+        int group = toxav_add_av_groupchat(tox, &Audio::playGroupAudioQueued, this);
+        emit emptyGroupCreated(group);
+        return group;
     }
     else
     {
         qWarning() << "createGroup: Unknown type "<<type;
+        return -1;
     }
 }
```
antis81, line 1084: @TheSpiritXIII Proposal: How about this ...

``` c++
int Core::createGroup(uint8_t type)
{
    // Note: I'd prefer an error code (enum => QTOX_ERR_UNKNOWN_TYPE)
    //  even better: a generic result class 
    // as we have a Group object providing the group's number.
    int group = -1;

    switch (type) {
    default: break;
    case TOX_GROUPCHAT_TYPE_TEXT:
        group = tox_add_groupchat(tox);
        break;
    case TOX_GROUPCHAT_TYPE_AV
        group = toxav_add_av_groupchat(tox, &Audio::playGroupAudioQueued, this);
        break;
    }

    if (group != -1) {
        emit emptyGroupCreated(group);
    } else {
        qWarning() << "createGroup: Unknown type "<<type;
    }

    return group;
}
```

src/persistence/settings.cpp
```diff
@@ -1379,6 +1409,48 @@ void Settings::setCircleExpanded(int id, bool expanded)
     circleLst[id].expanded = expanded;
 }
 
+void Settings::addFriendRequest(const QString &friendAddress, const QString &message)
+{
+    QMutexLocker locker{&bigLock};
+    QPair<QString, QString> request(friendAddress, message);
+
+    if (friendRequests.indexOf(request) != -1)
```
antis81: @TheSpiritXIII Please add brackets `{ }` for better code readability and less bugs. :)

src/persistence/settings.cpp
```diff
@@ -1379,6 +1409,48 @@ void Settings::setCircleExpanded(int id, bool expanded)
...
+        return;
+
+    friendRequests.push_back(request);
+    ++unreadFriendRequests;
+}
+
+unsigned int Settings::getUnreadFriendRequests() const
+{
+    QMutexLocker locker{&bigLock};
```
antis81, line 1426: @TheSpiritXIII Is this actually required on read operations?

Like I wrote in another context, getters/setters should not make use of any mutex in general. This might become a hard to find performance issue. Let me write down two (just logical, not usable) examples, to show what I mean:

``` c++
Settings s; // <--- a valied settings object
for (int i = 0; i < 100; ++i) {
     // mutex is activated 100 times, but we won't see it
     // means: stack is copied 100 times 
     s.getUnreadFriendRequests();

     // To create more drama, use other getters to read more settings.
     // assumption: Each getter has it's own mutex
     s.getFoo();
     s.getBar();
}
```

Without mutex in getter this would change to this (better) code:

``` c++
QMutexLocker(&bigLock);
Settings s; // <--- a valid settings object 
for (int i = 0; i < 100; ++i) {
     // assumption: mutex is NOT set in getter, but above instead
     s.getUnreadFriendRequests();
     s.getFoo();
     s.getBar();
}
```

If I'm not completely wrong, the mutex is accessed just a single time in second example ==> less to worry about ... :smile:

Feel free to prove me wrong (no offence taken). I like to learn for myself.

src/widget/form/addfriendform.h
```diff
@@ -32,6 +35,13 @@ class AddFriendForm : public QObject
 {
     Q_OBJECT
 public:
+    enum Mode
```
antis81, line 38: Proposal: Make use of `enum class`.

src/widget/friendwidget.cpp
```diff
@@ -68,17 +68,16 @@ void FriendWidget::contextMenuEvent(QContextMenuEvent * event)
...
     QMenu menu;
     QMenu* inviteMenu = menu.addMenu(tr("Invite to group","Menu to invite a friend to a groupchat"));
+    QAction* newGroupAction = inviteMenu->addAction(tr("To new group"));
+    inviteMenu->addSeparator();
     QMap<QAction*, Group*> groupActions;
 
     for (Group* group : GroupList::getAllGroups())
     {
-        QAction* groupAction = inviteMenu->addAction(group->getGroupWidget()->getName());
+        QAction* groupAction = inviteMenu->addAction(tr("Invite to group '%1'").arg(group->getGroupWidget()->getName()));
```
antis81, line 77: How about change the menu item names to `&Invite to`->`&new group`, etc.?

antis81:
@TheSpiritXIII :pencil2: [invalid] ~~Can you please cleanup and add the `{ }` brackets in if statements?~~

**EDIT:** Apart from the mutex thingy, which we need to clear, this looks good to me!

src/persistence/settings.cpp
TheSpiritXIII, line 1417: Omitting the brackets for single liners is the convention for qTox. See [the wiki](https://github.com/tux3/qTox/wiki/Coding).

antis81, line 1417: Ok, didn't know that. Thanks! @agilob FYI. sorry, my fault.

src/persistence/settings.cpp
TheSpiritXIII, line 1426: You are correct about the mutex overhead. I don't think you can get rid of them in getters/setters. Different threads may get/set at a time, for instance the audio is in a separate thread. Bad things will happen if they occur simultaneously.

antis81, line 1426: Hm ... not exactly sure, but how about using a `QSharedData` inheritance in combination with `QExplicitlySharedDataPointer`? Isn't that completely thread-safe? If that works, the technique could also be used for the audio data parser.

antis81:
@agilob @TheSpiritXIII I'm fine with merging this one - ignoring my own comments. Logically this is correct. If someone has a better solution to the mutex overhead, I'd be happy to hear about.

src/persistence/settings.cpp
TheSpiritXIII, line 1426: No, the aforementioned classes offer reference counting, which is not what the settings class needs. A possible alternative is to split up the settings into multiple classes, which the main settings class would access when it needs to, and using RAII, accessing theses classes could be  mutex protected. In general, I think qTox needs more looser coupling.

src/widget/form/addfriendform.h
TheSpiritXIII, line 38: Disagree, since the code uses the enum has indices for its tab widget.

agilob:
Huh we can't reopen PR -.-'

zetok:
@agilob yeah, you would have to revert the commit that reverted changes from this PR.

TheNain38:
Is someone working on getting it work?

PR #2106, opened 2015-08-23
fix #2103

### LINKED ISSUE #2103: ability to hide the window during a screenshot
when you send a screenshot by qtox that sometimes you would like that application window is minimized automatically as it takes a lot of space on your screen

PKEv:
ability to hide the window during a screenshot (by pressing spacebar)
fix #2103


PKEv:
I did not touch the old style if necessary I can everywhere change

PR #2110, opened 2015-08-25
Added toxme.io to the pinned servers list

alexbakker:


agilob:
Is any other client using it already?

PR #2130, opened 2015-08-28
White text on white bg on login screen is invisible

Fak3:
When used dark Qt theme, text color should be set explicitly for explicitly white background.


Fak3:
This is how it currently looks on my dark themed desktop.

plasma-desktopox1722

agilob:
I edited your comment to embed image. 

Thanks for your contribution!

Issue #1999, opened 2015-07-13
When outgoing ACTION messages, nick in the HTML tags will be parse.


quininer:
![snapshot86](https://cloud.githubusercontent.com/assets/6286761/8641897/cbd8f7aa-294e-11e5-80e2-dd08c994db0c.png)
![snapshot85](https://cloud.githubusercontent.com/assets/6286761/8641899/d0c2f2d4-294e-11e5-8d35-706196b320b6.png)

I think this and the previous title as parsing HTML is a need to fix the "feature"?

zetok:
Can reproduce.

agilob:
It's impossible to fix it as chat messages are printed using html, what we can do it to strip html tags from username.

ProMcTagonist:
There's no way to escape them before printing?

agilob:
I'll take a deeper look at this issue when rewriting themes...

TheNain38:
You could use `QString QString::toHtmlEscaped() const` http://doc.qt.io/qt-5/qstring.html#toHtmlEscaped before printing, so it would just escape them, not strip them

PR #2132, opened 2015-08-29
Fix tray icon on KDE5 and Gnome3

agilob:
On KDE5 qTox will use Qt backend instead buggy status-notifier, on KDE5, Qt 5.5 it looks like this now:

Commit fix some issues:
- SVG icons are used instead pixmaps
- transparent icon on Qt5.5 will be fixed (uses transparent SVG rendered qimage instead qpixmap)
- adds keyboard shourtcut to quit application, also KDE5 adds by default menu action row to Qt application - "Quit", so our Quit was renamed to Exit. Exit has system-default icon and system default keyboards shortcut. Difference will be visible only on plasma-shell (see screenshot) where `Quit` asks if you really want to quit application and exit just quits without asking.

Commits were tested on KDE5, Gnome3 by me and i3 by @zetok.

qtox9

Before:
snapshot2


agilob:
@antis81 could you test it on your desktop and merge?

src/widget/systemtrayicon.h
```diff
@@ -48,6 +49,8 @@ class SystemTrayIcon : public QObject
 private:
     SystrayBackendType backendType;
     QSystemTrayIcon* qtIcon;
+    Widget *parent;
```
antis81: Proposal: Rename to `mParent`. Thus, `this->parent = parent` would become `mParent = parent` in c'tor implementation [in cpp line 31](https://github.com/tux3/qTox/pull/2132/files#diff-1b85c272c0d41b3a9efecede73cfa60eR31).

src/widget/widget.h
```diff
@@ -160,6 +160,9 @@ private slots:
     void processOfflineMsgs();
     void friendListContextMenu(const QPoint &pos);
 
+private:
+    const int icon_size = 15;
```
antis81: Please move initialization to c'tor so order stays clear. Not every "feature" c++11/14 offers is a good one ...

agilob: `to c'tor`? 

what does it mean?

constructor -.-'

src/widget/systemtrayicon.cpp
```diff
@@ -26,8 +26,9 @@
 #include <QDebug>
 #include "src/persistence/settings.h"
 
-SystemTrayIcon::SystemTrayIcon()
+SystemTrayIcon::SystemTrayIcon(Widget *parent)
```
antis81: @agilob Uhm ... we seem to have a private implementation here ... so I think that ...
a) `parent` member should be moved there
b) private class should be initialized here

Right?

src/widget/systemtrayicon.h
```diff
@@ -22,6 +22,7 @@
 #define SYSTEMTRAYICON_H
 
 #include "systemtrayicon_private.h"
```
antis81, line 24: @agilob Though unrelated to your current changes in this PR. If I understand PIMPL right and not am completely blind, this line should be moved to cpp, to keep implementation **private**. :smile:

src/widget/widget.h
antis81: Yep :) - sorry, it's an older c++ name convention.

src/widget/systemtrayicon.cpp
agilob: that was used to experiment with qt implementation/status notifier, in fact is should be removed because parent is never used anyway. Thanks for spotting this o/

agilob:
@zetok why travis didn't build first and second commit?

antis81:
@agilob I still need to setup a working build environment. The qtox repo, compiler, etc. is installed. Anything I need to be aware of despite of the lib dependencies? Anyway, I'll set it up and report back. Not sure though, if I really can check the "system tray icon" as I don't have a "system tray". :smile: -> using `cairo-dock` panel replacement.

Anyone else running a common Linux distro and "system tray"?

agilob:
we just need to test it on unity.

antis81:
I'm using (LXDE/LXQt) for working environment. Anyone using a unity environment to check, if this works?

src/widget/widget.cpp
```diff
@@ -1540,12 +1554,20 @@ QString Widget::getStatusIconPath(Status status)
...
 }
 
-inline QIcon Widget::getStatusIcon(Status status, uint32_t w/*=0*/, uint32_t h/*=0*/)
+inline QIcon Widget::getStatusIcon(Status status, uint32_t w, uint32_t h)
 {
-    if (w > 0 && h > 0)
-        return getStatusIconPixmap(status, w, h);
-    else
-        return QIcon(getStatusIconPath(status));
+#ifdef __linux__
```
antis81: How about `#ifdef Q_OS_LINUX` instead?

src/widget/widget.cpp
```diff
@@ -1540,12 +1554,20 @@ QString Widget::getStatusIconPath(Status status)
...
+    QString desktop = getenv("XDG_CURRENT_DESKTOP");
+    if (desktop.isEmpty())
+        desktop = getenv("DESKTOP_SESSION");
+    desktop = desktop.toLower();
+    if (desktop == "xfce" || desktop.contains("gnome") || desktop == "mate") {
```
antis81: Maybe change this to "exclude" strategy:

``` c++
if (desktop != "unity") { ... }
```

src/widget/widget.cpp
```diff
@@ -1540,12 +1554,20 @@ QString Widget::getStatusIconPath(Status status)
...
-inline QIcon Widget::getStatusIcon(Status status, uint32_t w/*=0*/, uint32_t h/*=0*/)
+inline QIcon Widget::getStatusIcon(Status status, uint32_t w, uint32_t h)
 {
-    if (w > 0 && h > 0)
-        return getStatusIconPixmap(status, w, h);
-    else
-        return QIcon(getStatusIconPath(status));
+#ifdef __linux__
+    QString desktop = getenv("XDG_CURRENT_DESKTOP");
+    if (desktop.isEmpty())
```
antis81, line 1911: Can you please add the `{ }` brackets? Improves readability and reduces potential for mistakes.

antis81:
@agilob Apart from the minor proposals I wrote, it looks good to me. Would you mind cleaning them up?

src/widget/widget.cpp
agilob: not really, this if must cover window managers based on GTK2/3. `desktop!="unity"` will break support for i3, awesome and other wired managers.

agilob:
done.

src/widget/widget.cpp
antis81: Ok then, no problem.

zetok:
> @zetok why travis didn't build first and second commit?

travis builds only latest HEAD, with build triggered by push to repo/PR

ProMcTagonist:
Why does it have "Exit | Ctrl + Q" and "Quit" both?

agilob:
> Why does it have "Exit | Ctrl + Q" and "Quit" both?

Read my PR above...

ProMcTagonist:
It's standard to include your own on KDE?

antis81:
> Why does it have "Exit | Ctrl + Q" and "Quit" both?

@agilob Maybe setting `QAction::QuitRole` helps to avoid double entry.

agilob:
@antis81 yes 8) 
snapshot3

antis81:
@agilob hehe. Hm ... only an idea: Maybe you can check for an existing quit action and only add, if none was found.

Btw.: This also affects OSX. It's not so obvious though, as the quit action is bound to a special menu called "application menu".

agilob:
I'm not sure if OSX will be affected by those changes. I added quitrole to that action, but it doesnt change anything on KDE5 and I can't find any working example that doesn't reproduce this issue, I checked some bitcoin clients and torrent clients written in c++/qt. I'll check source code of KDE tomorrow. Isn't application menu related to menubar in qmainwindow? [That's what documentation says](http://doc.qt.io/qt-4.8/qmenubar.html#qmenubar-on-mac-os-x)

agilob:
I added logout action:
snapshot7

Another commit partially fixes #1999, we can't completely fix 1999 but username and status will be printed as plaintext:
snapshot8

agilob:
@antis81 I'm pretty sure OSX will not be affected by those changes, and if OSX would add second quit action, then it ALREADY is there, as my changes affect only KDE5 and Gnome3, nothing else. I changed used backend used on KDE5 from status notifier to qt and fixed rendering transparent tray icon. If OSX adds second quit action, it already is there, my changes won't cause it.

agilob:
The last commit here fixes conflict with origin/master, sorry for misleading commit message.

tux3:
I'm getting a crash at systemtrayicon.cpp:287 on a KDE5 system.

src/widget/systemtrayicon.cpp
```diff
@@ -95,28 +95,12 @@ SystemTrayIcon::SystemTrayIcon()
...
-                                 8, image->width(), image->height(),
-                                 image->bytesPerLine(), callbackFreeImage, image);
-
-        statusNotifier = status_notifier_new_from_pixbuf("qtox",
-                            STATUS_NOTIFIER_CATEGORY_APPLICATION_STATUS, pixbuf);
-        status_notifier_register(statusNotifier);
+        qDebug() << "Using the Qt backend";
+        qtIcon = new QSystemTrayIcon;
+        backendType = SystrayBackendType::Qt;
+        connect(qtIcon, &QSystemTrayIcon::activated, this, &SystemTrayIcon::activated);
```
tux3, line 102: That seems wrong, you're using the Qt backend inside an #ifdef ENABLE_SYSTRAY_STATUSNOTIFIER_BACKEND

The ifdef should just be removed if you don't actually use status notifier here.

agilob:
Hmm. how did you compile it? I used it for a week now without any problems

zetok:
@agilob looks like it fails to build: https://build.tox.chat/job/qTox-testing_build_linux_x86-64_release/10/console

agilob:
Fixed and merged with upstream master.

agilob:
@tux3 could you try this now? Why are you getting this crash?

agilob:
@tux3 could you try this now? Could you tell me where is the crash from?

src/widget/systemtrayicon.cpp
```diff
@@ -94,31 +94,13 @@ SystemTrayIcon::SystemTrayIcon()
...
     }
-    #endif
     else if (desktop == "kde"
              && getenv("KDE_SESSION_VERSION") == QString("5"))
```
antis81: Uhm, isn't this the same as previous condition above? Btw.: Better user QStringLiteral("5").

tux3:
@agilob 
It works for me now, with just a couple issues.
- [ ] When I hover the mouse over the exit icon, it becomes a completely different icon:

snapshot34 snapshot35
- As @antis81 noted, the `else if` line 104 can be removed:

```
    else if (desktop == "kde"
             && getenv("KDE_SESSION_VERSION") == QString("5"))
    {
        backendType = SystrayBackendType::KDE5;
        qWarning() << "Detected a KDE5 session, but we don't have Status Notifier support. Disabling the systray icon";
    }
```

and the `SystrayBackendType::KDE5` value can be removed from the enum too.

agilob:
Hah interesting, I'll take a look at it.

Isn't is much better now? ;)

zetok:
@agilob btw, merge my PR to this branch? (with `--ff`, so that there'll be no useless merge commit)

agilob:
When I fix the issue above I'll make a new branch from this diff to avoid having merge conflicts again...

agilob:
Could you try this? If this doesn't work I'll remove the icon. I suppose this is because there is style for onHover in .css and your icon is changed from KDE5 to GTK on mouse hover.

zetok:
@agilob err, you still got unneeded code there.. (which wouldn't be the case if you did push my commit)

agilob:
what code is unused? what commit?

zetok:
https://github.com/agilob/qTox/pull/2

agilob:
pls ;_;

agilob:
Closing because it's merged o/

Issue #2070, opened 2015-08-11
proxy settings can not be set before the login


boza111:
proxy settings can be set only after logging in ....

agilob:
Interesting, how would you like it to be implemented?

zetok:
it probably doesn't have to be implemented - what is needed though, is a way to launch qTox in an "offline" mode, where it doesn't connect to the network.

It could be implemented as command like switch.

With that, one could start qTox in offline mode, set proxy, and restart qTox.

NoxNoctium:
another option would be to add this setting to the already proposed "first run" tour.
zetoks idea would also fit as a status chooser shown on the login screen, one could also choose to sign in already marked as busy or away then.
WLM did it like this:
![WLM](http://qs.lc/4b0ko)

agilob:
@markus1540 that one is doable.

agilob:
Uppss! thing is we don't support going offline at all!

ovalseven8:
Important security issue especially for people who would like to use Tor. 

A solution is needed.

ghost:
> Uppss! thing is we don't support going offline at all!

Yes you're right. (Never really saw this)
So this need to be implemented.

agilob:
Hmm... this can be added to import profile widget, it's nearly empty now.

ProMcTagonist:
Perhaps Import could be renamed to More or Options and have importing, proxy settings, etc.

ovalseven8:
Offline-mode would be more user-friendly, I think.

ovalseven8:
@tux3 
Don't know how much work implementing offline mode is. Do you think you'll have time soon? It would be a very nice feature and is also important for Tor users.

kradan:
Currently (3d2d0e6a63894b1c36163ea0eddcf7e6fad2ec07) it is not possible to register a tox account via tor. It would be great to fix this for this guide to work: https://askubuntu.com/questions/700081/how-to-tunnel-tox-through-tor
```
$ torify ./qtox                                                                                
1531816378 WARNING torsocks[15192]: [syscall] Unsupported syscall number 383. Denying the call (in tsocks_syscal() at syscall.c:488)                                                                                             
[08:33:09.828 UTC] persistence/settings.cpp:129 : Debug: No settings file found, using defaults                  
[08:33:10.028 UTC] persistence/settings.cpp:133 : Debug: "Loading settings from :/conf/qtox.ini"                 
[...]

PR #2140, opened 2015-08-30
Import profile on login screen

agilob:
- adds third stack widget to import profile
- validates profile
- if added successfully resets loginwindow and allows login with new profile
- looks ugly

snapshot6

god i hate stackwidgets.

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -185,7 +185,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
                 "|"   // standard protocols OR tox ID:
-                "^tox:(([\\w]+@[\\S]+)" // tox:something@something.com
+                "tox:(([\\w]+@[\\S]+)" // tox:something@something.com
                 "|(\\w{76}))$" // tox:76 characters long tox id
                 );
 
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -186,7 +186,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "\\w+\\S+)" // any other character, lets domains and other
                 "|"   // standard protocols OR tox ID:
                 "tox:(([\\w]+@[\\S]+)" // tox:something@something.com
-                "|(\\w{76}))$" // tox:76 characters long tox id
+                "|([A-Fa-f0-9]{76}))$" // tox:76 characters long tox id
                 );
 
     int offset = 0;
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -184,8 +184,11 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)(www\\.|http[s]?|ftp)://" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(^tox:[@\\w]+$)"); // starts with `tox` then : and only alpha-digits till the end
-                // also accepts tox:agilob@net as simplified TOX ID
+                "|"   // standard protocols OR tox ID:
+                "^tox:(([\\w]+@[\\S]+)" // tox:something@something.com
+                "|(\\w{76}))$" // tox:76 characters long tox id
+                );
+
     int offset = 0;
     while ((offset = exp.indexIn(out, offset)) != -1)
     {
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -84,20 +84,20 @@ ProfileForm::ProfileForm(QWidget *parent) :
     profilePicture = new MaskablePixmapWidget(this, QSize(64, 64), ":/img/avatar_mask.svg");
     profilePicture->setPixmap(QPixmap(":/img/contact_dark.svg"));
     profilePicture->setClickable(true);
-    connect(profilePicture, SIGNAL(clicked()), this, SLOT(onAvatarClicked()));
     QHBoxLayout *publicGrouplayout = qobject_cast<QHBoxLayout*>(bodyUI->publicGroup->layout());
     publicGrouplayout->insertWidget(0, profilePicture);
     publicGrouplayout->insertSpacing(1, 7);
 
     timer.setInterval(750);
     timer.setSingleShot(true);
-    connect(&timer, &QTimer::timeout, this, [=]() {bodyUI->toxIdLabel->setText(bodyUI->toxIdLabel->text().replace(" ✔", "")); hasCheck = false;});
 
-    connect(bodyUI->toxIdLabel, SIGNAL(clicked()), this, SLOT(copyIdClicked()));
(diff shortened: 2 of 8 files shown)


src/widget/form/profileform.cpp
```diff
@@ -205,10 +206,10 @@ void ProfileForm::onAvatarClicked()
...
     };
 
-    QString filename = QFileDialog::getOpenFileName(0,
-        tr("Choose a profile picture"),
-        QDir::homePath(),
-        Nexus::getSupportedImageFilter());
+    QString filename = QFileDialog::getOpenFileName(this,
+                                            tr("Choose a profile picture"),
+                                            QDir::homePath(),
+                                            Nexus::getSupportedImageFilter());
```
antis81, line 212: Minor one: Should be aligned after opening brace `(` ...

src/chatlog/chatmessage.cpp
```diff
@@ -184,8 +184,11 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)(www\\.|http[s]?|ftp)://" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(^tox:[@\\w]+$)"); // starts with `tox` then : and only alpha-digits till the end
-                // also accepts tox:agilob@net as simplified TOX ID
+                "|"   // standard protocols OR tox ID:
+                "^tox:(([\\w]+@[\\S]+)" // tox:something@something.com
+                "|(\\w{76}))$" // tox:76 characters long tox id
```
zetok: hm, ID should be limited to `12345678890ABCDEF` characters, but `\w` is any character?

agilob:
I added logout action:
snapshot7

src/chatlog/chatmessage.cpp
agilob: `\\S+` is any printable character [A-Za-z0-9]+ special characters
`\\w+` is [A-Za-z0-9]

in fact it should be [A-Fa-f0-9] but that would be much slower, keep in mind this regexp is run on every message and action.

Qt doesnt have classes like [:hex:] or \h+.

zetok: I think that correctness should take a slight precedence over speed, unless performance is affected _really badly_.

agilob:
Could someone merge it?

src/chatlog/chatmessage.cpp
antis81: > Qt doesnt have classes like [:hex:] or \h+.

This should work: `\x{1,76}` (not sure: might be `\xh{1,76}`)

**EDIT:** Sorry, of course 76 is fix -> `\xh{76}`

agilob: It must be 76 long, not from 1 to 76 `\\x` is for UTF8 characters, not standard ASCII.

\xhhhh Matches the Unicode character corresponding to the hexadecimal number hhhh (between 0x0000 and 0xFFFF).

btw. `\x` is invalid as it tries to escape `x` character. Alternatively it can be written as [A-Fa-f0-9]{76} but it's not for better performance.

agilob: ok @zetok you won this time :>

antis81: @agilob Uhm ... forget what I wrote above. Did you know about `QRegularExpression`? Looks like this replaced the `QRegExp` class in Qt5 and provides what you are looking for.

Apart from that I agree with @zetok in this particular case. In general we do not want to use "hacking tricks" to improve speed. So `[0-9a-fA-F]{76}` would be ok.

Btw.: If reg. exp. are to slow, we can use the classic way and parse by ourselves:

``` c++
// pretty shitty sample parser - just to make clear what I mean :)

QString toxId;
QString destination;
bool ok;
if (string.startsWith("tox:"))
    int pos = string.indexOf('@');
    if (pos > -1) {
        toxId = string.mid(4, pos).toInt(&ok, 16);
        if (ok) {
            destination = string.right();
        }
    }
}
```

src/widget/tool/profileimporter.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <QMessageBox>
+
+ProfileImporter::ProfileImporter(QWidget *parent) : QWidget(parent)
+{
+
+}
+
+bool ProfileImporter::importProfile()
+{
+    QString path = QFileDialog::getOpenFileName(this,
```
PKEv, line 20: to preserve the style is better to use "0" instead "this"

agilob, line 20: > to preserve the style is better to use "0" instead "this"

I will fix that bug tomorrow.

ProMcTagonist:
Is it possible to have a file picker already open embedded there in place of the button?

agilob:
I wish there was something like that in Qt :<

src/widget/tool/profileimporter.cpp
antis81, line 20: @agilob Remember our first discussion? :smile:

@PKEv This is not a bug and totally correct. It's a wrong assumption, that the system style is used, just because a widget has no parent. This is simply untrue and setting parent to 0 introduces other effects we do not want. We can only preserve the system's style by actually correctly assigning it to the widget. If we want to use system default, the css file has to change like:

``` css
!QFileDialog, !QDialog { ... }
```

Like I wrote before, it is better to implement a `QStyle` class.

Or we just stick with qTox style, which is also fine and recently was agreed by "the team".

agilob, line 20: I learnt this "trick" when I was starting with Qt couple of years ago and it worked so far ;)

Your solution is better so I'm reverting those changes, I will try it on my private project and if worked well it will be ported to qtox.

agilob:
Don't merge yet. Import profile tab will contain 'set proxy' fields too.

zetok:
> Don't merge yet. Import profile tab will contain 'set proxy' fields too.

Wouldn't it be better to merge it now, and then eventually add proxy thing in a separate PR?

agilob:
as soon as someone tells me how to fix this ugly interface

PKEv:
may make imports of inner tab when you select "new profile"?
Or simply button on the page "new profile"....

agilob:
I think that tab is already bloated.

PKEv:
or like this
default

ProMcTagonist:
@PKEv +1

agilob:
Purpose of another tab is to add fields with proxy settings.

zetok:
I like `Import` on `Load Profile` page

PKEv:
Why should make in the window proxy settings?
inside qtox these settings are the same..

zetok:
> Purpose of another tab is to add fields with proxy settings.

Why would it be needed / wanted there?

What might be wanted, is switch to start in a "offline" mode, and then one could set proxy in settings…

agilob:
You set p-medium on https://github.com/tux3/qTox/issues/2070

zetok:
um, in my first comment there I wrote "it probably doesn't have to be implemented - what is needed though, is a way to launch qTox in an "offline" mode, where it doesn't connect to the network."

Sorry, I probably should have been more clear about it - i.e. I don't think that proxy setting on login screen is a good idea.

agilob:
OK, the layout proposed by @PKEv is easily doable, and it would cleanup this branch.

PR #2148, opened 2015-08-30
Fix confusing statement, compact instructions

ProMcTagonist:
It sounds like you can press enter to make a new selection.


agilob:
I think this text is already way too long :<
How about:
`Press space to show or hide qTox, Enter to capture screenshot, Esc to quit`?

ProMcTagonist:
Better?

Should the selection/non-selection tooltips match? One has linedowns and one doesn't, now.

agilob:
Thanks o/

> Should the selection/non-selection tooltips match? One has linedowns and one doesn't, now.

I dont really care...

ProMcTagonist:
It just looks weird for them to be different.

PR #2150, opened 2015-08-30
set one global initial file accept directory for all file transfers

### LINKED ISSUE #1818: [Feature request] Add an option to set a default download directory without autoaccept

agilob:
One initial dir will be shared between per-friend/auto-accept and manually accepted files

fixes #1818


zetok:
- [x] regression - opening place of automatically received file or file itself via qTox doesn't work

agilob:
well... when you ask qfiledialog for a path, it returns a path, which doesn't end with `/`...

src/widget/form/chatform.cpp
```diff
@@ -162,7 +162,8 @@ void ChatForm::onTextEditChanged()
 
 void ChatForm::onAttachClicked()
 {
-    QStringList paths = QFileDialog::getOpenFileNames(0,tr("Send a file"));
+    QStringList paths = QFileDialog::getOpenFileNames(this,
+                                                      tr("Send a file"));
```
tux3, line 166: The newline here is maybe a bit gratuitous, since it only saves 5 characters, and there are many longer lines around...

src/widget/widget.cpp
```diff
@@ -706,7 +706,7 @@ void Widget::setStatusMessage(const QString &statusMessage)
 void Widget::reloadHistory()
 {
     for (auto f : FriendList::getAllFriends())
-        f->getChatForm()->loadHistory(QDateTime::currentDateTime().addDays(-7), true);
+        f->getChatForm()->loadHistory(QDateTime::currentDateTime().addDays(-1), true);
```
tux3: This change is NOT mentioned in the PR description or any of the commit messages.

Changing a `0` into a `this` to center some message boxes without mentioning it just annoys me a bit, but this is going too far.
Please don't sneak in unrelated changes with pull requests, or at least mention it in the description.

So apparently the only thing this does is that when the password is changed, we would reload 1 day of history instead of 1 week. This doesn't make much sense since in `Friend::loadHistory` we still load 1 week.

What were you trying to do?

agilob:
I was experimenting with code and changed this to see what happens, I forgot to change it back...

src/widget/widget.cpp
agilob: I was experimenting with code and changed this to see what happens, I forgot to change it back...

PR #1882, opened 2015-06-17
Themes - updated

agilob:
WHAT DOES IT HAVE?
EVERYTHING WHAT  https://github.com/tux3/qTox/pull/1640 HAS + MORE:
- fixed css for wombat (tpyos)
- applied css for chat!
- reloading style without restarting app
- code formatting in some places
- CSS formatting
- merging to recent changes in qtox
- compiles now!
  nothing meaningful, just A LOT of small changes sorry about that, I lost track of time. I'm also sorry about those commit messages. first commit `2` includes all commit from old branch, nothing more, nothing less.

WHY?!
I had really bad time merging and updating code after recent changes, so I made .patch from https://github.com/tux3/qTox/pull/1640 and checrry-picked it to this branch, it was just much easier to manage +100 git conflicts! 

Themes:
light

Style is not perfect and I know it. Please keep in mind this is my FIRST EVER experience with CSS. There are shortcomings.

If you know css better than me (you probably do!) please fix it.

@tux3 if you think it's "good enough" please merge it before code conflicts again so I don't have to spend even more hours on solving them, I also won't have much time to fix issues soon.

PR #2151, opened 2015-08-30
Translate context menu

### LINKED ISSUE #1720: translate system menu don't work / some windows/buttons are not translateable
when i build qtox from src, system menu translated
default
when i start installed qtox, system menu do not stanslated

Edit: What could be the problem? What is so special assembly?

PKEv:
fix #1720 
translate context menu of ChatTextEdit

### DIFF
--- a/src/widget/tool/chattextedit.cpp
+++ b/src/widget/tool/chattextedit.cpp
@@ -20,6 +20,7 @@
 #include "chattextedit.h"
 #include "src/widget/translator.h"
 #include <QKeyEvent>
+#include <QMenu>
 
 ChatTextEdit::ChatTextEdit(QWidget *parent) :
     QTextEdit(parent)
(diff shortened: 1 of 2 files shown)


src/widget/tool/chattextedit.cpp
```diff
@@ -69,3 +70,54 @@ void ChatTextEdit::retranslateUi()
 {
     setPlaceholderText(tr("Type your message here..."));
 }
+
```
zetok, line 73: Could there be added a comment saying that it's needed because Qt lacks proper translations of context menus?

agilob:
that's not how you do it oO

> Could there be added a comment saying that it's needed because Qt lacks proper translations of context menus?

Qt does have it, but apparently it's not used in this class! Other context menus are translated!

src/widget/tool/chattextedit.cpp
PKEv, line 73: it's not a QT problem, problem apparently is that the translation of these lines need a library that is not supplied with the qtox

zetok:
> Qt does have it, but apparently it's not used in this class! Other context menus are translated!

Oh. I always thought that it was Qt at fault, like with many other things…

PKEv:
this strings don't work in installed qtox, but good work when i build it myself:

```
    // system menu translation
            QTranslator *qtTranslator = new QTranslator();
            QString s_locale = "qt_"+locale;
            if (qtTranslator->load(s_locale, QLibraryInfo::location(QLibraryInfo::TranslationsPath)))
```

Since an external file transfers, drag them all with the distribution probably does not make sense, because it is easier to do just that

agilob:
It's our fault. The issue was fixed once somewhere, I remember making issue that file pickers were using English language on KDE in Polish, some guy fixed it and left qtox after that. Now as you see, file pickers use language from system, some context menus are translated, some aren't. Right click on search text-edit, and try right click in invitation message box:

This uses wrong context and could be translated by us:
snapshot10

This has correct context and translations are loaded from operating system:
snapshot11
You can see not only language is copied from OS here, also style.

I was aware of this bug, but it's so minor I ignored it (and I use system in English what makes it harder to notice...)

PKEv:
I'll think about how to make it more correctly

agilob:
I'm pretty sure it's fixed in agilob/themess, I'll check it out.

agilob:
Yes, it's (partially) fixed in my branch themess, both screenshots are from that branch.

You can see how one context menu loads language from qTox settings, second from system, that' a bug actually... but we should never reinvent most basic translations.
snapshot12

snapshot13

This will be fixed soon (soon...)

PKEv:
Please tell me how you realized it?
I can not find anything similar to that branch... 

it may happen that the work here is this piece of code

```
  // system menu translation
            QTranslator *qtTranslator = new QTranslator();
            QString s_locale = "qt_"+locale;
            if (qtTranslator->load(s_locale, QLibraryInfo::location(QLibraryInfo::TranslationsPath)))
```

and distributing the distribution qtox nothing will not work due to lack of user libraries

agilob:
https://github.com/tux3/qTox/pull/1882/files#diff-6ea91ba076929a0ccf5de6bb9a5a2574

PKEv:
there is another way:
a). Add "qt.conf" file to qTox directory with the following contents:

```
[Paths]
TranslationsPath = translations
```

b). copy folder in the directory qtox
"c:\Qt\5.4\mingw491_32\translations\"
c). enjoy

In the distribution of course possible to provide the desired language selection and copy not whole catalog 

Thus we will solve all the issues with the translation standard dialogs

PKEv:
I do not know how to make it through the pull...
can you do it? @zetok @agilob @suhr

agilob:
I'll take a look at this when other PRs are merged. We have **12** PRs awaiting right now, when at least 10 of them are merged I will start working on another feature that will implement this. If we do it now we will just have more code conflicts to solve later and waste time on.

PR #2156, opened 2015-08-31
change logging, typing notification and compact layout to enabled

### LINKED ISSUE #2155: Typing notifications should be on by default
they really should

agilob:
closes #2155

keep in mind this will affect only new profiles.

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -302,7 +302,7 @@ void Settings::loadPersonnal(Profile* profile)
     ps.endGroup();
 
     ps.beginGroup("General");
-        compactLayout = ps.value("compactLayout", false).toBool();
+        compactLayout = ps.value("compactLayout", true).toBool();
     ps.endGroup();
 
     ps.beginGroup("Circles");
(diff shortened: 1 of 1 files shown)


DanTheBritish:
Hold on, this changes the default layout from full to compact? Why?

zetok:
@LapFox Compact layout is simply better.

agilob:
It's much better. I showed qtox to a few people, first thing they changed was layout.

I was also considering changing autostart in tray, close to tray and minimize to tray as default, but currently this is a bit broken, so it will be enabled after fix.

DanTheBritish:
Alright, fair enough, the voice of one can't outweigh the many, so I'll be just change back to the full layout when I update :rocket:

agilob:
This will affect only new profiles, your settings won't be changed.

handsomematt:
Why did you change logging to be enabled by default? What a horrible idea for a chat protocol based on the principles of privacy.

tux3:
@HandsomeMatt
Logs aren't really a problem, because if someone can get your logs, they can also get your Tox profile and private key, which is arguably much worse. The solution if your threat model includes your computer being compromised is to use a password so your profile and logs can be encrypted.

It's reasonnable to expect a chat client to keep logs by default, we're aiming to replace Skype, not PGP.

handsomematt:
Looks like I'm finding a new Tox client that cares about privacy.

DanTheBritish:
This only affects the default setting, you can turn it off.

agilob:
> Looks like I'm finding a new Tox client that cares about privacy.

 You also need to find new friends who will turn off logging off as well, if one of them has logging on, "you're compromised" as well. To state it again, this affect only new profiles, your will be untouched...

> What a horrible idea for a chat protocol based on the principles of privacy.

tox protocol != qTox

PR #2157, opened 2015-08-31
automatic call answering 

jacksz:
i have noticed that tox lacks one thing which honestly i depend on , which is automatic call answering , i know this shouldn't be hard to implement but i know many people who would love something other than skype that does that! on mac and windows! and you definitely gonna be gaining my vote :+1: 
and everyone who hates skype :100:


agilob:
Wow

Fix those commits, make new branch and new commits, we won't merge such mess.

PR #2164, opened 2015-09-02
OpenSUSE ffmpeg-devel

gmscs:
Compilation of qTox under OpenSUSE requires ffmpeg-devel, not ffmpeg.


agilob:
I stoppeding using OS some time ago, but don't you need both packages? ffmpeg and ffmpeg-devel?

gmscs:
it's working fine on my end without ffmpeg, i'll play around with it to make sure

[EDIT] Everything seems to be working without ffmpeg

PR #2166, opened 2015-09-02
Add OpenSUSE to simple_make.sh

gmscs:
Not too happy with: " -e /etc/redhat-release -o -e /etc/zypp " for the qmake-qt5. 
Something similar could be done for Arch and _maybe_ Void. I'll add to this when i get an Arch and a Void install up and running.

### DIFF
--- a/simple_make.sh
+++ b/simple_make.sh
@@ -19,12 +19,20 @@ elif which dnf; then
         openal-soft-devel libXScrnSaver-devel qrencode-devel \
         opus-devel libvpx-devel qt5-qttools-devel glib2-devel \
         gdk-pixbuf2-devel gtk2-devel
+elif which zypper; then
+    sudo zypper in \
+	git patterns-openSUSE-devel_basis libqt5-qtbase-common-devel \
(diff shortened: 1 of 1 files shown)


agilob:
What does it do now?

Arch is supported from AUR.

gmscs:
This is the simple make and install. All I did was add support for installing qTox deps and making on OpenSUSE.
It used to give an error " Makefile:1975: recipe for target 'audiofilterer.o' failed " because it used qmake by default, now since it recognizes "/etc/zypp/" it will do qmake-qt5 which wont return that build error. About Arch and Void, I was just saying that something similar could be done for Arch and Void since AFAIK they go through the same qmake problem.

agilob:
> It used to give an error " Makefile:1975: recipe for target 'audiofilterer.o' failed " because it used qmake by default, now since it recognizes "/etc/zypp/" it will do qmake-qt5 which wont return that build error. 

That's what I needed. Thanks. I don't have any problem on Arch, we also didn't have any report, but I'll take a look at it soon.

PR #2181, opened 2015-09-06
French translation (small update)

tetraf:


zetok:
btw, is there a reason why there are spaces before `:` ?

tetraf:
It's French typography. Cf. : https://fr.wikipedia.org/wiki/Ponctuation#Traitement_des_espaces_pour_les_symboles_bipoints
Actually it must be a thin unbreakable space.

zetok:
I see, thanks `:)`

PR #2196, opened 2015-09-10
disable logging by default.

handsomematt:
don't log; it's rude.

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -321,7 +321,7 @@ void Settings::loadPersonnal(Profile* profile)
 
     ps.beginGroup("Privacy");
         typingNotification = ps.value("typingNotification", true).toBool();
-        enableLogging = ps.value("enableLogging", true).toBool();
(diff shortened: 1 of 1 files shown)


tblah:
This seems like a good idea for a messenger for the security conscious. It seems unkind to survey the person you are talking to automatically when they have already made an effort to choose tox over conventional communications methods.

agilob:
A few day ago we had a request to enable logging by default. Also it was in plans to enable it when it becomes stable.

zetok:
qTox aims to be usable and just work™ for users - having bad default option (like disabled history) is **not** helping with that.

> don't log; it's rude.

You can always disable it in your settings - there's no point in making using qTox painful for other people just because you don't want to keep logs. Don't do that - it's rude.

> It seems unkind to survey the person you are talking to automatically when they have already made an effort to choose tox over conventional communications methods. 

Tox provides **private communication between 2 parties**. That's the difference from conventional methods.

# 

Anyway, this is not going to be changed - logging is, and is going to stay an option for a reason. By default though, users should be able to enjoy features.

Issue #659, opened 2014-11-07
qtox_history doesn't retain contact names


tsdko:
Let's suppose there's a chat:

```
[13:36:41] tsun: foo
(tsun has changed their name to dere)
[13:36:48] dere: bar
```

Later on, after loading the history, it will look like this:

```
[13:36:41] dere: foo
[13:36:48] dere: bar
```

zetok:
Why should it contain name other than your present?

tsdko:
Not sure how other people feel about that, but I think history should contain the chat state exactly as it was during the time of the chat, including names, status changes, etc.

ProMcTagonist:
The chat archive should be an archive. Name changes are relevant and often critical to understanding context.

dubslow:
Uh... wrong issue, @tux3?

tux3:
Oops! Indeed.

zetok:
Note from #1189:
Beside saving nicks, saving when change happened (and in what way) might be wanted.

PR #2223, opened 2015-09-13
save alias in history

### LINKED ISSUE #1189: Lookup previous aliases
qTox should provide a way to lookup previous aliases of a particular user. If a lot of users on your friend list change names often, things get messy quickly.

I have no idea what the best way to implement this in the UI is, but my immediate thought is to add an option to the context menu you get when you right-click an item in the friend list. Something like "Show Aliases" that brings up a dialog, perhaps?

PKEv:
fix #1189
 a way to lookup previous aliases of a particular user.

### DIFF
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -42,7 +42,7 @@ HistoryKeeper *HistoryKeeper::getInstance()
     {
         QList<QString> initLst;
         initLst.push_back(QString("CREATE TABLE IF NOT EXISTS history (id INTEGER PRIMARY KEY AUTOINCREMENT, timestamp INTEGER NOT NULL, ") +
-                          QString("chat_id INTEGER NOT NULL, sender INTEGER NOT NULL, message TEXT NOT NULL);"));
+                          QString("chat_id INTEGER NOT NULL, sender INTEGER NOT NULL, message TEXT NOT NULL, alias TEXT);"));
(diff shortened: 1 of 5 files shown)


Zer0-One:
Can you please describe this PR with more than two words? And possibly provide screenshots?

tux3:
Please keep all source code comments in english, I can't merge this as is.

Now the actual code looks ok but storing the nick on each message instead of on each nick change is a bit of a waste of space.

PKEv:
We spent a lot of space but otherwise much more complicated structure

tux3:
It certainly is simple this way.

Ok, so this clearly solves #659, but I don't think it really addresses #1189 

@Zer0-One ?

PKEv:
complete closure of #1189 will allow only the by creating user page (About)

tux3:
Let's start by just closing #659, and we'll see what to do with #1189 then. I think it needs a bit more discussion because I'm not seeing much of a consensus.

src/persistence/historykeeper.cpp
```diff
@@ -244,7 +256,7 @@ void HistoryKeeper::importMessages(const QList<HistoryKeeper::HistMessage> &lst)
     db->exec("BEGIN TRANSACTION;");
     for (const HistMessage &msg : lst)
     {
-        QList<QString> cmds = generateAddChatEntryCmd(msg.chat, msg.message, msg.sender, msg.timestamp, msg.isSent);
+        QList<QString> cmds = generateAddChatEntryCmd(msg.chat, msg.message, msg.sender, msg.timestamp, msg.isSent, QString()); //!!!
```
agilob, line 259: heh

PR #2238, opened 2015-09-15
French translation fix misunderstanding

TheNain38:


zetok:
cc: @aaannndddyyy,  @pwnsdx , @tetraf ?

aaannndddyyy:
I don't even know what this function does. "Don't group chat windows"  regroup chats?, open chats in separate windows ??
If it means not to open a new window for a chat - different form qTox's main window, which contains the friend list and settings, then the original version is better. If it somehow "gorups theem together", then this PR is better. Still I don't know what it does.

zetok:
Well, basically, when you have a detached chat window, you can have more than 1 contact in them:
detached window

IIRC, when you have enabled mode "detached windows", there's by default a "main contact window" with all your contacts, and a detached window, with "grouped" contacts - only those that you currently have a conversation with, or manually drag into the detached window.

That option changes that behaviour, so that current conversations aren't grouped in a detached window, but rather, a new window is opened for each new conversation.

TheNain38:
At first I misunderstand this option and this PR fix my mistranslation.

aaannndddyyy:
Maybe simply "Open each chat in a separate window" would be more intelligible?
Neither translation says that, though.
To me it wan't clear that "group chat windows" means make them one.
when I normally group things, I get different groups. like in the task bar, each group is one containing one or more instances of the same type, e.g. browser windows. but grouping windows would then still mean having different window, just that they are grouped together into different groups.
"Combine multiple application windows into a single button" is what my task bar calls grouping.

So I think, the original English source could need some improvement

aaannndddyyy:
@TheNain38 Your original translatation without the "ne pas" would be much easier understandable.
EDIT: much more easily understandable

TheNain38:
Ok but it should be `Ouvrir les discussions dans des fenêtres séparées.` not `groupes`

TheNain38:
Fixed too.

aaannndddyyy:
Yep. Looks good. :)

TheNain38:
There is a problem the other option is translated the same as that one now (line 1066 and 1071)

TheNain38:
The English source could be improved as `Separate main window from other windows` it would be clearly better. The line 1066 could then be translated as `Séparer la fenêtre principale des autres fenêtres`

aaannndddyyy:
yes, or just "separate windows". "open chats in separate windows" is wrong. It also applies to Settings window, and add firnds window. Andthen to avoid sond ing the same all the time, in 1066 it could be "indiviual" instead of separate.

zetok:
> Yep. Looks good. :)

Does that mean that it can be merged?

aaannndddyyy:
the translation for that string looks good.
However, the one that was now renamed to multiple window mode still has texactly the same wording now.

@TheNain38 Pourrais-tu changer ça encore? L'original anglais est déjà changé ( commit 9269b0f ).

TheNain38:
How can I translate `Multiple windows mode` ?

TheNain38:
`Mode fenêtres multiples` ?

TheNain38:
Is it good like that?

aaannndddyyy:
@TheNain38 Yes, just did a quick check and Samsung and Gimp use this translation too. Thx

@Zetok: all fine now

EDIT:
If there were a tooltip, I'd go with your `Séparer la fenêtre principale des autres fenêtres` / `Separate main window from other windows` for it. But there is not. Those two options lack tooltips.

zetok:
@TheNain38 could those changes be squashed into single commit?

TheNain38:
@zetok how do you do that? It's my first time using github.

aaannndddyyy:
@TheNain38 : and could you please fix 'arreté' - lack of circonflex. Not part of your current changes, but in the same file.

aaannndddyyy:
@TheNain38 Thank you.

To squash:
    cd /path/to/qTox
    git pull
    git checkout master
    git rebase -i e6e0279^

An editor opens up
Il te faut remplacer "pick" par "f" 
All but the very first one. 
This means all the subscequent commits will be squashed into the first one. 
So as a result you will have a single commit.

verify that it is all correct

then:
    git push --force

TheNain38:
How do you accept the changes in the editor?

aaannndddyyy:
ctrl+X 
it asks you then whether to save (Y)

TheNain38:
@aaannndddyyy Thanks, @zetok you can merge now, I squashed them!

PR #2241, opened 2015-09-16
Correct support channel

zetok:


stal888:
but we're a team zetok

Issue #1926, opened 2015-06-28
qTox insane CPU & RAM usage due to starting new processes from clicking on tox: links


PhilHypnox:
Running qTox, I noticed a significant slow-down in everything CPU-intensive that I do (playing movies suddenly became impossible in even 720p). This is the task manager screenshot:

![uirh171](https://cloud.githubusercontent.com/assets/3148759/9689929/2d94ac08-5331-11e5-9eb1-5674fb6e6a1a.png)

Why is qTox eating up that much RAM & CPU?

I'm W7 x64.

tux3:
Did you intentionally open 6 different qTox instances?

PhilHypnox:
Nope. I just had it running for... about 7 to 9 days. I don't shut down my PC at all in general. Can't imagine it being on my side, unless these instances were started in issue 1925 while I was testing, and didn't shut down.

Which I really doubt.

tux3:
Alright, I'll take a look at it, thanks.

Can you post the log (in folder %APPDATA%/tox/qtox.log), please?

towlie:
@PhilHypnox Did you use video calls in these 7 to 9 days? If yes, did some of your contacts use utox?

zetok:
Reason - qTox cannot handle clicking on `tox:` URI well:
![*click* *click*](https://cloud.githubusercontent.com/assets/3148759/8398076/21d4e4d2-1dda-11e5-9b1a-9d9f49f92e57.png)

PhilHypnox:
http://pastebin.com/9Rs0Tkt3

the log. And yes, it does appear to be the case. It only goes back to normal once all instances of qTox are closed.

tux3:
@PhilHypnox 

Thanks, I have a good idea of what the problem is now.

Although a big part of the log is qTox not being able to connect to the network, is that a problem you had with qTox, or just spotty internet?

PhilHypnox:
Spotty connection, yes.

suhr:
Add tag l-slow.

zetok:
From #2271, it would appear that `tox:` links don't work on windows - since example provided in that issue didn't have correct ID, could this be checked with correct ID in "link", that will have 76 hexadecimal characters, e.g. `tox:90F4C2558794EDC54C62585541DC7105583D9E20B5B29F0034B5032A15F4CE3CEC4927D6D779` ?

PKEv:
and new users can't use this link because link "tox:...." was registered by nsis setup
[...]

PR #2244, opened 2015-09-17
Added support for local links in messages

PKEv:
Added support for local links in messages and correct detect links like "www.123.com"
default

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -177,38 +177,43 @@ void ChatMessage::hideDate()
 
 QString ChatMessage::detectAnchors(const QString &str)
 {
-    QString out = str;
+    QString out;
 
     // detect URIs
     QRegExp exp("("
-                "(?:\\b)(www\\.|http[s]?|ftp)://" // (protocol)://(printable - non-special character)
+                "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(^tox:[@\\w]+$)"); // starts with `tox` then : and only alpha-digits till the end
+                "|(?:\\b)(file:///.*$)" //link to a local file, valid until the end of the line
+                "|(?:\\b)(tox:[a-zA-Z\\d]{76}$)" //link with full user address
+                "|(?:\\b)(tox:\\S+@\\S+)"); // starts with `tox` then : and only alpha-digits till the end
                 // also accepts tox:agilob@net as simplified TOX ID
-    int offset = 0;
-    while ((offset = exp.indexIn(out, offset)) != -1)
+    //support for multi-line text
+    QStringList messageLines = str.split("\n");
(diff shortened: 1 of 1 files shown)


agilob:
Is this needed? The only case when I can think it's need is when you're chatting with someone in the same (V)LAN  and you want to access common resource on LAN drive.

Anyone for or against this change?

PKEv:
i use this case
And while there is support for such links...
it should not interfere with other users

TheNain38:
I use this to indicate a directory that is common to windows to help others, like:
file:///C:/Windows/System32/drivers/etc/hosts

agilob:
PR introduces regressions:

doesn't detect `tox:user@doman.com` format

First TOX-URI is detected correctly, second not, TOX id have constant length, third ID is invalid as well.

parsin2

First line shouldn't be detected as URI, opening it results in:
parsin1

PKEv:
It does not introduce - saves
in my stable qTox link "tox:agilob@toxme.com" don't detected too
@agilob You sure that link as "www.123.com" should not be detected?
The lack of specified protocol is not so critical, it is strange that this is not happening correctly of association of links (in win i can't reproduce) 
I'll see what can be done..

agilob:
> in my stable qTox link "tox:agilob@toxme.com" don't detected too

But it should ;)

> @agilob You sure that link as "www.123.com" should not be detected?

qTox uses qdesktopservices to open link and if protocol is undefined link is opened as local resource. www without http never works.

SkyzohKey:
So maybe that you could just replace links like `www.example.com` by `http://www.example.com`, It would basically solve the problem without adding a lot of regression to the link system, right ?

zetok:
@SkyzohKeyx no. Messing with user input is a bad idea.

agilob:
> , It would basically solve the problem without adding a lot of regression to the link system, right ?

No, we can't modify user input/output...

PKEv:
I just did not understand, but it actually earlier replace the protocol for the type of links work "www.123.com",

PKEv:
fix all, please look

PKEv:
I restored the substitution of reference because the user does not have to know how to write link
As an example in Google when you copy a link from the address bar is added to the protocol

agilob:
tox:<id> doesnt work

PKEv:
how it should work?
in the stable branch, nothing happens by pressing

agilob:
on my setup it opens 'add contact' window

PKEv:
in my stable qTox it start new process

zetok:
@PKEv  #1926 ?

PKEv:
i create new #2271 , because the title (in  #1926) does not reflect the nature of the problem

PKEv:
then, please merge

agilob:
@PKEv this message:

```
tox:A5CB22B35202E91CE49CD2CE07AF036C1410179D47D2E110A1E1D9D7AD59042FF1C7D6685D57
```

should be clickable `tox:76 hex characters`

PKEv:
fix

TheNain38:
Can this be merged?

agilob:
Sure, I do merges of ready code in evenings.

PR #2250, opened 2015-09-17
fixes #2249

### LINKED ISSUE #2249: rename detach option
The option "Open chats in separate window" is wrongly named.
Not only chats are ooened in a separate window. And if you have more than one chat they don't get a separate window each, but share one, unless another option is cheked. 
This leads to confusion.

Actually this very text or "Open each chat in an individual window" should be used for the option underneath it, now called "Don't group chat windows". The latter is not wrong, but harder to understand. People may under tsnad that chat windows are groupo by some criterion, which is not the case. Actually what's meant is that all chat windows are luped together into one group if this option is not ticked. So if it is ticked, each chat gets its individual separate chat window
see #2238

aaannndddyyy:


src/widget/form/settings/generalsettings.ui
```diff
@@ -377,7 +377,7 @@ instead of system taskbar.</string>
           <item>
            <widget class="QCheckBox" name="cbSeparateWindow">
             <property name="text">
-             <string>Open chats in separate window.</string>
+             <string>Use multiple windows</string>
```
zetok: This (new) wording indicates that with this option unchecked it's not possible to open a detached window (or have multiple windows), which is not the case, since user still can do that by manually by r-clicking.

zetok:
Also, could your commit name be a little more descriptive?

src/widget/form/settings/generalsettings.ui
aaannndddyyy: it does not say, it is not possible. Same issue you would have with the original., according to you it suggests you cannot open a detached window, but you can.
This option changes the default behaviour. 
I don't think, making the working of that option unnecessarily long would be a good thing to do.
So you would be mor ehappy with something like this: "Automatically use multiple windows (detached mode)"

zetok: what about `Multiple windows mode` ?

aaannndddyyy: I can live with both.
Anyone else has an opinion or a different suggestion?

agilob: Zetoks options seems to fit more to already existing options

aaannndddyyy:
done

src/widget/form/settings/generalsettings.ui
aaannndddyyy: thx for feedback @agilob

Issue #1898, opened 2015-06-22
Change login screen to make it less confusing


zetok:
Current login screen makes people think that Tox requires creating an "account" somewhere.

Probably should be changed to resemble more closely:
![mockup](https://camo.githubusercontent.com/8f80d79ec1408767500d85f9b8893b738cf4d473/68747470733a2f2f662e636c6f75642e6769746875622e636f6d2f6173736574732f343635363232372f3838313734372f36363330333333362d663935342d313165322d393834332d3439393531376432666534642e706e67)

LittleVulpix:
I wholeheartedly agree; people are questioning me why they have to create an account somewhere

ProMcTagonist:
+1

Change the colors and button size to match too.

smc2000:
I think that since the local file encryption is an optional feature, the login screen should represent that by saying 'Password (optional)' or some other way of letting the user know they don't need a password.

zetok:
@fatek while it is optional, it's simply better if people do use encryption. The best way to encourage to use it is simply to provide password field, and not say that they don't have to type anything in. :3

sudden6:
@zetok Maybe we could write something like "It is recommended you encrypt your profile with a password to prevent missuse" in a tooltip or message box if the password is empty?

zetok:
@sudden6 adding tooltip to password field when it's empty may be helpful.. though password field shouldn't be visible for profiles that are not encrypted, so it would be used only when creating a profile

sudden6:
@zetok yep, I fully agree that the password field should not be visible for unencrypted profiles.

linux-modder:
as someone who DOES use encrypted profiles,   The only part I think should change is  whether it is grayed out or not,   coming focused  IF  the user  checkbox's  the  use encryption or  indicates in the above mentioned tooltip / msg box so it's clear its not required but comes useable if selected OR accessing an encrypted profile.  Possibly  do the same for auto login  for encrypted profiles as its no longer auto.

zetok:
![improved](https://cloud.githubusercontent.com/assets/3148759/9953458/e896112a-5dd8-11e5-8cd8-86679b5ae3de.png) ![improved2](https://cloud.githubusercontent.com/assets/3148759/9953469/08c9dd82-5dd9-11e5-9449-a42627433e81.png)

Ok, it was improved - should `Login` be changed to `Existing profile` or something, or is it good now?

LittleVulpix:
[...]

PR #2253, opened 2015-09-18
Reword login window: "Login" → "Load Profile"

zetok:
load profile

### DIFF
--- a/src/loginscreen.ui
+++ b/src/loginscreen.ui
@@ -784,7 +784,7 @@ margin-bottom:5px;</string>
             <string>If the profile does not have a password, qTox can skip the login screen</string>
            </property>
            <property name="text">
-            <string>Login automatically</string>
+            <string>Load automatically</string>
            </property>
           </widget>
          </item>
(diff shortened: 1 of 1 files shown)


ovalseven8:
'Login' is more userfriendly in my mind...

zetok:
@ovalseven8 that depends - "load" implies that something happens locally (like loading save in a game), whereas "login" appears to be used with something in regard to some online service, where you don't have your data locally.

To quote what @LittleVulpix wrote about it in #1898:

> How about "load/use existing" or something that makes it clear the profile is actually on your device, not "somewhere" ? Too often I meet people who say "weird I logged in with the same user and pass but I don't see any of my contacts" when in reality they just created a new tox profile on a different device.

TheNain38:
Yes, I think "load" is better for the same reason. This is also why they changed "Create Account" to "Create Profile".

ovalseven8:
Of course, I can live with @zetok's changes as well. :)

ProMcTagonist:
I like Load. The game save comparison is good.

Zer0-One:
"load" is more appropriate here; was expecting someone to do this at some point. 

+1

PR #2254, opened 2015-09-18
restore window after screenshot

PKEv:
Added recovery after a screenshot of the application window as the application window can be hidden during screenshot.

### DIFF
--- a/src/widget/tool/screenshotgrabber.cpp
+++ b/src/widget/tool/screenshotgrabber.cpp
@@ -47,8 +47,9 @@ ScreenshotGrabber::ScreenshotGrabber(QWidget* parent)
 
 void ScreenshotGrabber::reInit()
 {
+    delete scene;
     scene = new QGraphicsScene;
-    window = new QGraphicsView (scene); // Top-level widget
+    window = new QGraphicsView(scene); // Top-level widget
     setupWindow();
     setupScene(scene);
     showGrabber();
@@ -89,7 +90,7 @@ bool ScreenshotGrabber::handleKeyPress(QKeyEvent* event)
     {
         Widget *Widget = Widget::getInstance();
         blocked = true;
-        if ( Widget->isVisible())
+        if (Widget->isVisible())
             Widget->setVisible(false);
         else
             Widget->setVisible(true);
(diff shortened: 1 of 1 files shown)


PKEv:
thank @agilob

src/widget/tool/screenshotgrabber.cpp
```diff
@@ -115,6 +116,7 @@ void ScreenshotGrabber::acceptRegion()
     qDebug() << "Screenshot accepted, chosen region" << rect;
     emit screenshotTaken(this->screenGrab.copy(rect));
     this->window->close();
+    Widget::getInstance()->setVisible(true);; // show window if it was hidden
```
TheNain38: Why is there two `;;` at the end?

TheNain38:
Could someone merge it, please?

src/widget/tool/screenshotgrabber.cpp
```diff
@@ -47,6 +47,7 @@ ScreenshotGrabber::ScreenshotGrabber(QWidget* parent)
 
 void ScreenshotGrabber::reInit()
 {
+    delete scene;
     scene = new QGraphicsScene;
     window = new QGraphicsView (scene); // Top-level widget
```
TheNain38: This is not part of your changes but could you remove the space in `QGraphicsView (scene)` ?

TheNain38:
@zetok Can this be merged?

agilob:
No, I have one more objection. I don't think it's safe to create another copy of Widget pointer as it happens there: `Widget *Widget = Widget::getInstance();` someone might by accident mess with it in the method. Instead the if below I would suggest:

Line 93:

```
// set opposite visibility, visible if hidden, hide if visible
Widget::getInstance()->setVisible( ! Widget::getInstance()->isVisible());
```

sorry @PKEv I didn't notice it before.

PKEv:
I think that the visibility needed restores anyway, it was hidden or not
better then changed to

```
if ( ! Widget::getInstance()->isVisible())
Widget::getInstance()->setVisible(1);
```

agilob:
I don't have preference if you write it in one line or four like you suggested ;)

PKEv:
fix

agilob:
Merged in https://github.com/tux3/qTox/commit/a53f6f2adfff59b28fa27fb522ccff235d7a7e58

PR #2258, opened 2015-09-18
More consistent translation strings in groupchatform.cpp

TheNain38:
It should not be able to be translated differently as one update when the number of user in the group chat change and the other is when language change, so it should be translated together.

https://github.com/tux3/qTox/blob/f8352d63ff17bbb551fc575252b747ac95208b99/src/widget/form/groupchatform.cpp#L167
https://github.com/tux3/qTox/blob/f8352d63ff17bbb551fc575252b747ac95208b99/src/widget/form/groupchatform.cpp#L370

### DIFF
--- a/src/widget/form/groupchatform.cpp
+++ b/src/widget/form/groupchatform.cpp
@@ -164,7 +164,7 @@ void GroupChatForm::onSendTriggered()
 
 void GroupChatForm::onUserListChanged()
 {
-    nusersLabel->setText(tr("%1 users in chat").arg(group->getPeersCount()));
+    nusersLabel->setText(tr("%1 users in chat", "Number of users in chat").arg(group->getPeersCount()));
 
     QLayoutItem *child;
     while ((child = namesListLayout->takeAt(0)))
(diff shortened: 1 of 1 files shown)


zetok:
Thanks, though in the future, it would be good to use a bit more descriptive commit names `:)`

TheNain38:
Yes, but I thought about it, and couldn't find an appropriate one.

zetok:
Well, a ~fine for this one would be something like

```
More consistent translation strings in groupchatform.cpp
```

Also, in there's no reason to not put in commit message what you wrote in PR message, aside from links.

https://github.com/tux3/qTox/wiki/Coding#general-tips :)

TheNain38:
Ok, thanks I take note.

zetok:
@TheNain38 I'm about to push that merge on master - do you want to improve commit message, or should I merge it as it is?

TheNain38:
how do you change a commit message?

zetok:
You use commit with `--amend` to amend last commit:

``` bash
git commit -a --amend -m "More consistent translation strings in groupchatform.cpp

It should not be able to be translated differently as one update when the number of user in the group chat change and the other is when language change, so it should be translated together."

git push -f origin translation-regroup
```

TheNain38:
Yes! I have succeded. You can merge now.

zetok:
Thanks `:)`

Will merge shortly.

PR #2264, opened 2015-09-19
Fix incorrect copyright headers

zetok:


agilob:
There's no need to keep that license in each file, it could just say, `this program is on license, find more in repository.`

zetok:
Um, according to https://www.gnu.org/licenses/gpl-howto.html :

> This statement should go near the beginning of every source file, close to the copyright notices.

ProMcTagonist:
+1 @zetok. Rogue domain needs to be gone.

rugk:
> I doubt "The qTox Project" can be a copyright holder.

Indeed. This would have to be replaced with tux3 or something like this.

aaannndddyyy:
@tux3 I know what you mean it to be.
But what you mean and what a legal docoment acutally says are not always identical.
Don't get me wrong, I don't want to be anal about it, and ianal, but if someone were to create a legal entity call like that, not afiliated to you at all, you might run into problems.
You would still have good chances to win, by simply showing that your code predates the registration of the new entity, however, it is not the safe way.
As a rule of thumb: The copyright lies generally with the original authors, unless they transfer it to others, which could be done by releasing with a note saying copyrigh XYZ. 
Whenever you want to mention who holds the copyright, make sure it is a legal entity (natural person, company, …)
Do not use unregistered names of let#s say abstract communities. 
Who or what is "The qTox Project"? Is it only those who write the code, docs, and translations? Is it all who somehow contribute by hosting, building, PR, … ?
Is it the entire community ? All who use it, give feedback, love it, hate it, …?
Have you registered the name?
…
"Copyright" + year (even if this is not the way FSF recommends it, is possible), "Copyright" + years + names of individual contributors what FSF suggests, but this can become a loooong list, "Copyright" + years + "the authors of this software" is possible, but redundant (they hold it by default) and not helpful/informative.

I think you get my point

rugk:
@aaannndddyyy :+1: :+1: :+1: :pray: 
A license is a legal thing so you can't do everything (including inventing names) there just because you think that "sounds better" or you want to show that it's a community.
For the community it does not matter what name is used there - the only thing which matters is which license used.

The name only matters for the legal things. And as it's also not really clear whether a [pseudonym is allowed](https://web.archive.org/web/20150706043215/http://stackoverflow.com/questions/13497692/can-i-license-my-programs-under-a-pseudonym), but it is [used by book authors](https://web.archive.org/web/20121104190340/http://www.mindyklasky.com/index.php/writing-advice/to-have-and-have-not-a-pseudonym-that-is) and there is also a [statement in the US copyright](https://web.archive.org/web/20150427001911/http://www.copyright.gov/fls/fl101.html) about this I think it's possible.
However with completely fictional names (which are obviously not one of a single person, but should include an undefined number of unnamed persons) this is something else. I would not use it.

> Have you registered the name?

He could do it [like this](https://en.wikipedia.org/wiki/Doing_business_as) but the problem there would be whether this is applicable worldwide. (However I'm not sure if this is possible for natural persons and not only companies)

@tux3

> As long as "The Tox Foundation" or "The Tox Project" is not a copyright holder either, we're all safe.

Even that's not bad either. If "The Tox Foundation" or "The Tox Project" is a registered (valid/legal) name, because it's an registered organisation e.g. this should be possible.
Possibly "The Tox Foundation" could change the license in this case (to a proprietary one), but that would not do much: Users who have a (recent) version of the source forked can just use this for further developing. The copyright holder cannot recall it's old license. They can only release new versions under a new license.
Propably I would even prefer the latest option as you have a [legal personality](https://en.wikipedia.org/wiki/Legal_personality) there, which is not a single person (which is obviously what you want).

---

Disclaimer: Note that I'm no lawyer so what I say here should be treated with care.

aaannndddyyy:
@rugk  The Tox Foundation is registered but not owned by project members, but a rogue ex member. 
That is why this must be taken out asap. (You could not know that if you have not been around for longer)
I don't know if this is even actually legal, to take Tox Foundation out.
It could be argued that by submitting the code with that license and submitting PR's to it as is, the authors may have passed on their rights to the Foundation. If that was the case, they authors _lose_ their right to the code, and henceforth the right to decide over the license.
So this PR could technically be illegal.

So to be clear: In case it is illegal, I am not recomming to to the change (recommending it would be a crime in some jurisdictions). But my general peace of mind would be much more peacful without that name in the copyright note ;)

rugk:
> @rugk The Tox Foundation is registered but not owned by project members, but a rogue ex member.
> That is why this must be taken out asap. (You could not know that if you have not been around for longer)

I know about the one who stole the money... But he was the only who has the rights of the Tox Foundation?
I mean he managed the money, but he was also the legal representative of the Foundation?

In this case it's of course not desired to put "Tox Foundation" in the license file.

tux3:
It's still a bit unclear to me what the proper copyright notice should be, so I'll take this now for the sake of fixing the incorrect "Tox Project" mention, and feel free to send a new PR if you speak actual legalese.

tux3:
Awkwardly merge-rebased in 90d1f4be02ecb43638d5a0ce5cc9842b0a1c7fef, closing.

zetok:
@tux3 uuh, master is now not signed

zetok:
eh, I guess that way to solve it would be 

```
git checkout c5127e3654cb315e4301102a685ddccef12ce108 -b signed-master
git merge --no-ff master -m "Sign master"
# since `signed-master` branch is signed, it can be fast-forwarded to master
git checkout master
git merge --ff signed-master
# master signed
```

Now, I'm not sure if I should do this, since it would look ~like merged PR - is anyone else willing?

tux3:
Yes, rebasing after the merge instead of before wasn't my brightest idea, it deleted the signature.
You can always push an empty signed commit if you want:

```
git commit --allow-empty -m "Sign HEAD"
```

But since this commit only changes comments it'd be pretty hard to slip in a bad change without anyone noticing.

PR #2276, opened 2015-09-22
Make whole icon change colour to orange when blinking on new event

### LINKED ISSUE #2273: [Proposal] change of blinking icon on new message
hi all,
I've try some Tox client on different OS.
On windows, i cannot see new messages because the icon change (green to orange) is not so easy to notice.
When i look at µTox we have real changes to catch the new message (+notification), the icon become red, and that is easy to see.
sans titre 1

I think It will be better to do the same.

zetok:
Previous ones blinked with only small status dot, which was hardly visible.
Now blinking icon should be a lot more visible.

closes #2273 

https://d.maxfile.ro/wlyrvixwhi.webm

### DIFF
--- a/img/taskbar/dark/taskbar_event.svg
+++ b/img/taskbar/dark/taskbar_event.svg
@@ -1,10 +1,25 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!-- Generator: Adobe Illustrator 18.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
-<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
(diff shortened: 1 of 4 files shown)


agilob:
Test it with http://www.color-blindness.com/coblis-color-blindness-simulator/ before introducing potential regression.

zetok:
@agilob I'm not sure how am I supposed to test it - I've saved icon as .png and uploaded there, but when I've tried to change mode from `Normal Color Vision` to something else, only default images showed up.

jin-eld:
Would be great to have a more prominent notification, I keep overseeing it when only the circle inside the lock is blinking, so I'd be happy to see this one merged :)

zetok:
@agilob ok, so, basically, what is the regression here, given that currently even non-colour-blind people can't see the notification?

I.e. is there a reason to not merge it?

agilob:
Could change the commit and add this image as second? taskbar_icon_orange_green.svg (or something like that), I can clearly see current blinking icon. The one from this commit is too aggressive for me. I can add icon setter as setting parameter loaded on startup (but without GUI to change it).

zetok:
@agilob yeah, I guess - although I'm not so sure that adding yet another setting is a good idea.

zetok:
@agilob ok, done.

PR #2277, opened 2015-09-22
correct screenshot tooltip

### LINKED ISSUE #2267: screenshot mode note's transparency is ugly
The little yellow note informing about the options is rather transparent.
This looks bad as it is directly over the qTox title or any maximized window's title.
It does not get into the screenshot, but it looks not nice.

PKEv:
fix #2267
has made the message disappears when the cursor is over it


aaannndddyyy:
Thank you

PR #2279, opened 2015-09-23
fix missing window buttons in "multi-window" mode

### LINKED ISSUE #2270: No "X" button to exit in gnome after enabling multiple windows
There isn't an "X" button in the top right corner when I enable multiple windows. I am running gnome and it looks like this
screenshot from 2015-09-20 12-03-52

This does have a workaround however, if i put my mouse in the top-left corner of the screen, i can close it from there, like this:
screenshot from 2015-09-20 12-10-40

antis81:
fix #2270

### DIFF
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -46,7 +46,7 @@
 #include <QPushButton>
 
 GenericChatForm::GenericChatForm(QWidget *parent)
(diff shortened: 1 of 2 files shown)


agilob:
I don't really understand, author of the issue says it doesn't fix the problem?

zetok:
@agilob yes, after testing it

agilob:
oki

PR #2283, opened 2015-09-23
Made the 'remove friend dialog' user-friendly again

### LINKED ISSUE #2281: You have to type "YES" to remove a friend.
capture

I mean, really?

alexbakker:
#1961 changed the friend removal dialog in a way that hurts the UX, this pull request fixes that.

Here's what it will look like:

- 'No' does not take any action at all.
- 'No' is focused by default, making accidental deletion harder.
- 'Yes' removes the friend from the friend list but also removes chat history if the checkbox is checked.

This should also fix #2281.

### DIFF
--- a/src/widget/form/removefrienddialog.ui
+++ b/src/widget/form/removefrienddialog.ui
@@ -46,7 +46,7 @@
       <enum>Qt::Horizontal</enum>
      </property>
      <property name="standardButtons">
-      <set>QDialogButtonBox::No|QDialogButtonBox::Yes</set>
+      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
      </property>
     </widget>
    </item>
--- a/src/widget/form/removefrienddialog.ui
+++ b/src/widget/form/removefrienddialog.ui
@@ -7,8 +7,14 @@
     <x>0</x>
     <y>0</y>
     <width>300</width>
-    <height>180</height>
+    <height>110</height>
    </rect>
+  </property>
+  <property name="minimumSize">
+   <size>
+    <width>300</width>
+    <height>110</height>
+   </size>
   </property>
   <property name="windowTitle">
    <string>Remove friend</string>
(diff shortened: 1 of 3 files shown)


agilob:
It should be Yes|No instead of No|Yes. 

Can you replace ID with username?

ProMcTagonist:
"Yes" should say "Remove" or "Delete".

http://uxmovement.com/buttons/why-the-ok-button-is-no-longer-okay/

alexbakker:
@agilob 

> It should be Yes|No instead of No|Yes.

Why would you want that? Every dialog in qTox is like this: Negative|Affirmative. For example, it's Cancel|Remove in the friend remove dialog right now and I thought that made sense because we probably want the option that's selected initially to not have any effect.

> Can you replace ID with username?

The ID is only visible there because the friend I tried to delete in the screenshot didn't have a name. In other cases, it will display the name.

alexbakker:
@ProMcTagonist 
Does that apply here? Yes and No are valid answers to the question that's being asked whereas 'Ok' would be a lot more vague indeed.

Perhaps you're right though, I don't know much about UX. If more people feel it should say 'Remove', I'll make that change.

TheNain38:
I find it clearer to have the button say 'remove' than 'yes'

zetok:
@Impyy you're introducing regressions under pretence of improving things.

IMHO this shouldn't be merged.

alexbakker:
@zetok 
Can you be a little more specific? Which regressions did I introduce?

zetok:
@Impyy first, need to type in confirmation, and second one, making buttons less clear.

alexbakker:
@zetok 
The whole point of this PR is to remove the need to type in a confirmation, that's not a regression. 
Also, I've shown willingness to change the text of the buttons back. I thought Yes and No would be better, looks like I was wrong.

alexbakker:
Done.

agilob:
@Impyy I checked on window that should have two buttons:
weblate1

If other windows say Cancel|OK then it's wrong, it's a better practise to use OK|Cancel.

alexbakker:
@agilob 
That dialog looks like this for me:


I read up on how button boxes work and apparently the order of the buttons inside a button box differ on each platform: http://doc.qt.io/qt-4.8/qdialogbuttonbox.html#details. If you want to force the buttons in every dialog to be in the same order on all platforms, you'd have to get rid of every button box but that is outside the scope of this PR.

agilob:
Hah, I didn't know it might be changed dynamically.

weblate2

alexbakker:
You may also want to merge the last commit (25ff57ad23ae66ccd89a5aa4bf3e06ec434033b6), @agilob.

Issue #1595, opened 2015-05-09
White screen when trying to take a screenshot on Windows


alexbakker:
I use Windows 8.1 with two 1080p monitors.
If I try to take a screenshot, my second screen turns completely white. If I move qTox to my second screen and try to take a screenshot the same thing happens.

PKEv:
Win 7 32x
Really second monitor completely ignored when screenshoting

agilob:
@Papierkorb

Papierkorb:
Don't have a dual-screen windows machine where I'm currently, earliest for me would be this week-end.

PKEv:
closed

alexbakker:
I don't see how #1748 is a proper fix for this. My second monitor is positioned on the left so I'm still experiencing this issue. Can this be reopened?

zetok:
Could someone using windows test https://build.tox.chat/view/Clients/job/qTox-testing_build_windows_x86-64_release/lastSuccessfulBuild/artifact/qTox-testing_build_windows_x86-64_release.zip , which should have it fixed?

antis81:
@PKEv @Impyy Can you confirm this is fixed.

zetok:
@antis81 why remove `A-windows` though, since issue seems to affect only windows?

antis81:
@zetok No, it affects a multi-monitor setup with **primary screen on right side**. In this case the previous aproach uses wrong coordinates to take the snapshot. The code I wrote picks the right "panorama" of all screens now, no matter in which order they are. Now user picks the wanted area out of it and takes the screenshot. That's all… :)

zetok:
I see. Thanks for explanation `:)`

PKEv:
White screen on the second monitone
(primary screen on right side)
During selection:
![default](https://cloud.githubusercontent.com/assets/6053197/10096093/9795b4a2-6377-11e5-997b-b3b705959d70.png)
result:
![qtox-screenshot-olia2428](https://cloud.githubusercontent.com/assets/6053197/10096108/ac58663c-6377-11e5-864f-045164ecb730.png)

antis81:
Sorry, I was wrong. Problem only exists with multi-monitor setup on old super duper Windows systems…

Unfortunately I cannot reopen PR #2289 -> new one required. 

@PKEv
Had a look into the very Qt5 code and found the solution:
[...]

PR #2262, opened 2015-09-18
full support screenshot 2nd monitor

PKEv:
must fix #1595 
I do not have a second monitor please check: @Impyy @blobgo 
Check please in both configurations: the left and right sides of the main display

PR #2289, opened 2015-09-24
fix screen grabbing for screenshots (including multi-monitor setup)

antis81:
implements PR #2262 and closes issue #1595


agilob:
I can't test it, so I trust you with this one ;)

antis81:
@agilob Please don't merge untested code - it didn't work! This is somewhat special, because it's a bug in Qt (Windows actually :smile:). So I checked all the internal Qt code paths for `QPlatformScreen` and found a solution…

PR #2297, opened 2015-09-25
fix position/size of screen grabber window

antis81:
enhancement for #2289

### DIFF
--- a/src/widget/tool/screenshotgrabber.cpp
+++ b/src/widget/tool/screenshotgrabber.cpp
@@ -76,7 +76,14 @@ void ScreenshotGrabber::showGrabber()
     this->window->show();
     this->window->setFocus();
     this->window->grabKeyboard();
-    adjustWindowSize();
+
+    QRect fullGrabbedRect = screenGrab.rect();
+    qDebug() << "adjusting grabber size to" << fullGrabbedRect;
+
+    this->window->setGeometry(fullGrabbedRect);
+    this->window->scene()->setSceneRect(fullGrabbedRect);
+    this->overlay->setRect(fullGrabbedRect);
+
     adjustTooltipPosition();
 }
 
@@ -189,21 +196,6 @@ void ScreenshotGrabber::reject()
     qDebug() << "Rejected screenshot";
     this->window->close();
     Widget::getInstance()->setVisible(true); // show window if it was hidden
-}
-
-QRect ScreenshotGrabber::getSystemScreenRect()
-{
-    return QApplication::primaryScreen()->virtualGeometry();
-}
-
-void ScreenshotGrabber::adjustWindowSize()
-{
-    QRect systemScreenRect = getSystemScreenRect();
-    qDebug() << "adjusting grabber size to" << systemScreenRect;
-
-    this->window->setGeometry(systemScreenRect);
-    this->window->scene()->setSceneRect(systemScreenRect);
-    this->overlay->setRect(systemScreenRect);
 }
 
 QPixmap ScreenshotGrabber::grabScreen()
--- a/src/widget/tool/screenshotgrabber.cpp
+++ b/src/widget/tool/screenshotgrabber.cpp
@@ -34,8 +34,8 @@
 #include "toolboxgraphicsitem.h"
 #include "src/widget/widget.h"
 
-ScreenshotGrabber::ScreenshotGrabber(QWidget* parent)
-    : QWidget(parent)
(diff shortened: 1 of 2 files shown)


antis81:
@PKEv Can you please fetch this branch and check?

agilob:
Just add [WIP] in name [work in progress]

PKEv:
super, but not everything works
default
does not select the left screen

PKEv:
or not, on the second screen does not unfold widget
(on image of the second screen is not dimmed)

antis81:
Thanks - it's a bit hard to see on the above screenshot.

This is actually due to the grabber window's positioning (it appears on the right screen only). I can reproduce this behaviour on Linux, when deactivating some "magic" flag (look for `Qt::BypassWindowManagerHint`). :smile:

Qt defines the widget position `P(0,0)` at the top-left corner of the **primary screen's desktop** (in our case the one on the right side). This usually makes sense, because we naturally position windows like that. Thus, when maximizing a window on a single screen it always has position `P(0,0)`.

I'll fix it and push soon…

antis81:
@PKEv Same procedure. Should be ok now - I don't really understand why, but `window->moveTopLeft(QPoint(-1,-1));` does the trick on any platform… :)

Cleaned up the implementation a bit and found another bad bug:
The call from Chatform creates a mem-leak, which might hurt when creating several screenshots in a row…

src/widget/tool/screenshotgrabber.cpp
```diff
@@ -76,7 +76,14 @@ void ScreenshotGrabber::showGrabber()
     this->window->show();
     this->window->setFocus();
     this->window->grabKeyboard();
-    adjustWindowSize();
+
+    QRect fullGrabbedRect = screenGrab.rect();
+    QRect nr(QPoint(-1,-1), fullGrabbedRect.size());
```
zetok: um, since "magic numbers" are used, perhaps there should be added a comment saying that they're fixing a problem?

antis81: I'd like to really understand why this happens, being able to control it. :)

antis81:
@PKEv Can you fetch and check that on Windows please?

PKEv:
@antis81 Not early Monday...

antis81:
@PKEv Thanks. I'll happily wait for reply and hope it fixes your (and lots of others) problem.

PKEv:
I test, it didn't help. 
default

But your idea help me getting it done:

I returned the code, otherwise the screen shot is only one monitor

```
QPixmap ScreenshotGrabber::grabScreen()
{
    QRect rec = QApplication::primaryScreen()->virtualGeometry();
    return QPixmap::grabWindow(QApplication::desktop()->winId(),
                               rec.x(),
                               rec.y(),
                               rec.width(),
                               rec.height());
}
```

In the widget settings, pass the offset in variable rec

```
    QRect fullGrabbedRect = screenGrab.rect();
//    QRect nr(QPoint(-1,-1), fullGrabbedRect.size());
    QRect rec = QApplication::primaryScreen()->virtualGeometry();
    this->window->setGeometry(rec);
    this->scene->setSceneRect(fullGrabbedRect);
    this->overlay->setRect(fullGrabbedRect);
```

result:
default
It remains only to move the text label
@antis81  You can check it out at itself?

antis81:
@PKEv
If I got you right, a `QGuiApplication::primaryScreen()->grabWindow(0);` grabs only the primary screen on Windows? If so, there's a bug in Qt (Windows) code and I'll report it.

I check the code as soon as I can tomorrow.

PKEv:
@antis81 i don't think that is bug (virtual desktop != screen)

antis81:
@PKEv It is a bug, because: On Linux/Mac `QScreen::grabWindow(0);` grabs the virtual desktop, on (old?) Windows, it grabs the primary screen (if I understood you right). Btw. the underlying code for Linux/Mac is much more simple and straight forward than Windows code - but that doesn't matter here… :smile:

antis81:
@PKEv Also works on Linux like expected. :+1:

agilob:
Can I test it and merge? I wonder if it works on virtual desktops on KDE.

antis81:
@agilob Yes, please!

PKEv:
the text is not that in the middle of the screen
default

antis81:
@PKEv I'll do it in another PR after this one got merged.

agilob:
I'll test and merge this one in 4-6hours.

agilob:
It's OK on KDE5.

PR #2298, opened 2015-09-25
Load user status from data file

alexbakker:
Toxcore saves the user status, let's make use of that.

This also fixes a small issue where the status icon wouldn't be changed to offline when 'Reconnect' is pressed.


zetok:
oops, for a moment I thought it was an issue

PR #2304, opened 2015-09-26
Better translations with plural and fix something not being translated dynamically

TheNain38:
Could someone please check before merging.

### DIFF
--- a/src/widget/form/groupchatform.cpp
+++ b/src/widget/form/groupchatform.cpp
@@ -164,7 +164,11 @@ void GroupChatForm::onSendTriggered()
 
 void GroupChatForm::onUserListChanged()
 {
-    nusersLabel->setText(tr("%1 users in chat", "Number of users in chat").arg(group->getPeersCount()));
+    int peersCount = group->getPeersCount();
+    if (peersCount == 1)
+        nusersLabel->setText(tr("1 user in chat", "Number of users in chat"));
+    else
+        nusersLabel->setText(tr("%1 users in chat", "Number of users in chat").arg(peersCount));
 
     QLayoutItem *child;
     while ((child = namesListLayout->takeAt(0)))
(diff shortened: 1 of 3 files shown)


src/widget/groupwidget.cpp
```diff
@@ -145,7 +147,13 @@ void GroupWidget::onUserListChanged()
...
-        statusMessageLabel->setText(tr("%1 users in chat").arg(g->getPeersCount()));
+    {
+        int peersCount = g->getPeersCount();
+        if (peersCount == 1)
+            statusMessageLabel->setText(tr("1 user in chat"));
+        else
+            statusMessageLabel->setText(tr("%1 users in chat").arg(peersCount));
+    }
     else
         statusMessageLabel->setText(tr("0 users in chat"));
```
TheNain38: Is it possible to have 0 users in a group chat? If you are in, there is 1 user, you, and if you aren't then it doesn't even exist

src/widget/form/groupchatform.cpp
```diff
@@ -367,5 +371,9 @@ void GroupChatForm::keyReleaseEvent(QKeyEvent* ev)
 
 void GroupChatForm::retranslateUi()
 {
-    nusersLabel->setText(GroupChatForm::tr("%1 users in chat", "Number of users in chat").arg(group->getPeersCount()));
```
TheNain38, line 370: Why is there `GroupChatForm::tr` instead of `tr` ?
The `tr` is used in a `GroupChatForm` method.

src/widget/groupwidget.cpp
agilob: It shouldn't be possible to have 0 users in a chat.

TheNain38:
I will update this PR, wait before merging.

TheNain38:
You can merge now

PR #2309, opened 2015-09-27
Added dynamic translation to status and username

TheNain38:
Could someone check before merging?

PR #2311, opened 2015-09-27
Maybe fix crash because of #2309

TheNain38:
#2309

Could someone check if it fixes crash?

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -858,6 +858,8 @@ QSplitter:handle{
                  <property name="cursor">
                   <cursorShape>PointingHandCursor</cursorShape>
                  </property>
+                 <property name="text">
+                 </property>
                 </widget>
                </item>
                <item>
@@ -936,6 +938,8 @@ QSplitter:handle{
                   <font>
                    <pointsize>8</pointsize>
                   </font>
+                 </property>
+                 <property name="text">
                  </property>
                 </widget>
                </item>
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -2032,30 +2032,10 @@ void Widget::setActiveToolMenuButton(ActiveToolMenuButton newActiveButton)
 
 void Widget::retranslateUi()
 {
-    QString name = ui->nameLabel->text(), status = ui->statusLabel->text();
     Core* core = Nexus::getCore();
     ui->retranslateUi(this);
-    if (core->getUsername().isEmpty())
-    {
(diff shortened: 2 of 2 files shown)


tux3:
One second.
At first glance retranslateUi can be called even if Core isn't started, so you need to check that it's not a nullptr before calling getUsername/getStatusMessage on it

That would explain the crash. Did you try running it in a debugger?

TheNain38:
No, I didn't try.

TheNain38:
If it is `nullptr` it would set the username and status to empty "" ?

tux3:
Alright, I ran it and it indeed crashes because tox is a nullptr.
I'll push a fix, but please try to be more careful next time!

TheNain38:
@tux3 Yes, for sure, but I can't setup a build environement on my windows neither on my linux. Also I didn't had the time to see and understand all part of the code. Sorry, I will be more careful.
Also thanks for creating qTox. It's the best client ever. ;)

tux3:
@TheNain38 okay, I understand. Thanks for contributing :)

TheNain38:
@tux3 And also the `setUsername(core->getUsername());` and `setStatusMessage(core->getStatusMessage());` to avoid duplicated code?

tux3:
@TheNain38 Yes, I pushed that too in 14235c5a19fbd3c8484ec16068ea33a72d64ff0a

TheNain38:
@tux3 Ah, didn't see

TheNain38:
@tux3 you can also delete the line 2035 in 'src/widget/widget.cpp'

tux3:
@TheNain38 You're right, thanks!

zetok:
@TheNain38 

> I can't setup a build environement on my windows neither on my linux.

If you have problem(s) with that, please say what kind of problem those are - that is, if you use distro that is sufficiently "new" to build qTox, instructions in INSTALL.md should be sufficient - if they're not, it's an issue and should be reported.

Also, if instructions are hard to follow, they should be improved, so it would be good if you could point out what is problematic `:)`

TheNain38:
@zetok In my Windows 8.1 64-bits I have problems with missing filter_audio.h and vpx/vpx_image.h and some other I don't remember and on my linux mint I can't build toxcore.

zetok:
Well, dunno about building on windows, but what is the problem with building toxcore on mint?

What **exactly** is the problem?

TheNain38:
@zetok I will tell you more the next evening, I need to sleep and I'm on my phone, so I can't say much right now.

PR #2312, opened 2015-09-28
Update fr translation: 'Your name' and 'your status'

TheNain38:


zetok:
This PR doesn't seem to be needed, provided that the only change that is made aside from running `lupdate` is removal of `type="unfinished"` - change that is not needed for translation to be correctly applied:
works

It mostly only inflates git history - should it be merged?

TheNain38:
@zetok If so it would be better to update all translations with `lupdate` in one commit?

zetok:
@TheNain38 no, what I mean, this change is currently unneeded - i.e. in future, when there will be some real change to the source code which would render translations obsolete and/or add new translations, then new commits would be good.

As for updating all translations with lupdate - there is no need for it, since they're already "updated" like that just before compiling.

TheNain38:
@zetok Then the French translation isn't loaded on my qTox e1509a9f03d721fe9aa8741f63bdbaea7f3c5b6a
When opened with Qt Linguist before and after I do `lupdate` there is the new translations

zetok:
That's weird.. welp, looks like I'm going to merge this PR `;)`

TheNain38:
Merge a PR that update with `lupdate` all other languages too

TheNain38:
Wait, that's weird only Spanish is translated

zetok:
> Merge a PR that update with `lupdate` all other languages too

err, that shouldn't be needed

TheNain38:
@zetok maybe the `lupdate` proccess in compiling is broken? because it fixed the problem

zetok:
Well, if it was completely broken, qTox wouldn't compile - I wonder though whether it might have something to do with translations being updated concurrently to the source being compiled.

PR #2314, opened 2015-09-28
Fix dir to user profile

### LINKED ISSUE #2183: Open directory of current profile
When Make Tox portable is checked, clicking on 'Current profile location' it opens D:\ on my computer instead of the current directory.
Also the cursor is the pointer when on the same line as 'Current profile location' even when not on the words.

capture

Sorry if this is not clear, I'm not very good at speaking in English.

O.S.: Windows 8.1 64 bit
qTox commit: 9c65603e3da3c619dc2a285ab42bcf4f123e8aba

PKEv:
fix #2183 (for portable mode)

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -153,10 +153,12 @@ void ProfileForm::show(ContentLayout* contentLayout)
     head->show();
     QWidget::show();
     prFileLabelUpdate();
-    QString DirPath = QDir(Settings::getInstance().getSettingsDirPath()).path().trimmed();
+    QString DirPath = Settings::getInstance().getMakeToxPortable() ? QApplication::applicationDirPath() :
+                                                                    QDir(Settings::getInstance().getSettingsDirPath()).path().trimmed();
     bodyUI->dirPrLink->setText(bodyUI->dirPrLink->text().replace("Dir_Path",DirPath));
     bodyUI->dirPrLink->setOpenExternalLinks(true);
     bodyUI->dirPrLink->setTextInteractionFlags(Qt::LinksAccessibleByMouse | Qt::TextSelectableByMouse);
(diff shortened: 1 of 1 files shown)


tux3:
`getSettingsDirPath()` already checks whether qTox is in portable mode, what is this trying to do with `QApplication::applicationDirPath()`? Is it to get the absolute path?

PKEv:
yes, it check but return string like (QString(".")+QDir::separator())
move on such a link is not possible, and that would not have engaged in the analysis of the returned string, I do my check

src/widget/form/profileform.cpp
```diff
@@ -153,7 +153,8 @@ void ProfileForm::show(ContentLayout* contentLayout)
     head->show();
     QWidget::show();
     prFileLabelUpdate();
-    QString DirPath = QDir(Settings::getInstance().getSettingsDirPath()).path().trimmed();
+    QString DirPath =Settings::getInstance().getMakeToxPortable() ? QApplication::applicationDirPath() :
```
TheNain38: Why is there no space in `=Settings`

TheNain38:
@PKEv but the bug where when your mouse cursor is in the same line as the text but not on the words is not corrected? Because my bug report that too, but not well explained.

PKEv:
@TheNain38 i don't understand

TheNain38:
@PKEv If you hover with your mouse on the line where 'Current profile location: ' is, then it shows the 'pointer' mouse even when not on the words.
Sorry, if this is not explained well.

PKEv:
fix

TheNain38:
@PKEv Thanks. @tux3 Could it be merged?

agilob:
@PKEv is it ready? can I merge it?

PKEv:
Yes, please merge.

agilob:
ok, Thanks o/

PR #2319, opened 2015-09-28
Add dynamic translation to db write type in advanced settings

TheNain38:
Could someone check if this is right?


agilob:
works for me. Any objections?

zetok:
Well, that's the exact wording that is being used right now.

Edit: i.e. if PR makes those strings translatable, then why would there be any objections?

TheNain38:
@agilob Can be merged?

Issue #2308, opened 2015-09-27
Memory leak after closing main window


SmoothDude:
To reproduce this issue:
- Open your favourite task manager (to watch memory leak)
- In qTox's General settings enable these: Show system tray icon, Close to tray | Minimize to tray
- Now close main window
- Open it from system tray again
- Repeat last 2 steps above until you'll notice that memory leaks

Maybe related with: #2217

empirebuilder1:
Can confirm in Jenkins build 144 (Revision 14235c5a19fbd3c8484ec16068ea33a72d64ff0a). Closing and opening the program steadily increases memory usage by about 400-800k each time. However, if left idle and closed for about two minutes, memory usage drops to normal.

Issue #2217, opened 2015-09-13
Memory leak after closing new chat window


SmoothDude:
To reproduce issue:
- Open your favourite task manager (to watch memory growth)
- Launch qtox
- Open new chat window from any contact
- Close it
- Repeat 2 last steps until you notice that some memory still isn't freed after new chat window closed

zetok:
Could partially reproduce (after a lot of opening and closing detached window of a contacts).

Memory usage increased from ~77MB to ~81436 KB.

With that being said, when I've tried to continuously open & close same contact, memory stayed at ~81436 KB, (aside from +2MB increase during the time when detached window was open).

Possibly opening & closing other contacts would increase memory usage further.

dmshynk:
This issue is apparently caused by a memory leak in QClipboard. See this Valgrind output:
```
559 (48 direct, 511 indirect) bytes in 1 blocks are definitely lost in loss record 10,895 of 11,658
  in AddFriendForm::setIdFromClipboard() in /home/ds/Programming/qTox/src/widget/form/addfriendform.cpp:149
  1: operator new(unsigned long) in /var/tmp/portage/dev-util/valgrind-3.11.0/work/valgrind-3.11.0/coregrind/m_replacemalloc/vg_replace_malloc.c:333
  2: 0x1a5e57f9
  3: QClipboard::text(QClipboard::Mode) const in /usr/lib64/libQt5Gui.so.5.4.2
  4: AddFriendForm::setIdFromClipboard() in /home/ds/Programming/qTox/src/widget/form/addfriendform.cpp:149
  5: AddFriendForm::show(ContentLayout*) in /home/ds/Programming/qTox/src/widget/form/addfriendform.cpp:88
  6: Widget::onAddClicked() in /home/ds/Programming/qTox/src/widget/widget.cpp:640
  7: Widget::init() in /home/ds/Programming/qTox/src/widget/widget.cpp:339
  8: Nexus::showMainGUI() in /home/ds/Programming/qTox/src/nexus.cpp:177
  9: LoginScreen::onLogin() in /home/ds/Programming/qTox/src/widget/loginscreen.cpp:210
  10: QtPrivate::FunctorCall&lt;QtPrivate::IndexesList&lt;&gt;, QtPrivate::List&lt;&gt;, void, void (LoginScreen::*)()&gt;::call(void (LoginScreen::*)(), LoginScreen*, void**) in /usr/include/qt5/QtCore/qobjectdefs_impl.h:500
[...]

PR #2323, opened 2015-09-30
Fix memory leaks

dmshynk:
Got started on fixing various memory leaks possibly causing  #2308 and #2217. Several issues still exist, working on them.

### DIFF
--- a/src/widget/genericchatroomwidget.cpp
+++ b/src/widget/genericchatroomwidget.cpp
@@ -72,6 +72,7 @@ void GenericChatroomWidget::compactChange(bool _compact)
     // avatar
     if (isCompact())
     {
+        delete textLayout;  // Not needed
         setFixedHeight(25);
         avatar->setSize(QSize(20,20));
         mainLayout->addSpacing(18);
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -114,6 +114,8 @@ SystemTrayIcon::SystemTrayIcon()
                                  8, image->width(), image->height(),
                                  image->bytesPerLine(), callbackFreeImage, image);
 
+        delete image;   // Not needed anymore.
+
         statusNotifier = status_notifier_new_from_pixbuf("qtox",
                             STATUS_NOTIFIER_CATEGORY_APPLICATION_STATUS, pixbuf);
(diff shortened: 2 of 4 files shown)


src/widget/systemtrayicon.cpp
```diff
@@ -114,6 +114,8 @@ SystemTrayIcon::SystemTrayIcon()
                                  8, image->width(), image->height(),
                                  image->bytesPerLine(), callbackFreeImage, image);
 
+        delete image;   // Not needed anymore.
```
antis81: Why not change `QImage*` to `QImage`? Also solves the ugly `*image = …` in line 111 above.

agilob: @antis81 merge this when you think it's ready.

antis81:
@dmshynk Ready for merge? I have some suggestions here to make life easier, but those can be discussed in another PR, I think.

dmshynk:
All ready. I'm now struggling with the next leak, which seems to be out of my control (see my comment to #2217), so any suggestions are welcome.

antis81:
@dmshynk :+1:

src/widget/genericchatroomwidget.cpp
```diff
@@ -72,6 +72,7 @@ void GenericChatroomWidget::compactChange(bool _compact)
     // avatar
     if (isCompact())
     {
+        delete textLayout;  // Not needed
```
antis81, line 75: You can move recreation of `textLayout` (c'tor in line 63) here and replace the delete - this is safe. As I'm not sure what that nasty comment on the above delete (line 59) means I'd leave that untouched for now.

dmshynk:
@antis81, thanks for the valuable feedback, will do tomorrow morning. I am sort of anxious to mess with working code too much when it comes to memory management issues, but this indeed seems to be safe.

src/widget/genericchatroomwidget.cpp
antis81, line 75: Argh no, my proposed change will lead to crash (delete dangling pointer on subsequent calls). You would also need to set `textLayout = nullptr;` after line 59. Anyway, I'll happily help to clean this up in a new PR - merging this one later today.

PR #2330, opened 2015-10-02
Fix ACTION messages, status, username HTML parsing

### LINKED ISSUE #1999: When outgoing ACTION messages, nick in the HTML tags will be parse.

TheNain38:
Normally fixes #1999
Could someone check before merging?

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -58,7 +58,7 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     {
     case ACTION:
         senderText = "*";
-        text = wrapDiv(QString("%1 %2").arg(sender, text), "action");
+        text = wrapDiv(QString("%1 %2").arg(sender.toHtmlEscaped(), text), "action");
         msg->setAsAction();
         break;
     case ALERT:
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -858,6 +858,9 @@ QSplitter:handle{
                  <property name="cursor">
                   <cursorShape>PointingHandCursor</cursorShape>
                  </property>
+                 <property name="textFormat">
+                  <enum>Qt::PlainText</enum>
+                 </property>
                 </widget>
                </item>
                <item>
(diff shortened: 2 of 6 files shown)


zetok:
> Could someone check before merging?

@TheNain38 um, what prevents you from checking it?

TheNain38:
@zetok I can't setup a build environnement, I'm trying as hard as possible to setup one

zetok:
@TheNain38 yes, but you still didn't say what is causing you problems

antis81:
Looks good to me :+1:

TheNain38:
@zetok see irungentoo/toxcore#1443

TheNain38:
I've corrected another HTML parsing bug with the typing notification, good to merge.

agilob:
@antis81 pls merge if you tested it and it's ok, I don't want to be the only person who merges :<

TheNain38:
@agilob I've tested it in a VM and it works

agilob:
IT can't be merged because it modifies incoming message. The code is completely stripping HTML from message instead treating message as plaintext. When user with nick `<h1>kotez</h1>` is sending me an action message I should see it as `* <h1>kotez</h1> test` not `* kotez test`. See two examples how the same nickname is displayed in two different ways:
weblate6

weblate5

Another thing, your code introduces regression, nick is displayed in different way in groupchat than in chat. **AFAIR** those changes should be done in chatline.cpp or chatmessage.cpp.

TheNain38:
@agilob Look at your first screenshot, in the groupchat his name hasn't been updated, also I've noticed another place where parsing applies but shouldn't, I will update

agilob:
I was invited to the groupchat by zetok when kotez already had nick with html tags, so there was nothing to be updated.

TheNain38:
@agilob In your screenshot kotez name in groupchat isn't `<h1>kotez</h1>`, it is `kotez`
capture

zetok:
@TheNain38 yes, name was `<h1>kotez</h1>` - this looks like a regression caused by name being assigned as alias.

agilob:
Works for me now. Thanks!

PR #2331, opened 2015-10-03
Dynamic translation of call, video, mute, unmute

### LINKED ISSUE #2259: Tooltips on mute microphone etc... aren't translated on language change when in a call but are when muted/unmuted
How to reproduce:
- Start a call
- Change language
- look at tooltip

TheNain38:
fixes #2259
Could someone check if this is right?

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -1085,5 +1085,17 @@ void ChatForm::hideNetcam()
 
 void ChatForm::retranslateUi()
 {
+    QString volObjectName = volButton->objectName();
+    QString micObjectName = micButton->objectName();
     loadHistoryAction->setText(tr("Load chat history..."));
+
+    if (volObjectName == QStringLiteral("green"))
+        volButton->setToolTip(tr("Mute call"));
+    else if (volObjectName == QStringLiteral("red"))
+        volButton->setToolTip(tr("Unmute call"));
+
+    if (micObjectName == QStringLiteral("green"))
+        micButton->setToolTip(tr("Mute microphone"));
+    else if (micObjectName == QStringLiteral("red"))
+        micButton->setToolTip(tr("Unmute microphone"));
 }
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -512,12 +512,27 @@ bool GenericChatForm::eventFilter(QObject* object, QEvent* event)
 
 void GenericChatForm::retranslateUi()
 {
+    QString callObjectName = callButton->objectName();
+    QString videoObjectName = videoButton->objectName();
+
+    if (callObjectName == QStringLiteral("green"))
+        callButton->setToolTip(tr("Start audio call"));
+    else if (callObjectName == QStringLiteral("yellow"))
+        callButton->setToolTip(tr("Accept audio call"));
+    else if (callObjectName == QStringLiteral("red"))
+        callButton->setToolTip(tr("End audio call"));
+
+    if (videoObjectName == QStringLiteral("green"))
+        videoButton->setToolTip(tr("Start video call"));
+    else if (videoObjectName == QStringLiteral("yellow"))
+        videoButton->setToolTip(tr("Accept video call"));
+    else if (videoObjectName == QStringLiteral("red"))
+        videoButton->setToolTip(tr("End video call"));
+
     sendButton->setToolTip(tr("Send message"));
     emoteButton->setToolTip(tr("Smileys"));
     fileButton->setToolTip(tr("Send file(s)"));
     screenshotButton->setToolTip(tr("Send a screenshot"));
-    callButton->setToolTip(tr("Start an audio call"));
-    videoButton->setToolTip(tr("Start a video call"));
     saveChatAction->setText(tr("Save chat log"));
     clearAction->setText(tr("Clear displayed messages"));
 }


src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,40 @@ void ChatForm::hideNetcam()
 
 void ChatForm::retranslateUi()
 {
+    QString callObjectName = callButton->objectName();
+    QString videoObjectName = videoButton->objectName();
+    QString volObjectName = volButton->objectName();
+    QString micObjectName = micButton->objectName();
     loadHistoryAction->setText(tr("Load chat history..."));
+
+    if (callObjectName == "green")
```
antis81: Better use `QStringLiteral("green")` here (and for the following conditions).

src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,40 @@ void ChatForm::hideNetcam()
...
+        callButton->setToolTip(tr("Start audio call"));
+
+    if (callObjectName == "yellow")
```
antis81: `else if` :)

src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,40 @@ void ChatForm::hideNetcam()
...
+        callButton->setToolTip(tr("Accept audio call"));
+
+    if (callObjectName == "red")
+        callButton->setToolTip(tr("End audio call"));
+
+
+    if (videoObjectName == "green")
+        videoButton->setToolTip(tr("Start video call"));
+
+    if (videoObjectName == "yellow")
```
antis81: `else if`…

agilob: Yup, elseif of switch would work better here.

src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,47 @@ void ChatForm::hideNetcam()
 
 void ChatForm::retranslateUi()
 {
+    QString volObjectName = volButton->objectName();
+    QString micObjectName = micButton->objectName();
     loadHistoryAction->setText(tr("Load chat history..."));
+
+    switch(callButton->objectName())
+    {
+    case QStringLiteral("green"):
```
TheNain38: I can't use QStringLiteral in `case QStringLiteral("green"):` ?

antis81: That comes natural to switch-case statement. It doesn't work with `"literals"` (or other dynamic types) -> must be enumeration type (int, enum, …).

src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,47 @@ void ChatForm::hideNetcam()
...
+        videoButton->setToolTip(tr("Start video call"));
+        break;
+
+    case QStringLiteral("yellow"):
+        videoButton->setToolTip(tr("Accept video call"));
+        break;
+
+    case QStringLiteral("red"):
+        videoButton->setToolTip(tr("End video call"));
```
antis81: Just a general tip, when using switch-case statements:
Don't forget to `break;` in **every** case. And usually you want to define a `default: break;` as well.

src/widget/form/chatform.cpp
TheNain38: @antis81 How did I do that correctly, then? with a switch?

src/widget/form/chatform.cpp
TheNain38: Yes I forgot the `default:` one

TheNain38:
good to merge

src/widget/form/chatform.cpp
antis81: Short answer: You can't -> use `if (); else if(); else;`.

Long answer:
Switching the `objectName` to indicate the status is **bullshit**. We need to query the actual status of the underlying object. That means:

``` c++
// this is some pseudo code, just to give an idea
// tox API should send some kind of event (callback), if an audio-/video call starts/mutes/ends/…
// We react to the status change via an event handler and set the tooltip text according to the status:
void GenericChatForm::callStatusChanged(Call::Type t, Call::Status s)
{
    QString description;

    // assuming c++11 "enum class Type {Audio, Video, …}" here
    if (t == Call::Type::Audio) {
        switch (s) {
        default: break;
        case Call::Status::Active: description = tr("Hangup and end the current call."); break;
        case Call::Status::Ended: description = tr("Start a new  call."); break;
        case Call::Status::Pending: description = tr("Accept the call.!"); break;
        }
        callButton->setToolTip(description);
    }
    else if (t == Call::Type::Video) {
        switch (s) {
        default: break;
        case Call::Status::Active: description = tr("End the current video call."); break;
        case Call::Status::Ended: description = tr("Start a new video call."); break;
        case Call::Status::Pending: description = tr("Accept the video call.!"); break;
        }
        videoButton->setToolTip(description);
    } else {
            // undefined call type -> must not happen
            Q_ASSERT(false);
    }

    // undefined call status -> must not happen
    Q_ASSERT(!description.isEmpty());
}


```

src/widget/form/chatform.cpp
```diff
@@ -1085,5 +1085,33 @@ void ChatForm::hideNetcam()
...
+    else if (callObjectName == QStringLiteral("red"))
+        callButton->setToolTip(tr("End audio call"));
+
+    if (videoObjectName == QStringLiteral("green"))
+        videoButton->setToolTip(tr("Start video call"));
+    else if (videoObjectName == QStringLiteral("yellow"))
+        videoButton->setToolTip(tr("Accept video call"));
+    else if (videoObjectName == QStringLiteral("red"))
+        videoButton->setToolTip(tr("End video call"));
+
```
antis81: Don't you think, that moving the two if-blocks for `callObjectName` and  `videoObjectName` to the base class is the better approach?

TheNain38: @antis81 Moving them under line 1116?

antis81: No, I mean moving them to "genericchatform.cpp from line 520 downwards" :)

TheNain38: @antis81 Ok, when done it can be merged?

TheNain38:
@antis81 done, ready for merge

Issue #2275, opened 2015-09-21
Unknown Error on Profile Creation (OS can't handle "special" symbols in filenames)


Snowyiu:
When creating a new profile and using the characters '|', '<' or '>' in the Username field, there'll be an unknown error when the 'Create Profile' button is clicked.

zetok:
Well, the problem is with your OS - i.e. it can't even handle normal characters like `| < >` in filenames.
(profile name that you enter during creation of profile is used as a filename)

I guess that solution to it would be to tell the user that they can't use certain characters on error.

ghost:
Windows' unbeaten beauty and clarity
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx

Symbols that should be forbidden for names under Windows:
`<` `>` `:` `"` `/` `\` `|` `?` `*`

And there are more recommendations in the link.

Also important to notice is that Windows and Mac are case insensitive. In Linux at least I can create different profiles changing only the case.

@zetok on a side note: `/` is not forbidden in Linux, it can create a new user, but the profile is not saved to the disc. The same probably happens in Mac. Do you want a new issue?

zetok:
@kehugter actually, Linux can't also handle `\0` in filenames `;)`

I think that this issue could be used to track it for all systems - I'm not sure though to what issue title could be changes to reflect that.. ideas?

agilob:
We had the same issue in saving files sent from linux 20:12:40.jpg is not valid on Windows but is on Linux. One solution could be parse invalid filename to HTML... not pretty but will work or use boost library to do it for us... but that would increase build size and code wouldn't be nice. Remember we need to be able to save valid file and read it later, so we need both ways solution.

@antis81 ?

ghost:
@zetok I tried it too and worked (backslash + zero)...

![qtox profile](https://cloud.githubusercontent.com/assets/6382342/10023329/d5cd55ea-6151-11e5-9c69-d3a846bc31e5.png)

![ls](https://cloud.githubusercontent.com/assets/6382342/10023331/db097750-6151-11e5-992d-bbca253c1ceb.png)

agilob:
> @zetok I tried it too and worked (backslash + zero)...

Try the same on windows.

zetok:
@kehugter not literal backslash + zero, but null char that it represents.

ghost:
@zetok I see, I thought the string `\0` was meant to be interpreted as null

[...]

PR #2340, opened 2015-10-04
Make screenshot names use ~ISO 8601 (timestamp) instead of random string

zetok:
Now screenshots will be saved as: `qTox-Screenshot_YYYY-MM-DD hh-mm-ss.sss`

Since miliseconds are used, it shouldn't cause filename conflicts when
savign screenshots.

Change should make it easier for users to navigate through screenshots
chronologically.

E.g. names: 

```
qTox_Screenshot_2015-10-04 12-58-19.651.png
qTox_Screenshot_2015-10-04 13-01-32.421.png
```

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -904,16 +904,24 @@ void ChatForm::doScreenshot()
 }
 
 void ChatForm::onScreenshotTaken(const QPixmap &pixmap) {
-    QTemporaryFile file(Settings::getInstance().getSettingsDirPath()+"screenshots"+QDir::separator()+"qTox-Screenshot-XXXXXXXX.png");
-    if (!file.open())
+    // use ~ISO 8601 for screenshot timestamp, considering FS limitations
+    // https://en.wikipedia.org/wiki/ISO_8601
+    // Windows has to be supported, thus filename can't have `:` in it :/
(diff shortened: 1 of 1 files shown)


agilob:
Is `_` valid char on windows in file names?

zetok:
to quote @krobelus from #2275:

# 

Windows' unbeaten beauty and clarity
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx

Symbols that should be forbidden for names under Windows:
`<` `>` `:` `"` `/` `\` `|` `?` `*`

# 

I've added that to wiki, since it'll be helpful lots of times in the future: https://github.com/tux3/qTox/wiki/Coding#limitations

agilob:
You should use [QDateTime::toString(FORMAT)](http://doc.qt.io/qt-5/qdatetime.html#toString) to drastically simplify that change, even to one line.
Personally, I prefer when 5 minutes is printed are `05` not, `5`, but that's just my opinion.

zetok:
oh, looks like I've missed that; will do

zetok:
Ok, done.

agilob:
Is that milliseconds part necessary?

zetok:
> Is that milliseconds part necessary?

Yep, to avoid filename conflicts when user manages to take >1 screenshots per second.

PR #2341, opened 2015-10-04
Added a context menu to edit/remove the avatar

### LINKED ISSUE #2339: Implement functionality to remove avatar
At the moment there is no functionality to remove the avatar.

Just deleted my avatar for testing purposes directly in my .config/tox/avatars.
Worked, _but_ the friends can still see my old avatar. Only if they restart qTox it is updated. How should qtox handle it? Update immediately or after restart is enough?

alexbakker:
Should fix #2339.
I'm still fairly new to C++ and the qTox code base, let me know if I did anything stupid.

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -341,7 +341,8 @@ void Core::start()
     }
     else
     {
-        qDebug() << "Self avatar not found";
+        qDebug() << "Self avatar not found, will broadcast empty avatar to friends";
+        setAvatar({});
     }
 
     ready = true;
@@ -792,10 +793,17 @@ void Core::setUsername(const QString& username)
 
 void Core::setAvatar(const QByteArray& data)
 {
-    QPixmap pic;
-    pic.loadFromData(data);
-    Settings::getInstance().saveAvatar(pic, getSelfId().toString());
-    emit selfAvatarChanged(pic);
+    if (!data.isEmpty())
+    {
+        QPixmap pic;
+        pic.loadFromData(data);
+        Settings::getInstance().saveAvatar(pic, getSelfId().toString());
+        emit selfAvatarChanged(pic);
+    }
+    else
+    {
+        emit selfAvatarChanged(QPixmap(":/img/contact_dark.svg"));
+    }
 
     AvatarBroadcaster::setAvatar(data);
     AvatarBroadcaster::enableAutoBroadcast();
--- a/src/core/corefile.cpp
+++ b/src/core/corefile.cpp
@@ -56,12 +56,20 @@ void CoreFile::sendAvatarFile(Core* core, uint32_t friendId, const QByteArray& d
 {
     QMutexLocker mlocker(&fileSendMutex);
 
+    if (data.isEmpty())
+    {
+        tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, 0,
+                                         nullptr, nullptr, 0, nullptr);
+        return;
+    }
+
     static_assert(TOX_HASH_LENGTH <= TOX_FILE_ID_LENGTH, "TOX_HASH_LENGTH > TOX_FILE_ID_LENGTH!");
     uint8_t avatarHash[TOX_HASH_LENGTH];
     tox_hash(avatarHash, (uint8_t*)data.data(), data.size());
     uint64_t filesize = data.size();
     uint32_t fileNum = tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, filesize,
-                                     avatarHash, avatarHash, TOX_HASH_LENGTH, nullptr);
+                                         avatarHash, avatarHash, TOX_HASH_LENGTH, nullptr);
+
     if (fileNum == std::numeric_limits<uint32_t>::max())
     {
         qWarning() << "sendAvatarFile: Can't create the Tox file sender";
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -45,6 +45,8 @@
 #include <QMessageBox>
 #include <QComboBox>
 #include <QWindow>
+#include <QMenu>
+#include <QMouseEvent>
 
 ProfileForm::ProfileForm(QWidget *parent) :
     QWidget{parent}, qr{nullptr}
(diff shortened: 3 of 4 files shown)


src/core/core.cpp
```diff
@@ -792,10 +793,19 @@ void Core::setUsername(const QString& username)
...
+    {
+        QPixmap pic;
+        pic.loadFromData(data);
+        Settings::getInstance().saveAvatar(pic, getSelfId().toString());
+        emit selfAvatarChanged(pic);
+    }
+    else
+    {
+        QString selfPubKey = getSelfId().publicKey;
+        QFile::remove(Settings::getInstance().getSettingsDirPath()+"avatars/"+selfPubKey.left(64)+".png");
```
zetok: You probably should check whether `QFile::remove` returns `true`, and if it isn't, warn user that it was not possible to remove avatar.

Also, avatar doesn't change if it wasn't possible to remove it ↓

alexbakker:
Should show an error now if the avatar could not be removed.

zetok:
PR introduces regression - it's no longer possible to change avatar by l-clicking on it.

alexbakker:
That's what the 'Edit' item in the context menu is for. Isn't one way to change the avatar enough? Should I add a tooltip telling the user to right-click to edit/remove the avatar?

zetok:
> That's what the 'Edit' item in the context menu is for. Isn't one way to change the avatar enough? Should I add a tooltip telling the user to right-click to edit/remove the avatar?

No, since it's not user friendly. User should be able to change avatar by l-clicking on it.

Now, removing avatar by r-clicking isn't too user-friendly either, but it should do as a temporary work-around, until there will be a better way of doing it.

alexbakker:
I guess you're right. I'll remove the 'Edit' item and re-add the left-click.

zetok:
PR also makes `Could not remove avatar` error to be spammed every time on qTox start if avatar is not set.

alexbakker:
Both the regression and the 'Could not remove avatar' issue should be fixed now.

src/core/core.cpp
```diff
@@ -342,6 +342,7 @@ void Core::start()
     else
     {
         qDebug() << "Self avatar not found";
```
agilob: could you change this qdebug to "setting empty avatar"? so we don't print misleading debug messages.

src/core/core.cpp
```diff
@@ -792,10 +793,17 @@ void Core::setUsername(const QString& username)
...
+    if (data != nullptr)
+    {
+        QPixmap pic;
+        pic.loadFromData(data);
+        Settings::getInstance().saveAvatar(pic, getSelfId().toString());
+        emit selfAvatarChanged(pic);
+    }
+    else
+    {
+        emit selfAvatarChanged(QPixmap(":/img/contact_dark.svg"));
```
agilob, line 805: I'm not sure about this part. This is emitting qtox default avatar as users empty avatar, that means empty avatar for qtox will be different than empty avatar for µtox/another client. SO empty avatar might in some situation look like custom one.

src/core/core.cpp
alexbakker: Will fix.

src/core/core.cpp
alexbakker, line 805: I don't understand what you mean. The contact_dark.svg is never sent to friends, it only appears locally.

agilob, line 805: Ah sorry! If it's not sent to friends than it's ok and it's my bad!

alexbakker:
There, done.

src/core/core.cpp
```diff
@@ -792,10 +793,17 @@ void Core::setUsername(const QString& username)
 
 void Core::setAvatar(const QByteArray& data)
 {
-    QPixmap pic;
-    pic.loadFromData(data);
-    Settings::getInstance().saveAvatar(pic, getSelfId().toString());
-    emit selfAvatarChanged(pic);
+    if (data != nullptr)
```
antis81: Use `data.isEmpty()` here instead of implicit cast to pointer.

src/core/corefile.cpp
```diff
@@ -55,13 +55,24 @@ unsigned CoreFile::corefileIterationInterval()
...
-    static_assert(TOX_HASH_LENGTH <= TOX_FILE_ID_LENGTH, "TOX_HASH_LENGTH > TOX_FILE_ID_LENGTH!");
+    uint32_t fileNum;
     uint8_t avatarHash[TOX_HASH_LENGTH];
-    tox_hash(avatarHash, (uint8_t*)data.data(), data.size());
-    uint64_t filesize = data.size();
-    uint32_t fileNum = tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, filesize,
-                                     avatarHash, avatarHash, TOX_HASH_LENGTH, nullptr);
+    uint64_t filesize;
+
+    if (data != nullptr)
```
antis81: Same here - use data.isEmpty()

src/core/corefile.cpp
```diff
@@ -69,14 +80,17 @@ void CoreFile::sendAvatarFile(Core* core, uint32_t friendId, const QByteArray& d
...
 
-    ToxFile file{fileNum, friendId, "", "", ToxFile::SENDING};
-    file.filesize = filesize;
-    file.fileName = QByteArray((char*)avatarHash, TOX_HASH_LENGTH);
-    file.fileKind = TOX_FILE_KIND_AVATAR;
-    file.avatarData = data;
-    file.resumeFileId.resize(TOX_FILE_ID_LENGTH);
-    tox_file_get_file_id(core->tox, friendId, fileNum, (uint8_t*)file.resumeFileId.data(), nullptr);
-    addFile(friendId, fileNum, file);
+    if (data != nullptr)
```
antis81: Another one…

alexbakker:
Aaaaand fixed

src/core/core.cpp
```diff
@@ -341,7 +341,8 @@ void Core::start()
     }
     else
     {
-        qDebug() << "Self avatar not found";
+        qDebug() << "Self avatar not found, will broadcast empty avatar to friends";
+        setAvatar(nullptr);
```
tux3: Nope. You can't just pass a nullptr to a function that expects a reference, that's all kinds of **UNDEFINED BEHAVIOR**.

You probably want to pass an empty QByteArray instead:

```
setAvatar(QByteArray()); // Verbose way
setAvatar({}); // Unreadable way
```

subliun: # _UNDEFINED BEHAVIOUR_

src/core/corefile.cpp
```diff
@@ -55,13 +55,24 @@ unsigned CoreFile::corefileIterationInterval()
...
+        tox_hash(avatarHash, (uint8_t*)data.data(), data.size());
+        filesize = data.size();
+        fileNum = tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, filesize,
+                                         avatarHash, avatarHash, TOX_HASH_LENGTH, nullptr);
+    }
+    else
+    {
+        fileNum = tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, 0,
+                                         nullptr, nullptr, 0, nullptr);
+    }
```
tux3, line 64: Nitpick, but if you turn just that else into:

```
if (data.isEmpty())
{
    fileNum = tox_file_send(core->tox, friendId, TOX_FILE_KIND_AVATAR, 0,
                                         nullptr, nullptr, 0, nullptr);
    return;
}
```

You can turn back the two `if (!data.isEmpty())` blocks at lines 62 and 83 into unconditional code. It skips the error handling code, but that shouldn't be much of a problem if we're sending empty avatars (?).

src/core/core.cpp
agilob: Thanks @subliun now I see better!

alexbakker:
Fixed. I hate it when you're right.

agilob:
can be merged?

alexbakker:
I think so. I fixed the things tux3 pointed out.

agilob:
ok, works for me.

PR #2343, opened 2015-10-05
Fix memory leaks

### LINKED ISSUE #2217: Memory leak after closing new chat window

dmshynk:
Some more progress on memory leaks, one fixed use of an uninitialized variable. No longer able to reproduce #2308 with these changes. Still unable to fix #2217, any insights are welcome.

### DIFF
--- a/src/platform/statusnotifier/statusnotifier.c
+++ b/src/platform/statusnotifier/statusnotifier.c
@@ -1608,7 +1608,7 @@ get_prop (GDBusConnection        *conn,
     else if (!g_strcmp0 (property, "ToolTip"))
     {
         GVariant *variant;
-        GVariantBuilder *builder;
+        GVariant *pixmap;
 
         if (!priv->icon[STATUS_NOTIFIER_TOOLTIP_ICON].has_pixbuf)
         {
@@ -1621,13 +1621,13 @@ get_prop (GDBusConnection        *conn,
             return variant;
         }
 
-        builder = get_icon_pixmap (sn, STATUS_NOTIFIER_TOOLTIP_ICON);
+        pixmap = get_icon_pixmap (sn, STATUS_NOTIFIER_TOOLTIP_ICON);
         variant = g_variant_new ("(sa(iiay)ss)",
                 "",
-                builder,
+                pixmap,
                 (priv->tooltip_title) ? priv->tooltip_title : "",
                 (priv->tooltip_body) ? priv->tooltip_body : "");
-        g_variant_builder_unref (builder);
+        g_variant_unref (pixmap);
 
         return variant;
     }
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -225,7 +225,6 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
     if (av_opt_set_dict2(s, &tmp, AV_OPT_SEARCH_CHILDREN) < 0)
     {
         av_dict_free(&tmp);
-        avformat_free_context(s);
     }
     avdevice_list_devices(s, &devlist);
     if (!devlist)
@@ -243,6 +242,7 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
         devices[i].second = dev->device_description;
     }
     avdevice_free_list_devices(&devlist);
+    avformat_free_context(s);
     return devices;
 }
 
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -225,8 +225,11 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
     if (av_opt_set_dict2(s, &tmp, AV_OPT_SEARCH_CHILDREN) < 0)
     {
         av_dict_free(&tmp);
+        avformat_free_context(s);
     }
     avdevice_list_devices(s, &devlist);
+    av_dict_free(&tmp);
+    avformat_free_context(s);
     if (!devlist)
     {
         qWarning() << "avdevice_list_devices failed";
@@ -242,7 +245,6 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
         devices[i].second = dev->device_description;
     }
     avdevice_free_list_devices(&devlist);
-    avformat_free_context(s);
     return devices;
 }
 
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -226,6 +226,7 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
     {
         av_dict_free(&tmp);
         avformat_free_context(s);
+        return devices;
     }
     avdevice_list_devices(s, &devlist);
     av_dict_free(&tmp);
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -72,11 +72,12 @@ SystemTrayIcon::SystemTrayIcon()
         };
         QImage* image = new QImage(":/img/icon.png");
         if (image->format() != QImage::Format_RGBA8888_Premultiplied)
-        *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+            *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
         GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image->bits(), GDK_COLORSPACE_RGB, image->hasAlphaChannel(),
                                  8, image->width(), image->height(),
                                  image->bytesPerLine(), callbackFreeImage, image);
         gtkIcon = gtk_status_icon_new_from_pixbuf(pixbuf);
+        g_object_unref(pixbuf);
         gtkMenu = gtk_menu_new();
 
         void (*callbackTrigger)(GtkStatusIcon*, gpointer) =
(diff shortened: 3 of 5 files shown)


src/video/cameradevice.cpp
```diff
@@ -225,7 +225,6 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
     if (av_opt_set_dict2(s, &tmp, AV_OPT_SEARCH_CHILDREN) < 0)
     {
         av_dict_free(&tmp);
-        avformat_free_context(s);
```
tux3, line 228: You're introducing a new leak here if we return early at line 233
Instead, if av_opt_set_dict2 fails we should keep the avformat_free_context(s); at line 228 and return early here, and if it doesn't fail we should add:

```
av_dict_free(&tmp);
avformat_free_context(s);
```

right after the avdevice_list_devices (no need to wait until the end of the function).
That part of the code is pretty much a copy of `list_devices_for_context`, which isn't in the public API for some reason.

dmshynk:
@tux3, done and done. Thanks for pointing this out, guess I should really spend more time reading the docs.

dmshynk:
I don't even know what to say for myself:clap: It was literally the second thing you mentioned in the original message. Fixed

src/platform/statusnotifier/statusnotifier.c
```diff
@@ -1608,7 +1608,7 @@ get_prop (GDBusConnection        *conn,
     else if (!g_strcmp0 (property, "ToolTip"))
     {
         GVariant *variant;
-        GVariantBuilder *builder;
+        GVariant *pixmap;
```
antis81: Yeah, I've noticed the warning here… :+1:

src/widget/systemtrayicon.cpp
```diff
@@ -72,11 +72,15 @@ SystemTrayIcon::SystemTrayIcon()
         };
         QImage* image = new QImage(":/img/icon.png");
```
antis81: ~~Same procedure as last year ;) - Use stacked variable simplifies life.~~

Why are we're misusing QImage here to create the GdkPixbuf object? Isn't it able to load the image data from file directly instead?

Other than that, IMO the `StatusNotifier` type should change to either one of `QIcon`, `QImage` or `QPixmap`.

src/widget/systemtrayicon.cpp
```diff
@@ -107,16 +111,19 @@ SystemTrayIcon::SystemTrayIcon()
...
-        QImage image(":/img/icon.png");
-        if (image.format() != QImage::Format_RGBA8888_Premultiplied)
-        image = image.convertToFormat(QImage::Format_RGBA8888_Premultiplied);
-        GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image.bits(), GDK_COLORSPACE_RGB, image.hasAlphaChannel(),
-                                 8, image.width(), image.height(),
-                                 image.bytesPerLine(), callbackFreeImage, &image);
+        QImage* image = new QImage(":/img/icon.png");
+        if (image->format() != QImage::Format_RGBA8888_Premultiplied)
+            *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+        GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image->bits(), GDK_COLORSPACE_RGB, image->hasAlphaChannel(),
```
antis81: Anything against using `QImage` (or QPixmap?) all the way?

src/widget/systemtrayicon.cpp
```diff
@@ -233,13 +244,17 @@ void SystemTrayIcon::setContextMenu(QMenu* menu)
                 };
                 QImage* image = new QImage(a->icon().pixmap(64, 64).toImage());
                 if (image->format() != QImage::Format_RGBA8888_Premultiplied)
-                *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+                    *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
                 GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image->bits(), GDK_COLORSPACE_RGB, image->hasAlphaChannel(),
```
antis81: **important** Will need to `memcpy` the `image->bits()` here, as these are shared implicitly and become invalid, when QImage is deleted.

After doing so, QImage can be created on stack again.

src/widget/systemtrayicon.cpp
agilob: afair QPixmap had problem with transparent icons in a version of Qt.

agilob:
@dmshynk could you fix the issue pointed by @antis81  so I can merge it?

src/widget/systemtrayicon.cpp
antis81: Hm… this cannot be correct in this case. We use QImage here to load and initialize image data, but then transport the data to a gtk object. To me this looks like a pretty dirty hack -> and it crashes, because data isn't copied, but implicitly shared here (see docs for QImage::bits()). This requires a `memcpy(QImage::bits())`. Otherwise image data become invalid for `GdkPixbuf` when QImage is deleted.

agilob: That was the reason for using qimage,  don't know anything about gdkpixbuf.

src/widget/systemtrayicon.cpp
dmshynk: But the ```image``` won't get deleted until after ```pixbuf```'s reference count has dropped to zero, as per [docs](https://developer.gnome.org/gdk-pixbuf/stable/gdk-pixbuf-Image-Data-in-Memory.html#gdk-pixbuf-new-from-data). Note that the deletion only happens in the callback passed to ```gdk_pixbuf_new_from_data()```.

src/widget/systemtrayicon.cpp
dmshynk: Plus, using ```QPixmap``` is impractical in this case, as it doesn't provide an interface to extract raw data, like ```QImage::bits()```. Now, notice how the ```else```  block uses a ```QSystemTrayIcon```, the Qt-native solution, which doesn't really work — see [this bug](https://bugreports.qt.io/browse/QTBUG-31762), never fully resolved, AFAIK. However, for Plasma 5,we could go with a native solution starting from Qt5.4 (see [here](http://blog.martin-graesslin.com/blog/2014/06/where-are-my-systray-icons/)). I'll look into it.

src/widget/systemtrayicon.cpp
antis81: Phew, this is is total brain f*_k. You're right, the callback reinterpret_casts gpointer to QImage_ and free's it. But… we don't actually need the callback here don't you think? As this code is addressed to GTK+ backend, we should rather use completely their classes and replace the `QImage` code here IMO.

src/widget/systemtrayicon.cpp
```diff
@@ -366,11 +381,15 @@ void SystemTrayIcon::setIcon(QIcon &icon)
         };
         QImage* image = new QImage(icon.pixmap(64, 64).toImage());
         if (image->format() != QImage::Format_RGBA8888_Premultiplied)
-        *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+            *image = image->convertToFormat(QImage::Format_RGBA8888_Premultiplied);
         GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image->bits(), GDK_COLORSPACE_RGB, image->hasAlphaChannel(),
                                  8, image->width(), image->height(),
                                  image->bytesPerLine(), callbackFreeImage, image);
+        if(!pixbuf) // If pixbuf initialization fails, it won't trigger the callback.
```
antis81: This looks bullshit to me. If pixbuf initialization fails, the callback should be still called. Are you sure about this?

dmshynk: It should, but valgrind analysis suggests otherwise. I suppose the *correct* approach would be to switch to gtk-specific ```GBytes``` in place of ```QImage```, as you noted above, then we could use [```gdk_pixbuf_new_from_bytes```](https://developer.gnome.org/gdk-pixbuf/stable/gdk-pixbuf-Image-Data-in-Memory.html#gdk-pixbuf-new-from-bytes) and forgo the hassle of freeing in a callback and that all. I'll look into that as well.

antis81: That sounds better. Thanks a lot. :+1:

dmshynk:
Done what could be done, now working on getting rid of ```SystemTrayIcon``` and GTK+ bindings completely, replacing them with a plain ```QSystemTrayIcon```. Seems to work on at least Plasma 5.4 and Unity 7.3, more testing is needed. That will be a separate pull request, though.

zetok:
@dmshynk sounds great `:)`

Will removal GTK+ stuff be based on #2132?

dmshynk:
@zetok, definitely, as soon as it gets merged. By the way, Travis is apparently using an older version of ```libgdk-pixbuf``` (```gdk_pixbuf_new_from_bytes()``` was introduced in v.2.32, which Ubuntu doesn't get until 15.10). What do?

agilob:
Ubuntu is like Windows of Linuxes.

antis81:
@dmshynk A big thank you. Can you fix build please before merge (missing include?):

```
src/widget/systemtrayicon.cpp: In member function ‘void SystemTrayIcon::setContextMenu(QMenu*)’:

src/widget/systemtrayicon.cpp:176:96: error: ‘gdk_pixbuf_new_from_bytes’ was not declared in this scope

                                          8, image.width(), image.height(), image.bytesPerLine());

                                                                                                ^

src/widget/systemtrayicon.cpp:220:95: error: ‘gdk_pixbuf_new_from_bytes’ was not declared in this scope

                                         8, image.width(), image.height(), image.bytesPerLine());

                                                                                               ^

src/widget/systemtrayicon.cpp: In member function ‘void SystemTrayIcon::setIcon(QIcon&)’:

src/widget/systemtrayicon.cpp:352:87: error: ‘gdk_pixbuf_new_from_bytes’ was not declared in this scope

                                 8, image.width(), image.height(), image.bytesPerLine());

                                                                                       ^

src/widget/systemtrayicon.cpp:368:87: error: ‘gdk_pixbuf_new_from_bytes’ was not declared in this scope

                                 8, image.width(), image.height(), image.bytesPerLine());

                                                                                       ^
```

agilob:
What include is missing?

dmshynk:
@agilob, nothing is missing, all compiles just fine on my machine. The issue is with Travis CI's environment. Since it's Ubuntu 12.04 LTS ([prooflink](http://docs.travis-ci.com/user/ci-environment/)), ```libgtk-pixbuf``` is at v.2.26, which means that ```gdk_pixbuf_new_from_bytes``` is simply not there to include, as it is not introduced until v.2.32. There is no good way around it, aside from [pinning](https://help.ubuntu.com/community/PinningHowto), which is totally not worth it since we are planning on getting rid of GTK+ deps anyway. Meanwhile, I guess we'll have to resort to using a crude ```guchar*``` and the godawful deletion in callback to make it build.

agilob:
I will merge it today unless there are objections.

tux3:
LGTM

tux3:
Actually let's merge that right now, and then #2132

agilob:
Ok, so I will fix #2132 to remove unused code, it contains a fix that was already merged. It will be fixed in different PR. I'll to by best to do to it today!

tux3:
Don't worry, I'm rebasing and fixing up #2132 right now, then I can panic about new AV not being ready...

agilob:
ok, thanks

PR #2344, opened 2015-10-05
New AV qTox

tux3:
This implements single and group calls using the new AV API.

I can't resolve the merge conflicts at the moment (because my Debian unstable broke the merge resolution tool), but it's ready for testing.

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -545,4 +545,5 @@ HEADERS += \
     src/widget/contentlayout.h \
     src/widget/contentdialog.h \
     src/widget/tool/activatedialog.h \
-    src/widget/tool/removefrienddialog.h
+    src/widget/tool/removefrienddialog.h \
+    src/core/indexedlist.h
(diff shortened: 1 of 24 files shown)


zetok:
Generally, it appears to work.

The only real problem that I've managed to spot so far is that after I ended video call, window with video on the side that ended call was not closed.

With that being said, I didn't really test audio - i.e. I did test audio call, but with no actual sound being sent.

# 

On a side note - I hope that thing in toxav that causes enormous RAM usage will be fixed - though I wonder.

From what I understood, RAM usage is so high due to something in libvpx, while when I was testing, even during audio-only calls qTox used ~900MB of RAM.

@mannol ↑ care to comment?

zetok:
~~Crash when other side "rejected" call, and then side calling tried to turn call off on~~ Side that rejected call crashed 38b5bf889c9f2424a17109ae7c25c73fb732bd88: https://gist.github.com/zetok/883400a46536bc94c475

zetok:
When all peers exit an audio groupchat in which there was audio, (so that only 1 qTox instance is left in it) qTox starts spamming in stdout

```
[14:40:43.757] src/core/coreav.cpp:252 : Debug: toxav_group_send_audio error
```

and there's a ~constant ~15% CPU usage.

My other, "idle" qTox instance with >100 contacts uses ~3-5% CPU.

Not sure if it's related to new AV.

tux3:
Replaced by #2355

Issue #2288, opened 2015-09-24
[Feature request] avatars should be also affected by `Portable` mode


zetok:
From #1883 and #1642.

Related (depends on): #1656

zetok:
Fixed in #2346, thus closing.

Issue #1656, opened 2015-05-15
Gather all user data in a single file, so we can encrypt and keep track of it


tux3:
For each profile we currently have:
- A toxcore data save file
- A qTox preference file
- A chat history database file
- A cache of friend's avatars
- And possibly a file transfer DB soon?

That's not a good situation, especially if we want to support encrypted profiles (we can't just encrypt half the important files and leave the rest out there! Which we currently do!).

It'd be cleaner to keep that user data in a single file instead of scattered all over the place. And of course we still want our toxcore save file to be compatible with other clients.

agilob:
Ymm I wanted to add new table in chat history DB. Just one table with 4 or 5 columns. There is already some code for that. Is there any disadvantage of keeping it in one table? 
I see four advantages:
1. file transfer history and chat is encrypted/unencrypted at once
2. less underlying code/classes
3. one database
4. did I say 4?

tux3:
@agilob Where would you store e.g. the screenshots that are in transfer but broken, as a BLOB in the DB?

agilob:
In the same place where profiles are `.config/tox/`+`transfers/`. Screenshot will be removed after transfer is finished successfully.

tux3:
There's still the problem of encryption, then. We have all those files (screenshot, avatar caches, the qtox settings, etc) that sit around unencrypted.

A potential advantage of centralizing the user data in one place and enforcing that is that with this design we could do some sandboxing later on: only allow qTox to write to the current user's data file, and not touch anything else on disk, for security. That's not going to happen in the short term or any time soon, but if we start with the right design I think it'd help. At least with keeping all of a user's profile encrypted, at first.

ghost:
Just a theoretical idea: If you want to encrypt multiple files, why not make a database which includes a list of all files and some random encryption keys for these files and only encrypt this database with the user passphrase?

tux3:
@blobgo That's a way to do it, but it would make the code pretty complicated.

zetok:
It's a bad idea, given that data will be corrupted. AFAICT, when you store more data in one file, chances of data being corrupted increase, along with size of file + depending on FS they may increase further in case where file is being often used (one file for everything is going to be used often).

When you put all that data in 1 encrypted file, wouldn't just one corrupted bit make file unreadable and impossible to recover by normal means?

I mean, as long as user data wouldn't be at stake, I would like idea of stuff being in one file…

> And possibly a file transfer DB soon

[...]

PR #2346, opened 2015-10-05
Save avatar in DB

PKEv:
Partially addresses an issue #1656 

Question:
Whether it is necessary import old avatars?

### DIFF
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -36,9 +36,11 @@
 #include "src/persistence/db/encrypteddb.h"
 
 static HistoryKeeper *historyInstance = nullptr;
+QMutex HistoryKeeper::historyMutex;
 
 HistoryKeeper *HistoryKeeper::getInstance()
 {
+    historyMutex.lock();
     if (historyInstance == nullptr)
     {
         QList<QString> initLst;
@@ -67,6 +69,7 @@ HistoryKeeper *HistoryKeeper::getInstance()
         dbIntf = new PlainDb(path, initLst);
         historyInstance = new HistoryKeeper(dbIntf);
     }
+    historyMutex.unlock();
 
     return historyInstance;
 }
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -136,13 +136,11 @@ HistoryKeeper::HistoryKeeper(GenericDdInterface *db_) :
     }
 
     ans.clear();
-    bool needImportAvatar = false;
     ans = db->exec("PRAGMA table_info('aliases')");
     ans.seek(2);
     if (!ans.value(1).toString().contains("av_hash"))
     {
         //add collum in table
-        needImportAvatar = true;
         db->exec("ALTER TABLE aliases ADD COLUMN av_hash BLOB");
         qDebug() << "Struct DB updated: Added column av_hash in table aliases.";
     }
@@ -151,16 +149,9 @@ HistoryKeeper::HistoryKeeper(GenericDdInterface *db_) :
     if (!ans.value(1).toString().contains("avatar"))
     {
         //add collum in table
-        needImportAvatar = true;
         db->exec("ALTER TABLE aliases ADD COLUMN avatar BLOB");
         qDebug() << "Struct DB updated: Added column avatar in table aliases.";
     }
-
-    if (needImportAvatar)
-    {
-        importAvatar();
-    }
-
 
     updateChatsID();
     updateAliases();
@@ -173,34 +164,18 @@ HistoryKeeper::HistoryKeeper(GenericDdInterface *db_) :
         messageID = sqlAnswer.value(0).toLongLong();
 }
 
-void HistoryKeeper::importAvatar()
+void HistoryKeeper::importAvatar(const QString& ownerId)
 {
-    QSqlQuery sqlAnswer = db->exec("SELECT * FROM aliases;");
-    while (sqlAnswer.next())
+    QString puth (Settings::getInstance().getSettingsDirPath() +
+                  QString("avatars") + QDir::separator() +
+                  ownerId +".png");
+    qDebug() << QString("Try import avatar for: %1.").arg(ownerId);
+    getAliasID(ownerId);
+    if (QFile::exists(puth) && !hasAvatar(ownerId))
     {
-        QString userID = sqlAnswer.value(1).toString();
-        QString puth (Settings::getInstance().getSettingsDirPath() +
-                      QString("avatars") + QDir::separator() +
-                      userID +".png");
-        qDebug() << QString("Try import avatar for: %1.").arg(userID);
-        if (QFile::exists(puth))
-        {
-            QPixmap pic(puth);
-            saveAvatar(pic,userID);
-            qDebug() << QString("Import avatar for: %1.").arg(userID);
-        }
-        else
-        {
-            puth = Settings::getInstance().getSettingsDirPath() +
-                    "avatar_" + userID + ".png";
-            if (QFile::exists(puth))
-            {
-                QPixmap pic(puth);
-                saveAvatar(pic,userID);
-                qDebug() << QString("Import from profile_dir avatar for: %1.").arg(userID);
-            }
-        }
-
+        QPixmap pic(puth);
+        saveAvatar(pic,ownerId);
+        qDebug() << QString("Import avatar for: %1.").arg(ownerId);
     }
 }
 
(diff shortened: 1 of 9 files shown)


agilob:
Will this fix #2288 as well? Since avatar is saved in db, qtox will 'carry' avatars between systems in one place.

PKEv:
yes, it will be closed

agilob:
Ok, I will start using this for a few days and tell you the results.

PKEv:
Thanks, I'll wait

agilob:
So there is one issue I found so far
- [x] when loading first time, avatars from file should be read and saved in database, then removed from drive. So I lost all avatars when I started it first time:

weblate9

PKEv:
Now avatars are common to different accounts, and if they are crossed, I can delete avatars also belong to another user...

agilob:
Before running your branch 3/4 of the contacts had avatars.

PKEv:
They are not lost - when you roll back to the mainline they get back, I do not delete nothing yet

agilob:
Yup, that's the problem. When running qtox from your branch, avatars are empty (qtox is using default image). Avatars will not be set unless someone changes it in the client, which possibly means, never. Current avatars must be copied to database on start and deleted from drive after saving in DB.

zetok:
> Current avatars must be copied to database on start and deleted from drive after saving in DB.

No, avatars shouldn't be deleted from hdd, since other clients are using that data as well.

> Avatars will not be set unless someone changes it in the client,

Err, avatars should be received every time you connect to friend - aren't they saved then?

PKEv:
then may be import all the avatars that are in the table ALIASES  and only delete own avatar

ProMcTagonist:
> Err, avatars should be received every time you connect to friend - aren't they saved then?

They're only received if you don't have their current avatar.

agilob:
OK, there is only this one regression where avatars are not copied from drive to database on first start with this code. I think this should be fixed before merging so people don't lose avatars. It would go like that:

for each my contact -> check if contact has avatar in database -> check if contact has avatar on drive -> copy to database

Copying should happen only once, and that copying code will be removed after a few weeks when all users switch to the new code.

PKEv:
Added imports at the first start.
Own avatar appear only the second start.
If you have already run this branch (the database structure has changed) for the import, you must manually remove the right column av_hash & avatar from table aliases, or the import is not fulfilled

zetok:
c50e83c6f041d660c052241087754c14b9a5e8ee: Importing avatars doesn't really work - it imported 2 avatars, and rest was not touched.

zetok:
5582bbbf664f55202ae088bd641197bff9b07c65: 5 imported, 2 avatars failed to import

(though that's a small sample)

PKEv:
Need more information
try again, adding debugging code
and please send a log

zetok:
5050fb05301a88b5f8ebcc3633d6077155c497a3: 5 imported, 2 not: https://gist.github.com/zetok/5e5032a4a80ffbfe52cc

I think that importing fails for contacts that are not in DB(?)

agilob:
> I think that importing fails for contacts that are not in DB(?)

Which I think is ok.

zetok:
> > I think that importing fails for contacts that are not in DB(?)
> 
> Which I think is ok.

Why would removing avatars from people that aren't in DB be ok?

agilob:
Removing their avatars or not importing to db?

PKEv:
Apparently these contacts do not have an avatar in dir %profile%/avatar/... 
I do not see other reason
You can send a list of files from a directory %profile%/avatar/ ?

zetok:
@agilob 

> Removing their avatars or not importing to db?

From user perspective that's the same.

# 

@PKEv 
https://gist.github.com/zetok/ac07ba7fb09632557e7f

PKEv:
@zetok  You really do not have an avatar for these contacts

zetok:
> @zetok You really do not have an avatar for these contacts

Which is weird, since on qTox compiled from `master` missing avatars are shown

PKEv:
@agilob 
i do not remove imported avatars,
it is not convenient if you have a lot of contacts with similar profiles

zetok:
Wait - what if avatars are imported correctly, but for some reason qTox doesn't load them?

PKEv:
@zetok I do not know what to say, they should be stored in this folder
Judging from the log, they were not imported

PKEv:
@zetok You can run a search at a file with the name: DA6B2411E6880C6CE25DA59E4163F70C6963108DD61E2C71D725E2F59F4C7B2F
?
I wonder where he is...

zetok:
masterPR

`master | PR`

```
zetok@galaxor ~ $ find -name 'DA6B2411E6880C6CE25DA59E4163F70C6963108DD61E2C71D725E2F59F4C7B2F*' -type f
./working/utox2/.config/tox/DA6B2411E6880C6CE25DA59E4163F70C6963108DD61E2C71D725E2F59F4C7B2F.txt
```

Huh.

Edit: from `DA6B2411E6880C6CE25DA59E4163F70C6963108DD61E2C71D725E2F59F4C7B2F.txt`: `
Hello, welcome to the Load Testing Echo Bot. This is an automatic responder that will repeat anything you say back to you. You can also use it for load testing. Say 'spam' if you want to be spammed. Say 'stop' to stop the spam. To pause the spam, set your status to away or busy.`

PKEv:
@zetok please try again

zetok:
f6b20e07d4555b95d33e71e6388350af3d90efb2: still missing avatars: https://gist.github.com/zetok/9ecd64931c58a46cf834

PKEv:
:rage:
You can send a list of files from a directory %profile%?

zetok:
https://gist.github.com/zetok/e20553f2429e12da78d5

PKEv:
no Ideas more...

PKEv:
Someone has no idea where else can be stored avatars?

PKEv:
rebase and update 'delete avatar' action and add icon

PKEv:
+ add icon to 'delete avatar' action

agilob:
I think this code looks better and it can be merge. Could you make a final test and merge it @zetok ?

zetok:
afb3bf35ee506a4fa56e9f3ec88e235b6eb22978: still not all avatars are shown: https://gist.github.com/zetok/c19c832f5c6ea4a0550a

PKEv:
@zetok 
Please, tell me toxid that doesn't have avatar.
Please, try to find file with that toxid on your computer (mask: _toxid_)

zetok:
well, there is `~/.config/tox/avatars/FB91935C998C55B6FA6C495FBD1BFFB335394CF4874FB422AFF3703B918E6A04.png`.

> Please, tell me toxid that doesn't have avatar.

That's not really doable, since a way to copy contact's PK was removed.

ProMcTagonist:
What about groupbot?

zetok:
@ProMcTagonist what about it?

ProMcTagonist:
It doesn't have an avatar and its ID is public knowledge. Sorry if I misunderstood the situation.

PKEv:
@zetok  please try again

zetok:
@PKEv https://gist.github.com/zetok/5dcc6efc92492d5897d2 ← part of the stdout output before I've killed qTox

PKEv:
interesting ....

PKEv:
@zetok please try again

zetok:
@PKEv awesome \o/
AFAICT every avatar on my test profile was successfully imported.

Note: importing seems to be very slow

I'm also going to test with my normal profile, and see how whether all avatars will be imported.

zetok:
**very slow**: https://gist.github.com/zetok/be9e3edf2be23763691f

Good news, is that it would seem that every contact that had avatar, still has it `\o/`

src/persistence/historykeeper.cpp
```diff
@@ -146,6 +167,21 @@ HistoryKeeper::HistoryKeeper(GenericDdInterface *db_) :
         messageID = sqlAnswer.value(0).toLongLong();
 }
 
+void HistoryKeeper::importAvatar(const QString& ownerId)
```
agilob: could you rename this method to `importAvatarToDatabase`, just `importAvatart` seems to be misleading for me.

agilob:
This was my only comment about the branch, if all works it can be merged :)

zetok:
> (…) if all works it can be merged :)

Well, the problem is speed - it's a serious slowdown on every start for anyone using spinning rust (hdd) who has bigger contact list.

agilob:
but that's a one time thing, isn't it?

zetok:
It seems to be a problem on every start.

PKEv:
I  can make it run once.

agilob:
Yes, it should run only once.

PKEv:
all is ready

agilob:
@zetok I'm ok with merging it, please test it everything works and merge before we get code conflicts.

PR #2348, opened 2015-10-06
fix: save position when moving widget

### LINKED ISSUE #2063: qTox doesn't save window position
Windows 7 and 10
- It seems that it saves something but it is wrong. Every restart it moves it's position a bit more to the left and also moves a bigger step to the top.
- This is what it saves: https://gist.github.com/towlie/de3a3d270e111d6d83cb
- I saw that qtox starts in the middle of the screen on first start. but on second start it starts at coordinates 0;0
- qTox only saves a new position if you resize the window but it will move a few pixels to the left again after every restart

PKEv:
fix #2063

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -451,6 +451,16 @@ Widget* Widget::getInstance()
     return instance;
 }
 
+void Widget::moveEvent(QMoveEvent *event)
+{
+    if (event->type() == QEvent::Move)
+    {
+        saveWindowGeometry();
+        saveSplitterGeometry();
+    }
+    QWidget::moveEvent(event);
+}
+
 void Widget::closeEvent(QCloseEvent *event)
 {
     if (Settings::getInstance().getShowSystemTray() && Settings::getInstance().getCloseToTray() == true)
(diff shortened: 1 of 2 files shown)


towlie:
Nice nice. This fixes the issue. But one thing is still left:

"Every restart it moves it's position a bit more to the left and also moves a bigger step to the top."
I would create a seperate issue for this part when this gets merged.
Shall I ping you for this issue @PKEv ?

agilob:
> "Every restart it moves it's position a bit more to the left and also moves a bigger step to the top."

What DE do you use? I saw it on Windows, but never could reproduce on my machine with KDE4/5.

towlie:
@agilob I use/used Windows 7 and Windows 10 (x64)

towlie:
I documented this whole thing here: https://github.com/tux3/qTox/issues/2349

agilob:
It doesn't fix the issue for me - KDE5. Window is moved a few pixels each restart.

PKEv:
this PR fix only сorrects "not the preservation" of the position after moving

PR #2350, opened 2015-10-06
accelerate return from "Away" status

### LINKED ISSUE #2301: When closing qTox and you were AFK and your status switch to 'Away' if you close qTox and reopen it your status is away
It happens since #2298 was merged.
Reproduce:
- Set your status to 'Online'
- Wait for qTox to think you're AFK, so it sets your status to 'Away'
- Close qTox without moving mouse, use Alt+F4
- Reopen qTox and your status is set as if you set it to 'Away'

Expected behaviour:
qTox should load your status as if you were 'Online'

What happens:
qTox load the 'Away' status

PKEv:
fix #2301 
please test this

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1609,6 +1609,13 @@ bool Widget::event(QEvent * e)
             Nexus::getInstance().updateWindowsStates();
 #endif
 
+        case QEvent::KeyPress:
+            if (autoAwayActive) //to accelerate return from "Away" status
+            {
+                emit statusSet(Status::Online);
+                autoAwayActive = false;
+            }
+
         default:
             break;
     }


src/widget/widget.cpp
```diff
@@ -1608,7 +1608,12 @@ bool Widget::event(QEvent * e)
         case QEvent::WindowStateChange:
             Nexus::getInstance().updateWindowsStates();
 #endif
-
```
zetok, line 1611: I think that it would be preferable to keep that line

TheNain38:
@PKEv For me it works for keys, even when using a macro, but if you click to close the window then it saves the wrong status, and when moving the mouse it takes the same time as before to realize the mouse was moved and switch to good status.

And also the `void Widget::onUserAwayCheck()` https://github.com/tux3/qTox/blob/42659aefe8f7fd6f2299defec2edd1767cad8292/src/widget/widget.cpp#L1628 is broken I think.

And also, what happened if qTox is closed by another program after setting the status to "Away"

PKEv:
"onUserAwayCheck" I have no way to verify this is platform-specific section, in Win work but slowly 

where stored status when exiting?

PR #2355, opened 2015-10-07
New AV

tux3:
Implements A/V using the new API.

This is a rebased version of #2344


tux3:
@mannol
There's a race condition in toxav with the new code, we can deadlock if a client answers a call while the other end cancels it.

The problem happens when thread A calls toxav_answer while thread B calls tox_iterate.

If tox_iterate in thread B is handling a call invite cancel message, it first locks `session->mutex` at `msi.c:handle_msi_packet:772`, then tries to lock `toxav->mutex` at `toxav.c:callback_end:1100`

Meanwhile, if thread A calls `toxav_answer`, it first locks `toxav->mutex` at `toxav.c:toxav_answer:381`, and then it tries to acquire `session->mutex` at `msi.c:msi_answer:227`

And that's a deadlock, when toxav and toxcore are supposed to be able to run on two different threads. 
Is toxav_answer supposed to be not thread-safe and only callable from the tox_iterate thread?
This is the sort of things that should either be in the doc, or get fixed.

mannol:
toxav_answer is (supposed to be) thread safe. I'll fix it.

tux3:
Thanks! Sorry for the rant.

mannol:
This should be fixed now. Events triggered from tox thread will now have priority in synchronisation over actions executed from api thread.

tux3:
Superseded by branches newav_final and newav_final_for_realsies

PR #2357, opened 2015-10-07
Fixed #2294

### LINKED ISSUE #2294: When changing language setting 'Proxy type: ' reset to 'None' instead of the value it was before

alexbakker:
If someone can think of a cleaner way to do this, please tell me.

Note: the same issue as #2294 will occur again if we ever add another combobox with items we want to translate dynamically.


zetok:
Um, could you use a bit more descriptive commit message?

alexbakker:
Um, yes.

PR #2364, opened 2015-10-09
fix write to DB

### LINKED ISSUE #2292: [Regression] qTox lost all my sent messages from few last days
qTox 1673b43e26c853f6446f228fec083af166cbf446 , has happened also on some earlier version.

I don't have any of my sent sent messages in my history after restarting qTox (all received messages are still shown), for past 8 days.

Seems like a recent regression, I'll try to track it down later with bisect.

PKEv:
fix #2292, please test i don't sure

### DIFF
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -273,7 +273,7 @@ QList<QString> HistoryKeeper::generateAddChatEntryCmd(const QString& chat, const
     int sender_id = getAliasID(sender);
 
     cmds.push_back(QString("INSERT INTO history (timestamp, chat_id, sender, message, alias) VALUES (%1, %2, %3, '%4', '%5');")
-                   .arg(dt.toMSecsSinceEpoch()).arg(chat_id).arg(sender_id).arg(wrapMessage(message)).arg(dispName));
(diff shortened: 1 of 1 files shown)


tux3:
<strike>I'm pretty sure this won't fix the problem. A null byte encoded as UTF8 will still be a null byte...</strike>

So, this has the effect of truncating the message/alias at the first null byte.

zetok:
It improves it a bit - messages are partially saved - if they contain null char, everything after it is truncated.

tux3:
@PKEv Do you know if we can change the column type from TEXT to BLOB?

PKEv:
And there is a real need to support the recording null-char, all messengers support it?
Otherwise probably no other way - just go to BLOB

PKEv:
I think this is to merge the code to improve stability

agilob:
@zetok this is about your issue and you know best how to reproduce it. Please merge if that improves situation.

zetok:
@agilob it improves situation - thus I'm going to merge it.

Also, I'm going to open a new issue, since when message contains NULL, part after NULL doesn't get saved.

agilob:
How about now after e107de1 got merged?

PR #2367, opened 2015-10-10
Login page: make "Create" and "Load" buttons use darker green

### LINKED ISSUE #2333: Pale buttons on login window using Windows 10
Hello,

when I used qTox on Windows 10, the buttons on the login window are pale. So it seems you aren't able to press the 'Load' or 'Create Profile' button, but actually it is possible what is also the desired behaviour.

Screenshot: 
qtox

zetok:
Change should make it easier to read text.

fixes #2333

Before:
before

Change:
change

next to

### DIFF
--- a/src/loginscreen.ui
+++ b/src/loginscreen.ui
@@ -638,7 +638,7 @@ margin-bottom:45px;</string>
               <string notr="true">border-radius:5px;
 padding:5px;
 color:white;
-background-color:#6cc865;</string>
+background-color:#42a33a;</string>
              </property>
              <property name="text">
               <string>Create Profile</string>
@@ -1003,7 +1003,7 @@ margin-bottom:5px;</string>
               <string notr="true">border-radius:5px;
 padding:5px;
 color:white;
-background-color:#6cc865;</string>
+background-color:#42a33a;</string>
              </property>
              <property name="text">
               <string>Load</string>


alexbakker:
I suggest changing the color to #6bc260 instead, for the sake of consistency with the rest of the UI. Looks perfectly readable on my monitor.

ovalseven8:
> I suggest changing the color to #6bc260 instead, for the sake of consistency with the rest of the UI. Looks perfectly readable on my monitor.

But there is not really a difference compared to the current color.

alexbakker:
> > I suggest changing the color to #6bc260 instead, for the sake of consistency with the rest of the UI. Looks perfectly readable on my monitor.
> 
> But there is not really a difference compared to the current color.

I know, that's why I said "for the sake of consistency". Because in my opinion, the current color doesn't make the text unreadable at all.

ovalseven8:
On my Linux device everything works fine. But I installed qTox on a Windows 10 device a few days ago and the button was so pale there that you can think the button is not 'clickable' or active. That's the reason why I created the issue...

zetok:
@Impyy 

> I suggest changing the color to #6bc260 instead, for the sake of consistency with the rest of the UI. Looks perfectly readable on my monitor.

Thanks - I'm going to check it, and if it's good, I'll update this PR.

On a side note, perhaps you could add to wiki link (or something) that would point to which colours are being used?
https://github.com/tux3/qTox/wiki/Coding

zetok:
I'm not sure whether `#6bc260` is the way to go:
qtox login page buttons button
↑ with "my" colour + old one

There's hardly any difference between old colour and #6bc260:
my new + "consistent"
↑ the colour on top is my own, whereas colour on the bottom is #6bc260.

When compared to screenshot above, I'm not sure if I see the difference between #6bc260 and the old colour..

zetok:
> Because in my opinion, the current color doesn't make the text unreadable at all.

Sorry, but that's not really a good argument - it's actually a pretty shitty one.
I would like to suggest to you to read http://www.w3.org/WAI/eval/preliminary.html#contrast

Basically, just because **for you** it's not problematic, it doesn't mean that same applies to other people.

Now, I don't think that qTox is good enough when it comes to that area, so that everyone would feel comfortable using it - but hopefully in the future this will change `:)`

alexbakker:
Right. But as long as qTox doesn't offer a few different color schemes, that information is not very relevant to this discussion. Simply making the background color darker wouldn't satisfy all cases according to that page, unless I'm missing something.

Anyways, all I'd like is a consistent UI, not having a bunch of different shades of green all over the place.

zetok:
> Simply making the background color darker wouldn't satisfy all cases according to that page, unless I'm missing something.

Yes, it wouldn't improve in all cases, but it would improve at least some.

> Anyways, all I'd like is a consistent UI, not having a bunch of different shades of green all over the place.

Consistency is usually good, but it shouldn't have higher priority than making UI be better for users.

If one of already used in qTox would do the job - the better, it should be used, but since it appears that none of currently used colours is good enough..

agilob:
So this can be closed?

PR #2371, opened 2015-10-10
fix #2167

### LINKED ISSUE #2167: qTox should check if string in "Add friend" input field is 64 chars long
If it is, qTox needs to check whether it seems to be a public key (A-Z0-9), and if so, inform the user that this is not a Tox ID, since it's not 76 chars long.

BiTOk:
qTox should check if string in "Add friend" input field is 64 chars long. fixes #2167

### DIFF
--- a/src/widget/form/addfriendform.cpp
+++ b/src/widget/form/addfriendform.cpp
@@ -24,6 +24,7 @@
 #include <QErrorMessage>
 #include <QApplication>
 #include <QClipboard>
+#include <QRegularExpression>
 #include <tox/tox.h>
 #include "src/nexus.h"
 #include "src/core/core.h"
@@ -145,9 +146,12 @@ Ignore the proxy and connect to the Internet directly?"), QMessageBox::Yes|QMess
 void AddFriendForm::onIdChanged(const QString &id)
 {
     QString tId = id.trimmed();
-    bool isValidId = ToxId::isToxId(tId) || QRegExp("\\S+@\\S+").exactMatch(tId);
+    QRegularExpression dnsIdExpression("^\\S+@\\S+$");
+    bool isValidId = ToxId::isToxId(tId) || tId.contains(dnsIdExpression);
+
     toxId.setStyleSheet(isValidId ? QStringLiteral("") : QStringLiteral("QLineEdit { background-color: #FFC1C1; }"));
-    toxId.setToolTip(isValidId ? QStringLiteral("") : tr("Invalid Tox ID format. Is's must be 76 digital or alphabet characters length or similar name@domain.com."));
+    toxId.setToolTip(isValidId ? QStringLiteral("") : tr("Invalid Tox ID format. It must have length of 76 hexadecimal characters or similar name@domain.com."));
+
     sendButton.setEnabled(isValidId);
 }
 
--- a/src/widget/form/addfriendform.cpp
+++ b/src/widget/form/addfriendform.cpp
@@ -54,9 +54,12 @@ AddFriendForm::AddFriendForm()
     head->setLayout(&headLayout);
     headLayout.addWidget(&headLabel);
 
-    connect(&toxId,&QLineEdit::returnPressed, this, &AddFriendForm::onSendTriggered);
+    connect(&toxId, &QLineEdit::returnPressed, this, &AddFriendForm::onSendTriggered);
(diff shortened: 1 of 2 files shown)


zetok:
- [x] regression: input field is highlighted as invalid on start, when user didn't supply anything and field is empty:
  don't highlight when empty

src/widget/form/addfriendform.cpp
```diff
@@ -143,6 +143,18 @@ Ignore the proxy and connect to the Internet directly?"), QMessageBox::Yes|QMess
...
 }
 
+void AddFriendForm::onIdChanged(const QString &id)
+{
+    QString tId = id.trimmed();
+    QRegularExpression dnsIdExpression("^\\S+@\\S+$");
+    bool isValidId = ToxId::isToxId(tId) || tId.contains(dnsIdExpression);
+
+    toxId.setStyleSheet(isValidId ? QStringLiteral("") : QStringLiteral("QLineEdit { background-color: #FFC1C1; }"));
+    toxId.setToolTip(isValidId ? QStringLiteral("") : tr("Invalid Tox ID format. It must have length of 76 hexadecimal characters or similar name@domain.com."));
```
zetok: I don't think that setting tooltip is enough - e.g. I didn't even think to hoover over input field to get some more info, and when things "didn't work", only then I've looked at code, to find out that tooltip is used.

What needs to be done in order to clearly communicate to user that it's not an ID, is to show a warning dialog window.

BiTOk:
I did it on purpose, because the empty box really is not a valid identifier. Why do you think that an empty identifier correct? How should lead the field after entering and clearing value?

zetok:
> because the empty box really is not a valid identifier.

Validity of identifier only matters when user inputs something that can be sent - since nothing is being sent with an empty field, highlighting it in red makes it only confusing for the user.

zetok:
> How should lead the field after entering and clearing value?

If there is no input, there should be no highlighting - though, as soon as user inputs something, it can/should appear, if what was entered isn't valid.

When user clears field, highlight should disappear.

agilob:
Good point @zetok, @BiTOk instead using textChanged signal for QLineEdit use editingFinished()

This won't start validator for every pressed key, but when user stopped typing (delay after typing or user deactivated lineedit)

antis81:
@BiTOk Please add `fixes #2167` to your PR description. This will reference the issue and should close it, when done.

BiTOk:
@agilob editingFinished isn't emitted by delay, it's emitted after >Return or Enter key is pressed or the line edit loses focus. I'll use textChanged signal.

src/widget/form/addfriendform.cpp
```diff
@@ -143,6 +142,29 @@ Ignore the proxy and connect to the Internet directly?"), QMessageBox::Yes|QMess
...
+    bool isValidId = tId.isEmpty() || ToxId::isToxId(tId) || tId.contains(dnsIdExpression);
+
+    QString toxIdText(tr("Tox ID", "Tox ID of the person you're sending a friend request to"));
+    QString toxIdComment(tr("it must have length of 76 hexadecimal characters or similar name@domain.com", "Tox ID format description"));
+
+    if(isValidId)
+    {
+        toxIdLabel.setText(toxIdText + QStringLiteral(" (") + toxIdComment + QStringLiteral(")"));
+    } else
+    {
```
tux3, line 161: This could be more readable, either get rid of the braces or put them all on their own lines.

tux3:
The explanation text is way too long, it needs to fit in the window for all translations.
It's also not very well worded in English.

snapshot2

Otherwise it seems to work fine.

BiTOk:
Could be replaced by: "76 hexadecimal chars or follow the format name@domain.com"? @tux3

tux3:
@BiTOk Sounds good.

ProMcTagonist:
Maybe "(either 76 hexadecimal characters or name@domain.com)"

zetok:
> Maybe "(either 76 hexadecimal characters or name@domain.com)"

This.

agilob:
@tux3 please merge if it's ok now.

zetok:
@agilob it seems to work fine for me, and @tux3's problem (text length) doesn't appear to be a problem any longer:
text length
red
Merge?

antis81:
Nice! I say yes - please merge.

PR #2373, opened 2015-10-10
optionally hide all qTox windows in screenshot

### LINKED ISSUE #2303: Screenshoting while multiple windows mode
I set option multiple windows mode, and when i doing screenshot i press "space" to hide chat-window, but only contact-list window is hidden.
That's effect you can see in attached gif
qtox
.

antis81:
fixes #2303

There's an occasional problem, that the main window reappears. Probably `FriendListWidget::setMode` makes the main window re-visible on a mode change.

@agilob Here you are. :smile: Reparenting the `Widget` class had not the desired effect as well. Instead, I set up a stack of currently visible widgets to restore them. Further fixed some other related bugs here.

### DIFF
--- a/src/widget/tool/screenshotgrabber.cpp
+++ b/src/widget/tool/screenshotgrabber.cpp
@@ -37,22 +37,24 @@
 
 ScreenshotGrabber::ScreenshotGrabber(QObject* parent)
     : QObject(parent)
+    , scene(0)
 {
-    scene = new QGraphicsScene;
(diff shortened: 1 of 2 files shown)


agilob:
> Further fixed some other related bugs here.

Any reported that could be closed?

agilob:
Works for me, please close issues that were reported and fixed but not mentioned here.

antis81:
@agilob I don't think there are any other reported issues - I just cleaned up code a little, improved readability and reliability - that's all.

PR #2383, opened 2015-10-11
fix IPC::waitUntilAccepted

### LINKED ISSUE #2360: Can't start a second instance
@tux3 Why waiting less make qTox not able to start a second instance? 59e06920692a5966645bf60baa8753f440842651

Is the timeout too low?
https://github.com/tux3/qTox/blob/59e06920692a5966645bf60baa8753f440842651/src/ipc.cpp#L192

antis81:
Fixes #2360.

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -188,7 +188,6 @@ bool IPC::waitUntilAccepted(time_t postTime, int32_t timeout/*=-1*/)
     while (!(result = isEventAccepted(postTime)))
     {
         qApp->processEvents();
-        QThread::msleep(10);
         if (timeout > 0 && difftime(time(0), start) >= timeout)
             break;
     }
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -19,8 +19,9 @@
 
 #include "src/ipc.h"
 #include "src/persistence/settings.h"
-#include <QDebug>
 #include <QCoreApplication>
+#include <QDebug>
+#include <QThread>
 #include <random>
 #include <unistd.h>
 
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -186,11 +186,13 @@ bool IPC::waitUntilAccepted(time_t postTime, int32_t timeout/*=-1*/)
 {
     bool result = false;
     time_t start = time(0);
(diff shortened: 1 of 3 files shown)


antis81:
Hm… @tux3 could you please help in finding the reason for #2360? However the sleep halts qTox for 10 secs. on startup and is definitely **not** the way how things should be done.

antis81:
@tux3 Found it. Will add a fix here.

antis81:
@tux3 Can you please review/verify this?

src/main.cpp
```diff
@@ -233,15 +232,10 @@ int main(int argc, char *argv[])
...
-            dest = Settings::getInstance().getCurrentProfileId();
-
-        time_t event = ipc.postEvent("activate", QByteArray(), dest);
-        if (ipc.waitUntilAccepted(event, 2))
+        time_t event = ipc.postEvent("activate");
+        if (!ipc.waitUntilAccepted(event, 2))
         {
-            if (!ipc.isCurrentOwner())
-                return EXIT_SUCCESS;
+            return EXIT_SUCCESS;
```
antis81, line 238: We probably should return error code, shouldn't we?

src/ipc.cpp
```diff
@@ -188,7 +189,6 @@ bool IPC::waitUntilAccepted(time_t postTime, int32_t timeout/*=-1*/)
     while (!(result = isEventAccepted(postTime)))
     {
         qApp->processEvents();
-        QThread::msleep(10);
```
tux3, line 191: Isn't it going to spin at 100% CPU if you remove the sleep?

antis81, line 191: No, `qApp->processEvents();` prevents it. Rather than that, qTox starts immediately the second you open it… :)

antis81, line 191: Sorry, I was wrong -> CPU actually spins to 100%. No problem though - I'll fix it and after we should be ok here.

antis81:
@tux3 Should be good now. You can check by setting the timeout to -1 (infinite) and never set result == true.

antis81:
@tux3 Should I merge this?

tux3:
@antis81 Yes, it looks good to me.

PR #2282, opened 2015-09-23
checking new profile filename

PKEv:
add checking new profile filename
fix #2275

PR #2388, opened 2015-10-12
checking new profile filename

### LINKED ISSUE #2275: Unknown Error on Profile Creation (OS can't handle "special" symbols in filenames)

PKEv:
fix #2275

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -269,6 +269,7 @@ void ProfileForm::onRenameClicked()
 {
     Nexus& nexus = Nexus::getInstance();
     QString cur = nexus.getProfile()->getName();
+    QRegExp shablon ("[a-zA-Z_\\-\\d]+");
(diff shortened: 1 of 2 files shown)


agilob:
> You can only use the following characters: letters from 'A' to 'z', and the characters '-' and '_' 

why? why not digits? what happens if current profile is called 1234_567 and there is another one called 123_4567?

zetok:
Just.. No.

As I've said earlier, this is a regression, and IMHO in its current form PR shouldn't be merged.
https://github.com/tux3/qTox/pull/2282#issuecomment-146833871
https://github.com/tux3/qTox/issues/2275#issuecomment-146847576

tux3:
What zetok said. This is a regression and way too restrictive.

PR #2395, opened 2015-10-13
Fix performance issues detected with static analyze

BiTOk:


agilob:
Any measures? @antis81 opinion?

BiTOk:
Just simple c++ performance rules.

antis81:
@agilob Ok to me. Though it will not really boost performance. Sorry, no measures.

BiTOk:
So small fixes can't significantly affect to performance, but it's better and faster than original.

agilob:
PVS studio is a SCA for windows compilers and doesn't include g++... if anything you should use cppcheck for qTox. There of course will be issues that are shared between compilers (nearly all), but PVS is dedicated for VisualStudio compiler which in many cases behaves differently than clang or g++.

Lets start from latest read about it: http://blog.jetbrains.com/rscpp/microsoft-cpp-compiler-bugs-ambiguities/

BiTOk:
No, PVS officially compatible with g++: http://www.viva64.com/en/d/0323/
I knew this rules above even when developed under linux. This is common c++ rules, just google it. 

How described by the your refers for the contents of the commit? This in no way related to the peculiarities of the compiler.

PR #2406, opened 2015-10-16
Make text about not long enough password disappear if no password given

### LINKED ISSUE #2405: When changing password if you type a character and then delete it, it shows "The password is too short"
Change password problem

It shouldn't because there is no more password

zetok:
fixes #2405


zetok:
@agilob ok, done.

PR #2409, opened 2015-10-17
Autorun:xdg: create autostart dir if needed

matttbe:
Hello,

First, thank you for developing and maintaining qTox!

After a fresh installation of Ubuntu, `~/.config/autostart` dir is not available. It is mostly created by `gnome-session-properties` when adding the first extra application which will be launched at startup.

With this modification, `QDir::mkpath("~/.config/autostart")` (= `mkdir -p`) is now used before trying to open a new file in `~/.config/autostart` dir.

It then fixes a bug reported by @363734 to me: "_qTox is launched at startup after having enabled corresponding option from qTox's settings_".

### DIFF
--- a/src/platform/autorun_xdg.cpp
+++ b/src/platform/autorun_xdg.cpp
@@ -26,14 +26,19 @@
 
 namespace Platform
 {
-    QString getAutostartFilePath()
+    QString getAutostartDirPath()
     {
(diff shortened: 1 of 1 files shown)


zetok:
Thanks for the PR!

antis81:
@matttbe Nice to see you here. :smile:

matttbe:
@antis81 Nice to see you too! :-)

PR #2430, opened 2015-10-21
Add :focus selector for the login button in loginScreen.css

codedust:
The button now changes it's background color when the mouse is pressed ("onmousedown").


agilob:
:+1:

PR #2436, opened 2015-10-21
Fix #2433 (check blob to null)

### LINKED ISSUE #2433: Overloaded operator issue with historykeeper.cpp
On line 527 the '!=' operator get's flagged as an error with:
../qTox/src/persistence/historykeeper.cpp:527:62: error: use of overloaded operator '!=' is ambiguous (with operand types 'QByteArray' and 'long')
    if (sqlAnswer.first() && sqlAnswer.value(0).toByteArray()!=NULL)

I'm not even 60% sure what's going on here exactly, new to dealing with QT, but from what I could understand both sqlAnswer.first() and sqlAnswer.value(0) are calling the same result from the SQL query above (L:526) which might be the issue.

PKEv:
Fix #2433 
abaut "sqlAnswer.first()"
After the query is executed, the query is positioned on an invalid record and must be navigated to a valid record before data values can be retrieved (from help)

### DIFF
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -524,7 +524,7 @@ void HistoryKeeper::removeAvatar(const QString& ownerId)
 bool HistoryKeeper::hasAvatar(const QString& ownerId)
 {
     QSqlQuery sqlAnswer = db->exec(QString("SELECT avatar FROM aliases WHERE user_id= '%1'").arg(ownerId.left(64)));
-    if (sqlAnswer.first() && sqlAnswer.value(0).toByteArray()!=NULL)
+    if (sqlAnswer.first() && (sqlAnswer.value(0).toByteArray().length() > 0) )
     {
         return true;
     }


src/persistence/historykeeper.cpp
```diff
@@ -524,7 +524,7 @@ void HistoryKeeper::removeAvatar(const QString& ownerId)
 bool HistoryKeeper::hasAvatar(const QString& ownerId)
 {
     QSqlQuery sqlAnswer = db->exec(QString("SELECT avatar FROM aliases WHERE user_id= '%1'").arg(ownerId.left(64)));
-    if (sqlAnswer.first() && sqlAnswer.value(0).toByteArray()!=NULL)
+    if (sqlAnswer.first() && (sqlAnswer.value(0).toByteArray().length() > 0) )
```
antis81, line 527: Should simply work with `if (!sqlAnswer.isNull(0))…`. Why always so complicated? :)

**Even better:** `return !sqlAnswer.isNull(0);`

agilob:
merged in e107de1

PR #2441, opened 2015-10-22
Compile on Mac

RowenStipe:
Added missing includes to get qTox compiled on Mac OSX 10.10

Includes pull #2436 fixes suggested by @antis81


RowenStipe:
Because without it I wasn't able to get them to compile properly. Would say that functions pow and round where not defined.

agilob:
ok, thanks!

PR #2454, opened 2015-10-23
Remove ~duplicated code for tray backed selection

zetok:
`else` clause already handles KDE5 case


tux3:
Fine by me, I guess if someone reports a bug they'll tell what DE they are running on anyways.

agilob:
I remember I didn't agree with this change, but I don't remember why... (fails on KDE5.1 without SN maybe?) Let me test it before merging.

agilob:
ok, I dont know, it worked now, it didnt before.

PR #2462, opened 2015-10-24
Change '.com' to '.tld'

ovalseven8:


tux3:
I don't know, does this make it any clearer to the user what we mean?

ProMcTagonist:
No.

ovalseven8:
The PR makes the string more general. At the moment at first view it seems like you can only use .com domains.

zetok:
> The PR makes the string more general.

If the users knew what `.tld` means, that would be true. Check, it doesn't even make more sense to me.

On the other hand, `.com` is something that is an ~equivalent for the users to `.tld`…

ProMcTagonist:
Most users will not know what .tld means, whereas .com is obvious to everyone. No one is going to think it's somehow restricted to .com.

Zero gain, active damage, PR is harmful, close.

subliun:
it should be changed to example.com as per http://www.iana.org/domains/reserved

zetok:
> it should be changed to example.com as per http://www.iana.org/domains/reserved

Yeah - @ovalseven8 mind making PR with that change? `:)`

PR #2463, opened 2015-10-24
some cleanup for audio in "newav_final_for_realsies"

antis81:
- ~~rename `Audio::subscribeInput()` -> `Audio::subscribe`~~
- Open audio output device only when not already open when subscribing to input.
  - It is questionable, that the output device should be opened at all in `subscribeInput`

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -170,8 +170,7 @@ void Audio::unsubscribeInput()
 
     if (inputSubscriptions > 0)
         inputSubscriptions--;
-    else if(inputSubscriptions < 0)
-        inputSubscriptions = 0;
+    assert(inputSubscriptions >= 0);
 
     if (!inputSubscriptions)
         _cleanupInput();
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -166,15 +166,15 @@ If the input device has no more subscriptions, it will be closed.
 void Audio::unsubscribeInput()
 {
     qDebug() << "unsubscribing input" << inputSubscriptions;
+    QMutexLocker locker(&audioInLock);
+
     if (inputSubscriptions > 0)
         inputSubscriptions--;
     else if(inputSubscriptions < 0)
         inputSubscriptions = 0;
 
-    if (!inputSubscriptions) {
-        closeOutput();
-        closeInput();
-    }
+    if (!inputSubscriptions)
+        _cleanupInput();
 }
 
 /**
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -184,14 +184,7 @@ void Audio::openInput(const QString& inDevDescr)
 {
     QMutexLocker lock(&audioInLock);
 
-    if (alInDev) {
-#if (!FIX_SND_PCM_PREPARE_BUG)
-        qDebug() << "stopping capture";
-        alcCaptureStop(alInDev);
-#endif
-        alcCaptureCloseDevice(alInDev);
-    }
-    alInDev = nullptr;
+    _cleanupInput();
 
     /// TODO: Try to actually detect if our audio source is stereo
     int stereoFlag = AUDIO_CHANNELS == 1 ? AL_FORMAT_MONO16 : AL_FORMAT_STEREO16;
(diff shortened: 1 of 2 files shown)


tux3:
Imo Audio::subscribe() isn't as clear as Audio::subscribeInput() that it only concerns the mic

antis81:
Actually the whole subscribe concept is unclear to me and seems more like a dirty hack. The method opens both input and output devices. But… when closing an in/out device (which happens in "openInput" / "openOutput" too!!), in-/outputs start to interfere with subscriptions.

tux3:
Yes, it's pretty messy. My understanding was that the output was pretty much always open, and we closed the input whenever nobody was using it (subscribers) so the integrated webcam mics wouldn't stay open all the time, keeping the webcam led on.

Feel free to do some cleanups, but does subscribing/unsubscribing risk closing the main audio output now? I though it was only for inputs

antis81:
This was also my thought at first - that's why I messed up the comment too :smile:. Yes, performing the first "subscribeInput" will already close the main output, if it was opened before (see `Audio::openOutput`). Then it reopens it (wtf!?). I don't like when methods do more than they should do. Unfortunately I know nothing yet about OpenAL API. It looks fairly simple to learn and straightforward though. I don't make any promises, but I try my best to fix it.

antis81:
@tux3 Can you review and maybe check these changes to `Audio` class, please? I cannot test it currently.

src/audio/audio.cpp
```diff
@@ -134,17 +134,19 @@ void Audio::setInputVolume(qreal volume)
...
 {
-    qDebug() << "subscribing input" << inputSubscriptions;
-    if (!inputSubscriptions++)
+    qDebug() << "subscribing input" << subscriptions;
+    if (!subscriptions++)
     {
         openInput(Settings::getInstance().getInDev());
-        openOutput(Settings::getInstance().getOutDev());
+        if (!alOutDev)
+            openOutput(Settings::getInstance().getOutDev());
```
antis81: This is important!

antis81:
@tux3 Reverted the rename back to `Audio::subscribeInput`.

src/audio/audio.cpp
```diff
@@ -521,19 +519,16 @@ bool Audio::tryCaptureSamples(int16_t* buf, int samples)
 
     alcCaptureSamples(Audio::alInDev, buf, samples);
 
-    if (inputVolume != 1)
+    for (size_t i = 0; i < samples * AUDIO_CHANNELS; ++i)
     {
-        for (size_t i = 0; i < samples * AUDIO_CHANNELS; ++i)
-        {
             int sample = buf[i] * pow(inputVolume, 2);
```
TheNain38: You forgot deleting 4 spaces for the line to be aligned

antis81: Thanks! Done!

antis81:
@tux3 @agilob Could one of you review (and run) this please and merge back into `newav_final_for_realsies` branch, if ok?

src/audio/audio.cpp
```diff
@@ -464,6 +419,49 @@ void Audio::playAudioBuffer(ALuint alSource, const int16_t *data, int samples, u
...
+    if (alInDev)
+    {
+#if (!FIX_SND_PCM_PREPARE_BUG)
+        qDebug() << "stopping capture";
+        alcCaptureStop(alInDev);
+#endif
+
+        if (alcCaptureCloseDevice(alInDev))
+        {
+            alInDev = nullptr;
```
agilob, line 432: What is it doing? Why is it not calling delete on object?

src/audio/audio.cpp
```diff
@@ -464,6 +419,49 @@ void Audio::playAudioBuffer(ALuint alSource, const int16_t *data, int samples, u
...
+    if (inputSubscriptions)
+        _cleanupInput();
+
+    if (alOutDev) {
+        alSourcei(alMainSource, AL_LOOPING, AL_FALSE);
+        alSourceStop(alMainSource);
+        alDeleteSources(1, &alMainSource);
+
+        ALCdevice* device = alcGetContextsDevice(alContext);
+        if (!alcMakeContextCurrent(nullptr)) {
```
agilob: Minor: remove brackets in this `if`

agilob:
Can test on Sunday.

:+1:

src/audio/audio.cpp
antis81, line 432: Because it is not calling `new` either -> have a look at `alcCaptureOpenDevice` (line 196). Looks like it "owns" the instance, but I am no OpenAL expert. :innocent:

src/audio/audio.cpp
antis81: Will do it in the later today evening…

antis81: Done!

agilob:
What about this one ?

antis81:
I'll rebase onto master later today…

PR #2456, opened 2015-10-23
New AV API v3

tux3:
Implements the new toxav API.

I'm not going through merge conflict hell to rebase this again, so pending testing and bugfixes, this one is going in!

PR #2466, opened 2015-10-25
domain.com -> example.com

ovalseven8:
@zetok


zetok:
Looks ok to me, thanks.

Right now `master` is in a bit of freeze, so that #2456 could be merged without bigger problems.
After(?) that PRs will get merged.

ProMcTagonist:
Good :)

Issue #2438, opened 2015-10-22
Clicking on Audio/Video settings tab freezes GUI


dolugen:
Note that this only happens when a webcam is plugged in. Otherwise I can visit the tab fine.

Running it from a terminal shows some info. Clicking on the tab shows below info. It doesn't say anything about video input/output.

System: 64bit Ubuntu 14.04 with i3 WM

```
[12:43:42.651] src/audio/audio.cpp:145 : Debug: subscribing input 0
[12:43:42.661] src/audio/audio.cpp:208 : Debug: Opening audio input  "Built-in Audio Analog Stereo"
[12:43:42.661] src/audio/audio.cpp:234 : Debug: "Opening audio output Built-in Audio Analog Stereo"
[12:43:42.666] src/audio/audio.cpp:154 : Debug: starting capture
[12:43:42.668] src/audio/audio.cpp:168 : Debug: unsubscribing input 1
[12:43:42.668] src/audio/audio.cpp:308 : Debug: Closing output
[12:43:42.669] src/audio/audio.cpp:282 : Debug: Closing input
[12:43:42.669] src/audio/audio.cpp:287 : Debug: stopping capture
[12:43:42.670] src/audio/audio.cpp:145 : Debug: subscribing input 0
[12:43:42.675] src/audio/audio.cpp:208 : Debug: Opening audio input  "Built-in Audio Analog Stereo"
[12:43:42.675] src/audio/audio.cpp:234 : Debug: "Opening audio output Built-in Audio Analog Stereo"
[12:43:42.678] src/audio/audio.cpp:154 : Debug: starting capture
[12:43:42.680] src/audio/audio.cpp:168 : Debug: unsubscribing input 1
[12:43:42.680] src/audio/audio.cpp:308 : Debug: Closing output
[12:43:42.680] src/audio/audio.cpp:282 : Debug: Closing input
[12:43:42.680] src/audio/audio.cpp:287 : Debug: stopping capture
[12:43:42.680] src/audio/audio.cpp:145 : Debug: subscribing input 0
[12:43:42.681] src/audio/audio.cpp:208 : Debug: Opening audio input  "Built-in Audio Analog Stereo"
[12:43:42.681] src/audio/audio.cpp:234 : Debug: "Opening audio output Built-in Audio Analog Stereo"
[12:43:42.684] src/audio/audio.cpp:154 : Debug: starting capture
```

dolugen:
It's the same with Unity DE.

[...]

Issue #2058, opened 2015-08-06
Switching to 'Audio/Video' settings tab is blocked on camera initialisation


koobs:
_Audio/Video_ tab content display is blocked (and waits) on the thread that initialises the camera (and/or microphone) in any situation when that tab is made active. Examples:
- Switching from any other settings tab **to** the _Audio/Video_ tab
- Entering settings (via Gear icon) if the _Audio/Video_ tab was the last active tab

Camera initialisation should be made non-blocking and/or in a separate thread, such that the contents of the _Audio/Video_ tab can be displayed before the camera/microphone device has been enumerated and initialised.

While camera initialisation takes place, the camera preview area could be made:
- Empty (minimum)
- Empty with a user message "Waiting for camera initialisation ..."
- Empty with a progress animation "Initialising ..."

If the progress of initialization is known (definite), then the code could update the progress bar or message accordingly:
1. Searching devices
2. Identifying resolutions
3. Initializing camera

Or whatever the process/steps may be.

Environment:
- **OS Name**: Microsoft Windows 7 Ultimate (x86_64, amd64, 64-bit)
- **OS Version**: 6.1.7601 Service Pack 1 Build 7601
- **Processor**: Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz, 3401 Mhz, 4 Core(s) +HT
- **Physical Memory**: 16.0 GB
- **Camera**: Logitech HD Pro Webcam C920
- **Hardware ID**: USB\VID_046D&PID_082D
- **Driver Provider**: Microsoft
- **Driver Version**: 6.1.7601.18208 (12/07/2013) (UVC)
- **qTox Version**: https://github.com/tux3/qTox/commit/f10b4eaee7072bb9fbad5490be016455ba934bcf

zetok:
Based on what you wrote in other issue, I take that you use

> (…) Logitech C920 using Native UVC driver (not Logitech Webcam Software)

?

Your OS is ..?

Is issue reproducible on newest qTox builds https://wiki.tox.chat/binaries (if they're available for your OS) ?

koobs:
Also reproducible with latest qTox (Windows 64-bit) build: https://github.com/tux3/qTox/commit/aed6c25fda02c2aadeeb464cc7be2b79723cc641

zetok:
#2467 should help with it.

PR #2467, opened 2015-10-25
Adds button to enable camera preview when user order it

agilob:
adds a button in av settings tab which blocks automatic start
of camera preview in the tab. The camera/desktop preview will be
loaded after user explicitly ordered it.

Hopefully this will help with debugging freezing qtox. at least we will
know if camera or audio is broken.

snapshot2

Won't close, but will help with #2438 debugging.

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -46,6 +46,7 @@ AVForm::AVForm() :
 {
     bodyUI = new Ui::AVSettings;
     bodyUI->setupUi(this);
+    bodyUI->camFrame->hide();
 
 #ifdef QTOX_FILTER_AUDIO
     bodyUI->filterAudio->setChecked(Settings::getInstance().getFilterAudio());
@@ -59,6 +60,7 @@ AVForm::AVForm() :
     connect(bodyUI->outDevCombobox, qcbxIndexChangedStr, this, &AVForm::onOutDevChanged);
     connect(bodyUI->videoDevCombobox, qcbxIndexChangedInt, this, &AVForm::onVideoDevChanged);
     connect(bodyUI->videoModescomboBox, qcbxIndexChangedInt, this, &AVForm::onVideoModesIndexChanged);
+    connect(bodyUI->enableCameraButton, &QPushButton::clicked, this, &AVForm::onEnableCameraClicked);
 
     connect(bodyUI->filterAudio, &QCheckBox::toggled, this, &AVForm::onFilterAudioToggled);
     connect(bodyUI->rescanButton, &QPushButton::clicked, this, [=](){getAudioInDevices(); getAudioOutDevices();});
@@ -85,10 +87,9 @@ AVForm::~AVForm()
 
 void AVForm::showEvent(QShowEvent*)
 {
+    getVideoDevices();
     getAudioOutDevices();
     getAudioInDevices();
-    createVideoSurface();
-    getVideoDevices();
     Audio::getInstance().subscribeInput();
 }
 
@@ -212,8 +213,19 @@ void AVForm::onVideoDevChanged(int index)
     createVideoSurface();
 }
 
+void AVForm::onEnableCameraClicked()
+{
+    bodyUI->enableCameraButton->hide();
+    bodyUI->camFrame->show();
+    createVideoSurface();
+}
+
 void AVForm::hideEvent(QHideEvent *)
 {
+    bodyUI->camFrame->hide();
+
+    bodyUI->enableCameraButton->show();
+
     if (camVideoSurface)
     {
         camVideoSurface->setSource(nullptr);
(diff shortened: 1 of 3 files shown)


ovalseven8:
Please change 'Enable camera' to 'Preview'.
Because some people could think they first have to click 'Enable camera' to do a video call, what's wrong.

ProMcTagonist:
The whole thing is a button? Can you toggle it back off?

agilob:
How about,'start video preview'?

you can by changing tab. give it a try

ghost:
Will this new behaviour be removed at some point?

agilob:
I don't see a reason why it should be removed. I was supposed to be implemented since beginning.

BiTOk:
@agilob Can we open camera in separate thread without this button? This can resolve problem with qTox freezing.

agilob:
We can, but it's not the problem. The issue is not to start a camera when user wants to change audio settings or video settings.

ovalseven8:
Of course, qTox should solve the problem with freezing but additionally I like @agilob's solution.

agilob:
Feel free to merge ;)

tux3:
If this is just to help debugging, then it shouldn't go into master. Unless people actually want to have to click that button.

zetok:
> If this is just to help debugging, then it shouldn't go into master. Unless people actually want to have to click that button.

It's not really, at least IMHO. Letting qTox use webcam causes **huge** cpu usage, and I don't see why e.g. laptop users would want that to happen when they have limited battery time, and when they just go to the settings power consumption dramatically jumps up.

I mean, yes, this is a ~workaround, but if something can't really be ~fixed, then workaround is still better than nothing - at least in this particular case.

ghost:
What about enabling the button if started as `qtox --debug`? And throwing into that option any future debugging tweak.

koobs:
I agree with others, the preview functionality just shouldn't block on the main thread (See: #2058). 

Once the preview is enabled, does the state persist (preview remains enabled) between program launches and tab switches, or does it have to be turned on _every time_ ?

If the latter and not the former, I'm heavily -1 on this as a solution in production releases. 

Preview functionality (by default) is great, its the blocking that needs solving.

agilob:
This is a new functionality, not only for debugging purpose, I don't want my camera to start when I want to change audio device.

rugk:
I think this is unnecessary. When you click on "audio/video" it's quite clear that you expect something with audio/video to happen. A preview screen is nothing surprising.
And laptop user really won't always look at the tab "audio/video" the whole time just to drain their battery. :grin:

> and when they just go to the settings power consumption dramatically jumps up.

and half their battery is drain immediately... :wink: 
No, a few seconds access to the webcam won't be such bad. If laptop users would really be cared about this they would just quickly switch away and next time they'll remember what's there.
Additionally this settings tab is not something whcih is accessed regularly. And if it is accessed it is very likely the user wants to change something regarding audio or video settings and at least the latter one already makes clear that the user already is in a video conversion.

So I'm in favour of this:

> What about enabling the button if started as qtox --debug? And throwing into that option any future debugging tweak.

Alternatively you should not merge it into master - maybe a development branch would fit.

TheNain38:
I'm with @rugk's opinion on this. The user expect that something with audio/video will happen if they go on this tab.

ovalseven8:
> I don't want my camera to start when I want to change audio device.

Agree. What's wrong with that PR?

agilob:
Someone doesn't like it...

rugk:
> I don't want my camera to start when I want to change audio device.

Just: Why?

agilob:
> Why 
> 
> > I want to change audio device.

Why would I want to change oil every time I tank up a car?

rugk:
> Why would I want to change oil every time I tank up a car?

You do not change the oil. You just have a look on it.
Because when changing oil you have to do something, for looking at your camera picture you do not.

Additionally when driving to the gas station the chances are also high that you want to change your oil - and you do not have to press a button before you get to the oil can.

So expect of "What you want" (UX) arguments - are there any technical arguments for such an additional button?

TheNain38:
e.g skype does have a separate audio and a separate video tab, but you do not have to press a button to see the video

ghost:
I also think it's inconvenient to add steps to such trivial tasks. It makes the software feel less "friendly" for the non geeks. Takes away the _great-it-just-works_ factor.

If knowing _if camera or audio is broken_ is the main goal, then the logs should be more verbose or, as TheNain38 says, there should be separate audio/video tabs (which is worse than having to activate the camera, IMO).

TheNain38:
@kehugter In fact, I don't like to have separate tabs, I would prefer it to stay the same as it is right now

ProMcTagonist:
What if: If the user enables the camera by clicking the button, then leaves the settings/tab, then comes back within one minute, they shouldn't have to click to turn it on again?

PR #2468, opened 2015-10-25
[WIP] for crypt profile

### LINKED ISSUE #2432: qTox does not encrypt
1. Create account without password
2. Set password
3. qTox doesn't encrypt the history file immediately!
4. After restart it's apparently encrypted BUT there's still the unencrypted file!

Don't know if the tox profile file gets encrypted properly too.

PKEv:
fix #2432
fix change crypt/encrypt profile

Help, I can not understand why in an encrypted profile is not saved avatars

### DIFF
--- a/src/persistence/historykeeper.cpp
+++ b/src/persistence/historykeeper.cpp
@@ -26,7 +26,6 @@
 #include <QSqlError>
 #include <QFile>
 #include <QDir>
-#include <QSqlQuery>
 #include <QVariant>
 #include <QBuffer>
 #include <QDebug>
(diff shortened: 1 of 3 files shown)


PKEv:
@tux3 
I see an encrypted profile generally can not take the avatar of another user, is this normal? or there is a bug with encryption?

tux3:
@PKEv It doesn't work now, someone must have broken it.
When I implemented avatars, they still worked the same for encrypted and non-encrypted profiles, but that was a long time ago.

towlie:
Aren't avatars now saved into the db which gets encrypted? So how should others be able to see the content?
@tux3 You might have overlooked this https://github.com/tux3/qTox/pull/2346

tux3:
@towlie 
Now that I look at it #2346 sounds like a horribly bad idea due to how the encrypted database seems to work.
AIUI, it saves everything that ever happened as text SQL queries and encrypts that, so it's not an actual database on disk, it's going to be terribly slow to load/save, and it wastes a lot of space.

PKEv:
Yes for encrypted database runs all bad
I do not know the encryption algorithm and why he does not write data into the database
@tux3 please,tell me what is the essence of the algorithm

tux3:
I don't know why it doesn't work, I'll take a look at it once the new av is merged.

PR #2476, opened 2015-10-27
Starting point for possible OSX video functionality

RowenStipe:


RowenStipe:
I messed up some where with this... Sorry.

Issue #2424, opened 2015-10-19
Switch between video devices in new qTox video view


ovalseven8:
Hello,

the switches between the video devices during a video call aren't working properly at the moment.
Sometimes is works a little bit, sometimes it doesn't work at all.

For example if I had 'Desktop' and switched to 'None' it should display my avatar on the screen of the friend. But it doesn't change there. There is still the desktop, but only a picture of it that doesn't change.

agilob:
This is now easily doable.

ovalseven8:
Please check and you'll see that it doesn't work properly yet.

agilob:
I know the problem, it's not yet implemented, but what I say, it's easy to do it now. This will wait until new AV is merged so less code will conflict, and new AV is high priority.

ovalseven8:
@agilob: I see, sorry. :wink:

tux3:
Fixed by the new AV PR #2456

zetok:
Since it should be fixed now, closing.

If you can still reproduce it on new AV, please comment for reopen.

ovalseven8:
Please reopen.

Just tested it and it's even worse now.
It doesn't work properly and qTox hangs when changing the video device.

Please use two qTox instances and test it comprehensively. I think that would be most helpful because you can see the errors/failures.
For example the own video in the video view isn't updated and so on. Now, during a video call, qTox hangs in general.

zetok:
> It doesn't work properly and qTox hangs when changing the video device.
> 
> Please use two qTox instances and test it comprehensively. I think that would be most helpful because you can see the errors/failures.
> For example the own video in the video view isn't updated and so on. Now, during a video call, qTox hangs in general.

Can't reproduce.

tux3:
@ovalseven8 If you see qTox hanging, can you please try to take a backtrace in a debugger? It would help a lot to resolve the problem.

ovalseven8:
@tux3:
Here some information. Hope it will help a little. I just opened qTox (without making a video/audio call) and changed my video settings (one time using Desktop, one time using Webcam, ...).
Because of that qTox was unbelievably slow and hanged for many seconds again and again. So, not really usable at all.

```
(gdb) run
Starting program: /home/User/Downloads/qtox 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff27ef700 (LWP 3777)]
[...]

Issue #2431, opened 2015-10-21
Only turn on audio when a certain volume is hit


ghost:
Mumble has the functionality that the microphone only turns on when a certain volume is hit.
I think it would perhaps be a good idea for qTox as well.

Especially, for suppressing background noise.

```
<mannol> toxic had that feature
<mannol> it was in toxav for a while before i removed it
<mannol> its easy to make
<reselver> OK. Why was it removed?
<mannol> filter audio was introduced IIRC
<reselver> mannol: So, it should be implemented in client, right?
<mannol> yes
<reselver> mannol: Didn't test it in toxic - had it some positive effects?
<mannol> yeah
<mannol> like not using bandwidth when you are not talking and etc
```

agilob:
Related to push to talk?

mannol:
like push to talk but automatic

mannol:
It can easily be done. I can show you how if you are interested.

ProMcTagonist:
+1, especially if it really does save bandwidth. We need as much of that conserved as possible.

An-dy:
+1 (as a option when it not enable easy crypto analysis based on word length (silence "no" silence and silence "yes" silence need to be indistinguishable)).

sudden6:
Implemented by https://github.com/qTox/qTox/pull/4790

PR #2509, opened 2015-11-06
Improve audio based on "newav_final_for_realsies"

### LINKED ISSUE #2435: Output device isn't paused/closed after playing a ringtone/message beep
@antis81 
https://github.com/tux3/qTox/pull/2415#issuecomment-149213985

### LINKED ISSUE #2504: Filter audio in qTox is broken
Hello,

since the merge of ToxAV, the filter audio function is broken in qTox.
You can activate it but the audio is awfully ugly then.

Also tested it with uTox and it works fine - it seems only to be a qTox issue.

antis81:
Revival of #2463
- [x] Open audio in-/output device only, if not already open, when subscribing to in-/output.
- [x] use internal methods to cleanup & close audio devices
- [x] Change audio device during active call.
- [x] Fix: un-/subscribe to audio in/out
- [x] Fix: open/close audio device
  - [x] Fix #2435: Close audio device after playing a "notification" sound.
  - [x] Fix: Audio devices remain open after finished call.
- [x] Fix: AV-Settings interfers with active audio call
  - Close AV-Settings window accidently calls `unsubscribeInput` twice and closes audio devices when hiding the widget
- [x] Fix #2504: Remove regular breaks in filtered audio.
- [x] Improve situation with clipping noise when (re-)opening audio in/out device
  - Still appears when re-opening qTox (not the audio device itself), but highly improved now -> it's a 99% solution… :)
- [x] Implement non-blocking audio player for playback of "notification" alerts and sound buffers/files with a known length. (Yay, no audio timers anymore! :smile:)
- [x] Privatize audio handling and move completely to private implementation -> make it stable and reliable.
  - [x] Move every dependency of OpenAL related stuff to `Audio`.
- [x] Our `Audio` class should be aware of how many OpenAL sources are in use -> "sync" `ToxCall::source` with a private `ALuint*` array (**caution: big topic!**)…

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -75,12 +75,12 @@ Audio::Audio()
 
 Audio::~Audio()
 {
-    closeInput();
-    closeOutput();
     audioThread->exit();
     audioThread->wait();
     if (audioThread->isRunning())
         audioThread->terminate();
+    cleanupInput();
+    cleanupOutput();
 }
 
 /**
@@ -171,7 +171,11 @@ void Audio::unsubscribeInput()
 void Audio::subscribeOutput()
 {
     QMutexLocker locker(&audioOutLock);
+    internalSubscribeOutput();
+}
 
+void Audio::internalSubscribeOutput()
+{
     if (!alOutDev)
         initOutput(Settings::getInstance().getOutDev());
 
@@ -189,15 +193,6 @@ void Audio::unsubscribeOutput()
 
     if (!outputSubscriptions)
         cleanupOutput();
-}
-
-/**
-Open an input device, use before suscribing
-*/
-void Audio::openInput(const QString& inDevDescr)
-{
-    QMutexLocker lock(&audioInLock);
-    initInput(inDevDescr);
 }
 
 void Audio::initInput(const QString& inDevDescr)
@@ -264,14 +259,10 @@ void Audio::initInput(const QString& inDevDescr)
 }
 
 /**
-Open an output device
-*/
-bool Audio::openOutput(const QString &outDevDescr)
-{
-    QMutexLocker lock(&audioOutLock);
-    return initOutput(outDevDescr);
-}
+@internal
 
+Open an audio output device
+*/
 bool Audio::initOutput(const QString& outDevDescr)
 {
     qDebug() << "Opening audio output" << outDevDescr;
@@ -336,35 +327,13 @@ bool Audio::initOutput(const QString& outDevDescr)
 }
 
 /**
-Close an input device, please don't use unless everyone's unsuscribed
-*/
-void Audio::closeInput()
-{
-    QMutexLocker locker(&audioInLock);
-    cleanupInput();
-}
-
-/**
-Close an output device
-*/
-void Audio::closeOutput()
-{
-    QMutexLocker locker(&audioOutLock);
-    cleanupOutput();
-}
-
-/**
 Play a 44100Hz mono 16bit PCM sound
 */
 void Audio::playMono16Sound(const QByteArray& data)
 {
     QMutexLocker lock(&audioOutLock);
 
-    if (!alOutDev)
-    {
-        if (!initOutput(Settings::getInstance().getOutDev()))
-            return;
-    }
+    internalSubscribeOutput();
 
     ALuint buffer;
     alGenBuffers(1, &buffer);
@@ -484,6 +453,11 @@ void Audio::playAudioBuffer(ALuint alSource, const int16_t *data, int samples, u
         alSourcePlay(alSource);
 }
 
+/**
+@internal
+
+Close active audio input device.
+*/
 void Audio::cleanupInput()
 {
     mInputInitialized = false;
@@ -503,6 +477,11 @@ void Audio::cleanupInput()
     }
 }
 
+/**
+@internal
+
+Close active audio output device
+*/
 void Audio::cleanupOutput()
 {
     mOutputInitialized = false;
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -32,7 +32,6 @@
 #include "src/core/coreav.h"
 
 #include <QDebug>
-#include <QTimer>
 #include <QThread>
 #include <QMutexLocker>
 #include <QFile>
@@ -56,8 +55,6 @@ Audio& Audio::getInstance()
 
 Audio::Audio()
     : audioThread(new QThread())
-    , inputSubscriptions(0)
-    , outputSubscriptions(0)
     , alOutDev(nullptr)
     , alInDev(nullptr)
     , mInputInitialized(false)
@@ -138,15 +135,17 @@ void Audio::setInputVolume(qreal volume)
 
 If the input device is not open, it will be opened before capturing.
 */
-void Audio::subscribeInput()
+void Audio::subscribeInput(const void* inListener)
 {
     QMutexLocker locker(&mAudioLock);
 
     if (!alInDev)
         initInput(Settings::getInstance().getInDev());
 
-    inputSubscriptions++;
-    qDebug() << "Subscribed to audio input device [" << inputSubscriptions << " subscriptions ]";
+    if (!inputSubscriptions.contains(inListener)) {
+        inputSubscriptions << inListener;
+        qDebug() << "Subscribed to audio input device [" << inputSubscriptions.size() << "subscriptions ]";
+    }
 }
 
 /**
@@ -154,46 +153,44 @@ void Audio::subscribeInput()
 
 If the input device has no more subscriptions, it will be closed.
 */
-void Audio::unsubscribeInput()
+void Audio::unsubscribeInput(const void* inListener)
 {
     QMutexLocker locker(&mAudioLock);
 
-    if (inputSubscriptions > 0)
+    if (inListener && inputSubscriptions.size())
     {
-        inputSubscriptions--;
-        qDebug() << "Unsubscribed from audio input device [" << inputSubscriptions << " subscriptions]";
+        inputSubscriptions.removeOne(inListener);
+        qDebug() << "Unsubscribed from audio input device [" << inputSubscriptions.size() << "subscriptions left ]";
     }
 
-    if (!inputSubscriptions)
+    if (inputSubscriptions.isEmpty())
         cleanupInput();
 }
 
-void Audio::subscribeOutput()
+void Audio::subscribeOutput(const void* outListener)
 {
     QMutexLocker locker(&mAudioLock);
-    internalSubscribeOutput();
-}
 
-void Audio::internalSubscribeOutput()
-{
     if (!alOutDev)
         initOutput(Settings::getInstance().getOutDev());
 
-    outputSubscriptions++;
-    qDebug() << "Subscribed to audio output device [" << outputSubscriptions << " subscriptions ]";
+    if (!outputSubscriptions.contains(outListener)) {
+        outputSubscriptions << outListener;
+        qDebug() << "Subscribed to audio output device [" << outputSubscriptions.size() << "subscriptions ]";
+    }
 }
 
-void Audio::unsubscribeOutput()
+void Audio::unsubscribeOutput(const void* outListener)
 {
     QMutexLocker locker(&mAudioLock);
 
-    if (outputSubscriptions > 0)
+    if (outListener && outputSubscriptions.size())
     {
-        outputSubscriptions--;
-        qDebug() << "Unsubscribed from audio output device [" << outputSubscriptions << " subscriptions]";
+        outputSubscriptions.removeOne(outListener);
+        qDebug() << "Unsubscribed from audio output device [" << outputSubscriptions.size() << " subscriptions left ]";
     }
 
-    if (!outputSubscriptions)
+    if (outputSubscriptions.isEmpty())
         cleanupOutput();
 }
 
@@ -335,7 +332,8 @@ void Audio::playMono16Sound(const QByteArray& data)
 {
     QMutexLocker locker(&mAudioLock);
 
-    internalSubscribeOutput();
+    if (!alOutDev)
+        initOutput(Settings::getInstance().getOutDev());
 
     ALuint buffer;
     alGenBuffers(1, &buffer);
@@ -351,12 +349,14 @@ void Audio::playMono16Sound(const QByteArray& data)
     alGetBufferi(buffer, AL_SIZE, &sizeInBytes);
     alGetBufferi(buffer, AL_CHANNELS, &channels);
     alGetBufferi(buffer, AL_BITS, &bits);
-    int lengthInSamples = sizeInBytes * 8 / (channels * bits);
 
     ALint frequency;
     alGetBufferi(buffer, AL_FREQUENCY, &frequency);
 
     alDeleteBuffers(1, &buffer);
+
+    if (outputSubscriptions.isEmpty())
+        cleanupOutput();
 }
 
 /**
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -62,6 +62,8 @@ Audio::Audio()
     , outputSubscriptions(0)
     , alOutDev(nullptr)
     , alInDev(nullptr)
+    , mInputInitialized(false)
+    , mOutputInitialized(false)
     , outputVolume(1.0)
     , inputVolume(1.0)
     , alMainSource(0)
@@ -202,6 +204,7 @@ void Audio::initInput(const QString& inDevDescr)
 {
     qDebug() << "Opening audio input" << inDevDescr;
 
+    mInputInitialized = false;
     if (inDevDescr == "none")
         return;
 
@@ -256,6 +259,8 @@ void Audio::initInput(const QString& inDevDescr)
         alcCaptureStart(alInDev);
 #endif
     }
+
+    mInputInitialized = true;
 }
 
 /**
@@ -271,6 +276,7 @@ bool Audio::initOutput(const QString& outDevDescr)
 {
     qDebug() << "Opening audio output" << outDevDescr;
 
+    mOutputInitialized = false;
     if (outDevDescr == "none")
         return true;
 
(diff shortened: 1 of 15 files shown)


tux3:
Okay, so a basic test shows this is buggy, audio calls don't have sound right after starting qTox, until the A/V settings tab is opened.

src/audio/audio.cpp
```diff
@@ -144,7 +144,9 @@ void Audio::subscribeInput()
     if (!inputSubscriptions++)
     {
         openInput(Settings::getInstance().getInDev());
-        openOutput(Settings::getInstance().getOutDev());
+        if (!alOutDev)
+            openOutput(Settings::getInstance().getOutDev());
```
antis81: @tux3 Why do we try to open the output audio device in subscribeInput()? Could you check without? I currently cannot build… :(

tux3: Unfortunately I have a metric ton of things to do, so I can't afford to investigate audio code that is working fine in master.
Please find a way to test your code before sending PRs, I can't take in buggy untested code and I don't have the time to fix the bugs in every PRs.

antis81:
@tux3 It is not as buggy as one might think at first view - just required a slight modification… :smile: I finally managed to compile (will open new issue for that soon) and thus check it myself. Now listen to the sound… :smile:

You should notice, that there's no more nasty noises when re-calling a person. The only thing I'm not sure about is, if the open device is closed on hangup. I like to fix it in this PR, too.

@agilob @zetok Tested with ALSA. Can you confirm, this improves qTox audio with PulseAudio, please?

tux3:
@antis81 
It does work better now, but there's still a couple of problems :)

In `Audio::openInput` you call `cleanupInput` with sets `inputSubscriptions = 0`, so this means if we change the input source during a call the current subscribers are suddendly not accounted for and the the input gets closed instead of switched.

Also, the `assert(inputSubscriptions >= 0);` in `unsubscribeInput` is always true thanks two the two lines above, so it should be removed.

antis81:
> In Audio::openInput you call cleanupInput with sets inputSubscriptions = 0, so this means if we change the input source during a call the current subscribers are suddendly not accounted for and the the input gets closed instead of switched.

Yes, this is actually still more or less original behaviour and we need to get rid of that - now we can. We don't want to close an already opened device. However, this will affect switching in-/output devices e.g. in qTox settings -> but coreav correctly closes it now, before opening another one, does it?

tux3:
@antis81 
No, that was normal behavior. Having subscribers and having a device open are two different concepts. We must be able to change _which_ device is open without dropping all subscribers, that's what makes it possible to change the in/out devices used in the middle of a call.

Having subscribers means someone somewhere would like to use the microphone, when we have no subscribers we can close the input device (because nobody is using it), but when we change the input device we can't drop all subscribers (they still want to use _a_ microphone!).

antis81:
> Also, the assert(inputSubscriptions >= 0); in unsubscribeInput is always true thanks two the two lines above, so it should be removed.

Done!

antis81:
@tux3 Currently we have a single input & output device. The "inputSubscriptions" counter is like a reference counter -> actually this should be done in "openInput" then, instead of closing & re-opening. In short -> it's a total mess and - as you wrote - we need real management of "who is using which in-/output". I'll move some more code around to make situation better. I think you will understand soon… :)

antis81:
@tux3 Latest commits highly improve audio handling and fix most of the current open issues! Still works with single in/out audio device, but it actually makes things "just work"! Hope you understand me now… :smile:

tux3:
@antis81 
EDIT: Hm, I need to test this.

antis81:
> @antis81 Without even having tested, this looks like it'll fail an assert and crash as soon as you try changing the audio input device.

It doesn't! Fortunately I am able to build again (had a really hard time with ffmpeg!) -> In short: I checked this out in any of our current situation. But feel free to stress it. I want to get this really DONE :)

**EDIT:** Now basically the in/out device is **only** closed, when a) the device changes or b) no one is listening anymore.

tux3:
> It doesn't!

Indeed! I missed the changes in avform.cpp :)
Thanks for making it work, I'll test it out!

antis81:
The assert is for good reason. Maybe I should add a short comment on why it is there…

src/audio/audio.cpp
```diff
@@ -233,55 +250,49 @@ Open an output device
...
-                alcCloseDevice(alOutDev);
-                return false;
-            }
+            alGenSources(1, &alMainSource);
         }
         else
         {
-            qWarning() << "Cannot open output audio device " + outDevDescr;
+            qWarning() << "Cannot create output audio context";
+            alcCloseDevice(alOutDev);
```
antis81: This should change to `cleanupOutput` -> will change that…

antis81: Done!

antis81:
@agilob @tux3 @zetok and anyone interested in improving qTox audio: Please help review and test this PR, thanks!

agilob:
I had only chance to test it with echobot, audio filter adds incredible amount of noise and echo, disabling audio filter improves quality. Personally I use tox for audio so rarely, that I don't remember what audio quality was before ;)

After audio call was finished qTox was still holding 2 audio capture devices open:
screenshot9

I got an error message when quitting qtox:

```
[09:44:00.911] src/main.cpp:269 : Debug: Clean exit with status 0
AL lib: (EE) alc_cleanup: 2 devices not closed
```

I suppose one capture was from ~~AV tab~~ and one from the call (I opened AV tab to modify audio filter)

EDIT: When leaving AV tab audio capture is closed, so both were from one call?

antis81:
@agilob

> I had only chance to test it with echobot, audio filter adds incredible amount of noise and echo, disabling audio filter improves quality. Personally I use tox for audio so rarely, that I don't remember what audio quality was before ;)

See PR description… This behaviour is not introduced by this PR (actually it's better now, than in master… :))

towlie:
@antis81 Can you also fix https://github.com/tux3/qTox/issues/2435 with this PR?

towlie:
@antis81 
- Audio device pausing is broken (devices are never closed)
- "new message" beep doesn't work until you call someone

antis81:
@towlie Thanks for the feedback!

> "new message" beep doesn't work until you call someone

Fixed -> timer replaced!

> Audio device pausing is broken (devices are never closed)

That's what @agliob wrote. Don't have a solution yet. Seems the pointer from `ToxCall` is changed(??)

The biggest problem is, that the audio sources are spread "everywhere" -> see `ToxCall::source`. This is also the cause for the "jittering" when (re-)starting an a/v-call (throws a thousand errors). If the audio device is being opened (from the audio thread!!) and the static `Audio::createSource` is called, everything gets messed up. All those audio sources should be handled by the `Audio` class. Keeps going on and on… :)

Anyway: From now on I can merge any time…

antis81:
@agilob 

> EDIT: When leaving AV tab audio capture is closed, so both were from one call?

Yep: `1 input & 1 output = 2 devices` :)

antis81:
Uhm… why travis build fails? -> https://travis-ci.org/tux3/qTox/builds/91150291#L2975

towlie:
@antis81  qtox freezes until a call ring or message beep passed. I think we know what causes this. 
I found a way for you to check the play state without blocking the thread.
http://stackoverflow.com/questions/1046315/getting-notified-when-a-sound-is-done-playing-in-openal

AL_SOURCE_STATE Can have  AL_INITIAL, AL_PLAYING, AL_PAUSED or AL_STOPPED
http://wiki.delphigl.com/index.php/alGetSource

antis81:
@towlie Added a "real" non-blocking audio player… freeze no more! :)

antis81:
@tux3 @agilob  Hm… why travis doesn't want to build this PR? Changed OpenAL version?

``` text
src/audio/audio.cpp:588:42: error: ‘ALC_ALL_DEVICES_SPECIFIER’ was not declared in this scope
         pDeviceList = alcGetString(NULL, ALC_ALL_DEVICES_SPECIFIER);
```

**EDIT:** :bulb: Travis obviously builds with an "older" OpenAL lib. Moved compatibility code to the right places -> works! :+1:

ovalseven8:
Will this PR also close https://github.com/tux3/qTox/issues/2424#issuecomment-154532695?

antis81:
> Will this PR also close #2424 (comment)?

Nope, it would probably get really big and chaotic (having issues with this as well bro :frowning:).

agilob:
IDK, works for me.

towlie:
@antis81 I see that your https://github.com/tux3/qTox/issues/2435 fix kicks in after 2 - 3 seconds. How will it behave when we have custom notification sounds that will take 5 s and more? I mean will it still work? 
I heard that you can calculate the play duration (If you don't already do this)

towlie:
Oh I see you implemented this function I posted earlier, with a 2000ms wait time. So everything must be fine then.

antis81:
@towlie Yes, actually I'm waiting on a condition which never occurs. Thus, this is almost how to do it right. Anyway, works for this use case… :)

antis81:
Introducing a (smooth) threaded gain level meter with e97a4d2. However, it is not completely finished, but adds good value to A/V settings.

@agilob Would you review and help getting the threads right?

agilob:
Can try over the weekend

antis81:
Thanks @agilob. Would be great, if we can get that stable to a certain point.

antis81:
@agilob If there's a time slice available, please review. I improved the code for input gain metering. Filtered audio also works again now (though the filter concept is pretty much useless in current state). There's an open issue for that. Anyway, this is a future task. Devices are still is not closed properly -> that's the biggest open task.

@TheNain38 Feel free to check it out and maybe help review as well… :)

TheNain38:
@antis81 I'm not really good with audio, never work on that type of stuff, and I still have a lot to learn, but I will try to understand your code :)

antis81:
@TheNain38 That would be great! I wouldn't say exactly I'm an expert in audio programming either (probably you know more than I do) :). However, I want to get the audio concept into good shape and made quite some progress on that topic. As you can see from the growing list in the PR description above, this task is not 100% completed yet and some things are subject to change (e.g. the "subscription mechanism").

antis81:
@agilob Audio devices should be closed now correctly when finishing a call. Could you please verify?

@zetok Could you please do some usabilty checks?

@tux3 I'm not completely satisfied with the solution: Instead of counting "output subscriptions" we should actually keep the AL sources privatized and manage them completely within the `AudioPrivate` class. **The commit 0862b2d already addressed issue #2504 -> it reverts your latest changes to audio filtering in `master`. This should be correct though and works on "old & new" OpenAL implementations. Can you re-check, please?**

Though this is not finished 100%, this PR fixes a lot of problems with audio and I feel like this is worth a merge. Agree'd?

tux3:
@antis81 

> I'm not completely satisfied with the solution: Instead of counting "output subscriptions" we should actually keep the AL sources privatized and manage them completely within the AudioPrivate class. 

Why, and what would that look like? The audio sources are already just handles to resources managed by Audio/OpenAL, if you put the sources in AudioPrivate how does each ToxCall know what source to play sound on for example? And how would the AudioPrivate class know when it can open/close the output device if you don't keep a count of subscribers?

About the audio filter commit, are you sure you don't need to account for the two channels in Audio::getEchoesToFilter too?

agilob:
> @agilob Audio devices should be closed now correctly when finishing a call. Could you please verify?

Confirmed!

tux3:
One small bug, when the audio input device is switched from a working one to "None", the VU meter sticks to where it was before instead of going back to 0.

snap3

But I really love how smooth the VU meter is now!

TheNain38:
@antis81 the last item: `fix #2578: (@agilob Probably a mutex problem…)` can be removed because the issue was closed

antis81:
Thanks for all the feedback and glad it is mostly positive!

@tux3 According to the "privatizing" stuff.

> Why, and what would that look like?

I checked out some things with the PIMPL (private implementation) concept here and found it helps a lot with the related topic(s) and with understanding what's going on - it's just not consequently finished. The concept itself helps to encapsulate specific code (in this case OpenAL) and keep it where it belongs. By this PR, `Audio` is the only class knowing about the OpenAL audio sources -> that's how it should be. This should answer the "Why?" question. That being said, I have a rough picture on what it could look like: The AL sources should be stored in an array inside the AudioPrivate implementation. Actually this is very similar to the subscription mechanism, just that the calling class knows about being subscribed to an audio in/out, but not about the internal audio source id. I'm thinking about something inbetween the current `subscribeOutput(void* ptr_to_self)` and `createSource()`. To be true, if I knew how this completely looks like, I would have right off implemented it. Thus, I'm also asking for help to be able to create an overall "nice" concept. :)

antis81:
> About the audio filter commit, are you sure you don't need to account for the two channels in Audio::getEchoesToFilter too?

I'll check the code, but IIRC this is ok. Btw. I have only checked the "new" code path, so maybe you're right and we need to use both channels in both lines. :)

antis81:
> @antis81 the last item: fix #2578: (@agilob Probably a mutex problem…) can be removed because the issue was closed

Thanks, done.

tux3:
@antis81 I'm still not really convinced by idea of giving too much ownership to the Audio class. At the end of the day the Audio class doesn't know about the lifetime of the sources, that's something that only CoreAV and the ToxCalls know, so you can move the data around, but since it's _already_ a handle and not a heavy resource, I think it makes much more sense to give ownership of the audio source to whoever is actually using it.
As for the Pimpl, why did you decide to go with that? Pimpl is typically nice to improve compile times, but that was really not a problem here.
There are some things that don't give me a good feeling, like when you start adding QList<void*> lists that have to be managed, this seems more unsafe and more complex than it ought to be. I haven't had the time to look deeply at the code, but at first glance this is a bit worrying.

Overall it's definitely an improvement, but I'm a bit surprised by the design you went with.

antis81:
> As for the Pimpl, why did you decide to go with that? Pimpl is typically nice to improve compile times, but that was really not a problem here.

Usually pimpl is used to create solid API's and hide the private stuff behind them. Reaching out internal data is bad design. The reason here is actually to move code to where it belongs and hide complexity. Actually code is now split into several classes: Audio, AudioPlayer, AudioMeter & AudioMeterListener. The Audio class itself only manages the audio resources (open/close devices and source/subscription count). And … as I said this is not finished -> please keep reading… :)

> There are some things that don't give me a good feeling, like when you start adding QList<void*> lists that have to be managed, this seems more unsafe and more complex than it ought to be. I haven't had the time to look deeply at the code, but at first glance this is a bit worrying.

Agreed. It's some interims state and not my goal to keep track of unsafe pointers. Basically I just used this "technique" to track down, where opening and closing of audio devices is triggered from and fixed the buggy behaviour. Same thing is possible with "reference counting". However, for the output device this is exactly the same as counting how many AL-sources are in use. The `SID` (as the source ID is called in OpenAL) is private to Audio, but the "reference" is not. Thus, instead of keeping a `QList<void*>` we should keep a `QList<Audio::SID>` (Audio::SID matches the ALuint type). For input devices, there's no SID available and I suggest to "revert" and use simple reference counting. Now, that we know where the bugs are located. :)

antis81:
Well then… let me finish the PR and we can merge the final solution. I'll mark it WIP until done.

antis81:
@JacobHenner Any interest in helping out to finish this one?

antis81:
@tux3 Almost there - just a double unsubscribe of inputs to solve… :)

JacobHenner:
@antis81 I'll take a look at it tomorrow.

agilob:
Good job @antis81

antis81:
Thanks @agilob - always a pleasure working with you. :)

@JacobHenner Would you help check why `Audio::unsubscribeInput` is called twice when the settings dialog closes (hides actually) in parallel to an already active call? This mutes the microphone in the active call and I don't really have an explanation to why it happens.

JacobHenner:
@antis81 I'll take a look at it. It seems that there is a regression introduced from 1d9e89f, and the first available audio device is no longer chosen by default.

antis81:
@JacobHenner Fixed (workaround actually) the double `unsubscribeInput`. Any idea on how to implement a clean registration mechanism to audio inputs? Please open new PR for the device list stuff -> completely unrelated to this one.

@tux3 @agilob If nobody complains (I know it's not perfect) => I say it's ready to merge!!

agilob:
I can test it in two days

SmoothDude:
When audio output/input device selected to none, qTox aborts when it tries to play or record audio.
Log:

```
[16:19:29.050] ../src/widget/form/chatform.cpp:456 : Debug: onCallTriggered
[16:19:29.050] ../src/widget/form/chatform.cpp:510 : Debug: disableCallButtons
[16:19:29.052] ../src/core/coreav.cpp:177 : Debug: "Starting call with 0"
[16:19:29.052] ../src/audio/audio.cpp:407 : Debug: Opening audio input "Monitor of Built-in Audio Analog Stereo"
connect(2) call to /dev/shm/jack-1000/default/jack_0 failed (err=No such file or directory)
attempt to connect to server failed
[16:19:29.089] ../src/audio/audio.cpp:430 : Warning: Cannot open input audio device "Monitor of Built-in Audio Analog Stereo"
[16:19:29.089] ../src/audio/audio.cpp:384 : Debug: Subscribed to audio input device [ 1 subscriptions ]
[16:19:29.089] ../src/audio/audio.cpp:454 : Debug: Opening audio output "Built-in Audio Analog Stereo"
[16:19:29.090] ../src/audio/audio.cpp:513 : Warning: Cannot open output audio device "Built-in Audio Analog Stereo"
qtox: ../src/audio/audio.cpp:734: void Audio::subscribeOutput(Audio::SID&): Assertion `sid' failed.
Aborted (core dumped)
```

antis81:
@SmoothDude Thanks, fixed!

agilob:
@antis81 I made a call linux-windows and everything performs smooth! For me just audio filter works strange (as it always did) - if enabled I get a lot of noise, if disabled sound is clear, but I think this is ALSA related.

antis81:
@agilob Thanks for feedback! :+1: That's the same results I get with a call to echobot and linux-linux calls.

The filter behaves like it always did, because the concept by itself did not change. There's an open issue (#2431) -> I think this is more what we actually need instead of the current "filter". What should it even do? Filter echoes? Filter frequencies?

There's a few minor issues not solved, but the concept is "smooth" and it is important to get it into master, to keep distraction small…

agilob:
> The filter behaves like it always did

It does the opposite to what is should ;_;

towlie:
@antis81 It should cancel echoes but it never did work for most people (maybe all?). But it also raises the Volume of voice and suppresses noise which should work.

towlie:
Maybe you can use utox as reference?

antis81:
@tux3 I'm going to merge this today. Btw.: Build fails because Travis is missing the `libsqlcipher` library (same as in current master).

tux3:
LGTM.

antis81:
Thx - finally. :)

agilob:
Good job!

TheNain38:
@antis81 Really, thank you!

antis81:
@TheNain38 @agilob Same to you, really thank you for the feedback and patience! Keep contributing good stuff. :)

agilob:
Yea, what's our next milestone?

PR #2519, opened 2015-11-07
Don't make program name "qTox" traduisible

TheNain38:


src/loginscreen.ui
```diff
@@ -23,10 +23,10 @@
    </size>
   </property>
   <property name="windowTitle">
-   <string>qTox</string>
+   <string notr="true">qTox</string>
   </property>
   <property name="windowIcon">
-   <iconset resource="../../res.qrc">
+   <iconset resource="../res.qrc">
```
tux3, line 29: What are those path changes for?

TheNain38:
Qt creator atomatically added these

src/loginscreen.ui
TheNain38, line 29: Qt creator atomatically added these

TheNain38, line 29: @tux3 But these new paths are right?

tux3, line 29: Yep, the new paths are right actually.

Issue #2391, opened 2015-10-13
Allow viewing profile information of a contact


SlugFiller:
After the profile information of a contact is loaded, the list is updated with the name and (size-reduced) picture, instead of the ID. If you set an Alias (which is necessary to prevent people playing around with names), the name they set for themselves is no longer visible, as pointed out in #1195.

From my experience with certain other programs, I was expecting to be able to right click the contact, and pick "view profile", to see more details, such as the contact's ID, their current name (Not the alias I set), a zoomed-in version of their profile picture, etc. But no such option is present.

As an added bonus, it would be nice if there was a small textbox where I can add notes about a person. It's helpful in instances where you add more people than you are likely to remember in a month or two. This is a common occurrence to me, which I usually solve by going over our chat history, but sometimes I just want to add a little note specifying who this person is, and why I've added them.

zetok:
#1189

> As an added bonus, it would be nice if there was a small textbox where I can add notes about a person.

#1636

SlugFiller:
I was actually more concerned about a person's future names (since an alias would obscure those), than past ones. And also, more technical things, like the ID, which isn't a name at all (e.g. If I want to give a contact's ID to someone, so they can talk to each other). But I do see the mention of the user "About" page, which would definitely be needed for this.

ovalseven8:
Is done.

SlugFiller:
This isn't actually fixed, since you still can't see a contact's current name (Which they have set for themselves), only the alias you've set for them. This should remain open pending #1195 or #1189.

There is, however, a workaround now, in that you can copy the set alias to the notes, and set the alias to empty, resulting in the current name being shown.

However, I believe it should be shown in the dialog regardless of whether the current alias is set to empty.

wuniversales:
Why not add an option to see all the images, videos, files, video messages, audios ... of that specific contact?
That option is very useful.
![image](https://user-images.githubusercontent.com/48658161/72998700-93454780-3dfe-11ea-980f-b6bf9ea08d08.png)

Issue #2236, opened 2015-09-15
[Feature request] show friend public key


zetok:
Possibly on contact `About` page #2136

ovalseven8:
Is done.

Issue #1636, opened 2015-05-12
Contact "notes"?


tr37ion:
I would like to suggest adding a small notes input box for contacts. Here you can add some notes on each contact for better remembering them later on. Especially with many pseudonym contacts it will be nice keeping some small notes on each contact somewhere.

agilob:
It's very easy to implement, just follow 'alias' implementation.
Question is, do we really need this? I don't know any IM that supprts that.

tux3:
I'm not convinced we need this. This seems like feature creep to me.

tr37ion:
Well, if you don't store chat history this could be helpful to leave some small notes, too. Just a little undefined space for personal use.

TheLastProject:
Could you explain why the current alias system isn't good enough? Using "Mike@School" or so as alias should work well enough as note.

tr37ion:
I see, you want an example. To be honest, I think there is no _universal example_ of how people would use an added _notes_ field. Its the nature of notes, that people use them like they think they are helpful to them.
Just one more field would not compromise the software imho. Let the people decide, if and how they use it. And who knows, if you can later filter/search the note's content, people may use it for things I can't imagine.

zetok:
> Could you explain why the current alias system isn't good enough? Using "Mike@School" or so as alias should work well enough as note.

Because it's a goddamn **alias**. If you have to work with tons of people, what do you do when you have 2 Mikes? 5? 10? Perhaps 3 of them are from same school? Are you going to call them like `Mike the third` ? And how exactly does that help in telling you who the guy is? Or how much other info can you stuff in 128 character limited alias? Oh, and that comfort of editing this 1-lined field - who could even imagine something better…

For people who deal with a lot of people note feature is a **must**.

> It's very easy to implement, just follow 'alias' implementation.

I wonder though, aren't aliases saved in plain, unencrypted text file?

Zer0-One:
@zetok I'm thinking of adding something that lets you view someone's username history, so that you can track their username changes over time (to avoid the confusion of not knowing who half of your contacts are). Would a feature like that supersede this "notes" feature you're talking about?

ProMcTagonist:
@zetok 

> I wonder though, aren't aliases saved in plain, unencrypted text file?

Awful if true. Good thing I don't use them because you still can't see aliases and users' chosen display names at the same time.

@Zer0-One 

Like in Steam? A dropdown next to their name at the top of the chatform?

Zer0-One:
[...]

PR #2521, opened 2015-11-07
Added a context menu item to copy a friend's public key

alexbakker:
There were some requests for this in #tox. I think this is an important thing to have.

### DIFF
--- a/src/widget/friendwidget.cpp
+++ b/src/widget/friendwidget.cpp
@@ -41,6 +41,7 @@
 #include <QDebug>
 #include <QInputDialog>
 #include <QCollator>
+#include <QClipboard>
 #include <cassert>
 
 FriendWidget::FriendWidget(int FriendId, QString id)
@@ -81,6 +82,7 @@ void FriendWidget::contextMenuEvent(QContextMenuEvent * event)
         removeChatWindow = menu.addAction(tr("Remove chat from this window"));
 
     menu.addSeparator();
+    QAction* copyPublicKey = menu.addAction(tr("Copy public key"));
     QMenu* inviteMenu = menu.addMenu(tr("Invite to group","Menu to invite a friend to a groupchat"));
     QMap<QAction*, Group*> groupActions;
 
@@ -206,6 +208,10 @@ void FriendWidget::contextMenuEvent(QContextMenuEvent * event)
                 friendList->addCircleWidget(FriendList::findFriend(friendId)->getFriendWidget());
             else
                 Settings::getInstance().setFriendCircleID(id, Settings::getInstance().addCircle());
+        }
+        else if (selectedItem == copyPublicKey)
+        {
+            emit copyFriendIdToClipboard(friendId);
         }
         else if (groupActions.contains(selectedItem))
         {


towlie:
This was removed because it confuses users. So why do you reimplement this? (Publiic key != Tox ID)
Toxcore is not designed to share Tox IDs

alexbakker:
The reason it was confusing before was because it said "Copy tox id" which is why I made it say "Copy public key". If a user wants to know a friend's public key you can't expect them to extract the public key from the save file themselves.

ProMcTagonist:
> If a user wants to know a friend's public key

I can't imagine this ever happening for the vast majority of our target users. This will be even more confusing than before, because everyone will respond to confused users with "That's the public key, not the ID," and make them feel like they should have known that, when they were introduced to Tox with the premise that it's configurationless and crypto for everyone.

It would be better to adopt the protocol extension that @subliun is cooking up for Antox, with an opt out.

alexbakker:
I'm not saying it should be this way forever. But while subliun's protocol extension is still cooking I think we should at least provide this option for now. We can remove it later when someone comes up with a better solution.

tux3:
I don't see the point though, why would someone want to copy the public key? There's not much you can do with it, and it's easy to confuse it with a Tox ID.

alexbakker:
Say you have 2 people in your friend list with the same name, the same status message and the same avatar. How can you tell who's the impostor?

TheNain38:
There should also be a way to know when someone changes it's name, and what name it was before he changes

tux3:
@Impyy 
The simple solution would be to remove both, and re-add your friend. You need to know your friend's Tox ID anyways if you want to confirm its identity.

Technically you could also notice that one of your friend "disappeared", so the attacker would blow its cover immediately, and the order of the friend list corresponds to the order you added people in (unless you move people around) so you would know who's who if you remember what your friend list looked like yesterday.

I suppose that's a valid use case if you have a big friend list full of people you don't trust, but I'm not sure if we should trade real user confusion for a theoretical attack that can be solved by other means.

ProMcTagonist:
> Say you have 2 people in your friend list with the same name, the same status message and the same avatar. How can you tell who's the impostor?

Message history, name change history, status history. The last two are yet to be implemented but are critical for this very reason among others. Right clicking and checking a public key or ID that you've memorized the correct version of is not an acceptable solution.

alexbakker:
Having name history won't help if the contact you added uses the same name as another contact from the get go.

agilob:
> Say you have 2 people in your friend list with the same name, the same status message and the same avatar. How can you tell who's the impostor?

I dont even know what my toxid looks like... how would I know someones else?

This should be added to about windows of a user, not to context menu.

ProMcTagonist:
> Having name history won't help if the contact you added uses the same name as another contact from the get go. 

Which is why I also mentioned status history and message history. Your real friend has probably changed status and/or messaged you. It would also be helpful to have a small identicon for the key on friends' profile pages. Under no circumstances is it desirable for a user to be forced to memorize public keys to identify their friends.

alexbakker:
@agilob

> I dont even know what my toxid looks like... how would I know someones else?

If you notice someone is impersonating a friend on your friend list, the logical next step would be to contact that friend through a different channel and ask for their public key so you can compare. I'm not saying people should memorize each others public keys.

> This should be added to about windows of a user, not to context menu.

I agree, that would be a good way to do it. Like I said before, I don't think that this should be the permanent solution. But as long as there is no about window, we have to provide the option to obtain the public key in a different way.

@ProMcTagonist 

> Which is why I also mentioned status history and message history. Your real friend has probably changed status and/or messaged you.

Not everyone will have history enabled (assuming you want to make status history optional, similarly to message history). This is not a proper solution.

> It would also be helpful to have a small identicon for the key on friends' profile pages.

I agree, this would be nice to have.

> Under no circumstances is it desirable for a user to be forced to memorize public keys to identify their friends.

I agree, see my reply to @agilob.

zetok:
> I don't think that this should be the permanent solution.

You know what they say about temporary solutions.. `;)`

agilob:
I can try implementing today the about window for a user to stop this discussion. The window will be a place for those settings and feature. Do you like the solution @Impyy ?

alexbakker:
@agilob Yes that sounds good!

ovalseven8:
Related issues:
https://github.com/tux3/qTox/issues/1636 | Contact notes
https://github.com/tux3/qTox/issues/1189 | Lookup previous aliases
https://github.com/tux3/qTox/issues/2391 | Profile information
https://github.com/tux3/qTox/issues/2236 | Show friend public key

alexbakker:
Closing in favor of #2547

PR #2532, opened 2015-11-08
 Add an event filter so you can't modify sliders by scrolling, closes #1932

### LINKED ISSUE #1932: Remove mouse scroll ability for sliders (AV settings tab)
As it is now you can accidentally change the sliders when using mouse scroll to move the page.

TheNain38:
closes #1932
Now you can't modify a slider by scrolling on it.
But if you scroll on a slider it will propagate the event to the `AVForm` and it will make it scroll normally

I never worked with that type of stuff before, I would like that someone check if I've done it the right way

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -69,6 +69,9 @@ AVForm::AVForm() :
     bodyUI->playbackSlider->setValue(Settings::getInstance().getOutVolume());
     bodyUI->microphoneSlider->setValue(Settings::getInstance().getInVolume());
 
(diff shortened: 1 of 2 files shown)


TheNain38:
Could someone review my code? And say to me if there is a better way of doing this?

tux3:
@TheNain38 Instead of creating a FilterScroll class, you can just add `this` as the event filter, and you'll catch the events in `AVForm::eventFilter` (which already exists).
You just need to check in eventFilter that the object `QObject* o` passed in argument is either playbackSlider or microphoneSlider

TheNain38:
@tux3 All done

tux3:
LGTM

PR #2535, opened 2015-11-08
[WIP] Fix double sender name in action message

### LINKED ISSUE #2336: Bug, when receiving an action message, when reloading history, the sender name is displyed twice
Reproduce:
- Create two profile
- Send an action message to the other
- Close and reopen receiver's qTox
- See that there is the sender's name twice

TheNain38:
I've fixed the first bug, the second is if you send an action message and the OfflinemsgEngine deliver it, then it is sent with the sender's name twice

Should fix #2336 when finished
- [ ] Ready

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1081,7 +1081,7 @@ void Widget::onFriendMessageReceived(int friendId, const QString& message, bool 
     QDateTime timestamp = QDateTime::currentDateTime();
     f->getChatForm()->addMessage(f->getToxId(), message, isAction, timestamp, true);
 
-    HistoryKeeper::getInstance()->addChatEntry(f->getToxId().publicKey, isAction ? "/me " + f->getDisplayedName() + " " + message : message,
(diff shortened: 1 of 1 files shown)


TheNain38:
https://github.com/tux3/qTox/blob/f9c324c0f23e9e34b938178fabfd7edb63dc50c6/src/persistence/offlinemsgengine.cpp#L94 `val.msg->toString();` also contain the sender's name, and that is caused by the fact that the variable `ma` in https://github.com/tux3/qTox/blob/f9c324c0f23e9e34b938178fabfd7edb63dc50c6/src/widget/form/chatform.cpp#L952 comes from https://github.com/tux3/qTox/blob/f9c324c0f23e9e34b938178fabfd7edb63dc50c6/src/widget/form/chatform.cpp#L944 `addSelfMessage` and if `isAction` is true the returned message also contain the sender's name.

I would need to change the way it works to make it works correctly. I would need some help.
@tux3 Could you add the "help wanted" label please?

zetok:
Seems to be working again on ef5c094417b15db67b1b8736b5153a3cd221f488, thus closing.

PR #2538, opened 2015-11-09
Allow to scroll on QCheckBox and fix consistancy, fixes #2302

### LINKED ISSUE #2302: Can't scroll when mouse is on a desactivated checkbox

TheNain38:
fixes #2302

If later you add other QCheckBox in the settings, you don't have to worry anymore about this bug

### DIFF
--- a/src/widget/form/settings/advancedform.cpp
+++ b/src/widget/form/settings/advancedform.cpp
@@ -49,8 +49,13 @@ AdvancedForm::AdvancedForm() :
 
     for (QComboBox* cb : findChildren<QComboBox*>())
     {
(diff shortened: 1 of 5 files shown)


ProMcTagonist:
Thank you.

Tapsa:
Does this mean that when QScrollArea is scrolled and QCheckBox gets under the cursor, scrolling the QScrollArea is stopped? I consider that to be a bug.

TheNain38:
It means that when the scroll event gets to the QCheckBox it will be passed to the parent element, instead of taking effect on the QCheckBox which does nothing

PR #2543, opened 2015-11-09
Fix resources URI: change double slash to one and add one if missing

### LINKED ISSUE #2541: qTox crashes after clicking the login button
It doesn't crash always but often.
https://gist.github.com/towlie/cf23cfd75fed707e14fe

bt full:
https://gist.github.com/towlie/0394ba0ba93ac1552ebe

agilob:
closes  #2541

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -135,7 +135,6 @@ void Widget::init()
 
     actionLogout = new QAction(this);
     actionLogout->setIcon(prepareIcon(":/img/others/logout-icon.svg", icon_size, icon_size));
-    connect(actionLogout, &QAction::triggered, profileForm, &ProfileForm::onLogoutClicked);
 
     actionQuit = new QAction(this);
     actionQuit->setMenuRole(QAction::QuitRole);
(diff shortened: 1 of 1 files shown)


tux3:
No. This doesn't fix #2541

agilob:
It doesnt?

agilob:
ah ok

tux3:
#2541 is a crash because of using uninitialized data. 
Changing a resource's URI can't cause a crash. Worst case, some images will be missing.

Did you test your code before sending it?

agilob:
Ok  I didn't see the comment, sorry.

> Did you test your code before sending it? 

Yes, I did test it, but didn't crash for me. Instead tray icon was not responding to ~~left~~ right click

tux3:
It's not about not seeing the comment, you really need to be more careful. It's not the first time you send obviously broken code without bothering to test.
And then you send a broken fix that doesn't do anything!

I asked you before, please test your code if you're not 100% sure it's correct, that's the bare minimum.

agilob:
I test it now on my arch, ubuntu mint with xfce and windows7 64bit and only regression was on windows, it didn't show tray menu on right click, just nothing happened, but there was no crash.

Sorry about that.

tux3:
On Windows we're connecting to 0xBAADFOOD, this is ought to crash 9 out of 10 times.
I can see how this could not crash on Linux since it doesn't mark uninitialized data the same way windows does, so use valgrind, it'll **very** clearly catch this bug every time. 

The point remains that this is a crash _on startup_, those kind of stupid bugs shouldn't ever get through.

agilob:
I never used valgrind before, I'll give it a try next time :)

Thanks

PR #2569, opened 2015-11-13
update Lithuanian translation

minj:


zetok:
Could those changes be squashed?

Also, after squash, could you add commit which will remove unneeded lines via https://github.com/tux3/qTox/wiki/Translating#remove-obsolete-strings - the command has been slightly updated, so that it will remove unneeded in .ts line numbers.

minj:
Sure, I can but the point is to clearly shown what changed, is it not?

How about:
1) update+vanished+obsolete
2) actual translation changes

Or will you not accept anything but a single commit with unreadable diff?

Do you want me to submit a new PR or force push this one?

zetok:
> Sure, I can but the point is to clearly shown what changed, is it not?

Yes, when it is possible to have readable diff. Translation changes are hardly readable either way, and because of that it simply seems to be more beneficial to just put all "translation" changes in a single commit.

> Or will you not accept anything but a single commit with unreadable diff?

No, just given what I wrote above, a single commit seems like a best option.

> Do you want me to submit a new PR or force push this one?

Force push to this one. And thanks `:)`

minj:
The proposed command strips location information. Is this really what you want?

Personally I find it beneficial to see the source code in linguist

zetok:
> The proposed command strips location information. Is this really what you want?

Yep.

> Personally I find it beneficial to see the source code in linguist

Yes, the command above this one that should be used to update .ts is not stripping that info. At least that's how I think that it should work..

Stripping "useless" stuff / deprecated strings happens at the end, at least according to the "guide" on wiki.

minj:
Alright, I got what you mean. Done and squashed everything into a single commit. Should be good to merge now.

PR #2570, opened 2015-11-14
Allow adding friends using only public key, fixes #2551

### LINKED ISSUE #2551: [Proposal]: Add possibilty to add someone using only the public key
You would need to warn the user that it will not send a friend request.
And you will add the contact using

``` c
uint32_t tox_friend_add_norequest(Tox *tox, const uint8_t *public_key, TOX_ERR_FRIEND_ADD *error);
```

TheNain38:
fix #2551
- [x] Ready

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -562,6 +562,33 @@ void Core::acceptFriendRequest(const QString& userId)
     }
 }
 
+void Core::addFriendNoRequest(const QString& friendPubKey)
+{
+    if (hasFriendWithAddress(friendPubKey))
+    {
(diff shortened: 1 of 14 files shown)


agilob:
Anyone against?

ovalseven8:
Still don't think that qTox needs that feature.

tux3:
I'm strongly against this, unless you can first show in #2251 that many actual users need this feature. 
So far I haven't seen that _at all_, at best people are cautiously neutral about it.
We can't add an option for every feature. qTox should be the intersection of users' needs, not the union.

zoff99:
i don't think this should be closed

PR #2580, opened 2015-11-18
also rescan video devices

sudden6:
moved and renamed audio device rescan button to the bottom and renamed it to rescan devices, which it now does

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -63,7 +63,7 @@ AVForm::AVForm() :
     connect(bodyUI->videoModescomboBox, qcbxIndexChangedInt, this, &AVForm::onVideoModesIndexChanged);
 
     connect(bodyUI->filterAudio, &QCheckBox::toggled, this, &AVForm::onFilterAudioToggled);
-    connect(bodyUI->rescanButton, &QPushButton::clicked, this, [=](){getAudioInDevices(); getAudioOutDevices();});
+    connect(bodyUI->rescanButton, &QPushButton::clicked, this, [=](){getAudioInDevices(); getAudioOutDevices(); getVideoDevices();});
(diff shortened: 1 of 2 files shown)


ghost:
Maybe it's only a logging thing, but I only get feedback for the audio devices

> src/audio/audio.cpp:187 : Debug: stopping capture
> src/audio/audio.cpp:210 : Debug: Opening audio input  "Built-in Audio Analog Stereo"
> src/audio/audio.cpp:236 : Debug: "Opening audio output Built-in Audio Analog Stereo"
> AL lib: (WW) FreeContext: (0x12345aa) Deleting 1 Source(s)

Also, how about an extra empty line between the button an the video devices, to make it clearer that the button is not just for that group?

sudden6:
> Maybe it's only a logging thing, but I only get feedback for the audio devices

This is really a logging thing, as the current video device isn't reopened until you select it.

> Also, how about an extra empty line between the button an the video devices, to make it clearer that the button is not just for that group?

Honestly I think this button looks really ugly, but ui design isn't something I'm good at, so I only did minimal changes.

src/widget/form/settings/avform.cpp
```diff
@@ -63,7 +63,7 @@ AVForm::AVForm() :
     connect(bodyUI->videoModescomboBox, qcbxIndexChangedInt, this, &AVForm::onVideoModesIndexChanged);
 
     connect(bodyUI->filterAudio, &QCheckBox::toggled, this, &AVForm::onFilterAudioToggled);
-    connect(bodyUI->rescanButton, &QPushButton::clicked, this, [=](){getAudioInDevices(); getAudioOutDevices();});
+    connect(bodyUI->rescanButton, &QPushButton::clicked, this, [=](){getAudioInDevices(); getAudioOutDevices(); getVideoDevices();});
```
antis81: Can you please split the commands into separate lines (-> easier to read, debug etc.).

PR #2586, opened 2015-11-22
Fix behavior when clicking on the tray icon

### LINKED ISSUE #1542: Behavior of window when clicking tray/minimizing
Uncheck 'minimize to tray'
- click on tray icon should hide window (minimizes instead)
- click on minimize button should minimize windows

Same settings as above
- make qtox window active
- click on tray icon - qtox minimizes
- click on tray icon - qtox shows, minimizes and [shows again as active window](http://upload.agilob.net/qtox_minimize.webm)

tolikzinovyev:
This should fix #1542 and #1413
Also, please let me know what the expected behavior for the middle click is.
Thanks.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -757,7 +757,8 @@ void Widget::onIconClick(QSystemTrayIcon::ActivationReason reason)
     }
     else if (reason == QSystemTrayIcon::Unknown)
     {
-        forceShow();
+        if (isHidden()) 
+            forceShow();
     }
 }
 
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -738,31 +738,23 @@ void Widget::onIconClick(QSystemTrayIcon::ActivationReason reason)
     {
         case QSystemTrayIcon::Trigger:
         {
-            if (isHidden())
+            if (isHidden() || isMinimized())
             {
-                show();
-                activateWindow();
                 if (wasMaximized)
                     showMaximized();
                 else
                     showNormal();
+
+                activateWindow();
             }
-            else if (isMinimized())
+            else if (!isActiveWindow())
             {
-                forceShow();
                 activateWindow();
-                if (wasMaximized)
-                    showMaximized();
-                else
-                    showNormal();
             }
             else
             {
                 wasMaximized = isMaximized();
-                if (Settings::getInstance().getMinimizeToTray())
-                    hide();
-                else
-                    showMinimized();
+                hide();
             }
 
             break;
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -734,44 +734,30 @@ void Widget::confirmExecutableOpen(const QFileInfo &file)
 
 void Widget::onIconClick(QSystemTrayIcon::ActivationReason reason)
 {
-    switch (reason)
+    if (reason == QSystemTrayIcon::Trigger)
     {
-        case QSystemTrayIcon::Trigger:
+        if (isHidden() || isMinimized())
         {
-            if (isHidden() || isMinimized())
-            {
-                if (wasMaximized)
-                    showMaximized();
-                else
-                    showNormal();
-
-                activateWindow();
-            }
-            else if (!isActiveWindow())
-            {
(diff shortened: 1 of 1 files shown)


zetok:
> Also, please let me know what the expected behavior for the middle click is.

Well, it shouldn't do anything, AFAIK. (It doesn't do anything for me - does that behaviour differ on windows / osx?)

tolikzinovyev:
In arch linux with xfce it does what it says:

```
        case QSystemTrayIcon::MiddleClick:
            wasMaximized = isMaximized();
            if (Settings::getInstance().getMinimizeToTray())
                hide();
            else
                showMinimized();
            break;
```

What is your environment?

ghost:
Opening with left click works fine for me. Closing not so much:
- if qTox is maximised but not focused, the window goes to the front
- if qTox is minimised, the window is maximised
- if qTox is focused, the window is closed

The _normal_ (read _in most of my other programs_) behaviour is to close to tray in all of the described scenarios. This is specially true if you're using the `Multiple window mode`, where right now only the main window reacts to the tray icon. Or if you use "focus follows mouse" in your DE, you never know how many times you have to click the tray icon until the window is finally gone.

> expected behavior for the middle click

I think it usually does nothing, as zetok says. Banshee pauses the playback, but I can't think of any particular "special" action for qTox right now.

tolikzinovyev:
> if qTox is maximised but not focused, the window goes to the front
> if qTox is minimised, the window is maximised
> if qTox is focused, the window is closed

This is what skype and deluge do. We probably need more opinions on this.

ghost:
@ProMcTagonist you're the UX guy, right? What do you think about this?

ProMcTagonist:
@kehugter 

I'm by no means an expert.

In Xchat: 
clicking the tray icon toggles between focused and closed-to-tray, regardless of if the window is raised or not in the first place, or even if it's minimized. Controlling focus when the window isn't closed to tray is the job of your panel or mouse, not the tray icon.

In Deluge-gtk:
Clicking the tray icon focuses the window if it's open but not focused, focuses the window if it's minimized, closes to tray if it's focused, and focuses if it's closed to tray.

I'm initially inclined to go with Xchat's way because of the rationale provided there. Panel does panel things, tray toggles close to tray. It follows system logic.

But the deluge way might be more friendly. If a user isn't sure where the window is, because it's not visible, clicking on either the tray or panel icon will bring it to them. This might irritate users who already know how the system is supposed to work and want to send the window to the tray when it's not visible, costing them more clicks... But with an ease of use gain for an inexperienced user.

We conducted some informal surveys way back when and a popular opinion was that close-to-tray as a concept was extremely annoying for "typical" users (Skype on OEM Windows, etc). And close-to-tray is turned off by default in qTox, so it's likely that the people with it on understand the system logic.

As long as close-to-tray is off by default, I vote for the xchat method.

If close-to-tray is turned on by default, I'm leaning toward the deluge method, especially if skype already behaves like that. 

Further input should still be gathered.

ghost:
The current behaviour is almost a clone of what the task manager does and therefore not really useful IMO.

(disguised bump)

tux3:
This doesn't work on my Linux box, with this patch the middle click doesn't do anything with a minimized to tray, maximized, focused or unfocused qTox windows. It works for me on master.

ghost:
Right now is only for left click on tray, middle click action wasn't required by anyone yet.

tolikzinovyev:
yeah...

> Also, please let me know what the expected behavior for the middle click is.

tux3:
Alright, if everyone is ok with no middle click for now, then it's fine by me

tolikzinovyev:
Thanks for merging this. 

By the way, when compiling qtox with the unity backend enabled I had to use some patch I found here: https://build.tox.chat/view/Clients/job/patch-qtox-unity_src/lastSuccessfulBuild/artifact/qtox-unity-patch_src.patch/*view*/ . Why isn't it merged?

tux3:
@aszinovyev Since it's behind a #ifdef, I imagine most people don't run into this error, so you're the first person to actually report the problem.
Thanks, it's fixed now.

PR #2593, opened 2015-11-26
Select password field when it is wrong, Fixes #2584 using sudden6's idea

### LINKED ISSUE #2584: When entering a wrong password, the password field doesn't get emptied

TheNain38:
Fixes #2584 using @sudden6's idea

### DIFF
--- a/src/widget/loginscreen.cpp
+++ b/src/widget/loginscreen.cpp
@@ -214,6 +214,7 @@ void LoginScreen::onLogin()
         {
             QMessageBox::critical(this, tr("Couldn't load this profile"),
                                   tr("Wrong password."));
+            ui->loginPassword->selectAll();
             return;
         }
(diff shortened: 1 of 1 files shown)


ghost:
Works as expected, thanks!

zetok:
So, under i3 it works only when I use keyboard to load profile (enter) - if I use mouse, it doesn't work. I guess that's a thing specific to tiling window manager(s?). In any case, it seems like an upstream problem - is anyone else who doesn't use tiling wm able to reproduce that?

Edit: if it's not reproducible outside of tiling wm, IMHO it should be merged.

ghost:
As you said, is not reproducible. I tested keyboard and mouse in Plasma and Cinnamon.

zetok:
I see, thanks for testing :)

PR #2601, opened 2015-11-29
treat x-cinnamon the same as gnome fix  #2589

### LINKED ISSUE #2589: System tray backend on cinnamon 2.8.5
At least since version 2.8.5 Qt backend doesn't work well.
Passing XDG_CURRENT_DESKTOP="GNOME" triggers the use of the GTK backend and works correctly.

This should be the default in upcoming releases.

This page also needs updating: https://github.com/tux3/qTox/wiki/Tray-icon

sudden6:
fixes  #2589

### DIFF
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -60,7 +60,7 @@ SystemTrayIcon::SystemTrayIcon()
     }
     #endif
     #ifdef ENABLE_SYSTRAY_GTK_BACKEND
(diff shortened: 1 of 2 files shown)


ghost:
Can confirm that it doesn't break the tray icon in Cinnamon 2.6.13.

@BrunoMSantos could you check?

BrunoMSantos:
Yes I can, and I did. Works ok.
Now back to the workaround until this gets upstream.. what's the usual turnaround in qtox?

zetok:
So, there are plans to just get rid of GTK backend, but until that time comes, I think that it should be merged.

PR #2627, opened 2015-12-05
Small fix in translation file

ovalseven8:


tux3:
My german isn't very good, so I'll take your word for it that it's better.

Issue #2631, opened 2015-12-06
Playback and capture devices are set to 'None' by default


ovalseven8:
Hello,

Reproduce:
1. Download or compile latest qTox
2. Clear your ~/.config/tox directory so that the settings get reset to the default
3. Start qTox
=> Playback and capture devices are set to 'None'

Video device is also set to 'None' by default but that's okay for me and makes sense.

**Edit:** Playback device is done now -> https://github.com/tux3/qTox/pull/2633

antis81:
I'll have a look at this after #2509 got merged.

tux3:
I think this is fixed by #2633 actually.

ovalseven8:
@tux3: Well, capture device not.

JacobHenner:
@ovalseven8 I can fix that too, but I believe there were privacy implications of a default capture device mentioned on IRC. If there are no objections, I'll fix it and introduce a PR.

TheNain38:
@JacobHenner For capture device? Do you mean: if I disconnect my headphones then it would use my webcam? I don't really like that idea then... because skype is doing that and it's annoying... Sometimes I mute my microphone on hardware, and then I disconnect it, but then it uses my webcam and the sound isn't muted...

antis81:
@TheNain38 No, it's related to audio devices only (and to the qTox settings) - not to the webcam or any video devices at all. Even if your webcam has a microphone builtin, it's still a separate audio device in qTox. This is our beloved crazy modern world… :)

**EDIT:** "Capture device" == "microphone" :smile:

TheNain38:
Yes, I know this is a separate audio device, but when I disconnect my headphones I don't want my webcam's microphone to be used.

JacobHenner:
@antis81 @TheNain38 I think the idea was you might start a video or audio call, and the audio source automatically selected might not be the one you intended (i.e. perhaps a virtual audio device for your system's sound output instead of the microphone).

ovalseven8:
Somebody has broken it again...

JacobHenner:
I can investigate later.

antis81:
@JacobHenner Have a look at #2711 / #2743

ovalseven8:
P-high

antis81:
This is correct behavior by design. By default (e.g. when qTox is started for the first time), audio (and video) is simply disabled and has to be enabled in settings.

[...]

PR #2633, opened 2015-12-07
qtox will now attempt to default to the first audio output device

JacobHenner:


JacobHenner:
Fixes #2631 for playback devices. Leaving capture devices as-is (for now).

PR #2684, opened 2015-12-14
Update italian translation

ghost:


tux3:
The translation is good, but you're practically sending one commit per line changed!
Please try to regroup things a bit more next time so you don't spam the git history too much.

ghost:
Sorry, i will keep this advice in mind.

Issue #2686, opened 2015-12-15
Distinct profiles store avatars using same filename but different encryption key.


JacobHenner:
If the user is running multiple qTox instances on different profiles with different passwords, the avatars for mutual contacts will be stored using the same filename, which is generated in profile.cpp by `crypto_generichash((uint8_t*)hash.data(), hashSize, (uint8_t*)idData.data(), idData.size(), nullptr, 0);`. 

However, since each image is encrypted for the individual user, only one will be able to successfully open the avatar, and the other instances will emit `[22:11:49.024] src/core/coreencryption.cpp:88 : Warning: Decryption failed` and attempt to reobtain the image.

JacobHenner:
I can change the call which generates the hash to include a key based on the user, but that would invalidate all current avatars (and they would not be automatically deleted). Is that alright? @tux3

tux3:
> I can change the call which generates the hash to include a key based on the user, but that would invalidate all current avatars (and they would not be automatically deleted). Is that alright?

Sounds good to me.

PR #2687, opened 2015-12-15
Fixes #2629: avatars will be re-encrypted when the profile password c…

### LINKED ISSUE #2629: fix avatar issues with encrypted profiles
Works but the encrypted avatar is still there if you unencrypt the profile. Why not remove it?
Same with all the avatars if you change the password.

And if encrypting unencrypted profile there are still the unencrypted avatars.

JacobHenner:
…hanges. (Partial fix, see #2629 and #2686)

### DIFF
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -332,11 +332,14 @@ QString Profile::avatarPath(const QString &ownerId, bool forceUnencrypted)
         return Settings::getInstance().getSettingsDirPath() + "avatars/" + ownerId + ".png";
 
     QByteArray idData = ownerId.toUtf8();
-    constexpr int hashSize = TOX_PUBLIC_KEY_SIZE; // As long as an unencrypted hash
+    QByteArray pubkeyData = core->getSelfId().publicKey.toUtf8();
+    constexpr int hashSize = TOX_PUBLIC_KEY_SIZE;
     static_assert(hashSize >= crypto_generichash_BYTES_MIN
                   && hashSize <= crypto_generichash_BYTES_MAX, "Hash size not supported by libsodium");
+    static_assert(hashSize >= crypto_generichash_KEYBYTES_MIN
+                  && hashSize <= crypto_generichash_KEYBYTES_MAX, "Key size not supported by libsodium");
     QByteArray hash(hashSize, 0);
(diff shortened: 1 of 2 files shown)


src/persistence/profile.cpp
```diff
@@ -520,12 +525,19 @@ void Profile::setPassword(QString newPassword)
...
     HistoryKeeper::getInstance()->importMessages(oldMessages);
     Nexus::getDesktopGUI()->reloadHistory();
     saveAvatar(avatar, core->getSelfId().publicKey);
+
+    QVector<uint32_t> friendList = core->getFriendList();
+    QVectorIterator<uint32_t> i(friendList);
+    while (i.hasNext())
+    {
+        QString friendPublicKey = core->getFriendPublicKey(i.next());
+        saveAvatar(loadAvatarData(friendPublicKey,oldPassword),friendPublicKey);
```
tux3, line 544: If we previously had an empty password (unencrypted account), and we just set one, loadAvatarData will call avatarPath, which will see that we have a password set and return the path for an encrypted avatar, when we want to load the old unencrypted avatar here.

You could for example replace the "forceUnencrypted" parameter of avatarPath with an optional password parameter, and fix loadAvatarData to use it

If you touch avatarPath, it might be a good occasion to fix #2686 at the same time by hashing with the profile password as a salt.

JacobHenner, line 544: @tux3 If loadAvatarData calls avatarPath with an empty password (the old password should be empty), wouldn't it still return the path to the unencrypted avatars?

tux3, line 544: Currently loadAvatarData calls avatarPath without giving it a password, so it uses the profile password, and this happens _after_ setting the new profile password.
So if we used to have no password and we just set one, instead of loading the old unencrypted avatars avatarPath will see the new profile password and return the encrypted path.

JacobHenner, line 544: @tux3  So I just looked this over again, and when loadAvatarData is called, it is called with the password (even if that password was formerly empty). When this happens, loadAvatarData will call avatarPath once, which will generate the new path, but right afterwards, loadAvatarData checks if the password (the one which was passed) is empty. If it is, the path is changed to the result of the non-hashed avatarPath. Have I missed something?

tux3, line 544: Okay, I tried it again and it works fine, so I must have dreamed it, sorry about that

tux3:
I rebased it on master to fix conflicts and merged it manually. Thanks.

PR #2689, opened 2015-12-15
Support for theming qTox tray and window icons

ghost:
This patch adds ability to override qTox window and tray icons with ones from user's desktop theme. Some people prefer keeping tray icons consistent to achieve best looks.

Following theme icons used:
- qtox: general window icon which is also used in taskbar
- qtox-online, qtox-offline, qtox-busy, qtox-away, qtox-invisible, qtox-event: tray icons

If theme icon is not available then default built-in icon will be used.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -116,6 +116,10 @@ void Widget::init()
 {
     ui->setupUi(this);
 
+    QIcon themeIcon = QIcon::fromTheme("qtox");
+    if (!themeIcon.isNull())
+        setWindowIcon(themeIcon);
+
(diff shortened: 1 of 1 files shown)


tux3:
Why not.

ghost:
What should be the icon name corresponding to `taskbar_event_small.svg` in https://github.com/qTox/qTox/tree/master/img/taskbar/dark?
EDIT: Sorry, wrong place to ask. Should I raise a new issue?

zetok:
It's not actually used – see https://github.com/qTox/qTox/pull/2276#issuecomment-143006081

PR #2701, opened 2015-12-19
Smiley changes

ovalseven8:
- Rename 'cylgom' -> 'Basic'
- Rename 'krepa098' -> 'Classic'
- Rename 'TwitterEmojiSimple' -> 'ASCII-Universe'
- Rename 'TwitterEmojiSVG' -> 'Universe'
- Minor improvements

PR #2708, opened 2015-12-20
Change smiley character order

ovalseven8:
Removed some unicode characters that aren't displayed correctly for many users in the 'Basic' smiley pack. Goal is that at least in the 'Classic' and 'Basic' smiley packs there aren't any display problems.

**Edit:** Now they aren't removed but in another order.


Schlumpf:
:-1: for removing.
As I said in #2701, changing the order should solve the problem of wrong display in the input field on some systems. If you _remove_ the glyphs, all already written unicode character(e.g. `💣`) will not rendered anymore. Any if you are on a "problematic" system, you won't see them in the chat window, too.

ovalseven8:
@Schlumpf: OK, done. :)

ovalseven8:
@Schlumpf: Thanks, forgot to delete some lines. Should be fixed now.

zetok:
From what I understand, it's not the order that is a problem, but a missing font.

I wonder how feasible would be bundling with qTox font `Symbola`, which has support for all needed glyphs.

PR #2713, opened 2015-12-20
additional cleanup for audio

antis81:
@JacobHenner Now we also correctly log the default audio input/output device. :)


antis81:
Merged manually.

Issue #2705, opened 2015-12-20
Audio feedback widget doesn't work properly


ovalseven8:
That's all. Nothing changes:
![2015-12-20-125735_1366x768_scrot](https://cloud.githubusercontent.com/assets/8258609/11917848/4e7fd690-a719-11e5-8603-ede65622dd35.png)

tux3:
@antis81

antis81:
Thanks for reminding me of that not yet solved bug. :) You should see some error messages in the log as well.

When leaving and re-opening the A/V-Settings tab, it should work again, right?

ovalseven8:
@antis81:
Yeah, sometimes it is working if I reopen the A/V settings but not everytime.
**And: If you reopen the A/V settings a few times, qTox is crashing!**

ovalseven8:
Backtrace concerning the crash:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff99386700 (LWP 9688)]
0x00007ffff1f539f5 in QMutex::lock() ()
   from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
(gdb) thread apply all backtrace

Thread 15 (Thread 0x7fffd6ff3700 (LWP 9484)):
#0  0x00007ffff13b48dd in poll () at ../sysdeps/unix/syscall-template.S:81
#1  0x00007fffeebdbb72 in ?? () from /usr/lib/x86_64-linux-gnu/libxcb.so.1
#2  0x00007fffeebdd64f in xcb_wait_for_event ()
   from /usr/lib/x86_64-linux-gnu/libxcb.so.1
#3  0x00007fffd8953099 in ?? ()
   from /usr/lib/x86_64-linux-gnu/qt5/plugins/platforms/libqxcb.so
#4  0x00007ffff1f5bb0e in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#5  0x00007ffff1ca16aa in start_thread (arg=0x7fffd6ff3700)
    at pthread_create.c:333
#6  0x00007ffff13bfeed in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109

Thread 14 (Thread 0x7fffd4f3f700 (LWP 9486)):
#0  0x00007ffff13b48dd in poll () at ../sysdeps/unix/syscall-template.S:81
[...]

Issue #2703, opened 2015-12-20
New audio: When going in the audio/video settings 1/2 times, the audio feedback doesn't work


TheNain38:
The first time you go in the audio/video settings, the feedback works, when you click on another setting and go back to audio/video settings the feedback doesn't work anymore. And if you repeat this step, it works again...

tux3:
@antis81

antis81:
Yep -> relates to #2705

antis81:
Closing, as it is actually a duplicate of #2705.

PR #2719, opened 2015-12-21
Fix audio output subscriptions

### LINKED ISSUE #2706: Audio device is open after having a call
You see this on windows. I don't know which device it is. @antis81

antis81:
fix #2706

@agilob @tux3 @JacobHenner @TheNain38 @towlie (and all audio users on qTox…) Made a pretty silly little mistake in `Audio::unsubscribeOutput` (SID was invalidated **before** removing from the list) -> fixed! Please help  with **related** review and feedback. This should not leave any dangling audio output devices.

On Linux (probably OSX) we get a warning at the end of the logfile, when an audio device remains open reading `AL lib: ReleaseALC: 1 device not closed`. Not sure, but likely it is closed anyways.

Btw.: In Linux this is working with ALSA (using that by myself :)) and PulseAudio.

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -760,6 +760,8 @@ void Audio::unsubscribeOutput(SID& sid)
 {
     QMutexLocker locker(&d->audioLock);
 
+    d->outSources.removeAll(sid);
+
     if (sid) {
         if (alIsSource(sid)) {
             alDeleteSources(1, &sid);
@@ -771,8 +773,6 @@ void Audio::unsubscribeOutput(SID& sid)
 
         sid = 0;
     }
-
-    d->outSources.removeAll(sid);
 
     if (d->outSources.isEmpty())
         d->cleanupOutput();
--- a/src/core/toxcall.cpp
+++ b/src/core/toxcall.cpp
@@ -49,7 +49,6 @@ ToxCall::ToxCall(ToxCall&& other) noexcept
     // required -> ownership of audio input is moved to new instance
     Audio& audio = Audio::getInstance();
     audio.subscribeInput();
-    audio.subscribeOutput(alSource);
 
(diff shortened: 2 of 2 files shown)


towlie:
Yes, that's it. Works even with 2 instances on 1 pc. Thank you very much.

towlie:
Oh there is one thing. Closing the audio/video settings doesn't close the audio device. Until you cancel a call.

towlie:
For me it looks like you only free one device when I look into process explorer.
When I open the settings tab then you create 2 "ALPC Ports" and when I close it then only one of them gets removed.

In this example it closed it 3 times https://gist.github.com/towlie/e5ce49be18388beb9130
It even contains a warning.
This warning comes when I switch between the audio/video tab and another tab fast.
I think this was introduced with one of your latest PRs (It didn't happen before; using windows)

antis81:
Thanks for feedback.

> Oh there is one thing. Closing the audio/video settings doesn't close the audio device. Until you cancel a call.

The a/v settings tab only subscribes to audio input (to meter the mic gain). Unfortunately this is still done in show/hide events, which are called asynchronously. It's impossible to make that completely stable and code has to move to constructor/destructor. This implies some more changes and I stick with this workaround for now. 

> For me it looks like you only free one device when I look into process explorer.

That's exactly how it's supposed to be. :)

towlie:
Well yeah I could live with that. I rarely open that tab. What about that warning?

```
[01:01:06.612] :0 : Warning: QWaitCondition: Destroyed while threads are still waiting
```

antis81:
> What about that warning?

Well, that's yet to solve in another PR: see #2705 and #2703.

antis81:
@tux3 I'd like to merge this PR tomorrow to not cause much of a delay.

antis81:
Merged manually.

PR #2721, opened 2015-12-22
Disable tray icon is no is available to prevent looking for systray next time

agilob:
There is a bug in KDE latest version that does not create system tray immediately, qtox tries to load to system tray, fails and crashes after several seconds. The same happens after restart. This fix prevent loading to tray again after is was not available before.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -770,7 +770,7 @@ void Widget::onIconClick(QSystemTrayIcon::ActivationReason reason)
     }
     else if (reason == QSystemTrayIcon::Unknown)
     {
-        if (isHidden()) 
+        if (isHidden())
             forceShow();
     }
 }
@@ -1702,7 +1702,8 @@ void Widget::onTryCreateTrayIcon()
         disconnect(timer, &QTimer::timeout, this, &Widget::onTryCreateTrayIcon);
         if (!icon)
         {
-            qWarning() << "No system tray detected!";
+            Settings::getInstance().setShowSystemTray(false);
+            qWarning() << "No system tray detected, disabling system tray!";
             show();
         }
     }


agilob:
lol my commit message, I dont know what does it even mean. I think I wanted to say:

disable tray icon after tray icon was not detected to prevent detecting it next time

zetok:
Could you `git commit --amend -am "disable tray icon after tray icon was not detected to prevent detecting it next time"` ?

Also, while at rewriting stuff, could you add the PR description as a comment into commit?

src/widget/widget.cpp
```diff
@@ -770,7 +770,7 @@ void Widget::onIconClick(QSystemTrayIcon::ActivationReason reason)
     }
     else if (reason == QSystemTrayIcon::Unknown)
     {
-        if (isHidden()) 
+        if (isHidden())
             forceShow();
```
antis81, line 774: Indentation :)

agilob:
@zetok thanks ;)

agilob:
Bump?

zetok:
I think that user should be notified a little bit more promptly about behaviour that alters settings, even if it's "for their own good". I.e. some pop-up window with that info on start?

tux3:
I don't think this is a good idea. If there's some sort of fluke with the system tray, the user won't know why the settings changed themselves!
There's not really any harm in leaving the option on, so this isn't really needed.

agilob:
If user has no tray and trayicon is enabled, then there is no way to turn on qtox, it minimizes after launch, hides to tray, looks like it crashed because there is no window or any popup. Then I tried to turn it on again, and it said profile is locked...

tux3:
Then that's a different bug, it shouldn't stay minimized if it doesn't find a tray. But I don't think that's a reason to change the settings the user set.

PR #2725, opened 2015-12-22
Cleaned up all the ipc/startup code

ghost:
- Removed duplicate code of ipc event sending
- Removed duplicate code of profile loading
- Fixed bug where activating existing instance still started new instance of qtox asking to select a profile
- IPC messages are now profile-aware and are sent to instance that runs specified profile in -p flag or to ipc owner if -p is not specified.
- Specifying -p implies autologin

If -p flag is specified ipc events will be send to instance which runs specified profile. If instance using that profile does not run - new qTox instance will be started. This works with password-protected profiles too - new instance will handle "uri" or "save" events after accepting user password.

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -186,7 +186,8 @@ bool IPC::waitUntilAccepted(time_t postTime, int32_t timeout/*=-1*/)
 {
     bool result = false;
     time_t start = time(0);
-    forever {
+    forever
+    {
         result = isEventAccepted(postTime);
         if (result || (timeout > 0 && difftime(time(0), start) >= timeout))
             break;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -157,104 +157,84 @@ int main(int argc, char *argv[])
 #endif
 
 
+    QString profileName;
+    bool autoLogin = Settings::getInstance().getAutoLogin();
 #ifndef Q_OS_ANDROID
     // Inter-process communication
     ipc.registerEventHandler("uri", &toxURIEventHandler);
     ipc.registerEventHandler("save", &toxSaveEventHandler);
     ipc.registerEventHandler("activate", &toxActivateEventHandler);
 
+    uint32_t ipcDest = 0;
+    QString eventType, firstParam;
     if (parser.isSet("p"))
     {
-        QString profileName = parser.value("p");
-        if (Profile::exists(profileName))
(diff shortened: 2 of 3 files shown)


suhr:
> Specifying -p implies autologin

Why?

ghost:
Because if you select profile explicitly you obviously want to use that profile and not select another one.

EDIT:
Travis fails finding sqlcipher headers - not my fault o7

suhr:
> Because if you select profile explicitly you obviously want to use that profile and not select another one.

Not really. You usually have one main profile and some secondary. And you don't want to change your main profile to load a secondary one.
If you really want to change active profile you usually just exit then load an another one.

ghost:
You misunderstand purpose of -p flag. Its just like -P of firefox for instance.

suhr:
Ok, can you provide an option that does _not_ change the current profile?

antis81:
> Ok, can you provide an option that does not change the current profile?

Actually it is more like: Start qTox with the latest active profile. We could allow an empty attribute value to `-p`

@r-ku 

> Specifying -p implies autologin

That's a bad idea. Instead, one should be required to enter password to load the requested profile (just like with `ssh foo-user@bar-target` or `ssh -l foo-user bar-target`). In fact the meaning is `autologin == "remove password"` -> We should not allow that. @tux3 ??

ghost:
@antis81 starting qtox without `-p` starts last active profile. Point of `-p` is to pre-select that profile. There is no point in specifying `-p` without profile name as it is default behaviour already.

```
That's a bad idea. Instead, one should be required to enter password to load the requested profile
```

Obviously. And that is what happens if password is set. Password is used to encrypt profile remember? It is impossible to load profile without password if such password is set. Dont be confused. Autologin is implied in a sense that:
- If password is not set specified profile will be loaded without showing profile selection window. We selected profile with `-p` flag after all.
- If password is set it will select profile specified by `-p` in profile selection window. It still asks to enter password though because skipping this step is impossible.

I thought i clearly stated in my first post it works in this way. Apparently not, sorry about that.

antis81:
@r-ku I don't see a reason to hold this back any longer. Ready to merge?

ghost:
Yes the way i see it its complete. I dont want to add anything more.

tux3:
Looks good to me.

PR #2736, opened 2015-12-24
Fix default smiley pack. Necessary because of the name changes.

ovalseven8:


zetok:
@antis81 uh, this merge wasn't signed

antis81:
Hm, that's right, thanks. Probably because the commit was not made by me… :)

@ovalseven8
Please sign your qTox commits with PGP signature (https://git-scm.com/book/uz/v2/Git-Tools-Signing-Your-Work#Signing-Commits). Some help with that:

``` bash
cd /path/to/qtox-project-dir
git config --global user.signingkey FFFFFFFF  ## <=== insert your GPG key here as described in Git
git config --local commit.gpgsign true
```

Depending on your distro/OS, you may be required to install `gnupg-agent` (graphical passphrase-dialog).

ovalseven8:
A merge commit is missing, isn't it?

zetok:
> A merge commit is missing, isn't it?

Indeed.

antis81:
No, not really. I could have signed off (by using the --sign-off attribute), but still the commit wouldn't be signed by the original author, which is @ovalseven8. So please pgp-sign your commits by following the instructions above. Producing a merge commit is absolutely the same thing and every single commit within the merge will not be signed.

Actually we should introduce a push hook on Github to verify, that all commits are signed. Otherwise these incidents will happen and we live with that.

zetok:
@antis81 eh. It matters whether all stuff that gets merged gets signed, and not really whether every single commit in the merge gets signed. That's why people with write access to the repo should sign everything, and don't do `merge --ff`, while PR submitters don't have to sign anything.

And yes, **merge commit is missing**.

> I could have signed off (by using the --sign-off attribute)

Err, `--signoff` doesn't produce GPG signature, nor does it constitute merge commit.

You probably would want to adjust in your git config `merge.ff`, and set it to `false`.

In regard to signing things:
https://github.com/zetok/gentoo-overlay-tox/issues/49#issuecomment-55525109
↓

```
git show -q --pretty="format:%h %an %G?" $(git rev-list --first-parent 22ddef48b192e80f7159ea78e65aa606e914c950..f70b249ed8c101fbaf056336153bdf94bfcd2d29) | grep '.* [NBU]$'
```

PR #2741, opened 2015-12-28
add button to show qrcode of toxid of friends in show details view

dardevelin:
this makes it easier to give away someone's contact.
however this implementation still suffers from the fact that clients
don't get notified if the owner changes the toxid and therefor it
may be providing useless toxids through the qrcode. This should
be possible to fix by adding an extension to the toxcore to allow clients
to request the current toxid from the source akin to the irc CTCP requests.

### DIFF
--- a/src/widget/about/aboutuser.cpp
+++ b/src/widget/about/aboutuser.cpp
@@ -8,6 +8,8 @@
 #include <QFileDialog>
 #include <QMessageBox>
 
+#include "../qrwidget.h"
+
 AboutUser::AboutUser(ToxId &toxId, QWidget *parent) :
     QDialog(parent),
     ui(new Ui::AboutUser)
@@ -20,6 +22,7 @@ AboutUser::AboutUser(ToxId &toxId, QWidget *parent) :
     connect(ui->autoaccept, &QCheckBox::clicked, this, &AboutUser::onAutoAcceptClicked);
     connect(ui->selectSaveDir, &QPushButton::clicked, this,  &AboutUser::onSelectDirClicked);
     connect(ui->removeHistory, &QPushButton::clicked, this, &AboutUser::onRemoveHistoryClicked);
+    connect(ui->showQRCode, &QPushButton::clicked, this, &AboutUser::onShowQRCodeClicked);
 
     this->toxId = toxId;
     QString dir = Settings::getInstance().getAutoAcceptDir(this->toxId);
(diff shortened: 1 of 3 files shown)


ProMcTagonist:
Very nice. Is there a reason the user has to click a button to display the code?

> extension to the toxcore

Would be cool. How is that coming along, @subliun?

ovalseven8:
Well, I don't think this is needed.
There's a reason why the friends doesn't get the current Tox ID from the user. It's like a privacy feature and makes the NoSpam feature more reliable.

If a user really needs to know the full Tox ID, he just can ask the other user for it.

agilob:
I think it would be useful, but only if we had toxid, it's useless for public key.

dardevelin:
@ProMcTagonist No needed, just didn't feel like making the view details even bigger for an info that is only needed sometimes. I also though that by adding a button it would simplify if toxcore never landed the feature to make the clients implement this as a client feature if need be. Like i mentioned earlier, IRC does this very gracefully. CTCP VERSION and you get the version (if the client supports this feature). One could do similarly for ToxID.

@ovalseven8 well to be fair if you already accepted me. I am no longer your issue. This also would allow CTCP like functions to be able to 'black list' thus allowing for blocked users. 
Friends that are no longer such.

@agilob currently it uses the available ToxId, not the publicKey. If toxcore lands quickly enough with backend support for this, the button just needs to fetch using whatever toxcore does.
Otherwise the client can just implement CTCP like functionality. IRC has solved this issue and it doesn't seem to be too much of an issue to be honest.

I am not that familiar with the core, but if need be. I'll implement this in core. Even though I think CTCP client side would be a simple easy way to deal with this. if IRC clients can do it, I don't think it is an issue for tox clients.

agilob:
Yes, I think using public key instead full toxid introduces nothing more than just a confusion, in user profile you see Qr code to toxid, not public key, and that makes sense.

tux3:
Well, this is just like the "Copy Tox ID" feature we used to have. It's a great idea in principle, but the problem is that without toxcore support it's just not reliable, and it's going to cause more confusion than anything.

Since toxcore is probably not going to add support in the short term, and since we have no plan for protocol extensions at the moment, I'm going to close this for now. 
When there's a solid solution for getting a friend's Tox ID or for introducing one friend to another friend, it would make sense to resend this.

PR #2013, opened 2015-07-17
Chatform improvements

TheSpiritXIII:
I've made a bunch of changes. Listen up!

Changes I’ve made:
- You can now adjust the left and right columns of the chat form by dragging them. This will surely help when text size is able to be changed. Helps with #1288 #1974 
- Hovering over text on the chat form gives the I-beam cursor.
- You can now search and find text in the chat form using Ctrl+F.
- Whitespace bug is now fixed.
- Option to show group peer list on the side instead of on top. #677
- Chats always load with some history no matter how far back it was (previously, it would load history up to 7 days back). History is loaded by amount instead of days so it never loads too much or too little.
- History is loaded dynamically when scrolling up.
- The date of the messages you’re seeing are always shown on top. You don’t have to scroll down or up to see which day you’re looking at.
- Whenever the day changes, you will see a new message about it on the chat form when you receive your next message (previously, you would have to restart, which sounded like a bug).
- Fix missing psuedo Arabic characters. #1887 
- Notification when a new message has been made while you scrolled up. #1905

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/2013)

Issue #1974, opened 2015-07-08
Resize username column


ovalseven8:
Hello,

only a suggestion:
For some people it would be great, I think, if they could resize the username column. Especially, if people in Groupchats are using long usernames, it really could make sense if there would be more place for the names.

Screenshot: http://oi58.tinypic.com/984pba.jpg
(I shaded the column I am speaking about in yellow)

TheSpiritXIII:
It's already possible in my chatform branch. This will come soon.

Diadlo:
Seems, duplicate of #1190, so it will be closed. Please comment if it isn't

Issue #1905, opened 2015-06-24
New Messages indicator when you're scrolled up in a chat and get new messages in that chat


ProMcTagonist:
Like https://github.com/tux3/qTox/pull/1843 but in the chat form, just above the message box. Facebook does it, for reference, but it's hard to find a screenshot.

Hasshu:
Also, clicking on the indicator should scroll the chat to the first unread message, methinks.

Issue #1887, opened 2015-06-18
Arabic/Perso-Arabic text coming up blank


ghost:
I'm sending myself messages in perso-arabic, but none of the text shows up in qtox. Here's a screenshot:
![untitled](https://cloud.githubusercontent.com/assets/3895638/8230548/20158ab4-1576-11e5-9c32-b308345e4569.png)

the text that should be displayed is:
ثیس اس تِکست
(it's nonsense, btw)

zetok:
yep, doesn't work for me too on ce2c8309a28dc49ec005610791062ee55633ef80

Doneganai7:
I find that Esperanto accents work perfectly, and so do Japanese Kanji, but Kana show up as blank boxes.

I am using qTox on a 64 bit Linux Mint machine. I have also confirmed that while kana does not show properly on my computer, it does show correctly on Windows and Mac computers.

![qtox-problem](https://cloud.githubusercontent.com/assets/8411025/8638995/b7e45d76-2899-11e5-84e5-14502a41061c.png)

The Japanese with blank kana boxes should read これは日本語です。

Chiitoo:
The ثیس اس تِکست at first didn't come out here either, but I noticed that it might have something to do with being aligned to the right; right-to-left (just a guess without looking into anything deeper).

It will show up for example if either of the two conditions shown below are met:

![qtox_font_issue_1887_01](https://cloud.githubusercontent.com/assets/9367616/8655215/712f8bbc-2998-11e5-81a9-6bbae5b486bc.png)

After 'asd' the text will show up (left-to-right).  After that, there are a couple of lines of blank again (test-input), but once there is enough to new line(?), it will show up again.

In fact, it will show up with just the ثیس اس تِکست as long as the line becomes long enough (width of chat window seems to affect it somewhat).

As for Katakana and Hiragana, those show up fine here:

![qtox_font_issue_1887_02](https://cloud.githubusercontent.com/assets/9367616/8655216/7137789a-2998-11e5-9a52-85569bdfc069.png)

That might be a different issue (could qTox be using a weird font there for some reason?).

I'm using Gentoo Linux with all sorts of fonts and CJK support.

qTox was built is from: https://github.com/TheSpiritXIII/qTox/commit/4c040cacfc6b63d17b4c3cc57213a0b510cca707

(Not as a patch, but directly from that branch.)

zetok:
From #2315 thanks to @shake031:

# 

Issue:
[...]

PR #2744, opened 2015-12-28
Chatform improvements

antis81:
# :exclamation: For reference only! Don't merge this!

**Copied description from original PR #2013 by @TheSpiritXIII** 

I've made a bunch of changes. Listen up!

Changes I’ve made:
- You can now adjust the left and right columns of the chat form by dragging them. This will surely help when text size is able to be changed. Helps with #1288 #1974 
- **moved to #2752:** ~~Hovering over text on the chat form gives the I-beam cursor.~~
- **moved to #2751:** ~~You can now search and find text in the chat form using Ctrl+F.~~
- Whitespace bug is now fixed.
- Option to show group peer list on the side instead of on top. #677
- Chats always load with some history no matter how far back it was (previously, it would load history up to 7 days back). History is loaded by amount instead of days so it never loads too much or too little.
- History is loaded dynamically when scrolling up.
- The date of the messages you’re seeing are always shown on top. You don’t have to scroll down or up to see which day you’re looking at.
- Whenever the day changes, you will see a new message about it on the chat form when you receive your next message (previously, you would have to restart, which sounded like a bug).
- Fix missing psuedo Arabic characters. #1887 
- Notification when a new message has been made while you scrolled up. #1905

### DIFF
--- a/src/widget/form/groupchatform.h
+++ b/src/widget/form/groupchatform.h
@@ -28,6 +28,9 @@ class Group;
 class TabCompleter;
 class FlowLayout;
 class QTimer;
+class QSplitter;
+class QStringListModel;
+class QSortFilterProxyModel;
 
 class GroupChatForm : public GenericChatForm
 {
@@ -38,6 +41,9 @@ public:
 
     void onUserListChanged();
     void peerAudioPlaying(int peer);
+    void resetLayout();
+
+    virtual void show(Ui::MainWindow &ui);
 
 signals:
     void groupTitleChanged(int groupnum, const QString& name);
@@ -63,8 +69,10 @@ private:
     Group* group;
     QList<QLabel*> peerLabels; // maps peernumbers to the QLabels in namesListLayout
     QMap<int, QTimer*> peerAudioTimers; // timeout = peer stopped sending audio
-    FlowLayout* namesListLayout;
+    FlowLayout* namesListLayout = nullptr;
     QLabel *nusersLabel;
+    QSplitter* mainSplitter = nullptr;
+    QStringListModel* stringListModel = nullptr;
     TabCompleter* tabber;
     bool inCall;
     QString correctNames(QString& name);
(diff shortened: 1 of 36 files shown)


ovalseven8:
For the sake of completeness, here the open issues from the old PR:

---
- [ ] Font is broken (Windows) -> https://github.com/tux3/qTox/pull/2013#issuecomment-122373728
- [ ] Highlight the own user in peer list (maybe bold?)
- [ ] Another bug:
  - If you're in a groupchat the peer list is sorted alphabetically
  - Go to the Tox settings and deactivate 'Show group peer list on the side'
  - Activate it again
  -  > Peer list isn't sorted anymore
- [ ] Searching peers is not mature yet -> https://github.com/tux3/qTox/pull/2013#issuecomment-134936714
- [ ] Remove top view of peers in groupchats
- [ ] Review code and fix small code style issues

antis81:
Thanks @ovalseven8.

@agilob I guess this will become another rewrite and a whole bunch of fun. :smile:

src/widget/form/groupchatform.h
```diff
@@ -38,6 +41,9 @@ class GroupChatForm : public GenericChatForm
 
     void onUserListChanged();
     void peerAudioPlaying(int peer);
+    void resetLayout();
+
+    virtual void show(Ui::MainWindow &ui);
```
antis81: @agilob Here's the line where it crashes, but I don't get yet why… When looking into implementation you'll find this is complete crap and the first thing we need to clean up.

antis81:
@agilob Ok, made that compile again. Anyway the code requires a complete overhaul and is in bad shape.

ghost:
Is this testeable already? If yes, should it be locally merged to master first or doest it work on its own?

agilob:
Just clone that branch and compile, it works.

ghost:
:tada: Count me in for testing!

---

> the code requires a complete overhaul and is in bad shape

From a quick run, there seem to be more bugs than a PR thread could conveniently handle. What do you think about using your issue tracker to work on branch, @antis81?

antis81:
@kehugter Thanks! I think splitting this up into smaller chunks, which we can handle and review is the way to go.

agilob:
We agreed with @antis81 to close it and move commit to smaller PRs, some are already awaiting.

Issue #2377, opened 2015-10-11
Should be able to load chat history by date range, and search through whole history (not loaded).


dimgel:
Hi. Currently qTox looks all great except history view.

If I don't remember the day (or month, or even year) conversation happend on, I would never find it - loading history day by day is hell. Skype/linux, for example, can load history for last day/week/month/year or whole into memory, but it would be more convenient to allow user specify both boundary dates of loaded range. For example, if whole history is too large to fit into RAM but I need to find some ancient conversation, I may load it year by year to look through.

There should be ability to search through loaded history (case-insensitively, or with case sensivity option). Also, it would be absolutely fantastic to be able to search through whole history (or date range) without loading it into RAM.

zetok:
Searching loaded history is implemented in #2013.

Thus, issue is about:
- ability to load history by date range
- ability to search through whole history

dimgel:
Weird, nothing happens for me when I press Ctrl+F in chat window (non-empty, containing some messages) or anywhere. Also didn't find any clue about it; there should be some button or context menu item.

gentoo/amd64

zetok:
> Weird, nothing happens for me when I press Ctrl+F

Well, #2013 is not merged yet

dimgel:
Ah, sorry. :)

sudden6:
Search is implemented, second issue moved to:  #5501

PR #2751, opened 2015-12-30
[WIP] Chat: add text search

antis81:
implement #2377

@kehugter @agilob This PR serves as a review basis and is the first extraction of PR #2744 (I will abandon that one soon).

:warning: **Unstable**: Does crash as soon as `CTRL+F` is hit (Thus WIP).

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -500,6 +500,8 @@ SOURCES += \
     src/video/cameradevice.cpp \
     src/video/camerasource.cpp \
     src/video/corevideosource.cpp \
+    src/video/genericnetcamview.cpp \
+    src/video/groupnetcamview.cpp \
     src/core/toxid.cpp \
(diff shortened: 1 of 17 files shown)


ghost:
Crashes indeed.

It wasn't the case in #2744. There the search field was opened, you could search (the focus jumped to the chat input area almost immediately each time) and the results were shown. There was no way to close the search area.

tux3:
It's OK to send some WIP PRs that need some testing, but it looks like this doesn't work at all, and it has been inactive for over two weeks.
Please resend when it's a bit more developed, so we don't have too many inactive broken PRs sitting around for weeks.

antis81:
@tux3 Yes, this is good decision. Maybe i'm going to reimplement this in another PR.

tux3:
@antis81 Alright, looking forward to it

PR #2764, opened 2016-01-02
Fix travis build

agilob:
Trying to fix travis...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
 
 install:
 # install needed Qt stuff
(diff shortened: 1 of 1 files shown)


agilob:
(╯° °）╯︵ ┻━┻) ?? [sqlcipher?](http://packages.ubuntu.com/search?keywords=sqlcipher)

agilob:
It's not sqlcipher, libsqlcipher-dev and libsqlcipher0

zetok:
yeah, it's not available under 12.04.

Aside from that, there's a problem with building openal.

I think that it would be simpler to just remove travis altogether..

PR #2767, opened 2016-01-03
Set status icon on separate chat window

agilob:
closes #2210 
screenshot_20160103_122950

PR #2769, opened 2016-01-03
Set window icons on settings window

agilob:
Sets window icon for profile, settings etc when separate windows mode is enabled

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -1094,7 +1094,7 @@ QSplitter:handle{
              <x>0</x>
              <y>0</y>
              <width>775</width>
-             <height>284</height>
+             <height>279</height>
             </rect>
            </property>
            <layout class="QVBoxLayout" name="verticalLayout_5"/>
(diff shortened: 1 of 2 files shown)


ghost:
> settings

Works

> profile, settings etc

They all show the gear icon. Is that the expected behaviour? The `Add friend` window could display the `+` icon, the `File transfers` one the 2-arrows icon, etc.

> etc

No icon in chat window, unless that wasn't included in the changes.
Edit: probably not, saw #2767 after this comment.

---

Debian testing
Cinnamon 2.6.13
Qt 5.5.1

agilob:
Yes, it's expected. I think it would be "bad" to have 10 different icons for one application, but different windows. 

> No icon in chat window, unless that wasn't included in the changes.

It's in another PR -.-'

tux3:
Well, I think it would make more sense to always display the qTox icon.
Applications typically don't change their icon when you go in the settings, it would make it harder to find qTox in my taskbar if the icon is sometimes just a gear. If all applications did that my taskbar would be a mess!

agilob:
Done.

agilob:
Thanks

PR #2772, opened 2016-01-03
Remove useless travis

zetok:
Since it's useless anyway, there's no point in keeping it.

@RowenStipe, you've said something about trying to get it to work, are there any chances of succeeding?


RowenStipe:
I've ultimately come to the conclusion that setting up the Travis up on a docker image would be the best route to go. Every thing I've tried with just Travis will not work but does not look impossible.

Edit: I also plan on setting up the Travis to work with OSX testing but that's future plans.

zetok:
If it requires docker, then I think that it's a wrong way to go..

PR #2793, opened 2016-01-11
Remove empty src/main.h (closes #2792)

### LINKED ISSUE #2792: Why is there an empty file called "main.h"?
I have noticed in the source that the "main.h" file is empty, why is this?

zetok:
Looks like it was accidentally introduced in
99050424341bdbb8ef0fe218b5c78b578db73189


tux3:
Thanks, my bad.

Issue #2806, opened 2016-01-16
Sound gone after update (16.01.2015)


mastir:
Sound freezes, after audio settings open. If i switch setting tabs, microphone noise freezes and sound gone, forever (almost)

way to reproduce: open settings->audio, open any user or settings tab, return to settings->audio.
how to fix: close settings, close qtox window (to tray) open qtox, sound works again.

ubuntu 14.04, qtox-unity
now i've got qtox commit: c576a1485a6d0a1fed8911f1341709bff08bd96c

tux3:
@antis81 Any ideas? I'd like to work on other things, but there's quite a lot of audio bugs left unsolved and that's going to block the next stable release.

antis81:
Yes, I have a very exact idea of why this happens to user's. If there's a problem with the release, there's a ~~simple~~ quick solution: Remove (deactivate) the WIP feature "gain meter widget" within the A/V settings tab.

tux3:
@antis81 Okay, do you know how to fix this reasonably soon? Otherwise I think it would be better to kill the gain meter (or whatever is causing all those bugs) and pull it back when it's ready.

antis81:
@tux3
I'll setup two PR's:
1. Deactivate the "gain meter widget" feature -> closes this issue (#2806) and #2703.
2. Finish the feature implementation in a new PR.

Is that a plan?

tux3:
@antis81 Alright. If that solves it, it's all good; if not we'll have to go a bit further.

ElLamparto:
For me the sound always works (except that the microphone choice sometimes disappears from the settings. Linux/Windows - no difference).

The gain meter usually works under Linux and usually doesn't under Windows. The audio is not affected by the meter's caprices.

If audio is accompanied by video, then there is a lot of freezing (there is another ticket for that).

antis81:
> For me the sound always works (except that the microphone choice sometimes disappears from the settings. Linux/Windows - no difference).

Thanks for this positive feedback. :) Please post an issue for your mentioned problem. I never noticed this behaviour, but there's probably a logical reason for that.

> The gain meter usually works under Linux and usually doesn't under Windows. The audio is not affected by the meter's caprices.

This feature is currently deactivated, because there's multiple problems with that:
1. The concept needs to be implemented less resource hungry and actually use available data from the audio input.
2. Stopping the "measuring thread" leads to a crash, which is not easily reproducible and not worth fixing, before 1 is solved.

[...]

PR #2809, opened 2016-01-16
temporary remove audio gain metering

### LINKED ISSUE #2703: New audio: When going in the audio/video settings 1/2 times, the audio feedback doesn't work

antis81:
Remove unstable feature (see # for fix).

Closes #2703 and #2806.

### DIFF
--- a/src/widget/form/settings/avsettings.ui
+++ b/src/widget/form/settings/avsettings.ui
@@ -48,7 +48,7 @@
             </property>
            </widget>
           </item>
-          <item row="7" column="0">
+          <item row="6" column="0">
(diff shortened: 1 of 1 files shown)


tux3:
This doesn't remove the widget from the GUI, it would look like a bug to users if the VU meter is still there but doesn't work.

antis81:
@tux3 It's good to merge.

**EDIT:** You're right. I actually removed the widget now from the GUI without further changes to code. Thus, it is not constructed, but fully available as "temporary dead code".

Issue #2817, opened 2016-01-18
No password on startup


Granitoid:
The Function of autosave password was very useful.
![2016-01-18_13-09-15](https://cloud.githubusercontent.com/assets/16450834/12388577/bd7e298e-bde4-11e5-9a33-47d6aa96ee46.jpg)

zetok:
Saving password is not doable, given that it's used for encryption.

agilob:
Click on your avatar and disable encrypted profile.

Issue #2814, opened 2016-01-16
The setting of "Autostart" is disabled on OS X


ddobrev:
If it's disabled by mistake, please fix it. If not, please add auto-starting on OS X, it's an extremely useful setting for chat applications.

linux-modder:
Even on my fedora install the FIRST time I compiled it was disabled --which makes sense  autostarting a newly  first time install makes no sense to me.  If this is  post  initial launch  what  OSX version |  Mac model?

ddobrev:
The machine I tested on is MacBook Pro early 2009 with OS X 10.11 "El Capitan".

ddobrev:
The setting is now enabled but not persisted so it still cannot be used.

PR #2818, opened 2016-01-18
Login window: hide `Load automatically` if profile is encrypted

zetok:
E.g. see #2817 

Before:
before
now:
now - no button

### DIFF
--- a/src/widget/loginscreen.cpp
+++ b/src/widget/loginscreen.cpp
@@ -172,11 +172,18 @@ void LoginScreen::onLoginUsernameSelected(const QString &name)
     {
         ui->loginPasswordLabel->show();
         ui->loginPassword->show();
+        // there is no way to do autologin if profile is encrypted, and
+        // visible option confuses users into thinking that it is possible,
+        // thus disable it (and hope that users won't think that it's a bug)
+        ui->autoLoginCB->setEnabled(false);
+        ui->autoLoginCB->setToolTip(tr("Password protected profile can't be loaded automatically."));
     }
     else
     {
         ui->loginPasswordLabel->hide();
         ui->loginPassword->hide();
+        ui->autoLoginCB->setEnabled(true);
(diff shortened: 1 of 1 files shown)


agilob:
This might be confusing when functionality disappears. Wouldn't it be better to disable the checkbox?

zetok:
Dunno.

SkyzohKey:
Disable the checkbox and change the text to something like : « You cannot load an encrypted profile automatically ». It seems to be a nice solution.

zetok:
@agilob https://github.com/tux3/qTox/issues/2814#issue-127057512 ← this case seems to show that if setting is disabled, there will be question of whether that isn't a bug that should be fixed.

@SkyzohKey question is, is showing that needed at all?

What you're proposing has benefit of informing user about _functionality they can't use_. Only if user does further "research" into the matter, they would know that they could have an unencrypted profile, with all that it entails. Without that, it's only an annoying reminder that there's something that isn't available for them.

And downside of your proposal is that it will take space on the login window - resource that is really sparse there.

IMHO login window shouldn't be filled with things that distract user from what they want to do - i.e. type in password to load the profile.

tux3:
I'm with @agilob and @SkyzohKey 
I think it would be better to disable it than to hide it. If it's disabled it's clear that it exists and you just can't use it right now.
But if features start appearing and disappearing, new users will think they are going crazy! 

We could add an explanation text on hover saying "Profiles with a password can't be loaded automatically"

zetok:
> But if features start appearing and disappearing, new users will think they are going crazy! 

It wouldn't _really_ disappear, considering that it never worked..

Anyway, new look & feel:
disabled

tux3:
LGTM, although I think there should be a s in "Profiles"

agilob:
And I would change wording to "Password protected password can't be...

zetok:
@agilob right, will do.

zetok:
Done.

PR #2822, opened 2016-01-18
[OSX]Touchups

RowenStipe:
This commit has several touch ups that only effect the OS X clients.

`info.plist` and `qt.pro` are now configured to create an automatic
short version to help distinguish between clients clearly and include a
referenced file not put into the build when compiled.
`./img/icon/qtox_profile.icns’

`info.plist` also had it’s bundle version to be  updated to the 1.2.2
tag and supported languages added and fixed referenced file.

The default save location has also been moved to comply with the Tox
Client Standard. (Previous location was `~/Library/Prefrences/tox` )

A basic bash profile mover has also been included to show I did think
of that.


osx/info.plist
```diff
@@ -9,7 +9,7 @@
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleSignature</key>
-	<string>????</string>
+        <string>toxq</string>
```
zetok: Shouldn't tabs be used instead of spaces?

src/persistence/settings.cpp
```diff
@@ -537,6 +537,9 @@ QString Settings::getSettingsDirPath()
 #ifdef Q_OS_WIN
     return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::HomeLocation) + QDir::separator()
                            + "AppData" + QDir::separator() + "Roaming" + QDir::separator() + "tox")+QDir::separator();
+#elif defined(Q_OS_OSX)
+    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::HomeLocation) + QDir::separator()
+                           + "Library" + QDir::separator() + "Application Support" + QDir::separator() + "Tox")+QDir::separator();
```
tux3, line 542: If you change the settings dir path without writing compatibility code, people will lose their profiles when they update!

RowenStipe:
NOTE: This pull is still a work in progress. I want to get the profile migration code to work even if the new directory exists.

src/platform/install_osx.cpp
```diff
@@ -82,3 +84,27 @@ void osx::moveToAppFolder()
         }
     }
 }
+
+void osx::migrateProfiles()
```
tux3, line 88: Also, you could leave a comment saying that this is temporary migration code, so that when it gets forgotten about 10 years later we'll know it can be safely removed :)

RowenStipe:
Going to see if I can add: #2814 into this pr as well.

RowenStipe:
Unless there's something major I'm not seeing that needs to be fixed. This PR is ready.

PR #2825, opened 2016-01-19
Update Spanish translation

ghost:
@aaannndddyyy @DavidWolfgang FYI

### DIFF
--- a/translations/es.ts
+++ b/translations/es.ts
@@ -109,6 +109,15 @@ se pueden producir problemas con las videollamadas.</translation>
         <source>Qt version:</source>
         <translation>Versión de Qt:</translation>
     </message>
+    <message>
+        <source>Restart qTox to install version %1</source>
+        <translation>Reinicia qTox para instalar la version %1</translation>
+    </message>
+    <message>
+        <source>qTox is downloading update %1</source>
+        <comment>%1 is the version of the update</comment>
+        <translation>qTox está descargando la actualización %1</translation>
+    </message>
 </context>
 <context>
     <name>AboutSettings</name>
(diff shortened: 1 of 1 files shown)


tux3:
I'll merge this later if there are no objections.

translations/es.ts
```diff
@@ -109,6 +109,15 @@ se pueden producir problemas con las videollamadas.</translation>
         <source>Qt version:</source>
         <translation>Versión de Qt:</translation>
     </message>
+    <message>
+        <source>Restart qTox to install version %1</source>
+        <translation>Reinicia qTox para instalar la version %1</translation>
```
aaannndddyyy: versión

translations/es.ts
```diff
@@ -1867,6 +1864,86 @@ Será instalada cuando reinicies qTox.</translation>
         <comment>Desktop as a camera input for screen sharing</comment>
         <translation>Pantalla</translation>
     </message>
+    <message>
+        <source>Incorrect responce</source>
```
aaannndddyyy, line 1868: Here the original is wrong. Your translation is fine, though. 
@tux3 someone should change the English version here

translations/es.ts
```diff
@@ -1867,6 +1864,86 @@ Será instalada cuando reinicies qTox.</translation>
...
+        <source>Name not found</source>
+        <translation>Nombre no encontrado</translation>
+    </message>
+    <message>
+        <source>Tox ID not sent</source>
+        <translation>No se envió la Tox ID</translation>
+    </message>
+    <message>
+        <source>Lookup failed because the other server replied with invalid data</source>
+        <translation>La consulta falló debido a datos invalidos enviados por otro servidor</translation>
```
aaannndddyyy: inválidos, el (otro) servidor. otro - another

ghost: I'm omitting "el" on purpose here, to keep it generic. The user doesn't know which one "the other server" is, how many there are or what do they do. IMO the original gives no relevant info to the average user and should be improved @Diadlo @tux3

aaannndddyyy: the simply means the one you queried. if you just say other server without the, to me it sounds like a third party is somehow interfering, yet another server, not the one I queried. But it's a minor thing. with or without is ok for me.

tux3: The source is a bit ambiguous, but it's talking about a specific server, so I'd keep the "the".
The original could be changed from "the other server replied with" to "the server replied with" (since we know which one replied and we're not one ourselves).

aaannndddyyy: good, so el servidor - therefore i had put 'otro' in parenthesis right from the start. it was not needed.

ghost: OK!

translations/es.ts
tux3, line 1868: I'll push a patch after this is merged.

PR #2828, opened 2016-01-20
fix toxme error handling

sudden6:
fixes a bug in Toxme::deleteAddress returning a QNetworkReply::NetworkError
adds some warning messages

### DIFF
--- a/src/net/toxme.cpp
+++ b/src/net/toxme.cpp
@@ -47,6 +47,10 @@ QByteArray Toxme::makeJsonRequest(QString url, QString json, QNetworkReply::Netw
         qApp->processEvents();
     }
 
+    error = reply->error();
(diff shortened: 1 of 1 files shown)


src/net/toxme.cpp
```diff
@@ -70,6 +74,10 @@ QByteArray Toxme::getServerPubkey(QString url, QNetworkReply::NetworkError &erro
         qApp->processEvents();
     }
 
+    error = reply->error();
+    if (error)
+        qWarning() << "getServerPubkey: A network error occured:" << error.errorString();
```
tux3: Wouldn't it make sense to return early if we run into an error at this point?

sudden6: You are right, will change

Issue #2823, opened 2016-01-18
the size limit of the log file


PKEv:
At the moment there is no limit on the log file, so for example I have he has grown already to 30 MB.
Offer to limit its size and overwrite the file with new data

tux3:
Related: #1615

gzip compression reduces the log size on the order of 200x times, but makes it harder for people to open it so they can put it in bug reports. So I'm going to change my mind and say it's better to limit the size instead of what #1615 says.

I don't think we'll ever want to sift through more than 1MB of logs, or more than the last 15 runs of qTox.
Running some stats on my own log file says 15 runs of qTox is ~300kB, so that leaves some margin in case of log spam too.

We could just run that function every time qTox starts, first truncate the file to 1MB then search for the line "qTox file logger starting" and keep only the last 15.

sudden6:
> We could just run that function every time qTox starts, first truncate the file to 1MB then search for the line "qTox file logger starting" and keep only the last 15.

This seems like the best thing to do.
Implementing this sounds like an easy thing, maybe give it the "easy" tag?

PR #2829, opened 2016-01-21
Added log rotation functionality.

anoadragon453:
This is to prevent log files from ever exceeding a limit of 1MB each. Only the current and one other log file are kept, giving hopefully enough history for any necessary debugging.

Addresses #2823


src/main.cpp
```diff
@@ -127,8 +128,27 @@ int main(int argc, char *argv[])
...
+        qDebug() << "Log file over 1MB, rotating...";
+		
+        QDir dir (logFileDir);
+		
+        // Check if log.1 already exists, and if so, delete it
+        if (dir.exists(logFileDir + "qtox.log.1")) {
+            dir.remove("qtox.log.1");
+        }
+
+        dir.rename("qtox.log", "qtox.log.1");
```
tux3: Edit: You sometimes do `logFileDir + "qtox.log.1"` and sometimes just `"qtox.log.1"`, that's a bit inconsistent.

Also you probably shouldn't do `if(exists) remove()` because of TOCTOU issues, just do the remove and see if it worked.

tux3:
What's the advantage of keeping two 1MB file instead of one 2MB file though?
It seems like it just makes things a bit more complex.

anoadragon453:
I attempted that functionality at first, but couldn't find a way to trim a file from the beginning to a specific filesize with either reading in the entire file or reading line by line. Both of which, according to the QT5 reference for QFile will use quite a bit of memory, which is something we want to avoid, especially for code that is running on startup.

This method was quick and only required a single file size check to perform.

src/main.cpp
```diff
@@ -127,8 +128,27 @@ int main(int argc, char *argv[])
     sodium_init(); // For the auto-updater
 
 #ifdef LOG_TO_FILE
+    QString logFileDir = Settings::getInstance().getSettingsDirPath();
     logFileStream.reset(new QTextStream);
-    logFileFile.reset(new QFile(Settings::getInstance().getSettingsDirPath()+"qtox.log"));
+    logFileFile.reset(new QFile(logFileDir + "qtox.log"));
```
tux3, line 133: You open the file before moving, so if it is already > 1MB, it will be moved while still open, so qTox will be writing in "qtox.log.1".

tux3:
Fair enough, please just fix the couple issues above and then it should be good to merge.

anoadragon453:
Will do, thank you for the corrections.

zetok:
Just gotta say, this PR isn't too friendly in cases where one does incremental backups with btrfs snapshots.

I.e. with btrfs snapshots, when log file is removed, there's `free_space - size_of_removed_log`. That is, every time it's removed, there would be >1MB less free space.

Now, if file was only trimmed when it exceeds 1MB, then ~only those few KBs that were modified would be removed from pool of free space.

_Not that I care personally, given that the way I do backups mitigates stuff like that_ - though most of people who use btrfs snapshots would be affected by it, so it's something that should be kept in mind. It may not sound like much, but over time it gets to quite a big number.

With that being said, I think that log rotating is probably better than trimming, given what @anoadragon453 wrote about performance hit of trimming.

# 

On a side note, I wonder how big performance hit it would have been to parse my log file to trim it :3

```
197M    /home/zetok/.config/tox/qtox.log
```

tux3:
@zetok
This is the standard behavior of logrotate, so I don't think it's really an issue. 
BTRFS can do deduplication in case this becomes a problem for someone with many snapshots, (although this currently still has to be done manually)

zetok:
@tux3 yeah, I just wonder if that behaviour shouldn't be mentioned somewhere in user docs. Not sure where to put it though, since not a lot of people would be affected, and thus it's not something important.

tux3:
@zetok 
I think we should try to keep the qTox docs simple so people who need help can actually read them without being overwhelmed.
Technical details on how to save space for Btrfs snapshot users on Linux is maybe a bit too specialized :)

anoadragon453:
Fixed the issues above, and now works correctly in my testing.

tux3:
Okay, it looks good to me. Except you commited your qtox.log file :)

anoadragon453:
Haha, fixed.

Issue #2497, opened 2015-11-04
Markdown for messaging


rugk:
I think it would be a nice thing to be able to use some basic markdown to format your messages.
I think you all like it and it is very easy to use too. :)

E.g. Skype also allows simple formatting like `**bold**`, `__underlined__`, ... if I remember correctly. That's similar to markdown, but I think for (q)Tox real markdown should be used of course.
You can also see how nice this is in Discourse forums.

ovalseven8:
Personally, I don't necessarily need it.
Additionally, it's more complicated in Tox compared to Skype or internet forums because of the compatibility between the different clients.

agilob:
3 weeks ago there was html and people complained!

rugk:
Html(!)?

So you could send `<a href="tox.chat">my link</a>`?
Also `<script type="text/javascript">alert("Hi!");</script>`? 😉
In this case I would also have complained about this...

---

The cross-compatible issue is true - however markdown was designed to (also) look good when displaying the plain text, so it won't look so bad.
Additionally I can of course also ask this on toxcore so that this possibly will be adopted by the tox core.

ProMcTagonist:
> markdown was designed to (also) look good when displaying the plain text, so it won't look so bad.

Agreed, lots of people **already*\* emphasize like this.

Markdown gracefully degrades and shouldn't cause problems for anyone.

agilob:
This is not a toxcore issue but µtox, antox, toxic and other clients.

zetok:
@rugk

> The cross-compatible issue is true - however markdown was designed to (also) look good when displaying the plain text, so it won't look so bad.

No.

There was an issue on STS on how to do it properly - if you're interested in why, you may want to go to there and read it.

@agilob

> This is not a toxcore issue but µtox, antox, toxic and other clients.

Eh, not quite. This ~is partially a toxcore issue, given how toxcore API works and what it is supposed to do.
But yes, it's also a client issue.

quininer:
It is a good idea, I want to have a tag no longer make qTox eat my code indentation. ;)

zetok:
> I want to have a tag no longer make qTox eat my code indentation. ;)

Well, this is a bug that is fixed in #2013 - I kinda wonder when remaining issues with it could be fixed, so that it could be merged.

TheNain38:
This is a very good idea, markdown's design makes it looks good even when in plain text, so I think this would be cool to implement :)

An-dy:
+1

[...]

Issue #1069, opened 2015-01-30
Settings rethink


dcharles525:
I have seen talk about redoing the settings section so I am just going to post some ideas here. 

Tox is aimed at being a popular, secure, and easy to use chat service for everyone. The settings section in qTox does not reflect that. Now don't get me wrong, the client is in its early stages still and isn't stable but I think the settings need some design and simplification. 

One issue is there is a massive amount of stuff in the general tab which isn't really general. A lot of these settings won't ever be used by the basic consumer. This is where I propose a fix for all of the problems with the settings. 

Have a tab called "General" with the bare minimum settings, then have a check box that says enable "advanced mode". When the user enables this, make sure a warning pops up saying "HEY DON'T ENABLE THIS UNLESS YOU KNOW WHAT YOU ARE DOING!" or something, you get the idea. This keeps things a lot cleaner and simple. 

As for the design I'm not sure how much you can do with QT but hopefully you can style everything in the settings with some help from the original designer. 

Anyways, this is just a general issue with the settings, I am totally open for constructive conversation about improving the qTox settings.

ProMcTagonist:
I agree that the settings should be reworked to be in line with the core goals of the Tox project.

The settings right now look a bit cluttered and claustrophobic. A few of the tabs require scrolling with standard window size and that's annoying because you have to use the scrollbar or else you randomize all your combobars and sliders.

suhr:
cast @ItsDuke

ItsDuke:
@dcharles525 any chance you could take a screenshot of how it looks for you currently?

suhr:
@ItsDuke This is how it looks for me:
https://img.bi/#/FhpkmIh!hQvjGZi3opTTBrazJUcfKE4hW95kpmqBLHNl6vZ9
https://img.bi/#/S2J0moL!8wIcOWEbzwui2hW8IdeNNyJCsiG9tCYiBj3RgY9r
https://img.bi/#/GM2MzHP!YHkPjGJt0BcUY9KHbEiDk3YC71TD29x81CUGOxiI
https://img.bi/#/QxY97Ei!X2X6Ml71HkmxNJ5FSgplzBW9BHVGR9JHilkiEnal (video preview is black because I have no camera)
https://img.bi/#/tnCYfto!YfIBHYLgFyT7XWXiOfzp0j90cuQ7AFEtn4zvfz53

zetok:
[...]

PR #2832, opened 2016-01-21
Add Markdown support. 

anoadragon453:
Underline, Italics, Strikethrough and Bold supported. Addresses #2497 

example

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -54,6 +54,9 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     //quotes (green text)
     text = detectQuotes(detectAnchors(text), type);
 
+    //markdown
+    text = detectMarkdown(text);
+
     switch(type)
     {
     case ACTION:
@@ -173,6 +176,58 @@ void ChatMessage::hideDate()
     ChatLineContent* c = getContent(2);
     if (c)
         c->hide();
+}
+
+QString ChatMessage::detectMarkdown(const QString &str)
+{
+    QString out;
+
+    // Create regex for certain markdown syntax
+    QRegExp exp("(\\*\\*)([^\\*\\*]*)(\\*\\*)"   // Bold    **text**
+                "|(\\*)([^\\*]*)(\\*)"           // Italics *text*
+                "|(\\_)([^\\_]*)(\\_)"           // Italics _text_
+                "|(\\_\\_)([^\\_\\_]*)(\\_\\_)"  // Italics __text__
+                "|(\\-)([^\\-]*)(\\-)"           // Underline  -text-
+                "|(\\~)([^\\~]*)(\\~)"           // Strike  ~text~
+                "|(\\~~)([^\\~\\~]*)(\\~~)"      // Strike  ~~text~~
+                );
+
+    // Support for multi-line text
+    QStringList messageLines = str.split("\n");
+    QStringList outLines;
+    for (int i = 0; i < messageLines.size(); ++i)
+    {
+        out = messageLines.at(i);
+        int offset = 0;
+        while ((offset = exp.indexIn(out, offset)) != -1)
+        {
+            QString snippet = exp.cap();
+
+            QString htmledSnippet;
+
+            // Match captured string to corresponding md format
+            if (exp.cap(1) == "**") // Bold **text**
+                htmledSnippet = QString("<b>%1</b>").arg(snippet.mid(2,snippet.length()-4));
+            else if (exp.cap(4) == "*" && snippet.length() > 2) // Italics *text*
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(7) == "_" && snippet.length() > 2) // Italics _text_
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(10) == "__"&& snippet.length() > 4) // Italics __text__
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(2,snippet.length()-4));
+            else if (exp.cap(13) == "-" && snippet.length() > 2) // Underline -text-
+                htmledSnippet = QString("<u>%1</u>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(16) == "~" && snippet.length() > 2) // Strikethrough ~text~
+                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(19) == "~~" && snippet.length() > 4) // Strikethrough ~~text~~
+                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(2,snippet.length()-4));
+            else
+                htmledSnippet = snippet;
+            out.replace(offset, exp.cap().length(), htmledSnippet);
+            offset += htmledSnippet.length();
+        }
+        outLines.push_back(out);
+    }
+    return outLines.join("\n");
 }
 
 QString ChatMessage::detectAnchors(const QString &str)
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -55,7 +55,8 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     text = detectQuotes(detectAnchors(text), type);
 
     //markdown
-    text = detectMarkdown(text);
+    if (Settings::getInstance().getMarkdownPreference() != 0)
+        text = detectMarkdown(text);
 
     switch(type)
     {
@@ -183,13 +184,13 @@ QString ChatMessage::detectMarkdown(const QString &str)
     QString out;
 
     // Create regex for certain markdown syntax
-    QRegExp exp("(\\*\\*)([^\\*\\*]*)(\\*\\*)"   // Bold    **text**
-                "|(\\*)([^\\*]*)(\\*)"           // Italics *text*
-                "|(\\_)([^\\_]*)(\\_)"           // Italics _text_
-                "|(\\_\\_)([^\\_\\_]*)(\\_\\_)"  // Italics __text__
-                "|(\\-)([^\\-]*)(\\-)"           // Underline  -text-
-                "|(\\~)([^\\~]*)(\\~)"           // Strike  ~text~
-                "|(\\~~)([^\\~\\~]*)(\\~~)"      // Strike  ~~text~~
+    QRegExp exp("(\\*\\*)([^\\*\\*]{2,})(\\*\\*)"   // Bold    **text**
+                "|(\\*)([^\\*]{2,})(\\*)"           // Italics *text*
+                "|(\\_)([^\\_]{2,})(\\_)"           // Italics _text_
+                "|(\\_\\_)([^\\_\\_]{2,})(\\_\\_)"  // Italics __text__
+                "|(\\-)([^\\-]{2,})(\\-)"           // Underline  -text-
+                "|(\\~)([^\\~]{2,})(\\~)"           // Strike  ~text~
+                "|(\\~~)([^\\~\\~]{2,})(\\~~)"      // Strike  ~~text~~
                 );
 
     // Support for multi-line text
@@ -205,21 +206,34 @@ QString ChatMessage::detectMarkdown(const QString &str)
 
             QString htmledSnippet;
 
+            // Check for surrounding spaces and/or beginning/end of line
+            if (!((snipCheck.startsWith(' ') || offset == 0) && (snipCheck.endsWith(' ') || offset + snipCheck.trimmed().length() == out.length())))
+            {
+                offset += snippet.length();
+                continue;
+            }
+
+            int mul = 0; // Determines how many characters to strip from markdown text
+
+            // Set mul depending on markdownPreference
+            if (Settings::getInstance().getMarkdownPreference() == 2)
+                mul = 2;
+
             // Match captured string to corresponding md format
             if (exp.cap(1) == "**") // Bold **text**
-                htmledSnippet = QString("<b>%1</b>").arg(snippet.mid(2,snippet.length()-4));
+                htmledSnippet = QString("<b>%1</b>").arg(snippet.mid(mul,snippet.length()-2*mul));
             else if (exp.cap(4) == "*" && snippet.length() > 2) // Italics *text*
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(7) == "_" && snippet.length() > 2) // Italics _text_
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(10) == "__"&& snippet.length() > 4) // Italics __text__
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(2,snippet.length()-4));
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul,snippet.length()-2*mul));
             else if (exp.cap(13) == "-" && snippet.length() > 2) // Underline -text-
-                htmledSnippet = QString("<u>%1</u>").arg(snippet.mid(1,snippet.length()-2));
+                htmledSnippet = QString("<u>%1</u>").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(16) == "~" && snippet.length() > 2) // Strikethrough ~text~
-                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(1,snippet.length()-2));
+                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(19) == "~~" && snippet.length() > 4) // Strikethrough ~~text~~
-                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(2,snippet.length()-4));
+                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(mul,snippet.length()-2*mul));
             else
                 htmledSnippet = snippet;
             out.replace(offset, exp.cap().length(), htmledSnippet);
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -187,9 +187,9 @@ QString ChatMessage::detectMarkdown(const QString &str)
 
     // Create regex for certain markdown syntax
     QRegExp exp("(\\*\\*)([^\\*\\*]{2,})(\\*\\*)"   // Bold    **text**
-                "|(\\*)([^\\*]{2,})(\\*)"           // Bold *text*
+                "|(\\*)([^\\*]{2,})(\\*)"           // Italics *text*
                 "|(\\_)([^\\_]{2,})(\\_)"           // Italics _text_
-                "|(\\_\\_)([^\\_\\_]{2,})(\\_\\_)"  // Italics __text__
+                "|(\\_\\_)([^\\_\\_]{2,})(\\_\\_)"  // Bold    __text__
                 "|(\\-)([^\\-]{2,})(\\-)"           // Underline  -text-
                 "|(\\~)([^\\~]{2,})(\\~)"           // Strike  ~text~
                 "|(\\~~)([^\\~\\~]{2,})(\\~~)"      // Strike  ~~text~~
@@ -225,12 +225,12 @@ QString ChatMessage::detectMarkdown(const QString &str)
             // Match captured string to corresponding md format
             if (exp.cap(1) == "**") // Bold **text**
                 htmledSnippet = QString(" <b>%1</b> ").arg(snippet.mid(mul,snippet.length()-2*mul));
-            else if (exp.cap(4) == "*" && snippet.length() > 2) // Bold *text*
-                htmledSnippet = QString(" <b>%1</b> ").arg(snippet.mid(mul/2,snippet.length()-mul));
+            else if (exp.cap(4) == "*" && snippet.length() > 2) // Italics *text*
+                htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(7) == "_" && snippet.length() > 2) // Italics _text_
                 htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul/2,snippet.length()-mul));
-            else if (exp.cap(10) == "__"&& snippet.length() > 4) // Italics __text__
-                htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul,snippet.length()-2*mul));
+            else if (exp.cap(10) == "__"&& snippet.length() > 4) // Bold __text__
+                htmledSnippet = QString(" <b>%1</b> ").arg(snippet.mid(mul,snippet.length()-2*mul));
             else if (exp.cap(13) == "-" && snippet.length() > 2) // Underline -text-
                 htmledSnippet = QString(" <u>%1</u> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(16) == "~" && snippet.length() > 2) // Strikethrough ~text~
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -26,6 +26,8 @@
 #include "content/image.h"
 #include "content/notificationicon.h"
 
+#include <QDebug>
+
 #include "src/persistence/settings.h"
 #include "src/persistence/smileypack.h"
 #include "src/widget/style.h"
@@ -202,7 +204,8 @@ QString ChatMessage::detectMarkdown(const QString &str)
         int offset = 0;
         while ((offset = exp.indexIn(out, offset)) != -1)
         {
-            QString snippet = exp.cap();
+            QString snipCheck = out.mid(offset-1,exp.cap(0).length()+2);
+            QString snippet = exp.cap(0).trimmed();
 
             QString htmledSnippet;
 
@@ -219,21 +222,26 @@ QString ChatMessage::detectMarkdown(const QString &str)
             if (Settings::getInstance().getMarkdownPreference() == 2)
                 mul = 2;
 
+            if (snippet.length() > 0)
+            {
+                qDebug() << "Offset: " << offset << " Grabbed snippet: " << snippet << " exp.cap(7): " << exp.cap(7);
+            }
+
             // Match captured string to corresponding md format
             if (exp.cap(1) == "**") // Bold **text**
-                htmledSnippet = QString("<b>%1</b>").arg(snippet.mid(mul,snippet.length()-2*mul));
+                htmledSnippet = QString(" <b>%1</b> ").arg(snippet.mid(mul,snippet.length()-2*mul));
             else if (exp.cap(4) == "*" && snippet.length() > 2) // Italics *text*
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul/2,snippet.length()-mul));
+                htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(7) == "_" && snippet.length() > 2) // Italics _text_
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul/2,snippet.length()-mul));
+                htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(10) == "__"&& snippet.length() > 4) // Italics __text__
-                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(mul,snippet.length()-2*mul));
+                htmledSnippet = QString(" <i>%1</i> ").arg(snippet.mid(mul,snippet.length()-2*mul));
             else if (exp.cap(13) == "-" && snippet.length() > 2) // Underline -text-
-                htmledSnippet = QString("<u>%1</u>").arg(snippet.mid(mul/2,snippet.length()-mul));
+                htmledSnippet = QString(" <u>%1</u> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(16) == "~" && snippet.length() > 2) // Strikethrough ~text~
-                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(mul/2,snippet.length()-mul));
+                htmledSnippet = QString(" <s>%1</s> ").arg(snippet.mid(mul/2,snippet.length()-mul));
             else if (exp.cap(19) == "~~" && snippet.length() > 4) // Strikethrough ~~text~~
-                htmledSnippet = QString("<s>%1</s>").arg(snippet.mid(mul,snippet.length()-2*mul));
+                htmledSnippet = QString(" <s>%1</s> ").arg(snippet.mid(mul,snippet.length()-2*mul));
             else
                 htmledSnippet = snippet;
             out.replace(offset, exp.cap().length(), htmledSnippet);
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -222,11 +222,6 @@ QString ChatMessage::detectMarkdown(const QString &str)
             if (Settings::getInstance().getMarkdownPreference() == 2)
                 mul = 2;
 
-            if (snippet.length() > 0)
-            {
-                qDebug() << "Offset: " << offset << " Grabbed snippet: " << snippet << " exp.cap(7): " << exp.cap(7);
-            }
-
             // Match captured string to corresponding md format
             if (exp.cap(1) == "**") // Bold **text**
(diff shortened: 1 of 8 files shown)


zetok:
In its current form it interferes with user input.

The proper way to do it was discussed in https://github.com/Tox-Archive/Tox-STS/issues/74 - in short, new message type is needed in toxcore, and using that there can be a proper markdown support done.

anoadragon453:
Hmm, I see. So there needs to be some sort of differentiation between markdown and non-markdown text.

Which we could technically add with something similar to Wikipedia's "nowiki" tag, essentially cancelling out md detection within those brackets, but a user with a different client would see those brackets and be confused.

Interesting. Well, when we do get a separate message type, this code'll be here to parse it :)

Zer0-One:
@anoadragon453 You might be interested in catching up on the discussion here: https://github.com/Tox/Tox-Client-Standard/pull/25

anoadragon453:
Thank you for the link, interesting read. A standard spec is definitely required if we are to go through with it.

In my personal experience, some users turn down Tox specifically because it doesn't feature basic text formatting. 

IMO Toxcore should have a separate message type, each client should have a standard, non-instrusive way to switch between the two, and clients that can't support certain features should support as much as possible, and no more.

Hopefully we figure out what definite route we're taking soon so we can get some implementation going.

src/chatlog/chatmessage.cpp
```diff
@@ -175,6 +178,58 @@ void ChatMessage::hideDate()
...
+
+            // Match captured string to corresponding md format
+            if (exp.cap(1) == "**") // Bold **text**
+                htmledSnippet = QString("<b>%1</b>").arg(snippet.mid(2,snippet.length()-4));
+            else if (exp.cap(4) == "*" && snippet.length() > 2) // Italics *text*
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(7) == "_" && snippet.length() > 2) // Italics _text_
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(1,snippet.length()-2));
+            else if (exp.cap(10) == "__"&& snippet.length() > 4) // Italics __text__
+                htmledSnippet = QString("<i>%1</i>").arg(snippet.mid(2,snippet.length()-4));
```
agilob: Italics 3 times but bold once?

anoadragon453: @agilob What other pattern for bold should be added? :P

It'll have to be adapted to the STS' spec eventually anyways. It's pretty simply to edit/add new syntaxes though.

agilob: Your code:

```
** - bold
* - italics
_ - italics
__ - italics
```

*\* - should make bold otherwise is mega inconsistent.

zetok: > *\* - should make bold otherwise is mega inconsistent

Err, `__`.
According to github editor, `__` results in bold.

zetok:
@anoadragon453 https://github.com/irungentoo/toxcore/pull/1502 here you go, you can checkout branch and make needed stuff on qTox side :)

anoadragon453:
Thanks, I'll start on the framework for differentiating between different message types.

Edit: Nevermind, already implemented. Just need to add a new type.

Edit 2: Of course we use a bool to determine message type... 

Edit 3: Perhaps we should put a new message type on hold and consider other options... see https://github.com/Tox/Tox-Client-Standard/pull/25#issuecomment-173741331

zetok:
> Edit 3: Perhaps we should put a new message type on hold and consider other options...

Well, you're the one implementing it in qTox, so it's up to you when you decide to work on it. :)

The way I see it (perhaps biased a bit) is that it's going to be done via `MARKDOWN` message type.

Also, I don't see other options..?

ProMcTagonist:
@zetok https://github.com/Tox/Tox-Client-Standard/pull/25#issuecomment-173820390

anoadragon453:
I think implementing it in toxcore is fine, but we have to do two things to implement it:
- Be aware that MD messages may potentially break clients that have not been updated to work with them yet.
- Figure out a standardized way for the client to determine that the user wants to send a normal or MD message - be it either a button toggle or something within the message itself (think '/me', /md perhaps?).

Once those two things are figured out and agreed upon, I'll be more than happy to crank out the code to make it work.

ProMcTagonist:
needing to manually switch to markdown mode is stupid and a pain

the point of markdown is being readable whether it's styled or not and being integrated into plaintext

anoadragon453:
@ProMcTagonist Good point, if we're not doing another message type, then there's no need to check.

tux3:
@ProMcTagonist 
I 100% agree, I think it's not natural to have to switch modes when you just want to chat.
And if people really feel it's important to not remove anything from the message another option is to render the markdown while keeping the markings, like GMANE does:

http://thread.gmane.org/gmane.text.docutils.user/4104/focus=4105
http://thread.gmane.org/gmane.mail.mutt.user/43722/focus=43750

But in my opinion this is not a problem.

zetok:
> the point of markdown is being readable whether it's styled or not and being integrated into plaintext

That doesn't matter. What matters, is whether you can send a non-marked by markdown message.

@tux3 how do you plan on doing this, when there'll be only md-marked messages?

ProMcTagonist:
People sending codeblocks through tox should use the MD code formatting because otherwise it'll get all stripped in most clients anyway so codeblocks are an argument pro-constant-markdown

@tux3 That looks like a neat solution, no characters are removed -> no problem

anoadragon453:
@zetok https://github.com/Tox/Tox-Client-Standard/pull/25#issuecomment-173741331 ?

zetok:
@anoadragon453 No.

tux3:
@zetok 
I don't see the case for sending non-markdown messages, though. 
If you simply don't like Markdown, we will probably offer an option to not render it.
But do you need a special messages that forces other people to not use markdown?

TheNain38:
Yes, it will be best to have an option to not render it if they don't like it. @tux3 +1

ProMcTagonist:
I vote for constant markdown parsing, in the normal message type, keeping signal characters like gmane, and an option not to render styling

TheNain38:
The problem with not removing signal characters is that you have the obligation to have them visible to have the styling.
And in the case of Markdown signal characters is the "replacement" to non rendered styiling

zetok:
@tux3 

> I don't see the case for sending non-markdown messages, though.
> If you simply don't like Markdown, we will probably offer an option to not render it.

It's not about liking it or not - it's about preserving information, with which markdown interferes.

Depending on the degree of "support" for markdown, it wouldn't interfere much, or would interfere to a degree where receiving (& rendering) side would have some info from message simply stripped.

Option to not render it (some quick switch [under r-click?]) is probably the solution. :)

ProMcTagonist:
@zetok 

> It's not about liking it or not - it's about preserving information, with which markdown interferes.

This was my concern too, until:

@tux3 

> And if people really feel it's important to not remove anything from the message another option is to render the markdown while keeping the markings, like GMANE does:
> http://thread.gmane.org/gmane.text.docutils.user/4104/focus=4105
> http://thread.gmane.org/gmane.mail.mutt.user/43722/focus=43750

So everything is preserved.

tux3:
@zetok 
We will have an option to not render Markdown then.
Now if markdown is enabled, should we do it the GMANE way or strip signaling characters?
Personally I'm okay with the GMANE option, although I don't think it's really needed.

anoadragon453:
I'd like to point out that copying a MD'd message perseveres the non-parsed text.

ProMcTagonist:
@anoadragon453 it should be viewable without pasting to a text editor, which gmane solves nicely

anoadragon453:
@ProMcTagonist I'm unfamiliar with Gmane, how does it do this exactly?

ProMcTagonist:
@anoadragon453 
http://thread.gmane.org/gmane.text.docutils.user/4104/focus=4105
http://thread.gmane.org/gmane.mail.mutt.user/43722/focus=43750

TheNain38:
> I'd like to point out that copying a MD'd message perseveres the non-parsed text.

+1
Also the problem in keeping signal characters is that the non-parsed text is the "replacement" to parsed text with Markdown...

anoadragon453:
@ProMcTagonist Ah I see, so it would bold without removing the surrounding characters. This is a great in-between, but I feel some users - myself included - would prefer the standard stripping way. Can we keep this as one of the options, i.e.
- No Markdown
- Markdown with no character stripping
- Markdown with stripping

TheNain38:
I'm okay with what @anoadragon453 said, but when it has stripping, should it also copy signal characters?

ProMcTagonist:
I strongly suggest that the default should be gmane so you can be assured your messages are being seen properly by your friends.

zetok:
> Markdown with no character stripping

Um. Just use it as `No Markdown` mode, and in message edit box? I.e. `No Markdown` mode is when no characters are being stripped when rendering, so..

anoadragon453:
@zetok So only two modes then? Yeah that's fine.

@TheNain38 That's what I'm envisioning.

TheNain38:
I would prefer:
- No Markdown
- Markdown (with stripping)

And when copying, it should always copy all the characters.

tux3:
I'm for Markdown with or without stripping by default, with No Markdown being an option.

zetok:
@anoadragon453 Yeah. Now that I read what I wrote earlier, it might not have been that clear, so clarifying:
- "No Markdown" - rendered like Gmane does it, no characters stripped
- Markdown - real markdown

anoadragon453:
@TheNain38 @zetok 

> - No Markdown
> - Markdown (with stripping)
> 
> And when copying, it should always copy all the characters.

I agree with this behavior.

ProMcTagonist:
> "No Markdown" - rendered like Gmane does it, no characters stripped

But that's lying to the user. Someone clicking on "No Markdown" doesn't want to see any text formatting at all. I don't understand why they would want that, but we can't just provide an option that doesn't do what it says, that's a recipe for angry users.

anoadragon453:
@ProMcTagonist Hrm... back to three modes then? It's won't be very hard to do, just an if.

tux3:
I don't think we need both real markdown and Gmane markdown, so let's just pick one and then have an option to disable it entirely (for people who just don't want markdown at all)

zetok:
> >  "No Markdown" - rendered like Gmane does it, no characters stripped
> 
> But that's lying to the user. Someone clicking on "No Markdown" doesn't want to see any text formatting at all.

If you say so :)

I proposed it, since I think that it looks & feels better than just "plain" plaintext, deprecating it, while not interfering with data. Though if users don't want that kind of behaviour.. well, there could be just `plain` and `markdown` modes.

anoadragon453:
If we're going for just two modes, I'd vote for just `plain` and `markdown`, considering I wasn't even aware of 'Gmane-like' markdown until a few minutes ago.

ProMcTagonist:
I would prefer to make Gmane the default so no information is lost. It's a very clever md implementation that has the best of nonstripped and md-formatted text.

TheNain38:
I vote for what I said

> - No Markdown
> - Markdown (with stripping)
> 
> And when copying, it should always copy all the characters.

Also, the problem in keeping signal characters is that: in non-parsed text signal characters are the "replacement" to parsed text with Markdown (but without signal characters)...

tux3:
So if I don't count myself, that looks like 3 votes for `plain` and `markdown` plus 1 vote for `gmane`and `plain` ?

ProMcTagonist:
Gmane solves the problem of characters being lost to markdown. We want markdown formatting to be the default, and no one is going to check every message by pasting it around. To avoid lost characters and misunderstandings while having markdown formatting, the default should be gmane.

tux3:
I'm okay with either by default.
We could always add some heuristics to prevent misunderstandings. For example only allow bold/italics/underline/strikethrough within a single line, to avoid unrelated characters or smileys applying style to half the message.
The gmane implementation will only render markdown applied to a single word to avoid misrendering pointers and underscores in code (but we don't have to be this strict because we don't parse emails full of code)

anoadragon453:
@ProMcTagonist I only really see that when applied to codeblocks being posted, which isn't a very regular occurrence. And like I mentioned earlier, we could always have a char sequence to cancel out markdown parsing within them, it could even be 

```
specifically designed to hold code.
```

anoadragon453:
@tux3 Smilies are parsed before Markdown parsing begins, so they shouldn't interfere.

TheNain38:
Also, fenced codeblocks are my favorite style.

tux3:
@anoadragon453 That's good, although that of course doesn't apply to smileys that aren't in the user's current smiley pack, for example two `-_-` or a `*-*` might add some unwanted formatting.

anoadragon453:
@tux3 All the more reason for a tag to prevent md parsing.

TheNain38:
@tux3 Yes, but with Gmane style of parsing md, it also have this problem, even if they aren't stripped

ProMcTagonist:
@anoadragon453 We should have codeblocks, but that's a separate thing and users shouldn't have to use a code tag to make sure they aren't misunderstood, that's like toggling markdown modes plus they get stuck with code formatting. Gmane is simpler, and less headache for everyone.

@TheNain38 gmane always shows what the user originally input so there can be no misunderstandings, even if something is bolded or italicized.

gmane

tux3:
@anoadragon453 But then we have to teach users the joys of escaping text, and not all our users are coders.
It might be safer to use heuristics when possible. I know I rarely apply emphasis accross multiple lines for example, and I rarely use more than one smiley per line.

TheNain38:
Gmane also have the parsing problem, it doesn't solve it...

anoadragon453:
Casual conversation with a user:

screenshot from 2016-01-22 18-01-03

I think the Gmane way would confuse/frustrate some people.

ProMcTagonist:
> ease of access
> not everyone is savvy as you programmers

Sounds like a vote for Gmane as opposed to forcing users to put everything in code blocks and hiding their normal input from them.

tux3:
@anoadragon453 
That image makes a good case for my heuristics! Without them both of Felix's underscores disappear and half the message gets underlined.

anoadragon453:
@tux3 Ah, but you forget we parse on a line-by-line basis.

tux3:
@anoadragon453 Ah you're right, I though that was one message on three lines, not three messages.
Still, I maintain my support for the magic heuristics.

TheNain38:
I'm with @tux3

anoadragon453:
Keep in mind Skype also doesn't parse unless the formatting is also surrounded by spaces, thus `*-*` would produce a bold -, but `(*-*)` would remain as-is.

tux3:
@anoadragon453 +1 for this, good idea on Skype's part.

ProMcTagonist:
> thus *-* would produce a bold -, but (*-*) would remain as-is.

Many people send smilies without parentheses.

With Gmane the smilies will show in all situations, regardless of spaces or if magic heuristics don't catch something.

A vote against gmane is a vote against smilies.

TheNain38:
Again, with @tux3 for spaces surrounding smiley parsing

tux3:
@TheLastProject 
That's a good point, so I'm in favor of Gmane to save the smileys + some heuristics to not surprise people with wild formatting. I don't have a good solution for *-* becoming **-** otherwise.

anoadragon453:
@tux3 I'm still in favor of having all three to keep everybody happy :)

ProMcTagonist:
I'm fine with all three for user choice as long as Gmane is default.

tux3:
Well, we could simply have a dropdown with the three options instead of a enable/disable checkbox.

Still need to pick a default though :)

zetok:
> I'm fine with all three for user choice as long as Gmane is default.

Agreed. Are there good reasons to not use it as default?

TheNain38:
I'm in favor of Gmane as a default, because I think not having formatting isn't a good idea, and "real" markdown could be too cumbersome for casual users.

anoadragon453:
@ProMcTagonist @tux3 @zetok The default is an interesting question. I know people moving from Skype might be confused initially with _,*'s still showing even when they intend to send without them, and they would probably have to be told by someone to change it in settings (normal users don't change settings, they just expect everything to work). Then again, I completely understand the argument for not parsing-out certain characters by default.

zetok:
> (…) they would probably have to be told by someone to change it in settings (…)

Err, no, putting that in settings IMHO is a bad idea. Wouldn't it be better to have it as a r-click setting in chatform?

anoadragon453:
If accidental parsing did happen, don't you think that would be a time the users would chat about it, and thus wold be more likely to switch to the option that works for them, rather than the casual user who's used to Skype's implementation?

tux3:
@zetok Nobody will find it if it's in the right click, all the other options are in the settings

TheNain38:
@tux3 +1 or having the switch next to the chatform

anoadragon453:
@TheNain38 We don't need to muddy up the interface if users are only going to change this option ~once.

zetok:
@tux3 I wonder about that. I thought that it would be a lot harder to find it in settings, as opposed to r-click that people e.g. use to select & copy all text?

(yes, people do that, I witnessed people not knowing ctrl+a, ctrl+c plenty of times)

@anoadragon453 

> @TheNain38 We don't need to muddy up the interface if users are only going to change this option ~once.

Yes, but on the other hand, why would you hide in settings an option that user might want to quickly switch back and forth to see formatted/unformatted text?

tux3:
@anoadragon453 YES. Agreed. This. +1.

@zetok 

> @tux3 I wonder about that. I thought that it would be a lot harder to find it in settings, as opposed to r-click that people e.g. use to select & copy all text?

I think people will expect it to be with the other settings, if we scatter options around it's a bit messier.

> Yes, but on the other hand, why would you hide in settings an option that user might want to quickly switch back and forth to see formatted/unformatted text?

If we use Gmane by default, there's no need to switch back and forth.

zetok:
> If we use Gmane by default, there's no need to switch back and forth.

Maybe. Problem arises with _settings other than default_. Which is quite a lot of people, and I imagine that someone who set rendering to markdown would want at times to see cleartext.

tux3:
Presumably someone who changes the settings and sticks with it understands how Markdown works, so it's pretty easy to see that if there's **bold** there used to be a pair of asterisks around.
The information isn't lost, just transformed.

anoadragon453:
@tux3 Exactly, and if the user is receiving a block of code anyway, they can still copy it out without any information lost.

TheNain38:
@tux3 +1

anoadragon453:
So is everyone good with:
- No Markdown
- Markdown with characters (default)
- Markdown

In settings. With spaces required around characters for them to be parsed + some other magical heuristics?

ProMcTagonist:
> So is everyone good with:
> -   No Markdown
> -  Markdown with characters (default)
> -  Markdown
> 
> In settings. With spaces required around characters for them to be parsed + some other magical heuristics?

Perfect :+1:

tux3:
@anoadragon453 Sounds good to me.
The only magical heuristic I would like is to ignore markdown that spans multiple lines, because that's probably a mistake.

TheNain38:
Good for me

anoadragon453:
@tux3 Yep, and that's already implemented.

anoadragon453:
Alright, I'm off to dinner. If you guys want to pass this off to the other threads, I'll start on implementing it in qTox after I get back.

zetok:
> So is everyone good with:
> - No Markdown
> - Markdown with characters (default)
> - Markdown

Yep.

> In settings.

It would be good to put it in UI tab.

> With spaces required around characters for them to be parsed + some other magical heuristics?

I'm not too happy with a requirement to put something around emoticons in order to not have them parsed, since it in a significant degree makes them more bothersome to use, and as a result, less useful for users.
Perhaps magical heuristics can help? `;)`

zetok:
related to UI tab: https://github.com/tux3/qTox/issues/1069#issuecomment-72308411

ProMcTagonist:
@zetok, I believe he's saying spaces are required to trigger formatting, not to keep smilies from being formatted

zetok:
> @zetok, it's backward, spaces are required to trigger formatting not to keep smilies from being formatted

Right, I meant `something` = `something other than space`.

tux3:
@zetok emoticons still work the same way, it might only be parsed as markdown if it's not recognized anyways, and then only if there's a space around.

ProMcTagonist:
Ah, I see. What can we do to try and keep smilies from triggering formatting?

Possible string that could trigger formatting:

*_* that's a little... extreme *o*

tux3:
If we use Gmane markdown that just means \*o\* becomes \***o**\*

tux3:
If we use Gmane markdown that just means *o* becomes ***o***, but another heuristic could be to not apply markdown to only one character, since that's likely to be an emoticon and not a word.

zetok:
@tux3 

> (…) parsed as markdown if it's not recognized anyways, and then only if there's a space around.

Yes, which in combination `markdown + ASCII+Universe` would be almost all the time when using ascii emoticons.

tux3:
@zetok The 1-character heuristic should fix this for most smileys.

ProMcTagonist:
@tux3 But I have seen people do "*I* need this" etc with one character

If the problem can only ever bold a single smiley, that's fine in my opinion. Bold/italic single smiley not a problem.

 But if we can unintentionally bold/unbold whole bits of text that would be undesirable. Maybe like

*_< wow, really? *_*

It's sort of hard to imagine lines that could cause this, maybe it's less big of a problem than I thought, that would be nice

tux3:
@ProMcTagonist If we use the 1-character thing, then in your example the second smiley is seen as a smiley and not markdown, and so the first smiley has nothing to match with and can't do markdown either.

I don't have a solution for *I* not being recognized, but I think it might be better to have false negatives than false positives.

ProMcTagonist:
Sounds good.

anoadragon453:
Alright markdown preference now implemented. 'Show Characters' is the default.

screenshot from 2016-01-23 00-13-41

(Small UI is on account of HiDPI display)

Currently only new messages follow the updated rule, however (without a restart). Do we have a reliable way to reload all messages shown without restarting the app?

anoadragon453:
Spaces detection up and running.

zetok:
Aww, yiss. Perhaps except for the last one, which is `~.~`.

Also, I wonder if `*     *` shouldn't be parsed when it's between `~ ~`? (same goes for other variants)

# 
- [x] characters stripped from markdown version should be copied
- [x]  rename needed ← needs to be renamed, it's not clear what those are supposed to do.

Perhaps
- Plaintext? ← better suggestion needed
- Don't hide formatting characters
- Hide formatting characters

?

# 

Tested on 0343d70.

anoadragon453:
@zetok Thanks for the thorough testing.

> Aww, yiss. Perhaps except for the last one, which is `~.~`.

Single-character parsing is going to be prohibited, so this should cover cases such as this.

> Also, I wonder if \* \* shouldn't be parsed when it's between ~ ~? (same goes for other variants)

So the asterisks remain non-struckthrough while the rest of the text is, or does the asterisks cancel all text parsing?

> characters stripped from markdown version should be copied

Meaning one should be able to copy them out even when they are parsed away? This is normal behavior if you select the **message** and then hit copy, otherwise yes it does just copy the plaintext. I'm assuming you mean you want both? 

> needs to be renamed, it's not clear what those are supposed to do.

I'll work on this.

zetok:
> > Also, I wonder if \* \* shouldn't be parsed when it's between ~ ~? (same goes for other variants)
> 
> So the asterisks remain non-struckthrough while the rest of the text is, or does the asterisks cancel all text parsing?

Well, the way gh parses it:
- italics: ~~_これは日本語です。_~~ 
- bold: ~~**これは日本語です。**~~ 
- "plain": ~~これは日本語です。~~ 

Current parsing in qTox allows to have only "plain" one.

> > characters stripped from markdown version should be copied
> 
> Meaning one should be able to copy them out even when they are parsed away? This is normal behavior if you select the message and then hit copy, otherwise yes it does just copy the plaintext. I'm assuming you mean you want both? 

Um, If I understood correctly what you wrote; yes, parsed out characters should be copied along with the plaintext when user hits copy.

# 

On a side note, I wonder if some lib can't be used for parsing? I mean, in linked STS issue there were mentioned (AFAIR) several parsing libs, perhaps it would be good to use one of them?

anoadragon453:
@zetok 

> On a side note, I wonder if some lib can't be used for parsing?

Would be nice, since parsing all this manually is kinda sucky.

> Um, If I understood correctly what you wrote; yes, parsed out characters should be copied along with the plaintext when user hits copy.

Try selecting text until you get a blue box around the message, not just the blue highlight.

screenshot from 2016-01-23 03-56-27

You'll notice that this copies the markdown, whereas only copying part of a message does not.

zetok:
> > Um, If I understood correctly what you wrote; yes, parsed out characters should be copied along with the plaintext when user hits copy.
> 
> Try selecting text until you get a blue box around the message, not just the blue highlight.

Yes, that didn't work.

Actually, what didn't work is copying message at all.

Copying appears to work only when selected message has a visible nick.
I.e.
nope
↑ copying doesn't work

↓ copying works
works

anoadragon453:
@zetok Wait wha..? Try right-click copy instead of just Ctrl+C.

zetok:
> @zetok Wait wha..? Try right-click copy instead of just Ctrl+C.

Ok, that did work.

Well, not a problem with the PR, but an issue with how focus is being handled in chatform :/

zetok:
@anoadragon453 could you just edit via interactive rebase 6e2fc81d35201497af4b7a13acedfe961307d46b, so that the unneeded file wouldn't be introduced?
Also, while doing interactive rebase, d1fa6be710790c300f96d04a75815777a488b60e won't be needed any longer, so it can be removed.

anoadragon453:
@zetok Thanks for the suggestion, looking cleaner now.

anoadragon453:
Current Status:
- Everything seems to work how a user would expect it, only thing that doesn't work at the moment is combined syntax's, i.e. `-*hello*-`
- I need suggestions what what syntax's we're doing, and whatever is decided should be standard across all clients. If you want an example, you can take a look at [gtox's implementation](https://github.com/tux3/qTox/issues/2497#issuecomment-158461590). 

List of what I believe should be added:

``````
* Bold - *, **
* Italics - _,__
* Underline - -
* Strikethrough - ~
* Code Blocks (Cancels all other parsing) - `,``` (for multi-line)
``````

Thoughts?

**Comment by ProMcTagonist** [comment]
codeblocks should be displayed in a monospaced font with a grey background, and should NOT strip any spaces, even leading and multiple

**Comment by anoadragon453** [comment]
@ProMcTagonist Can do.

**Comment by zetok** [comment]
@anoadragon453 err, 17323c42e1943319ad4ef785bdbdb4ff823062c7 still has unneeded files introduced.

**Comment by zetok** [comment]
What I'm missing in this PR is turning markdown on/off on the messages that were already received and are being displayed.

**Comment by anoadragon453** [comment]
That's there, but you need to restart. That's what I was asking earlier with a safe way to reload all chat messages.

**Comment by ProMcTagonist** [comment]
@anoadragon453 will the new chatform in #2013 and #2744 allow a way to do that?

**Comment by anoadragon453** [comment]
@ProMcTagonist The markdown is parsed on chat message load. We essentially need to reload all messages, which it doesn't seem like any of these PRs cover.

**Comment by zetok** [comment]
> - Code Blocks (Cancels all other parsing) - `,``` (for multi-line)

Why not. Although, I think that it would be better to first get this PR to merge with just the basic stuff, and then make another PR which would add code blocks.

**Comment by anoadragon453** [comment]
@zetok Agreed, figuring out how to set a grey background for individual pieces of text is holding everything up.

**Comment by anoadragon453** [comment]
Sounds good to me.

**Comment by anoadragon453** [comment]
Unless anyone wants to add any more rules, this PR should be good to go. :+1:

**Comment by tux3** [comment]
I'll try to test and merge it this WE

File: src/chatlog/chatmessage.cpp [hunk 2]
```diff
@@ -175,6 +181,72 @@ void ChatMessage::hideDate()
...
 
+QString ChatMessage::detectMarkdown(const QString &str)
+{
+    QString out;
+
+    // Create regex for certain markdown syntax
+    QRegExp exp("(\\*\\*)([^\\*\\*]{2,})(\\*\\*)"   // Bold    **text**
+                "|(\\*)([^\\*]{2,})(\\*)"           // Bold *text*
+                "|(\\_)([^\\_]{2,})(\\_)"           // Italics _text_
+                "|(\\_\\_)([^\\_\\_]{2,})(\\_\\_)"  // Italics __text__
```
zetok: Hmm, github-flavoured markdown parses `*text*` as italics - _text_`

Additionally, `__text__` is being parsed as bold - **text**

anoadragon453: Yeah, then I had someone else tell me having *\* and \* for bold would make more sense, rather than having three Italics rules.

Though it would balance if we switched the __'s which we hadn't considered. 

I was just going off [this](https://github.com/tux3/qTox/pull/2832#issuecomment-174231662) lol, sans the codeblocks.

anoadragon453: People coming from Skype expect single *'s to be bold though, so it's probably best not to upset that muscle memory even if Github happens to do that.

quininer:
If possible, I look forward to support code highlighting, for example [highlight](http://andre-simon.de/doku/highlight/highlight.php).

anoadragon453:
@quininer We've decided to move codeblocks to a future PR, but syntax highlighting will definitely be considered :)

src/chatlog/chatmessage.cpp
zetok: Well, problem is that it's supposed to be a markdown support. As long as qTox and other clients actually do the markdown, there'll be no problem with how formatted text looks. If qTox will start to parsing text in arbitrary way.. eh.

So, either you do the real markdown, and call it basic markdown support, or you're venturing into area where there are no standards, and there is lots of confusion and non-standard behaviour.

Also, note that while people who used skype might be used to that (I didn't even know that skype supported text parsing, spare for emoticons), note that large parts of the internet actually do support markdown - e.g. reddit.

anoadragon453: Fixed.

anoadragon453:
As much as I want to simplify the process of moving from Skype, actual markdown does dictate it's own rules.

src/chatlog/chatmessage.cpp
```diff
@@ -54,6 +56,10 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     //quotes (green text)
     text = detectQuotes(detectAnchors(text), type);
 
+    //markdown
+    if (Settings::getInstance().getMarkdownPreference() != 0)
```
tux3: You should use an enum for the preference instead of a magic value

src/persistence/settings.cpp
```diff
@@ -1028,6 +1030,23 @@ void Settings::setDateFormat(const QString &format)
...
+    return markdownPreference;
+}
+
+void Settings::setMarkdownPreference(int newValue)
+{
+    QMutexLocker locker{&bigLock};
+    if (newValue < 0)
+        newValue = 1;
+    else if (newValue > 2)
+        newValue = 2;
```
tux3: If either of those branches are taken, this is an internal bug that we shouldn't silently ignore.
This wouldn't be as much of a problem with an enum instead of an int (use the type system!), but if you're going to check for internal interface violations, please assert() or at least log the bug.

src/widget/form/settings/generalform.cpp
```diff
@@ -201,6 +208,7 @@ GeneralForm::GeneralForm(SettingsWidget *myParent) :
     connect(bodyUI->showWindow, &QCheckBox::stateChanged, this, &GeneralForm::onShowWindowChanged);
     connect(bodyUI->showInFront, &QCheckBox::stateChanged, this, &GeneralForm::onSetShowInFront);
     connect(bodyUI->notifySound, &QCheckBox::stateChanged, this, &GeneralForm::onSetNotifySound);
+    connect(bodyUI->markdownComboBox, SIGNAL(currentIndexChanged(int)), this, SLOT(onMarkdownUpdated()));
```
tux3: Use the new connection syntax (the &Class::method one) if possible, it'll allow the compiler to catch errors at compile time instead of runtime.

sudden6:
@anoadragon453 could you please write a few words about this new feature in the user manual?

anoadragon453:
@sudden6 Sure, just in the wiki?

sudden6:
The user facing things are documented in /doc/user_manual_en.md

anoadragon453:
@sudden6 Added, thanks.

anoadragon453:
@tux3 Addressed the issues.

tux3:
I'll take a look at it as soon as I can

anoadragon453:
@zetok Good catch, fixed.

anoadragon453:
@tux3 Anything update on this?

anoadragon453:
:hourglass: ...

quininer:
:hourglass: ... +1

Deleetdk:
Any update on this? I downloaded the most recent version and I don't see any markdown options. Presumably, this branch wasn't merged yet?

ProMcTagonist:
@Deleetdk This PR has not yet been merged. You can see this indicator next to the title in Github's web interface that lets us know it's still open: 

open

When it turns purple and says "Merged," you'll know a PR was merged. (Red with "Closed" means it's rejected.)

anoadragon453:
Are we still waiting for the Tox Standard to decide the md rules?

tux3:
Due to the inability of the Tox Standard to actually standardize things, we will move forward with Markdown support now and deal with standard conformance when there is something to conform to :)

Deleetdk:
How long does it take for the builds to be updated? Specifically, the Windows 64-bit one.

tux3:
For the nightly builds, usually around ~30 minutes but since Jenkins is really busy at the moment count an hour or two.

anoadragon453:
Thanks for merging, I'll keep an eye on how the standard progresses and update accordingly :+1:

zetok:
> Thanks for merging, I'll keep an eye on how the standard progresses and update accordingly :+1:

"standard" is IMO currently a ~joke, and thus waiting for it is most likely a waste of time.

Issue #2600, opened 2015-11-28
Missing username in group chat


Vort:
I have encountered strange behavior while entered #tox-main-groupchat:
Nickname for one of the users was displayed incorrectly:
![qtox_missing_username](https://cloud.githubusercontent.com/assets/1242858/11452373/76004c04-95ee-11e5-8884-bfb675563f70.png)
This is not an empty name, as it is alphabetically preceded by "Andrew".
Also this name was not available with Tab-cycling.

ghost:
Yup, you can't see those in uTox. They are not users with empty aliases, those are labelled as `Tox User`.

Vort:
If this is a 'Tox User', why he is placed between 'A' and 'D'?

ghost:
No, I meat it's _not_ a `Tox User`.

The problem is repeated aliases. In that snapshot there were 2 `Andrew`s. Here I duplicated `nintendo` twice

![snapshot2](https://cloud.githubusercontent.com/assets/6382342/11456733/f6cdafda-9692-11e5-97a5-77981a894cbc.png)

ProMcTagonist:
So the bug is that repeated identical names aren't displayed?

kehugter:
That's right.

ghost:
That's right.

Issue #1847, opened 2015-06-09
Name not updated in groupchats


PKEv:
If you change the name, it will not be renewed in the active group chat

zetok:
Should be fixed once new groupchats will be used.

Issue #781, opened 2014-11-16
Groupchats do not honour friend aliases


TheLastProject:
From IRC:

PrW | If I am in groupchat I can not change the names of the people who are in group chat. Although one person is "Person A" in my friend list - in group chat he has another name

suhr:
Should we actually have the same name in groupchat and anywhere else?

zetok:
Honouring aliases only introduces bugs
![screenshot](https://cloud.githubusercontent.com/assets/3148759/5384055/7d2053f4-80ae-11e4-8c6b-18f84c98c042.png)
![ s screenshot](https://cloud.githubusercontent.com/assets/3148759/5384058/81f51c98-80ae-11e4-9086-92c6ccd9ca3f.png)
↑ contact removed me from friend list, changed its nick, and even though was shown as offline, qTox still displayed old nick in groupchat, after re-adding still old nick was shown in groupchat. Only nick change did help with situation.

Please fix

dubslow:
fjreegman wasn't an alias on your part, it was simply the last known name of the contact from your qTox's perspective?

@apprb

zetok:
From what I see qTox seems to use nick as an alias (i.e. it is shown as default alias when setting alias). Thus I thought that it's a problem with aliases..

Yes, that's the last name that was shown; I didn't explicitly set an alias to it.

apprb:
As part of resolving names in a group chat procedure qTox tries to find friend with the same ToxID as a group peer, on success it uses displayed friend name.
If somebody after deleting you from his friendlist changes nick your client still uses old nick because in your client hasn't received this changing.
Also group peers names list updates only if somebody leave/enter a group or if somebody changes own nick.

update 1: thing that really should be fixed is regenerating group peer names after friend alias changing for group where this friend is in.

anoadragon453:
Not reproducible with the latest build.

zetok:
Actually, I can still reproduce on newest qTox, one just has an alias set for friend; reopening.

anoadragon453:
To be clear one cannot show an alias to one contact and a separately-named alias to different users, one possible solution is to show both. See #1195

sudden6:
Alias and name in group now always matches.

PR #2839, opened 2016-01-22
Groupchats: fixing userlist updating/generating issues

apprb:
This PR contains fixes for #2600, #1847, and looks like for #781. Not top priority but pretty annoying.

### DIFF
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -54,9 +54,11 @@ void Group::updatePeer(int peerId, QString name)
     toxids[toxid] = name;
 
     Friend *f = FriendList::findFriend(id);
-
     if (f != nullptr && f->hasAlias())
     {
+        peers[peerId] = f->getDisplayedName();
+        toxids[toxid] = f->getDisplayedName();
+    } else {
         widget->onUserListChanged();
         chatForm->onUserListChanged();
         emit userListChanged(getGroupWidget());
@@ -91,8 +93,9 @@ void Group::regeneratePeerList()
 
         QString toxid = id.publicKey;
         toxids[toxid] = peers[i];
+
         Friend *f = FriendList::findFriend(id);
-        if (f)
+        if (f != nullptr && f->hasAlias())
         {
             peers[i] = f->getDisplayedName();
             toxids[toxid] = f->getDisplayedName();
--- a/src/group.cpp
+++ b/src/group.cpp
@@ -93,6 +93,8 @@ void Group::regeneratePeerList()
 
         QString toxid = id.publicKey;
         toxids[toxid] = peers[i];
+        if (toxids[toxid].isEmpty())
+            toxids[toxid] = tr("<Empty>", "Placeholder when someone's name in a group chat is empty");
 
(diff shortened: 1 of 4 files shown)


src/widget/widget.cpp
```diff
@@ -1462,13 +1462,9 @@ void Widget::onGroupNamelistChanged(int groupnumber, int peernumber, uint8_t Cha
         g = createGroup(groupnumber);
     }
 
-    QString name = Nexus::getCore()->getGroupPeerName(groupnumber, peernumber);
     TOX_CHAT_CHANGE change = static_cast<TOX_CHAT_CHANGE>(Change);
     if (change == TOX_CHAT_CHANGE_PEER_ADD)
     {
-        if (name.isEmpty())
-            name = tr("<Unknown>", "Placeholder when we don't know someone's name in a group chat");
-
```
zetok, line 1471: Given that toxcore can fail to provide a name, and qTox wraps around it by giving an empty string, I think that it would be better to change `<Unknown>` to `<Empty>` instead of removing it, especially given that toxcore can actually return an empty name.

I.e. as far as toxcore / qTox are concerned, the peer name is "empty" in either of those cases, and it would be good if user was notified about it in some way.

apprb, line 1471: In this function `name` variable was used even before my changing only in 

```
else if (change == TOX_CHAT_CHANGE_PEER_NAME) // core overwrites old name before telling us it changed...
{
    g->updatePeer(peernumber, name);
}
```

block. So checking for emptiness in the `f (change == TOX_CHAT_CHANGE_PEER_ADD)...` is useless and never worked before.

update: empty name check added in c43b38e

apprb:
Looked into toxcore, according it if username is empty in a groupchat toxcore returns "Tox User"

ProMcTagonist:
@apprb Well that's wrong. Hopefully new groups have fixed that by abstaining from doing it

apprb:
https://github.com/irungentoo/toxcore/blob/master/toxcore/group.c#L757-L801

The only way to get an empty nickname is to catch the moment of connecting/disconnecting of another user, but this will be automatically fixed by processing that user connecting/disconnecting event.

agilob:
Welcome back!

zetok:
pr
↓ after changing nick
changed nick
↓ after leaving & joining again
left & joined

↑ top (Away one) instance is with the PR applied, appears to just work™ :)

src/group.cpp
```diff
@@ -54,9 +54,11 @@ void Group::updatePeer(int peerId, QString name)
     toxids[toxid] = name;
 
     Friend *f = FriendList::findFriend(id);
-
     if (f != nullptr && f->hasAlias())
     {
+        peers[peerId] = f->getDisplayedName();
+        toxids[toxid] = f->getDisplayedName();
+    } else {
```
sudden6, line 61: Can you please fix the coding style, according to https://github.com/tux3/qTox/wiki/Coding

Else this is fine for me.

PR #2859, opened 2016-01-26
Add travis (again)

zetok:


zetok:
So, because of https://github.com/irungentoo/toxcore/issues/1508 it fails to build with clang - I guess that it should be removed until toxcore will be fixed?

zetok:
Ok, I got the osx part working.

Now the only thing left is to check whether my changes didn't break building & deploying qTox osx outside of travis. @RowenStipe ?

Main change is that now building script itself doesn't require `sudo`.

Issue #2852, opened 2016-01-24
Upside down video with UVC Webcam


Aleksandra13:
Hello, I found a problem with UVC Webcam on linux client of qTox, the video is captured upside down.
It is a known issue, and there is a common fix for that described here : http://ubuntuforums.org/showthread.php?t=838210&page=28
But when I tried the same (LD_PRELOAD=/usr/lib/i386-linux-gnu/libv4l/v4l2convert.so qtox) I got  following error:
[19:29:36.023] src/video/camerasource.cpp:209 : Debug: Opening device  "/dev/video0"
[video4linux2,v4l2 @ 0x4d1c9e0] Cannot find a proper format for codec 'mjpeg' (id 8), pixel format 'none' (id -1)
Assertion *codec_id != AV_CODEC_ID_NONE failed at libavdevice/v4l2.c:821
Is there any solution for that ? 
I have nightly build of qtox Tox Git commit: e93c35062114ba741f9d3831f6e1a57f55904f72.
Thanks, Aleksandra

zetok:
@Aleksandra13 you probably will want to read https://github.com/tux3/qTox/wiki/Video, it has instructions on how to vertically flip camera output.

If that "works" for you, feel free to close the issue.

In a case where instructions there may not be clear enough, could you point out problematic parts, so that they could be improved?

tux3:
> [video4linux2,v4l2 @ 0x4d1c9e0] Cannot find a proper format for codec 'mjpeg' (id 8), pixel format 'none' (id -1)

That looks like a fallout of the fix to #2826, turns out v4l will not fallback to an alternative if it can't provide what we ask. So we should try again without asking for a specific codec if mjpeg fails.

Aleksandra13:
Thank you! That helped. It would be nice to have such feature integrated int client.

PR #2867, opened 2016-01-26
generic pixel format selection

seankhl:
Now in the a/v drop down menu, it will show which pixel format it's using for taking data off the webcam. When you select either from the drop-down menu, it will use the correct one. So far, mjpeg, yuyv (raw), and h264 have been added, but only the first two have been tested since my webcam doesn't do h264 input. I googled around and I can't find any other formats listed, these three seem by far the most common. The interface is extensible so in the future, this can be done for more formats. And, I designed it so that it could easily be done for Windows/Mac in the future if necessary/beneficial. Finally, if the format is undetected, it will default to the same format it did before any of these changes were made, so the regression discussed in #2852 should be fixed.

Let me know if you have any questions or want any other changes to be made. As far as I can tell, the "short names" that the AVDictionary wants for the pixel formats are more or less arbitrary. Unfortunately there doesn't seem to be a good programmatic way of getting them. So I basically just made a lookup function in `v4l2.h/c`, which is where they belong, because they are v4l2/linux specific.

### DIFF
--- a/src/platform/camera/v4l2.cpp
+++ b/src/platform/camera/v4l2.cpp
@@ -177,7 +177,7 @@ QString v4l2::getPixelFormatString(uint32_t pixel_format)
     {
         return QString("h264");
     }
-    if (pixel_format == V4L2_PIX_FMT_MJPEG)
+    else if (pixel_format == V4L2_PIX_FMT_MJPEG)
     {
         return QString("mjpeg");
     }
--- a/src/platform/camera/v4l2.cpp
+++ b/src/platform/camera/v4l2.cpp
@@ -28,12 +28,33 @@
 #include <sys/ioctl.h>
 #include <linux/videodev2.h>
 #include <dirent.h>
+#include <map>
 
 /**
  * Most of this file is adapted from libavdevice's v4l2.c,
  * which retrieves useful information but only exposes it to
  * stdout and is not part of the public API for some reason.
  */
+
+static std::map<uint32_t,uint8_t> createPixFmtToQuality()
+{
+    std::map<uint32_t,uint8_t> m;
+    m[V4L2_PIX_FMT_H264] = 3;
+    m[V4L2_PIX_FMT_MJPEG] = 2;
+    m[V4L2_PIX_FMT_YUYV] = 1;
+    return m;
+}
+const std::map<uint32_t,uint8_t> pixFmtToQuality = createPixFmtToQuality();
+
+static std::map<uint32_t,QString> createPixFmtToName()
+{
+    std::map<uint32_t,QString> m;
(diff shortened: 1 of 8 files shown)


tux3:
I think that this is getting a bit too deep into technical details to be on the GUI.
The code looks good, but qTox should use the most convenient format automatically instead of asking the users to pick a codec. Users will probably not know the difference between YUYV, MJPEG and H264, and they shouldn't have to learn.

seankhl:
I thought you might say that...

To be honest, I think the fact that people can choose any frame rate at any pixel res already clutters the GUI a lot. For example, why would anyone pick a framerate below 10? It's much better to lower the resolution if you have a bad connection. And, many resolutions are redundant. I have over a dozen possible resolutions. maybe we should make it find the closest ones to the "standards" (360p, 480p, 720p, 1080p, etc) and display only those?

My solution for fixing this is going to be to code in an "ordering" for the pixel formats: h264>mjpeg>yuyv. Then, at any given resolution, it will use the best pixel format available.

agilob:
It could be a silent option, available in Settings class and file.ini but not in GUI so more technical users can experiment on their own. Such options must be properly documented.

tux3:
I agree about the framerates and resolutions too, I also think it's more complicated than it should be (some cameras list two pages worth of possible video modes!). 
Initially it was so people could pick a resolution that roughly matches their home connection (don't try to stream 1080p from the middle of nowhere!), but this is actually something we could and should handle transparently.

This is a separate issue, though.

seankhl:
They are somewhat related, since if I'm going to address the pruning of the video mode options list, I may as well do it in one fell swoop (i.e. with regard to the pixel formats as well as the FPSes, resolutions, etc).

I'm coming up with a solution. I can make a new pull request, or just commit here if you want. Let me know.

tux3:
Please fix the pixel format part for this PR, and you could start working on a more generic solution for FPS/resolution in another PR if you want. 
For FPS we can just always pick the maximum, but for the resolution things get a bit more complicated WRT dynamic video bitrate adjustements, so it would be best separated.

sudden6:
@seanlaguna Are you going to fixup this PR or should I take over?

seankhl:
Here, I've committed everything I've done. Feel free to take over.

It basically works, there are a few kinks to iron out. I don't they'll accept this though because it combines the format filtering (always pick h264 over mjpeg over yuyv) with the resolution filtering (find closest resolution to 1080p, 720p, 480p, 360p, 240p and only list those). I don't really feel like separating two perfectly fine contributions into separate PRs, so feel free to go ahead and do it, taking whatever you want from my code.

PR #2894, opened 2016-01-29
Proxy everything

### LINKED ISSUE #2643: Updates should be proxyfied
If we have a proxy set, put all our QNetworkAccessManagerS under proxy, including updates, toxme, etc.

sudden6:
Use the set proxy also for autoupdater and toxme lookups.

fixes #2643


src/net/autoupdate.cpp
```diff
@@ -108,6 +108,7 @@ AutoUpdater::VersionInfo AutoUpdater::getUpdateVersion()
 
     QNetworkAccessManager *manager = new QNetworkAccessManager;
     QNetworkReply* reply = manager->get(QNetworkRequest(QUrl(checkURI)));
+    manager->setProxy(Settings::getInstance().getProxy());
```
tux3, line 110: No! 
manager->get() will immediately start the request, at best you're racing to set the proxy, but it'll probably just not be proxied at all.

sudden6:
@tux3 fixed thanks

zetok:
@sudden6 btw, to avoid commits like 98f49c4494bb3c9a88ea2bb2fabde21d3e463e16, you can just simply rebase your branch - e.g. in this case it would be `git rebase master proxy-everything`

sudden6:
@zetok Thanks, does this also work if I apply it after a merge commit appears? e.g if I now do `git rebase master proxy-everything` the merge commit will disapear?

tux3:
No, the merge commit won't disappear now. You could attempt to remove the commit and then rebase again, but that's nontrivial.

sudden6:
ok, so this PR is ok now?

tux3:
LGTM

PR #2902, opened 2016-01-31
docs: put in INSTALL.md info about installing qTox from Arch repo

zetok:
Instruction should be tested by someone who runs arch.


zetok:
tested by @Impyy , thus pushing on master

PR #2923, opened 2016-02-04
Almost full translation into Ukrainian

azymohliad:
I translated almost everything into Ukrainian using Qt Linguist and [this guide](https://github.com/tux3/qTox/wiki/Translating). 

There are few things worth pointing out:
- **"Profile"** in Ukrainian is **"Профіль"**, but it has few more senses. I think more appropriate translation for "Profile" in this case is **"Обліковий запис"** (which exactly means **"Account"** in English). But I leaved **"Профіль"** everywhere because somebody already starts to translate this way and because the strings are shorter (it is matter for buttons like "Create profile" and so on). I can change it to **"Обліковий запис"** if needed.
- **"Load"** in Ukrainian is **"Завантажити"**, but it also means **"Download"**. So in case of "Log in" more appropriate would be **"Увійти"** (which is exactly **"Log in"**). But I also put **"Завантажити"** everywhere for the same reason - somebody has already started to translate it as **"Завантажити"**.

P.S. I am new to opensource contributing, so if I did something wrong please tell me:

### DIFF
--- a/translations/uk.ts
+++ b/translations/uk.ts
@@ -1,105 +1,88 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="uk_UA">
+<TS version="2.1" language="uk_UA">
 <context>
     <name>AVForm</name>
     <message>
-        <location filename="../src/widget/form/settings/avform.cpp" line="33"/>
         <source>Audio/Video</source>
         <translation>Аудіо/Відео</translation>
     </message>
     <message>
-        <location filename="../src/widget/form/settings/avform.cpp" line="83"/>
-        <source>Initializing Camera...</source>
-        <translation>Ініціалізація камери...</translation>
+        <source>%1x%2</source>
+        <translation></translation>
+    </message>
+    <message>
+        <source>Default resolution</source>
+        <translation>Роздільна здатність за замовчуванням</translation>
+    </message>
+    <message>
+        <source> at %1 FPS</source>
+        <translation> з %1 FPS</translation>
+    </message>
+    <message>
+        <source>None</source>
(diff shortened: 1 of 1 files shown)


zetok:
> **"Profile"** in Ukrainian is **"Профіль"**, but it has few more senses. I think more appropriate translation for "Profile" in this case is **"Обліковий запис"** (which exactly means **"Account"** in English). But I leaved **"Профіль"** everywhere because somebody already starts to translate this way and because the strings are shorter (it is matter for buttons like "Create profile" and so on). I can change it to **"Обліковий запис"** if needed.

Well, thing is, `Account` seems to mean that there's some account on a remote server, to which user logs in. This isn't the case with qTox - all of the profile data is located only on your local machine on which you are using qTox. I guess that `Профіль` should do fine?

> **"Load"** in Ukrainian is **"Завантажити"**, but it also means **"Download"**. So in case of "Log in" more appropriate would be **"Увійти"** (which is exactly **"Log in"**). But I also put **"Завантажити"** everywhere for the same reason - somebody has already started to translate it as **"Завантажити"**.

Basically same as above - at some point qTox had those buttons named `Log in`, but since it was causing confusion, it was renamed to `Load` - e.g. like when loading a local save in a game.

Dunno, do you think that `Завантажити` would cause confusion (i.e. make people think that they are downloading something from the internet when they're loading their profile)?

And thanks for your contribution, I'm going to merge it in a moment `:)`

azymohliad:
I thought about "Account" and "Log in" as terms of centralized model (with server), but wasn't sure. If English version avoid these words it already means someting. So yeah, you're right. 
And I think none of "Завантажити" and "Профіль" will cause the confusion. 
Thank you for quick reply and merge :)

Issue #2932, opened 2016-02-06
Refactoring: move Qt generated files into subdirectory


Talkless:
When compiling qTox _not_ in "shadow build" mode (Qt Creator), root project dir gets filled with ui*.h, moc*.cpp and *.o files, which is annoying.

I have prepared patch to use qmake's MOC_DIR, UI_DIR and others to move these generated files into tmp subdirectory.

Talkless:
#2933

tux3:
Closing, since the PR was rejected.

PR #2933, opened 2016-02-06
Move Qt generated files into subdirectory

Talkless:
When compiling qTox _not_ in "shadow build" mode (Qt Creator), root project dir gets filled with ui*.h, moc*.cpp and *.o files, which is annoying.

I have prepared patch to use qmake's MOC_DIR, UI_DIR and others to move these generated files into tmp subdirectory.

Addresses https://github.com/tux3/qTox/issues/2932

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ qtox
 build-*-Release
 build-*-Debug
 .DS_Store
+tmp
--- a/qtox.pro
+++ b/qtox.pro
@@ -320,6 +320,14 @@ contains(ENABLE_SYSTRAY_GTK_BACKEND, NO) {
 }
 }
 
+CONFIG(debug, debug|release):GENERATED_FILES_DIR=tmp/debug
+CONFIG(release, debug|release):GENERATED_FILES_DIR=tmp/release
+
+MOC_DIR=$${GENERATED_FILES_DIR}/moc
+OBJECTS_DIR=$${GENERATED_FILES_DIR}/obj
+RCC_DIR=$${GENERATED_FILES_DIR}/rcc
+UI_DIR=$${GENERATED_FILES_DIR}/ui
+
 !android {
     RESOURCES += res.qrc \
         smileys/smileys.qrc


agilob:
I made a PR once to fix it and @tux3 didn't like it :<
I'm for this change

sudden6:
I too like this idea

Zer0-One:
:+1:

antis81:
I'm against this option. The so-called "out of source" build setups are the only way of keeping your project dirs clean. Thus, I consider this request bad practice. However you can easily place a "build" subdirectory in project's main directory - this is, what lots of modern projects do like this. Btw. this can easily be achieved without using Qt-Creator, but just a terminal (or cmd on windows) instead:

``` bash
cd /path/to/project/qtox
mkdir build && cd build
qmake ../qtox.pro
```

Talkless:
@antis81 Now that you have said it, I remember using _build_ directory for number third party CMake projects, but not mine Qt projects for a long time. Old habits die hard I guess :-) .

antis81:
@Talkless Cool thing you understand me. :). I like to close it then, ok?

sudden6:
can this also be automated? So that qmake automatically creates a build subdirectory?

antis81:
@sudden6 The keyword is the qmake variable `OUT_PWD`. Documentation states not (never!) to alter this - unfortunately not "why?". Same is valid for cmake (CMAKE_BINARY_DIR)… Well, after some digging I found at least a bit of explanation from [cmake mailing list](https://cmake.org/pipermail/cmake/2009-November/033508.html).

However we can put the three lines in a `configure` shell script (I don't care about windows -> feel free to write a configure.bat or whatever :)). This is what some big projects do (like Qt for example :)). Probably we could provide a nice `-C ./build` option (like `make -C my-build-dir`). This can also be used to set the `QT_DIR` and `PATH` variables to locate the right qmake executable (especially when dealing with multiple Qt version setups).

Talkless:
@antis81 Well.. I guess it can be closed. Though looking at comments it seems that there is need for "automagics", but this is not _the_ solution.

Issue #2942, opened 2016-02-08
Remove “Filter audio” option from the settings


suhr:
Remove “Filter audio” option from the settings or make it inactive.

**Rationale:** At the present time (February 2016) filter_audio is badly broken. The purpose of filter_audio is improving audio quality, that means not only normalization&compression but also echo cancellation. But instead of removing echo it actually _adds_ it. Added echo is so loud, so it even causes a feedback effect with the Echobot. But though the only thing filter_audio does is ruining audio quality, this option still available in the settings.

**Advantages:** UX improvment (no more a confusing option to damage quality)

**Disadvantages:** None.

**Unresolved questions:** Can we have a reasonable alternative to filter_audio?

See also: https://github.com/tux3/qTox/issues/1655, https://github.com/tux3/qTox/issues/2126, https://github.com/tux3/qTox/issues/2756 and many other issues.

ovalseven8:
:+1:

sudden6:
Can't we just disable the echo cancelation algorithm of filter_audio?

Also I read on IRC that filter_audio works, if the input latency is set correctly, maybe add a slider in the ui?
Of course this will only be useful for power users and we should find a proper solution for this problem.

ProMcTagonist:
> if the input latency is set correctly

Why is it being set incorrectly? Is filter_audio broken or is qTox using it wrong? How does uTox use it?

zetok:
Well, IMHO until it gets fixed it should be disabled.

# 

@sudden6 

> Can't we just disable the echo cancelation algorithm of filter_audio?

Isn't that, like the only thing filter_audio is supposed to do?

sudden6:
@zetok 

quoting @suhr 

> The purpose of filter_audio is improving audio quality, that means not only normalization&compression,...

It seems to have other capabilities, altough I didn't check if they can be separately enabled/disabled.
EDIT: It seems to be possible, to only disable echo cancellation.

antis81:
Unfortunately I don't have the time to look deeper into that topic in near future. However, I had a quick look into code and it looks like the filter is not initialized correctly and as a consequence, it is used wrong by qTox.

If you're interested in fixing this issue, have a look at this function (and related code):

``` c++
bool AudioFilterer::enableDisableFilters(int echo, int noise, int gain, int vad)
```

[...]

PR #2951, opened 2016-02-12
Disable echo cancelation

sudden6:
It actually adds echo if the input latency is set incorrectly.
Should fix echo created by qTox, see also #2942

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -499,8 +499,6 @@ void Audio::doCapture()
         getEchoesToFilter(filterer, AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS);
 #endif
         filterer.filterAudio(buf, AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS);
-        // disable echo cancelation, because we don't know the delay
-        filterer.enableDisableFilters(0, 1, 1, 1);
     }
 #endif
 
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -66,6 +66,8 @@ Audio::Audio()
 
 #ifdef QTOX_FILTER_AUDIO
     filterer.startFilter(AUDIO_SAMPLE_RATE);
+    // disable echo cancelation, because we don't know the delay
+    filterer.enableDisableFilters(0, 1, 1, 1);
 #endif
 
     connect(&captureTimer, &QTimer::timeout, this, &Audio::doCapture);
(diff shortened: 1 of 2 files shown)


zetok:
> Why everybody hates doxygen?

Probably because it's hard to use.

mannol:
Doing echo cancellation with openAL and webRTC actually seems impossible because the delay is often changing and impossible to predict (tried it). Our best shot is using portaudio because it provides precise enough latency information about the device for echo to work (also tried it).

sudden6:
@mannol this patch should just be a temporary workaround, since this problem bothers many users.

antis81:
@zetok Really? It's pretty easy to use, widely spread and makes life easier for both a programmer to understand what code he/she actually wants to write (especially when writing an api) and the users of the api. Plus, it provides a great way to generate a technical documentation, that actually keeps itself up-to-date.

@sudden6 Remove the "filter audio" checkbox in a/v settings then. Please open a new PR, that can be merged into `audio-next` branch.

@mannol Yeah, nice. I was looking for an alternative to openAL, because focus there is on audio sources to easily produce 3D game sounds. Thus, recording API has just very basic support. To me it seems, this is the same situation with SDL. Looks like a low-level cross-platform C API suits our needs best.

agilob:
> Probably because it's hard to use.

@zetok I was using it before I learnt how to write a class.

zetok:
@antis81 Yes. What you've listed, is not why it's easy to use, but its features. Just because it has some features, it doesn't mean that it's easy to use.

sudden6:
@antis81 I think it would be better to rename the checkbox to "Optimize audio for speech" and just disable the faulty echo cancelation.

suhr:
> Our best shot is using portaudio because it provides precise enough latency information about the device for echo to work (also tried it).

I guess RTAudio can be uses as well.

mannol:
@suhr Wow, never heard about RTAudio before but, looks really good and provides latency estimations where possible. We should definitely jump on to it.

sudden6:
RTAudio looks very good, I also did some research and found libsound.io, which pretty much has the same featurese, but it doesn't look as mature as RTAudio.

Echo cancelation isn't the only problem, sometimes feedback happens even with this pull request applied. See my last comment in #2951 .

mannol:
I'm sure those issues are all connected with the audio engine and the way it's used.

sudden6:
Maybe, but I think we should open a new issue for this.

antis81:
@mannol Exactly. The issues were addressed to the `audio-next` branch and some of them got fixed and backported to master already. By changing the underlying API, the situation changes and it makes sense to evaluate RTAudio / PortAudio in another branch based on current master.

@sudden6 I'll remove this nasty checkbox in a/v settings, which only bothers so many people right now. In current state, audio calls work much better (or at all) without any filter applied. We now should evaluate the two libraries mentioned above and implement a real solution in a branch separated from `master`.

sudden6:
@antis81 good idea, so should I close this PR and you remove the checkbox?

antis81:
That's fine, yes.

PR #2957, opened 2016-02-15
Fixed the status tootlip issue #2868

### LINKED ISSUE #2868: &quot; is displayed instead of "" in tooltips
I have recently noticed that in the tooltips for user statuses it replaces any `"`s with `&quot;`. And I don't know if it does the same for any other characters as this is the only thing I've seen it for, though I haven't seen that many characters like this used in statuses or tooltips.

---

**OS Information:**

```
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 15.10
Release:    15.10
Codename:   wily
Flavour: GNOME
GNOME Version: 3.18
```

**Package Information:**

```
qtox:
  Installed: 469-c82e41f
  Candidate: 469-c82e41f
  Version table:
 *** 469-c82e41f 0
        500 https://pkg.tox.chat/debian/ nightly/release amd64 Packages
        100 /var/lib/dpkg/status
```

swrdfish:
Removed the `toHtmlEscaped()` method call to fix #2868

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,9 +2,9 @@
 
 ### Must read
 * If you aren't sure, you can ask on the [**IRC channel**](https://webchat.freenode.net/?channels=qtox) or read our [**wiki**](https://github.com/tux3/qTox/wiki) first.
-* Do a quick **search**. Others might already reported the issue.
+* Do a quick **search**. Others might have already reported the issue.
 * Write in **English**!
-* Provide **version** information: (You can find version numbers at menu `Settings → About`)
+* Provide **version** information: (You can find version numbers in menu `Settings → About`)
   ```
 qTox: 
 Commit hash: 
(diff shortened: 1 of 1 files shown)


TheNain38:
@ghoshbinayak Did you test the result? Because this `toHtmlEscaped()` method call is because the tooltips can't be set to `Plain` mode. If you set the status to: `<h1>TEST</h1>` it would display it as HTML and not just text.

swrdfish:
Sorry, I didn't know that. The tooltip actually does render the HTML tags. I'll look into it a bit more. 
Thanks for pointing it out :smiley:

swrdfish:
@TheNain38 So I looked into the issue a bit more.. and found something really strange. If there are any HTML tags in the QString then the tool tip renders the quotes just fine else it displays the QString as plain text. So I've just wrapped the message within <span> tags. Its a dirty fix though. Can you suggest any other way

TheNain38:
@ghoshbinayak It would be possible to use `toHtmlEscaped()` only when it contains HTML, if there was a way to know that...

swrdfish:
@TheNain38 Yes that seems a better option. I've made the necessary changes and things seem to work fine. Check the last commit and let me know if it needs further modifications.

TheNain38:
@ghoshbinayak And then they are other places in the code that would need this, each time `setToolTip()` is used. And also you could use the same code style for the ifs that only have one line in the condition (e.g. you could remove the `{`, `}`), like here: https://github.com/tux3/qTox/commit/9ace65444de554362c40949d303192f5267a1339#diff-39d71983c1fc16a6b3191ee87174baaeR893
And you could squash all the commits when you have finished :)

swrdfish:
@TheNain38 Ok I have made the necessary changes and have squashed the commits. But I had to force the push and that closed this pull request. 
Should I send a new pull request or, is there a way to reopen this request?
Sorry for the inconvenience, this is my first time contributing to an open source project.

swrdfish:
@TheNain38 it has been a few days since I created the new pull request and there are no comments yet. Did i do any thing worng?

TheNain38:
@ghoshbinayak No, it's all right, the only thing you could add is something like "fixes #2868" or "fix #2868" like it was here, so it would auto close the issue when it's merged :)

swrdfish:
@TheNain38 I have mentioned the issue no in the titlel of the new request like " fixes the tooltip issue #2868 " is that right or should i just use "fixes #2868" exactly. 

Btw the new request is here #2959

TheNain38:
@ghoshbinayak If you write "fix" "fixes" "close" "closes" with an issue number (#2868 for exemple) just after the word, then it will auto close the issue when the pull request is merged

PR #2958, opened 2016-02-16
Added Estonian translation

kaurman:
It's definitely not perfect but I'd say it is good enough to start using it. I apologize for the "false" commits, it's just that I'm new to using github and got a bit careless.


zetok:
Could you perhaps squash those commits?

And when squashing, also https://github.com/tux3/qTox/wiki/Translating#remove-obsolete-strings ?

kaurman:
Here's hoping that I didn't mess things up completely... Commit 60b02a9 should be the squashed one. Let me know if there's anything else I should/could do.

zetok:
@kaurman um, it's not squashed. Squashed commits would mean that instead of lots of commits there is only one single commit.

kaurman:
Hm... I thought that you can just cherry-pick 60b02a9 that should contain all the relevant commits. However, if this is not the case or just isn't the way things are done and I should get rid of the rest, I'll try and do that. Hints are welcome.

zetok:
> Hm... I thought that you can just cherry-pick 60b02a9 that should contain all the relevant commits.

I guess that I could see about that..

> However, if this is not the case or just isn't the way things are done and I should get rid of the rest, I'll try and do that. Hints are welcome.

It's nicer to not have to cherry-pick. If you want, to try.. :)

Squashing is done by `git rebase -i <commit before commits you want to squash>` and during interactive rebase you choose which commits you want to squash.

Though, with how currently things are, there is a merge commit in the way - namely 2328e15d08a71f2afd2128a66b915cc22c0650f7, which can mess up rebase, and this PR.

Probably the best way to go about it, is to checkout a new branch based on upstream master, apply changes there, and make a new PR:

```
git checkout original/master -b translation  # `translation` branch based on upstream master
# diff with changes before merge commit
git diff translation..60b02a9914f6268c592e1d3c09580adbe1cfe324 > translation.diff
patch -p1 < translation.diff  # apply changes
git add translations/et.ts
git commit -a
```

zetok:
> Hm... I thought that you can just cherry-pick 60b02a9 that should contain all the relevant commits. 

Actually, yeah, cherry-picking would be way faster, thanks for the PR `:)`

zetok:
Pushed dba6ebe262cbdce7bcab024ef4d3f0e41d826dc1 to master, thus closing this PR.

PR #2959, opened 2016-02-16
fixes the tooltip issue #2868

swrdfish:
Closed pull request #2957 and opened a new one after making the changes in a single commit.

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -138,7 +138,7 @@ ChatForm::~ChatForm()
 void ChatForm::setStatusMessage(QString newMessage)
 {
     statusMessageLabel->setText(newMessage);
-    statusMessageLabel->setToolTip(newMessage.toHtmlEscaped()); // for overlength messsages
+    statusMessageLabel->setToolTip(Qt::convertFromPlainText(newMessage, Qt::WhiteSpaceNormal)); // for overlength messsages
 }
 
 void ChatForm::onSendTriggered()
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -272,7 +272,7 @@ QDate GenericChatForm::getLatestDate() const
 void GenericChatForm::setName(const QString &newName)
 {
     nameLabel->setText(newName);
-    nameLabel->setToolTip(newName.toHtmlEscaped()); // for overlength names
+    nameLabel->setToolTip(Qt::convertFromPlainText(newName, Qt::WhiteSpaceNormal)); // for overlength names
 }
 
 void GenericChatForm::show(ContentLayout* contentLayout)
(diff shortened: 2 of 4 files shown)


zetok:
This PR doesn't fix it:
not fixed

swrdfish:
@zetok so, I found the problem, actually there are a few more places that require the same fix. I've now made the same fix to all other places I could find, except in the `src/widget/tool/croppinglabel.cpp` file where I'm getting the following error:

```
src/widget/tool/croppinglabel.cpp:129:12: error: ‘mightBeRichText’ is not a member of ‘Qt’
         if(Qt::mightBeRichText(origText))
```

Need some help here.

swrdfish:
@zetok I've found the solution to that problem and have pushed the required changes. Check if this fixes the issue.

zetok:
Yep, seems to be fixed.

Now someone who can into C++ should look at the code & merge it.

sudden6:
I didn't test this, but from reading the docs I think it would be easier to do something like this:

```
statusMessageLabel->setToolTip(convertFromPlainText(newMessage.toHtmlEscaped(), Qt::WhiteSpaceNormal)
```

So first, escape the message and then make it HTML again, so it does wordwrap.

swrdfish:
@sudden6 yes that should work as well, infact my first quick fix was somethng like this:
`statusMessageLabel->setToolTip( "<span>" + newMessage.toHtmlEscaped() + "</span>" )`

But @TheNain38 suggested that I first check if the message has HTML tags or not and then do the conversion.

sudden6:
@ghoshbinayak I think that `mightBeRichText()` isn't a good idea, because

> It mainly checks whether there is something that looks like a tag before the first line break. Although the result may be correct for common cases, there is no guarantee.

As this is a user controlled input, we should not trust it, to prevent possible attacks.

swrdfish:
@sudden6 yes that is true, but I think that this is what Qt uses before rendering a string in RichText. 

Consider the followinig two situations: 

 with only `&quote;`
screenshot from 2016-02-28 03-33-54

with `&quote;` plus `&lt;` and `&gt;`
screenshot from 2016-02-28 03-35-26
`mightBeRichText()` returns true for strings having `&lt;` or `&gt` but not `&quote;` 

One drawback with my method, as you have pointed out earlier, is that there is no text wrapping. And attacks.. I hadn't even thought of that.

I can change it to your proposed method if that is seems more appropriate to you.

sudden6:
@ghoshbinayak yes, please change it, because it displays exactly what the user entered + text wrapping.

swrdfish:
@sudden6 ok, I have made the required changes. See if it's ok. :)

zetok:
@ghoshbinayak could you remove that merge commit?

sudden6:
@ghoshbinayak For me, it's good.

swrdfish:
@zetok I have squashed the commits into one. I hope thats ok.

zetok:
@ghoshbinayak yep, thanks :)

tux3:
LGTM

PR #2974, opened 2016-02-23
Logging handler and log rotation cleanup.

### LINKED ISSUE #2920: qTox "Not responding" after update on OS X version 10.11.3
Application cannot start after Mac was updated to OS 10.11.3

qTox was installed from `qTox-OSX-1.2.4.307.dmg`

skotopes:
Bare metall logging handler which uses Qt formatting and plain C file IO. Minor additions in bootstrap-osx.sh

Fixes #2920.

### DIFF
--- a/bootstrap-osx.sh
+++ b/bootstrap-osx.sh
@@ -28,6 +28,8 @@ cp /usr/local/lib/libav* libs/lib
 cp /usr/local/lib/libswscale* libs/lib
 cp /usr/local/lib/libqrencode* libs/lib
 cp /usr/local/lib/libsqlcipher* libs/lib
+cp /usr/local/lib/libfilteraudio* libs/lib
+cp /usr/local/lib/libtox* libs/lib
 echo Copying include files...
 cp -r /usr/local/include/vpx* libs/include
 cp -r /usr/local/include/sodium* libs/include
@@ -35,5 +37,7 @@ cp -r /usr/local/include/qrencode* libs/include
 cp -r /usr/local/include/libav* libs/include
 cp -r /usr/local/include/libswscale* libs/include
 cp -r /usr/local/include/sqlcipher* libs/include
+cp -r /usr/local/include/filter_audio.h libs/include
+cp -r /usr/local/include/tox libs/include
 echo Done.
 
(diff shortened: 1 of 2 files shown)


bootstrap-osx.sh
```diff
@@ -28,6 +28,8 @@ cp /usr/local/lib/libav* libs/lib
 cp /usr/local/lib/libswscale* libs/lib
 cp /usr/local/lib/libqrencode* libs/lib
 cp /usr/local/lib/libsqlcipher* libs/lib
+cp /usr/local/lib/libfilteraudio* libs/lib
+cp /usr/local/lib/libtox* libs/lib
```
zetok, line 32: No.

filter_audio isn't even installed in the system, so trying to copy it from there is simply wrong.

On older version of scripts might have used to install it system-wide, I don't really remember.

src/main.cpp
```diff
@@ -130,36 +106,25 @@ int main(int argc, char *argv[])
...
         QDir dir (logFileDir);
-		
         // Check if log.1 already exists, and if so, delete it
-        if (dir.remove(logFileDir + "qtox.log.1"))
+        if (dir.remove(logFileDir + "qtox.log.1")) {
             qDebug() << "Removed log successfully";
-        else
+        } else {
             qDebug() << "Unable to remove old log file";
+        }
```
zetok: No. This too isn't needed: https://github.com/tux3/qTox/wiki/Coding#coding-style

bootstrap-osx.sh
skotopes, line 32: It is.
Anyway, process described in INSTALL.md is broken for mac os, so I'll recall this changes.

zetok, line 32: Perhaps - does https://github.com/tux3/qTox/blob/ef5b9b8c1e1a934d200743fbdc4ff0ed02cde023/osx/qTox-Mac-Deployer-ULTIMATE.sh work?

zetok:
Um, why closing PR?

skotopes:
What should I do instead?

zetok:
Make corrections?

PR #2977, opened 2016-02-25
fix(osx-script): remove unnecessary usage of `sudo`

zetok:
@RowenStipe does it work without `sudo` in those places?


RowenStipe:
@zetok initally I wasn't able to get it to build without `sudo` on my machine but now it does work without `sudo`

PR #2979, opened 2016-02-27
disable filteraudio, because it makes audio quality worse

### LINKED ISSUE #2942: Remove “Filter audio” option from the settings

sudden6:
Fixes  #2942 until we have a better audio backend

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -1239,7 +1239,8 @@ void Settings::setOutVolume(int volume)
 bool Settings::getFilterAudio() const
 {
     QMutexLocker locker{&bigLock};
-    return filterAudio;
+    // temporary disable filteraudio, as it doesn't work as expected
+    return false;
 }
 
 void Settings::setFilterAudio(bool newValue)
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -46,12 +46,15 @@ AVForm::AVForm() :
 
     bodyUI->btnPlayTestSound->setToolTip(
                 tr("Play a test sound while changing the output volume."));
-
 #ifdef QTOX_FILTER_AUDIO
     bodyUI->filterAudio->setChecked(Settings::getInstance().getFilterAudio());
 #else
     bodyUI->filterAudio->setDisabled(true);
 #endif
+
+    // temporary remove audio filtering, because it makes audio quality worse
+    Settings::getInstance().setFilterAudio(false);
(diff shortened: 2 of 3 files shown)


antis81:
@sudden6
This currently looks like a half-hearted attempt and the code actually introduces bad things…

Please remove the checkbox from the UI file instead of hiding it and make the settings flag always return `false` in `Settings::getFilterAudio()` - you can move the comment there.

sudden6:
@antis81 I didn't remove the checkbox, because I thought we want to reintroduce this at some point?

antis81:
Life is a process of creation and destruction - same with software. :) Right now I don't see anyone, who really cares about implementing this feature and - more important - the concept underneath. Why bother about a dead horse rather than care for the living one?

sudden6:
ok, I removed it. Or did you mean I should eliminate all traces of libfilter_audio?

zetok:
Err.

Wouldn't it be better to simply change `qtox.pro` so that by default filter_audio support won't be compiled?

sudden6:
@zetok without my changes the checkbox would still be displayed, but greyed out. Not user friendly.

antis81:
@zetok This leaves the checkbox as an artifact and is probably not the best solution at all.

zetok:
Right. I just thought that disabling filter_audio is supposed to be ~temporary, until it gets fixed. If that's not the case, nvm what I've said about just disabling it in `qtox.pro`.

tux3:
LGTM.
Although if nobody really wants to fix filteraudio, eventually we can start thinking about removing it entirely.
Even on uTox I haven't noticed a huge improvement with it on, so I wouldn' t be opposed.

ovalseven8:
> eventually we can start thinking about removing it entirely.

:+1:

PR #2980, opened 2016-02-28
fix #2976

### LINKED ISSUE #2976: Stores non-config data in ~/.config/tox: screenshots, log
qTox: v1.2-353-gef5b9b8
Commit hash: ef5b9b8c1e1a934d200743fbdc4ff0ed02cde023
toxcore: 72b1a17cc28c9a5a653823336bf0a495f6f271b1
Qt: 5.5.1
OS version: Fedora 23 x86_64

As per https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html,

> There is a single base directory relative to which user-specific configuration files should be written. This directory is defined by the environment variable $XDG_CONFIG_HOME.
> ...
> There is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME. 

As of today qTox creates the following files in ~/.config/tox:

```
qtox.ini
qtox.log
screenshots/qTox_Screenshot*.png
```

The log and screenshots are not configuration data, so they should be moved to `XDG_CACHE_DIR` or `QStandardPaths::CacheLocation`.

michals:
on linux logs goes to ~/.cache/...
and screenshots goes to ~/.local/share/...

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -129,7 +129,8 @@ int main(int argc, char *argv[])
     sodium_init(); // For the auto-updater
 
 #ifdef LOG_TO_FILE
-    QString logFileDir = Settings::getInstance().getSettingsDirPath();
+    QString logFileDir = Settings::getInstance().getAppCacheDirPath();
+    QDir(logFileDir).mkpath(".");
     logFileStream.reset(new QTextStream);
     logFileFile.reset(new QFile(logFileDir + "qtox.log"));
 
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -568,8 +568,7 @@ QString Settings::getAppDataDirPath()
     // TODO: change QStandardPaths::DataLocation to AppDataLocation when upgrate Qt to 5.4+
     //       For now we need support Qt 5.3, so we use deprecated DataLocation
     //       BTW, it's not a big deal since for linux AppDataLocation and DataLocation are equal
-    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::DataLocation)
-                           + QDir::separator() + "tox")+QDir::separator();
+    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::DataLocation))+QDir::separator();
 #endif
 }
 
@@ -587,8 +586,7 @@ QString Settings::getAppCacheDirPath()
     return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::HomeLocation) + QDir::separator()
                            + "Library" + QDir::separator() + "Application Support" + QDir::separator() + "Tox")+QDir::separator();
 #else
-    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::CacheLocation)
-                           + QDir::separator() + "tox")+QDir::separator();
+    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::CacheLocation))+QDir::separator();
 #endif
 }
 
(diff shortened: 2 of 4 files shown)


michals:
Hi guys, any update on this one? Is there any change needed or comment or tests?

zetok:
Could you make the code a bit more "vertical" ?

I haven't looked into what code does, but given that it's about file locations on all platforms, IMHO it would be good if PR was tested by users on those platforms.

zetok:
RE: testing
cc @RowenStipe , @towlie

sudden6:
I think we should use the `qtox` subdirectory for screenshots and logs.
Also testing is needed on Linux, Windows and Mac

michals:
I tested on Linux and it works correctly as requested in #2976 . 
It also does not break anything, because profile location is preserved.
Only consequences of update should be for Linux users: after update, new logs and new screenshots goes to new location (as recommended by freedesktop specs).

It should not affect other platforms, because for Windows and Mac exactly same code is executed (although, in other method). I didn't verify it on Windows nor Mac.

Previous code to determine path for screenshots and logs:
https://github.com/tux3/qTox/pull/2980/files#diff-5661f6b02abca6e3b7c75796b9a94ca8L535 

New code to determine path for screenshots:
https://github.com/tux3/qTox/pull/2980/files#diff-5661f6b02abca6e3b7c75796b9a94ca8R554
and logs:
https://github.com/tux3/qTox/pull/2980/files#diff-5661f6b02abca6e3b7c75796b9a94ca8R576

RowenStipe:
@zetok It works on OSX, also I agree with @sudden6 in that a qtox subdir should be used for things that are more specific to qTox; like screenshots, logs as well as, `.db`, and `.ini` files used only by qTox.

towlie:
This PR is linux only right? (the changes)

michals:
@towlie yes, it's Linux only. 

It makes linux version of qTox follow rules of storing app files in proper location as described in specs (referenced in issue #2976 )

sudden6:
@michals If you change the logs and screenshots directorys to `qtox` instead of `tox`, I think this would be ready to merge.

For the `.db` and `.ini` files a new PR would be better, becaus this potentially breaks things.

towlie:
I saw no problems on windows.

michals:
@sudden6 changed logs and screenshots directories to `qTox` instead of `tox`. Please note, that `/qTox` is added by `QStandardPaths`, so I just removed extra `/tox`.

sudden6:
With this merged, I can't take screenshots anymore.
qTox says "Failed to open temporary file" and then crashes. I suppose this is because the directory `.local/share/qTox` doesn't exist yet.

zetok:
> `.local/share/qTox`

I wonder if `qtox` wouldn't be better.

sudden6:
I wonder if there's some standard like the FHS for these directories.

zetok:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html

sudden6:
This document doesn't say anything about case-sensitivity.

zetok:
yep, that's why I'm wondering

michals:
For example, Gnome uses `~/.local/share/Trash/` with uppercase 'T'. 

The `qTox` is added automagically by `qstandardpaths`. It will require some quirks or changing the app name to have `qtox` instead of `qTox` in paths.

sudden6:
@zetok I see it the same as @michals , every program seems to use its name for this directory, so we should use `qTox`.

PR #2981, opened 2016-02-28
Fix toxURI parsing.

### LINKED ISSUE #2118: qTox shouldn't make invalid tox links
Tox links can be either `tox:‹toxid›` or `tox:‹name›@‹domain›`. Everything else should not be interpreted as links.

VittGam:
It was always eating two characters more than needed.

A toxURI passed to handleToxURI always starts with "tox:", it's checked in the handleToxURI callers. So here we need to remove that prefix, that's long 4 characters and not 6.

The previous code was meant to handle tox:// and tox: cases, but the tox:// case is invalid and should not be handled (see #2118).

Fixes: d6a0910fc8098ecb37d7f7fdfcdd2592adaac0ad ("fix uri parsing and tox URI detection, closes #2118")

### DIFF
--- a/src/net/toxuri.cpp
+++ b/src/net/toxuri.cpp
@@ -57,11 +57,7 @@ bool handleToxURI(const QString &toxURI)
     while (!core->isReady())
         qApp->processEvents();
 
-    QString toxaddr;
-    if (toxURI.startsWith("tox:"))
-        toxaddr = toxURI.mid(6);
-    else
-        toxaddr = toxURI.mid(4);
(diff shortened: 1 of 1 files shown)


zetok:
Thanks for the fix :)

On a side note, qTox does horrible, horrible stuff with URI `;_;`

VittGam:
You're very welcome :)

Yeah, maybe it should use a library for that... and for json too (in toxme.cpp) :sweat_smile:

PR #2994, opened 2016-03-08
Chatlog: Adjust whitespace behaviour/handling.

Chiitoo:
Essentially, this is just an undo of an undo: TheSpiritXIII@410277f4dc779d6801bc81b5b2b47b3d0fe2d6da

I've been building qTox with this for some weeks(?), using it daily on Linux, and tested somewhat on Windows.

The only issue thus far that I have discovered is:
- [ ] Tab-spacing acts odd in that, when copied in from outside qTox, they become twenty (20) characters wide (tested 2, 4, 8, and 16 characters wide tabulations).  I have no idea how simple that will be to fix.

Thoughts?

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -247,5 +247,5 @@ QString ChatMessage::detectQuotes(const QString& str, MessageType type)
 
 QString ChatMessage::wrapDiv(const QString &str, const QString &div)
 {
-    return QString("<div class=%1>%2</div>").arg(div, str);
+    return QString("<p class=%1>%2</p>").arg(div, /*QChar(0x200E) + */QString(str));
 }
--- a/ui/chatArea/innerStyle.css
+++ b/ui/chatArea/innerStyle.css
@@ -1,14 +1,18 @@
-div.msg {
+p {
+  white-space: pre-wrap;
+}
+
+p.msg {
   color: @black;
   font: @big;
 }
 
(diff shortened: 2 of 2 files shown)


zetok:
From what I remember, "fix" was reverted since @towlie experienced some problems with fonts with it.

Would that still be the case?

Chiitoo:
I remember that issue, but I have not bumped into it.

I wonder if it was related to the arabic/other fix that came with this one at the time?  In fact, the first time I tried this on Windows, it didn't seem to do anything at all (it didn't even fix the whitespace character, but I might have messed up something with the build that time).  Some time ago, when I tried it again, everything seemed to work fine there as well, and still do.

I personally tested it on Windows 7, and had a friend test it on Windows 10.

towlie:
Abou that font rendering issue. We didn't know what the issue is. Reverting did not help. It is something else from the whole PR.

zetok:
> Tab-spacing acts odd in that, when copied in from outside qTox, they become twenty (20) characters wide (tested 2, 4, 8, and 16 characters wide tabulations). I have no idea how simple that will be to fix.

They behave exactly the same in the input field on master qTox.

And actually, it would seem that it is master qTox that doesn't handle things correctly, since it modifies user input by converting tabs into spaces.

From what little testing that I did, it would seem that things are working right.

ProMcTagonist:
Thank you @Chiitoo and @sudden6.

Issue #390, opened 2014-10-06
Getting closer! qTox fails to compile!


dcharles525:
In file included from ./ui_mainwindow.h:28:0,
                 from widget/form/addfriendform.cpp:22:
./widget/adjustingscrollarea.h:28:46: error: expected ‘;’ at end of member declaration
./widget/adjustingscrollarea.h:28:48: error: ‘override’ does not name a type
./widget/adjustingscrollarea.h:29:30: error: expected ‘;’ at end of member declaration
./widget/adjustingscrollarea.h:29:36: error: ‘override’ does not name a type
widget/form/addfriendform.cpp: In constructor ‘AddFriendForm::AddFriendForm()’:
widget/form/addfriendform.cpp:40:13: error: ‘class QTextEdit’ has no member named ‘setPlaceholderText’
widget/form/addfriendform.cpp: In member function ‘bool AddFriendForm::isToxId(const QString&) const’:
widget/form/addfriendform.cpp:72:39: error: variable ‘const QRegularExpression hexRegExp’ has initializer but incomplete type
widget/form/addfriendform.cpp: In member function ‘QString AddFriendForm::getMessage() const’:
widget/form/addfriendform.cpp:88:43: error: ‘const class QTextEdit’ has no member named ‘placeholderText’
widget/form/addfriendform.cpp:89:1: warning: control reaches end of non-void function [-Wreturn-type]
widget/form/addfriendform.cpp: In member function ‘bool AddFriendForm::isToxId(const QString&) const’:
widget/form/addfriendform.cpp:74:1: warning: control reaches end of non-void function [-Wreturn-type]
make: **\* [addfriendform.o] Error 1

dubslow:
Huh, this is strange. What distro?

At a glance, that looks like you have the wrong Qt version. You need 5.2+. `qmake --version` will tell you what version of Qt is in use.

http://qt-project.org/doc/qt-5/qtextedit.html#placeholderText-prop

> This property was introduced in Qt 5.2.

dcharles525:
QMake version 3.0
Using Qt version 5.0.2 in /usr/lib/x86_64-linux-gnu

I do have 5.3 installed though (sorry for being so noob) but it doesn't seem to be registering that.

Oh and eOS

dubslow:
Hmm, some how you need to override 5.0.2 with 5.3.

`ls -l $(which qmake)`, and viewing the help on the linked-to file, suggests that 

[...]

PR #3000, opened 2016-03-12
docs(INSTALL.md): remove instructions for unsupported by qTox ubuntu

zetok:
Regarding https://github.com/tux3/qTox/issues/390#issuecomment-195701732

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,16 +20,13 @@
     - [Fedora](#fedora-other-deps)
     - [openSUSE](#opensuse-other-deps)
     - [Slackware](#slackware-other-deps)
-    - [Ubuntu <15.04](#ubuntu14-other-deps)
     - [Ubuntu >=15.04](#ubuntu-other-deps)
-  - [FFmpeg (Ubuntu <15.04)](#ffmpeg)
   - [toxcore dependencies](#toxcore-dependencies)
     - [Arch](#arch-toxcore)
     - [Debian](#debian-toxcore)
     - [Fedora](#fedora-toxcore)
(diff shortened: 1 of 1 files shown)


NicoHood:
Maybe you should clearly state that 14.04 is too old to compile the source and not only make it disappear. Thx for clarifying.

zetok:
> Maybe you should clearly state that 14.04 is too old to compile the source and not only make it disappear.

Well, technically, given enough of effort, it's possible to compile qTox on 14.04 - e.g. that's what is being done on travis. The problem that 14.04 has is that some libs are too old, and some are missing, which would make it really hard for someone to compile qTox there, since they would have to compile all missing dependencies first.

Thus, 14.04 is not really "supported" when it comes to compiling, but if someone wanted to, they could compile it there.

I.e. 14.04 is old enough to make compiling qTox there really problematic (thus unsupported), but not too old to completely exclude a possibility of qTox being compiled there.

With that being said, I'm in for removing it since there's crucial part that is needed for compiling not provided in INSTALL.md, and I don't think that anyone is willing to write & maintain one.

sudden6:
@zetok Maybe you could add the above in a shortened form into INSTALL.md? i.e. it is possible, but officially unsupported?

zetok:
@sudden6 I wonder if it's needed – i.e. no instructions for <15.04 should pretty much say that – and it's always possible to compile qTox on $distro, provided that there's enough effort put into compiling deps – and that falls into "unsupported" category by default.

sudden6:
that makes sense, ok then we just remove it

PR #3005, opened 2016-03-12
Update Bootstrap Nodes

ovalseven8:
- Removed 'noisykeyboard' because node is offline
- Added 'strngr' to have another node with port 443

(The node 'strngr' is from https://nodes.tox.chat)


zetok:
You might want to use https://github.com/zetok/nodes-parser-for-qtox

zetok:
Boostrapping from 104.167.101.29 33445 5918AC3C06955962A75AD7DF4F80A5D7C34F7DB9E1498D2E0495DE35B3FE8A57 noisykeyboard ...  FAILED ✗
Boostrapping from 195.93.190.6 443 FB4CE0DDEFEED45F26917053E5D24BDDA0FA0A3D83A672A9DA2375928B37023D strngr ...  FAILED ✗

Checked with https://github.com/zetok/tox-check-online-bootstraps

ovalseven8:
Strange.

@Impyy's [nodes.tox.chat](https://github.com/Tox/ToxStatus) says that strngr's node is working.

At least, the port is open:

```
PORT    STATE SERVICE
443/tcp open  https
```

alexbakker:
If nodes.tox.chat says a node is online, it means it was able to send a getnodes request and receive a sendnodes response in return.

zetok:
@Impyy how do you ensure that you don't have false positive results?

What & how you're using it?

The tool that I've written uses toxcore to bootstrap - if bootstrapping was successful within 20s from the moment when toxcore with given bootstrap nodes was started, bootstrap node is treated as working.
It can provide false positives if there's other tox instance running on the local network, due to toxcore using also lan discovery to bootstrap.

sudden6:
We will update the nodes list before the next release with a fresh list of working nodes.

Issue #2999, opened 2016-03-12
Markdown monospace support


zetok:
“It would be very cool if qTox could support code highlighting, or at least monospace and indentation. So, +1.”

by @xenohunter in https://github.com/tux3/qTox/issues/2497#issuecomment-164678436

e.g. some `monospace` markdown

sudden6:
wasn't there a "code" formatting already planned?

zetok:
@sudden6 

Kinda.

This issue is specifically for tracking just monospace support, since this is currently the only mode that can work without problems, considering that message splitting can mess up code blocks, and some sort of plan for handling that case should be in place for code blocks to be implemented.

I.e. for codeblocks, there should be opened a new issue, and there plan for handling things could be discussed.

tl;dr: planned - yep; concise and ready plan doesn't exist yet though.

ProMcTagonist:
How does message splitting affect code blocks in a way it doesn't affect other markup?

zetok:
> How does message splitting affect code blocks in a way it doesn't affect other markup?

Affects to higher degree, which makes it ~unacceptable. I.e. codeblock is way more likely to have thousands of characters which would trigger the message split than just a normal piece of text with some basic markdown.

All those problems stem from toxcore not doing the message splitting :/

PR #3008, opened 2016-03-13
Added codeblock support with ` char.

anoadragon453:
Changes font to monospace with a gray color. Wanted to do a different background as well but requires QString be embedded in one of Qt's [rich text processing APIs](https://doc.qt.io/qt-5/richtext.html). 

Markdown inside codeblocks is not triggered.

Addresses #2999.

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -193,6 +193,7 @@ QString ChatMessage::detectMarkdown(const QString &str)
                 "|(\\-)([^\\-]{2,})(\\-)"           // Underline  -text-
                 "|(\\~)([^\\~]{2,})(\\~)"           // Strike  ~text~
                 "|(\\~~)([^\\~\\~]{2,})(\\~~)"      // Strike  ~~text~~
+                "|(\\`)([^\\`]{2,})(\\`)"           // Codeblock  `text`
                 );
 
     // Support for multi-line text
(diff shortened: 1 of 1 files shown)


anoadragon453:
This however does not protect against long lines of code triggering a message split, as mentioned [here](https://github.com/tux3/qTox/issues/2999#issuecomment-195777960).

zetok:
line

Doesn't work when line ends with a `\n`. Other modes are affected by it as well.

anoadragon453:
Thanks for pointing that out, md checks for either an end-of-line or space to be parsed properly, if there's an `\n` being inserted, it doesn't match either of those criteria. I'll try to patch those up for all modes.

anoadragon453:
@zetok Newline issue fixed, I did notice that the parsing does accidentally succeed if we type something like:

```
*vv*'
```

when really it should fail as the apostrophe is there. 

Essentially it fails if a character is not present to the left of the markdown, while something _is_ present to the right.

src/chatlog/chatmessage.cpp
```diff
@@ -209,8 +210,10 @@ QString ChatMessage::detectMarkdown(const QString &str)
 
             QString htmledSnippet;
 
-            // Check for surrounding spaces and/or beginning/end of line
-            if (!((snipCheck.startsWith(' ') || offset == 0) && (snipCheck.endsWith(' ') || offset + snipCheck.trimmed().length() == out.length())))
+            qDebug() << "snip: " << snipCheck << " off+leng: " << offset + snipCheck.trimmed().length() << " out.len: " << out.length() << " offset: " << offset;
```
zetok: This debug statement isn't needed any longer

src/chatlog/chatmessage.cpp
```diff
@@ -209,8 +210,10 @@ QString ChatMessage::detectMarkdown(const QString &str)
...
+
+            // Check for surrounding spaces, newlines and/or beginning/end of line
+            if (!(((snipCheck.startsWith(' ') || snipCheck.startsWith('>')) || offset == 0) || ((snipCheck.endsWith(' ') || snipCheck.endsWith('<')) && offset + snipCheck.trimmed().length() == out.length())))
```
zetok: I wonder if that could be done differently

src/chatlog/chatmessage.cpp
anoadragon453: Good catch, I'll remove it.

src/chatlog/chatmessage.cpp
anoadragon453: It is quite confusing and scary, no?

I'll look into revising it.

anoadragon453: After trying to convert this to a singular QRegExp for the past three hours I can say this is the easier route.

sudden6:
@anoadragon453 can you fix this so your case from above doesn't parse as md?

anoadragon453:
@sudden6 Will get on that now.

src/chatlog/chatmessage.cpp
```diff
@@ -209,38 +210,39 @@ QString ChatMessage::detectMarkdown(const QString &str)
...
-            // Check for surrounding spaces and/or beginning/end of line
-            if (!((snipCheck.startsWith(' ') || offset == 0) && (snipCheck.endsWith(' ') || offset + snipCheck.trimmed().length() == out.length())))
+            qDebug() << snipCheck.toHtmlEscaped() << " offset: " << offset << "snipLength: " << snippet.toHtmlEscaped().trimmed().length() << " out.length: " << out.length() << " out: " << out;
+
+            // Only parse if surrounded by spaces, newline(s) and/or beginning/end of line
+            if ((snipCheck.startsWith(' ') || snipCheck.startsWith('>') || offset == 0) && ((snipCheck.endsWith(' ') || snipCheck.endsWith('<')) || offset + snippet.toHtmlEscaped().length() == out.length()))
             {
+                int mul = 0; // Determines how many characters to strip from markdown text
+                // Set mul depending on markdownPreference
+                if (Settings::getInstance().getMarkdownPreference() == 2)
```
sudden6: please replace this magic number

anoadragon453: Thanks, fixed.

anoadragon453:
All good to go :+1:

anoadragon453:
_Now_, all good to go for merge :+1:

ProMcTagonist:
It doesn't seem to be preserving whitespace.

nonpreserved whitespace

New issue or broken on my end?

zetok:
@ProMcTagonist well, e.g. gh markdown also doesn't seem to preserve whitespaces, although in a slightly different manner

` t e s t  ` → `t e s t`

Although in a mode without stripping it probably shouldn't strip them.. so new issue it is.

bignaux:
Confirmed bug here.
Interesting stuff :+1: 
https://doc-snapshots.qt.io/qt5-5.6/qtwebengine-webenginewidgets-markdowneditor-example.html
It would be neat to have preview in chatlog during typing.

anoadragon453:
@bignaux That would be nice. Not exactly sure where we'd put it in the current chat view. Move the chatlog up and display it just above the input box?

bignaux:
Yes, i was thinking to the same location as when you decide to press enter... with possibly the typing icon at right.

anoadragon453:
Would it automatically come up upon detecting markdown or would it be manually invoked? We already have two buttons next to send.

ProMcTagonist:
If there's a preview, it should appear above the textarea and update automatically.

But is a preview necessary? That leans toward a full formatting bar, and more clutter.

KISS. I'm against a preview.

anoadragon453:
It's easy enough to set a bit when markdown is successfully detected.

sudden6:
I'm with @ProMcTagonist I also don't really like the idea of a preview.
@anoadragon453 A new issue about not preserving whitespace is probably appropriate.

zetok:
@sudden6 it was fixed in #2994.

sudden6:
I thought this was about something else?

zetok:
> I thought this was about something else?
> 
> @anoadragon453 A new issue about not preserving whitespace is probably appropriate.

Um, #2994 was about fixing trimming of whitespaces.

PR #2983, opened 2016-02-29
Add direct link with template for submitting new issues

ghost:
It uses the template in https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md, filling all fields except "OS version".

There could be a more prominent place for that link as long as qTox is alpha.

PR #3010, opened 2016-03-13
Add issue template

rugk:
Using GitHubs [new feature](https://github.com/blog/2111-issue-and-pull-request-templates).

The template is based on the one included in the wiki, but I [styled](https://github.com/tux3/qTox/wiki/Writing-Useful-Bug-Reports/8bc3443e15e85fb5f23196c0c9d4fa481a77b871) it a bit so that it looks good in Markdown.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,29 @@
+Brief description:
+
+Operating System: Windows / OS X / Linux (include version and/or distro)  
+qTox version: (shown on `About` page in settings)  
+Hardware: 
+…
+
+Reproducible: Always / Almost Always / Sometimes / Rarely / Couldn't Reproduce
+
+
+**Steps to reproduce:**  
+1.  
+2.  
+3.  
+…
+
+
+**Observed Behavior:**  
+
+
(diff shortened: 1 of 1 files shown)


ghost:
There could be an official standard template, to coordinate this and #2983.

zetok:
@kehugter I wonder how (whether?) it would work together well..

For now, I think that it would be good to merge this PR, and once your PR gets needed changes, it could be somehow integrated.

What do you think?

rugk:
> I wonder how (whether?) it would work together well..

Did not tested, but I hope it works like this:
- if a user just clicks "New issue" on GitHub my template here is used
- if a user uses the link in qTox the content of this template is overwritten

ghost:
~~Actually, wouldn't this PR make the template in the link unnecessary?~~

rugk:
> make the template in the link unnecessary? 

Not really as the link template would auto fill some values AFAIK.

ghost:
Indeed, I remembered my previous reasoning 5 seconds after hitting "Comment". I'm a bit slow today.

ghost:
> Did not tested, but I hope it works like this:

For the record, it does.

PR #3033, opened 2016-03-21
update friend activity on groupchat events

### LINKED ISSUE #2848: Sorting "By Activity" isn't updated on invite to groupchat
Description:
When contacts are being sorted `By Activity`, and contact that is in `yesterday` or "older" activity group invites you to a groupchat, they aren't being moved to `today` group.

qTox 67620f96617312ccbb27c721c2c3090f036d52e3

Steps to reproduce:
1. Have a contact that is in `yesterday` or "older" activity group
2. Contact invites you to a groupchat

What should happen:
Contact is being moved to `today` activity group

What happened:
Contact was not moved from the old activity group

sudden6:
fixes #2848

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1418,6 +1418,8 @@ void Widget::copyFriendIdToClipboard(int friendId)
 
 void Widget::onGroupInviteReceived(int32_t friendId, uint8_t type, QByteArray invite)
 {
+    updateFriendActivity(FriendList::findFriend(friendId));
+
     if (type == TOX_GROUPCHAT_TYPE_TEXT || type == TOX_GROUPCHAT_TYPE_AV)
     {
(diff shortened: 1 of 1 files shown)


src/widget/widget.cpp
```diff
@@ -1508,6 +1513,10 @@ void Widget::onGroupPeerAudioPlaying(int groupnumber, int peernumber)
     if (!g)
         return;
 
+    ToxId author = Core::getInstance()->getGroupPeerToxId(groupnumber, peernumber);
+    if(!author.isSelf())
+        updateFriendActivity(FriendList::findFriend(author));
+
```
zetok: Actually, this is rather invalid.

In current groupchats it's doable, but in new groupchats it will not be possible, thus it's better to not introduce functionality that later on will be removed.

When we've talked about it yesterday, I've thought that groupchats are also being sorted `By Activity` when `Place groupchats at top of friend list` is not set. Since this is not the case, and groupchats aren't a subject to sorting `By Activity`, docs need to be corrected.

Also, I wonder how `FriendList::findFriend()` would handle PKs that aren't actually in the friendlist.

zetok:
Fixes #2848 for me.

staplemac:
Thanks for your contribution! This application is better than without you guys. :+1:

PR #3045, opened 2016-03-23
Stop emoji window disappearing upon emoji selection.

### LINKED ISSUE #3043: [Feature Request] Keep emoticon option open until manually deselect
Brief description: I and some of the people I chat with enjoy fooling around with streams of emoticons/emojis in lieu of actual text. To make this fun more easy, it would be nice to have the dialog remain open so I can spam a whole bunch to my buddy instead of having to open the dialog and find the page I was on again.

Operating System: Arch Linux (Linux 4.4.5-1)  
toxcore (pacman package) version: 3747-1 
qTox (pacman package) version: 1.3.0-1
Qt version: 5.5.1
Hardware: HP G61 laptop
AMD Athlon(tm) II Dual-Core M300 × 2 
Gallium 0.4 on AMD RS880 (DRM 2.43.0, LLVM 3.7.1)

Reproducible: Always

**Steps to reproduce:**  
1.  Click smiley face beside the text window to open emoticon dialog
2.  Click any icon

**Observed Behavior:**  
Emoticon dialog closes and emoticon is pasted into the text window 

**Expected (Desired) Behavior:**  
Emoticon dialog remains open, and the selected emoticon is pasted into the text window

**Additional info:**  
This is the default functionality in Skype, and makes it quite fun to play around with the emoticons. Currently, the emoticon dialog can be deselected by clicking outside of it, so there is already an escape from the potential "problem" posed by keeping the dialog open implemented.

anoadragon453:
Fixes #3043.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,7 @@ OS version:
 # Opening a pull request
 
 ### Must read
+* Use **[commit message format](#commit-message-format)**.
 * Read our [**coding guidelines**](https://github.com/tux3/qTox/wiki/Coding).
 * Keep the title **short** and provide a **clear** description about what your pull request does.
 * Provide **screenshots** for UI related changes.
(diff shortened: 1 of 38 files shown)


zetok:
While it does close the issue, I'm not sure if "clicking away" should be the only way in which emoticon selector can be closed.

I think that being able to select some emoticons and press "enter" to close selector & send message would be better.

anoadragon453:
Indeed, as the cursor in the chat window is still blinking.

anoadragon453:
@zetok That could be another issue though.

zetok:
> @zetok That could be another issue though.

Not really, given that it's about modifying current workflow, and not adding a new yet-to-be-extended functionality.

zetok:
Actually, I wonder if it shouldn't disappear if any key is being pressed, without catching they key - i.e. whatever key would be pressed, emoticon selector would close, and the key press would be forwarded to the input field.

anoadragon453:
Since the cursor is still blinking I believe users would expect that behavior, yes.

sudden6:
@anoadragon453 can you implement the behavior @zetok described?

anoadragon453:
@sudden6 KeyEvents passed to the emoji window, other than key_ESC, should be passed to the text entry widget. I've been pretty busy lately, so if someone else wants to implement the rest of this PR don't hesitate to, otherwise I'll finish it up eventually™

zetok:
Note: required changes were done in #3182

zetok:
Closing, since #3182 was merged.

Issue #2985, opened 2016-03-01
l10n needs to be abstracted out of HTML


minj:
Please prevent useless style changes from affecting l10n. Otherwise I have to deal with such nonsensical diffs:

``` diff
diff --git a/translations/lt.ts b/translations/lt.ts
index 3886cb8..0fedef3 100644
--- a/translations/lt.ts
+++ b/translations/lt.ts
@@ -138,26 +134,6 @@ Jei Jūsų interneto ryšys yra per prastas, turėsite keblumų su vaizdo skambu
         <translation>Licencija</translation>
     </message>
     <message>
-        <source>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
-&lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
-p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:&apos;Noto Sans&apos;; font-size:10pt; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-family:&apos;Sans Serif&apos;; color:#000000;&quot;&gt;Copyright © 2014-2015 by The qTox Project&lt;/span&gt;&lt;/p&gt;
-&lt;p style=&quot; margin-top:12px; margin-bottom:12px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-family:&apos;Ubuntu&apos;;&quot;&gt;qTox is a Qt-based graphical interface for Tox.&lt;/span&gt;&lt;/p&gt;
[...]

PR #3059, opened 2016-03-26

ovalseven8:


translations/de.ts
```diff
@@ -8,7 +8,7 @@
         <translation>Audio/Video</translation>
     </message>
     <message>
-        <source>%1x%2</source>
+        <source>%1p</source>
```
sudden6, line 11: This should probably be not translateable.

translations/de.ts
```diff
@@ -275,6 +254,22 @@ Viele Grüße,
 %1</translation>
     </message>
     <message>
+        <source>Add a friend</source>
+        <translation>Einen Kontakt hinzufügen</translation>
```
sudden6: don't we normally use "Freund" over "Kontakt"?

translations/de.ts
```diff
@@ -780,6 +775,14 @@ Disabling chat history now will leave the encrypted history intact (but not usab
         <translation>Chat aus diesem Fenster entfernen</translation>
     </message>
     <message>
+        <source>To new group</source>
+        <translation>Zu einer neuen Gruppe</translation>
+    </message>
+    <message>
+        <source>Invite to group &apos;%1&apos;</source>
+        <translation>Einladen in Gruppe „%1“</translation>
```
sudden6: "In Gruppe "%1" einladen" ?

translations/de.ts
```diff
@@ -1259,6 +1262,33 @@ werden übertragen, sobald ihr beide online seid.</translation>
     </message>
 </context>
 <context>
+    <name>GroupInviteForm</name>
+    <message>
+        <source>Invited by &lt;b&gt;%1&lt;/b&gt; on %2.</source>
```
sudden6, line 1268: HTML removal needed

#2985

translations/de.ts
ovalseven8: Not necessarily, it depends on the phrase. But I changed it. :)

translations/de.ts
ovalseven8: Ok...

PR #3060, opened 2016-03-26
added missing dependency to libsqlcipher-dev in simple_make.sh

quatrox:


sudden6:
Could you add `sqlcipher-devel` for openSUSE also?

zetok:
> But, that fix will come in a separate commit, as I have to figure out the correct dependency first.

Well, fedora doesn't seem to package sqlcipher. That's why INSTALL.md has a section for compiling it.

quatrox:
Please note that I do not have any computers running OpenSUSE at the moment, but a quick search on the Internet indicated that the package name was correct.

sudden6:
I should have added that this comes from my openSUSE system.

PR #3068, opened 2016-03-27
Update The qTox Project copyright years

ghost:
Update all to 2016. Add year span if missing.

Third party copyrights left untouched, including those directly involved
with the qTox Project (eg Tox Foundation, RowenStipe)


zetok:
> Update all to 2016. Add year span if missing.

No. I.e. copyright doesn't work like that.

What can be done, is increase year in a case where source file was actually edited in 2016.

Related: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

anoadragon453:
...what if editing the year _is_ editing the file?

Additionally we could just write a line to update the year if the was modified in 2016, but we would have to change the year constantly if we edited a file without it. Might just be easier to edit all at once.

zetok:
> ...what if editing the year is editing the file?

Doesn't apply. What matters, is the change to copyrightable stuff - i.e. the code. Also note that copyright date might not be "updateable" if change to the content wasn't "significant". Gotta love the copyright ;s

ghost:
> increase year in a case where source file was actually edited in 2016

I don't know... wouldn't that depend on the triviality of the edit? Will that supersede the previous year in some cases, leaving files without copyright? I'm unsure about everything right now.

> Gotta love the copyright ;s

Law, in general. It lacks of any logic and beauty, unlike surrealism.

This PR can be closed, I think. The year spans in some files are probably wrong, but it might be worse to _fix_ them. I'm too ignorant in copyright issues to get it right by myself alone anyway, so it'll be faster if someone else starts from scratch (if needed).

zetok:
> I don't know... wouldn't that depend on the triviality of the edit?

Yep, it kinda does. Generally "edit" is most likely enough, since almost always source gets modified in a way beyond "trivial" (`[citation needed]`). With that in mind, e.g. fix to a typo in a comment wouldn't really count as such.

> Will that supersede the previous year in some cases, leaving files without copyright?

Nope. Sorry, what I wrote was unclear - "increase" is not the right word - you just need to add the new year in which something was modified. Range of years (e.g. 2000-2010) is just used for simplification. Copyright date is applied on per-file basis.

e.g.
© 2010, 2012-2014 + modified in 2015 → © 2010, 2012-2015
↑ equal to ↓
© 2010, 2012, 2013, 2014, 2015

another example
© 2014 + modified in 2016 → © 2014, 2016

> Law, in general. It lacks of any logic and beauty, unlike surrealism.

:D

> I'm too ignorant in copyright issues to get it right by myself alone anyway, so it'll be faster if someone else starts from scratch (if needed).

No one knows about everything ;)
You know, I've learned some of that stuff exactly by sending PR(s?) with update to copyright dates. :D

Dunno, perhaps it would be faster to start from scratch.

With that being said.. sorry, generally updating copyright dates isn't really that important, or needed, given that it's more beneficial to have the oldest possible copyright claim.

Just gotta say, I don't agree with your opinion that you're too ignorant to do it. So if you feel like you want to do it - please do so. If you'll have a problem, would want an opinion, or something, you can always pop in on IRC, or make an issue (or something) :)

As a general rule of thumb, if you're unsure if change is significant, you probably should error on the side of not updating the date.

ghost:
Everything clear, thanks for the detailed explanation. And there's no need to be sorry, it took only a couple of minutes + trivial regex, and I even learnt something.

PR #3020, opened 2016-03-17
Update screenshot in README

Doom032:
I think it's more appropriate to have a screenshot without the ex-developer who is featured in the current one. He doesn't exactly bring the best of thoughts.

PR #3071, opened 2016-03-28
change picture

ProMcTagonist:
#3020

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Builds other than installer/packages don't receive updates automatically, so mak
 Note: The screenshots may not always be up to date, but they should give a good idea of the general look and features.
 
 
-![Screenshot 01](https://i.imgur.com/VRqXMD7.png)
+![Screenshot 01](https://i.imgur.com/wq3WpID.png)
 ![Screenshot 02](https://i.imgur.com/tmX8z9s.png)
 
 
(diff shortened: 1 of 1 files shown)


zetok:
There's no `Unread messages` that screen in #3020 had, and no groupchat shown.

Additionally, with recent merge that changed how groupchat invites are treated, groupchat invite also got "notification".

I think that it would be good to add both to screenshot?

And remove that `null` from my nick, since usually users won't get to ever see that? I.e. even when talking with me..
no null

Edit: and when file transfer widget will be fixed, screenshot will have to be fixed again, but that'll happen some time later.

ProMcTagonist:
gimme screenshots of both  in your same-sized qtox window and I'll gimp them in

zetok:
> gimme screenshots of both in your same-sized qtox window and I'll gimp them in

Er, I don't have the same size of the window anymore.. and AFAIK it can done without that by just copypasting (at least in krita).

ProMcTagonist:
I just prefer not having to resize because it's cleaner. But it'll be fine. 

Post some screenshots of the features you want featured. I can't compile latest qTox at the moment.

zetok:
> I just prefer not having to resize because it's cleaner.

No idea about that, when I was doing stuff in krita copypasting was identical to the pixel.

Anyway;

groupchat invite
and the rest of "features" (groupchat, 2 types of a new message notification (etc?)) are in https://gist.github.com/zetok/6af729e201a1fae5145a

ProMcTagonist:
qtox zetok mockup

ProMcTagonist:
qtox zetok mockup

PR #3044, opened 2016-03-23
Removed uneccessary for loop in markdown and url parsing.

anoadragon453:
We don't need to parse every line, just do one pass over.

Works in my testing, please test before merging.

PR #3077, opened 2016-03-29
Prevent regex capture from extended past line boundaries.

### LINKED ISSUE #3075: [regression] No longer correctly processed messages to a local folder
Brief description:

Operating System: Windows / OS X / Linux (include version and/or distro)  
qTox version: 1.3
Reproducible: Always 
after #3044

**Steps to reproduce:**  

send a multiline message, for example

```
text 
file:///D:/QtProgects 
text
```

**Additional info:**  
1. link to the catalog will not work because you do not handle the end of line
2. the reference was obtained by dragging the folder to set field message

anoadragon453:
Fixes #3075.

Thanks to @PKEv for the test case.

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -249,7 +249,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(?:\\b)(file:///.*$)" //link to a local file, valid until the end of the line
+                "|(?:\\b)(file:///)([\\S| ]*)" //link to a local file, valid until the end of the line
                 "|(?:\\b)(tox:[a-zA-Z\\d]{76}$)" //link with full user address
                 "|(?:\\b)(mailto:\\S+@\\S+\\.\\S+)" //@mail link
                 "|(?:\\b)(tox:\\S+@\\S+)"); // starts with `tox` then : and only alpha-digits till the end


zetok:
Hm, that doesn't seem to fix it for me
not fixed

anoadragon453:
That's odd, that's the first case I tested. I'll check it in a bit.

anoadragon453:
screenshot_20160331_072418

@zetok Can you post the exact text you're entering?

zetok:
I'll just copypasted text from the #3075

Edit:

```
text 
file:///D:/QtProgects 
text
```

anoadragon453:
@zetok Yeah working fine here. It says 252c9c8fae35f0258381537fee444b5eb02a9d7c as your commit hash in settings, correct?

zetok:
Nope. I've merged in in local branch on top of current master.

zetok:
Huh, weird, looks like I might have tested the wrong build. Whoops, sorry. Now it works.

nurupo:
@zetok sorry, don't have time for that. I'd suggest playing with the regex in https://regex101.com . It can break regex them down explaining what each part does and do many other helpful thing.

PKEv:
found a mistake:
link type:
file://192.168.3.15/box/!soft/MyPhoneExplorer_Setup_v1.8.7.exe
don't work too

anoadragon453:
Probably the exclamation point, I can add that to the regex I suppose. Should also check if it's the period, since it's looking for non-word characters.

Will probably rewrite it to just look for EOLs and let everything else through. I also noticed in my own testing that it doesn't allow @ signs, which the Linux fs does allow in filenames.

Edit: Nevermind it works, see below.

anoadragon453:
@PKEv Seems you're missing a `/` after `file:`. Works fine on my end with that added.

src/chatlog/chatmessage.cpp
```diff
@@ -249,7 +249,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(?:\\b)(file:///.*$)" //link to a local file, valid until the end of the line
+                "|(?:\\b)(file:///)([\\S| ]*)" //link to a local file, valid until the end of the line
                 "|(?:\\b)(tox:[a-zA-Z\\d]{76}$)" //link with full user address
```
zetok, line 253: remove also `$`, that'll fix another regression caused by #3044

Issue #2989, opened 2016-03-05
Bug: Audio call button is disabled for group on Linux


1KoT1:
1) Create group.
2) Add some people to group.
Actual: Audio call button is disabled
Expectation: Audio call button is enabled and I can call to gruop.

Additional info:
1) Audio call of one person is valid.
2) Audio call of group on Windows is valid.
3) When star qTox from QtCreator I see in output: [22:09:05.287] :0 : Warning: QMetaObject::invokeMethod: No such method Audio::playGroupAudio(int,int,const int16_t*,unsigned,uint8_t,unsigned)
I check Audio class. It actually hasn't method playGroupAudio

anoadragon453:
Looks like Audio::playGroupAudio was removed in ce2f8fd1d5948b6df6769ece4568d52040ce36fb. Correct me if I'm wrong, but it looks like [this line](https://github.com/tux3/qTox/blob/c82e41ff6f029dacb92ca32e8eac144505339082/src/audio/audio.cpp#L365) is still trying to call it.

1KoT1:
> Correct me if I'm wrong, but it looks like [this line](https://github.com/tux3/qTox/blob/c82e41ff6f029dacb92ca32e8eac144505339082/src/audio/audio.cpp#L365) is still trying to call it.

You are right. I debug qTox. And I get Warning on this line.

> Looks like Audio::playGroupAudio was removed in ce2f8fd

Thank you for information.

feeef:
I confirm this bug. It was working on a previous release but after a recent upgrade of qtox it stopped working for me as well.

This is a feature that I really need as I use tox to do audio conferencing with my collaborators.

thanks for your help!

gustavklopp:
+1
@anoadragon453  How come it's working for windows and not linux since this function was deleted also for Windows code?

anoadragon453:
@gustavklopp You sure you're running the latest nightly build on Windows?

gustavklopp:
@anoadragon453  So it's disabled for Windows too? Because No, I don't have qtox installed on Windows platform, only Linux

anoadragon453:
I've got a windows partition, I'll test it when I get home.

anoadragon453:
I'm getting some odd behavior. Here's my setup: 
- qTox 791d871 Win8.1
- qTox 49b8a48 Arch Linux
- qTox db812fe Arch Linux

Putting them all into a groupchat together, I get the following:

[...]

PR #3084, opened 2016-03-30
fix groupcalls

sudden6:
change the condition for when the audiocall button is enabled

affects #2989

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -354,20 +354,6 @@ void Audio::playMono16Sound(const QByteArray& data)
     playMono16Timer.start(durationMs + 50);
 }
 
-/**
-@brief May be called from any thread, will always queue a call to playGroupAudio.
-
-The first and last argument are ignored, but allow direct compatibility with toxcore.
-*/
-void Audio::playGroupAudioQueued(void*,int group, int peer, const int16_t* data,
-                        unsigned samples, uint8_t channels, unsigned sample_rate, void* core)
-{
-    QMetaObject::invokeMethod(&Audio::getInstance(), "playGroupAudio", Qt::BlockingQueuedConnection,
-                              Q_ARG(int,group), Q_ARG(int,peer), Q_ARG(const int16_t*,data),
-                              Q_ARG(unsigned,samples), Q_ARG(uint8_t,channels), Q_ARG(unsigned,sample_rate));
-    emit static_cast<Core*>(core)->groupPeerAudioPlaying(group, peer);
-}
-
 void Audio::playAudioBuffer(ALuint alSource, const int16_t *data, int samples, unsigned channels, int sampleRate)
 {
     assert(channels == 1 || channels == 2);
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -364,6 +364,49 @@ void CoreAV::volMuteToggle(uint32_t callId)
         calls[callId].muteVol = !calls[callId].muteVol;
 }
 
+/**
+@brief Called from Tox API when group call receives audio data.
+
+@param[in] tox          the Tox object
+@param[in] group        the group number
+@param[in] peer         the peer number
+@param[in] data         the audio data to playback
+@param[in] samples      the audio samples
+@param[in] channels     the audio channels
+@param[in] sample_rate  the audio sample rate
+@param[in] core         the qTox Core class
+*/
+void CoreAV::groupCallCallback(void* tox, int group, int peer,
+                               const int16_t* data, unsigned samples,
+                               uint8_t channels, unsigned sample_rate,
+                               void* core)
+{
+    Q_UNUSED(tox);
+
+    Core* c = static_cast<Core*>(core);
+    CoreAV* cav = c->getAv();
+
+    if (!cav->groupCalls.contains(group)) {
+        qWarning() << "Audio for invalid group id" << group
+                   << "will not be played back.";
+        return;
+    }
+
+    ToxGroupCall& call = cav->groupCalls[group];
+
+    emit c->groupPeerAudioPlaying(group, peer);
+
+    if (call.muteVol)
+        return;
+
+    Audio& audio = Audio::getInstance();
+    if (!call.alSource)
+        audio.subscribeOutput(call.alSource);
+
+    audio.playAudioBuffer(call.alSource, data, samples, channels,
+                          sample_rate);
+}
+
 VideoSource *CoreAV::getVideoSourceFromCall(int friendNum)
 {
     if (!calls.contains(friendNum))
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -386,9 +386,8 @@ void CoreAV::groupCallCallback(void* tox, int group, int peer,
     Core* c = static_cast<Core*>(core);
     CoreAV* cav = c->getAv();
 
-    if (!cav->groupCalls.contains(group)) {
-        qWarning() << "Audio for invalid group id" << group
-                   << "will not be played back.";
+    if (!cav->groupCalls.contains(group))
+    {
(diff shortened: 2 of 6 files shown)


zetok:
Well, that does fix #2989 for me.

Now, the problem that I have is that client's interface hardlocks when I'm trying to end the call. 

```
[16:22:47.235] src/audio/audio.cpp:292 : Debug: Opened audio output "Built-in Audio Analog Stereo"
[New LWP 5818]
[16:22:47.236] src/audio/audio.cpp:561 : Debug: Audio source 2 created. Sources active: 1
[16:22:47.236] src/audio/audio.cpp:181 : Debug: Subscribed to audio input device [ 2 subscriptions ]
[16:22:47.236] src/audio/audio.cpp:197 : Debug: Unsubscribed from audio input device [ 1 subscriptions left ]
[16:22:47.238] src/video/groupnetcamview.cpp:42 : Debug: Created expanding?  false
[16:22:47.238] src/video/groupnetcamview.cpp:42 : Debug: Created expanding?  true
[16:22:47.241] src/video/groupnetcamview.cpp:42 : Debug: Created expanding?  true
[16:22:53.196] src/core/coreav.cpp:388 : Debug: "Leaving group call 0"
[16:25:40.882] src/core/core.cpp:522 : Debug: "Group namelist change 0:1 1"
[16:26:17.182] src/core/corefile.cpp:285 : Debug: "Received avatar request 2:65536, reject, since we have it in cache."
```

anoadragon453:
This does not always fix the problem. I'm getting audio button enabled on groupchats intermittently now. When it is enabled I have the same issue as @zetok above where the client crashes upon ending the call with no useful debug information.

Additionally I noticed when a user leaves a chat, their avatar does not disappear in the avatar/visual-overview area, though that can be reserved to another issue. 

Edit: Huh, when I exited the Windows client, the audio button became enabled. Windows is not on this PR commit, but a couple back. The Linux are of course.

I also got a `src/core/core.cpp:955 : Warning: getGroupPeerName: Unknown error` message.

---

After a bit more testing if I only add clients that are on this commit then the call enabled button seems to work properly.

anoadragon453:
While this enables the call button, calls are still not functioning correctly, the other user does not even get an indication that someone has initiated a group call.

Step in the right direction though.

GitCop:

 * Commit: 49c8034fdf0afac4ab4522e08f20d11ab87c9382
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 487e6f597349c46dab25c4479052ca3b527f3103
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

 * Commit: 987cb16cc8bb4a204e11d752a6794142de3217ad
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at

---

src/audio/audio.cpp
```diff
@@ -368,6 +368,35 @@ void Audio::playGroupAudioQueued(void*,int group, int peer, const int16_t* data,
...
+{
+    assert(QThread::currentThread() == audioThread);
+    QMutexLocker locker(&audioLock);
+
+    if (!CoreAV::groupCalls.contains(group))
+        return;
+
+    ToxGroupCall& call = CoreAV::groupCalls[group];
+
+    if (call.inactive || call.muteVol)
```
antis81: I'll copy the `call.inactive` into #3183, thus it does the same thing - just the right way. :)

src/audio/audio.cpp
```diff
@@ -368,6 +368,35 @@ void Audio::playGroupAudioQueued(void*,int group, int peer, const int16_t* data,
...
+        return;
+
+    qreal volume = 0.;
+    int bufsize = samples * 2 * channels;
+    for (int i = 0; i < bufsize; ++i)
+        volume += abs(data[i]);
+
+    emit groupAudioPlayed(group, peer, volume / bufsize);
```
antis81: Hm… not sure if we actually need this event anymore - to me it looks wrong. I'll do some more research and see what it was actually used for.

sudden6: it was probably intended to indicate who is currently speaking.

sudden6:
@antis81 With these commits it actually works! Great work :+1:

src/core/coreav.cpp
```diff
@@ -364,6 +364,49 @@ void CoreAV::volMuteToggle(uint32_t callId)
...
+        qWarning() << "Audio for invalid group id" << group
+                   << "will not be played back.";
+        return;
+    }
+
+    ToxGroupCall& call = cav->groupCalls[group];
+
+    emit c->groupPeerAudioPlaying(group, peer);
+
+    if (call.muteVol)
```
antis81: You need to probably add the `call.inactive` here, but not sure.

antis81:
Looks like either a commit was not signed or there is a typo in the travis script. It tells:

``` bash
Merge unsigned commits
refractor: abstract HTML out of `new issue` GitHub link
Prevent regex capture from extended past line boundaries.
Add direct link with template for submitting new issues

Above ↑ commits don't conform to commit message format:

Pls fix.
```

It says `refractor` -> should be `refactor`. Anyway, this looks a bit strange…

sudden6:
this is probably because this branch was started before new commit format and not yet rebased.

sudden6:
@antis81 am I seeing this right that toxcore keeps sending the audio data of a groupchat to all peers, and there's no way of unsubscribing from the audio stream? Because this would be a huge bandwith waste.

TheNain38:
@sudden6 Normally, there are a "text groupchat" and an "audio groupchat" which are joined when joining a groupchat with both in it, instead you could just leave the audio group, and rejoin it when you actually enter the call? (If there are no way to unsubscribe from the audio stream)

sudden6:
I saw functions to make an av groupchat from a text groupchat, but not vice versa. How do other clients handle this?

zetok:
@sudden6 

> I saw functions to make an av groupchat from a text groupchat, but not vice versa.

Err? AFAIK there are no functions like that – i.e. you specify groupchat type at the time of creation, and go from there.

> How do other clients handle this?

Well, time ago, Toxic was the only client that allowed users to choose which groupchat type should be started, i.e. audio or text type. What other clients did, is defaulting to audio groupchats, and "forgetting" about text-only groupchats.

As for bandwidth - IIRC one simply doesn't send audio data if there is no need.

# 

@antis81 
Well, yeah, due to how travis works, it kinda will include commits that are later-on (after PR creation) included on master if in meanwhile PR gets an update.

E.g. with

```
 A---B---C  master
  \
   D-------E  PR
```

once `E` commit gets pushed to PR, travis gets all commits `A, B, C, D, E`, even if PR was based on `A`.
And yeah, there were a few commits in meanwhile before verification script was introduced: #3137.

antis81:
@zetok Ah ok, thanks for clarification. Seems like this PR is rebased now correctly on master and everyone can live happily ever after. :)

@sudden6 

> am I seeing this right that toxcore keeps sending the audio data of a groupchat to all peers, and there's no way of unsubscribing from the audio stream? Because this would be a huge bandwith waste.

At this very point, this is the wrong question. You ask for the goals we want to reach with qTox, which  goes far beyond the purpose of this little PR.

sudden6:
@zetok 

> > I saw functions to make an av groupchat from a text groupchat, but not vice versa.
> 
> Err? AFAIK there are no functions like that – i.e. you specify groupchat type at the time of creation, and go from there.

Sorry, I confused some toxcore internal function with the public API.

@antis81 

> At this very point, this is the wrong question. You ask for the goals we want to reach with qTox, which goes far beyond the purpose of this little PR.

Probably, yes. We can't change this anyway, because toxcore doesn't provide a way to controll this.

zetok:
Rebased commits are on master, thus closing.

PR #3090, opened 2016-04-01
Fixed incorrect desktop video resolution when using HiDPI scaling

initramfs:
When attempting to use "Desktop" as the video source under HiDPI scaling, the resolution returned by Qt's `QApplication::desktop()->screenGeometry().size();` would return a resolution in device-indepedent pixels (which would be less than the actual physical number of pixels being projected by the monitor). As such, toxav ends up clipping the Desktop feed to only the given region, resulting in a subset of the desktop being displayed. This issue is examplified below:
clipped

The problem is resolved by usage of the function `QScreen::devicePixelRatio()` which returns the scaling factor between the DPI pixels and the actual pixel pixels.

The one caveat with ths solution is that `devicePixelRatio()` is only available >= Qt 5.5, causing breakage when used with an older verison of Qt.

### DIFF
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -20,6 +20,7 @@
 #include <QDebug>
 #include <QApplication>
 #include <QDesktopWidget>
+#include <QScreen>
 extern "C" {
 #include <libavformat/avformat.h>
 #include <libavdevice/avdevice.h>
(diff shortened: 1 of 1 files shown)


GitCop:
- Commit: 598bfe5a9030b7e330fb3815d8233d5138bf2687
  - Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at

---

Issue #3089, opened 2016-03-31
Add client-side hook to validate commit messages


ghost:
GitCop can be too noisy (see #3079). To avoid this, qTox could ship a client-side `commit-msg` hook ([docs](https://git-scm.com/book/en/v2/Customizing-Git-An-Example-Git-Enforced-Policy#Client-Side-Hooks)).

TODO, if the idea is accepted:
- [ ] write the `commit-msg` script
- [ ] update `CONTRIBUTING.md`

ghost:
http://addamhardy.com/2013/06/05/good-commit-messages-and-enforcing-them-with-git-hooks.html looks like a good start. Using python is a safe bet for most systems, but I couldn't tell if it's Windows-proof.

`.git/hooks/commit-msg` (created by the user, instructions in the wiki) should just call the actual script in qTox's repo. Will the script (`commit-msg-validator.py`?) be buried in a new/existing dir?

Side note from #3070: why

> Any line of the commit message cannot be longer 100 characters!

instead of the "standard" 50/72?

zetok:
> > Side note from #3070: why
> 
> Any line of the commit message cannot be longer 100 characters!

copypasting.
I guess that 72 chars are preferable.

# 

> instructions in the wiki

I think that it would be better to include instructions for commit hook in the CONTRIBUTING.md.

> Will the script (`commit-msg-validator.py`?) be buried in a new/existing dir?

Yeah sounds good to me; I was thinking of re-organizing scripts anyway, and putting them under `scripts/` dir.

ghost:
> copypasting

Me too, from https://github.com/torvalds/linux/pull/17 and http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

I'm at ~100 lines in the validation script, and I think that some things in the message specifications should be reviewed:

> <BLANK LINE>
> <footer>

Has no real meaning, it would be simpler to specify an order for the body. Summary - breaking changes - solved issues, for example. Parsing the footer is unreliable in the script, because the body can contain line breaks.

> `BREAKING CHANGE:` with a space or two newlines

2 opposite standards? Why not just the simplest to parse, new line after `BREAKING CHANGE:`?

> Any line of the commit message cannot be longer 100 characters

The length shouldn't be mandatory for the body. Length can be over the limit in the header if a `revert:` is applied to a `fix:` that had already the max length.

> `This reverts commit <hash>.`

Should read `Revert commit <hash>.`

Issue #1786, opened 2015-06-04
Desktop sharing in multi-display


Diadlo:
With multi-display I can't choose which display I want share. In preview I see 1.5 displays left + half of right

PKEv:
@tux3 this + #1595   - need to monitor configuration settings selection?

Diadlo:
Fixed by #3392

PR #3091, opened 2016-04-01
Fixes incorrect desktop video resolution when using HiDPI

initramfs:
When attempting to use "Desktop" as the video source under HiDPI scaling, the resolution returned by Qt's `QApplication::desktop()->screenGeometry().size()` would return a resolution in device-indepedent pixels (which would be less than the actual physical number of pixels being projected by the monitor). As such, toxav ends up clipping the desktop feed to only the given region, resulting in a subset of the desktop being displayed. This issue is examplified below:
clipped

The problem is resolved by usage of the function `QScreen::devicePixelRatio()` which returns the scaling factor between the DPI pixels and the actual pixel pixels.

The one caveat with ths solution is that `devicePixelRatio()` is only available >= Qt 5.5. A temporary solution is to assume a default of no scaling when compiling with Qt < 5.5.0 (matches current behaviour).

### DIFF
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -20,6 +20,7 @@
 #include <QDebug>
 #include <QApplication>
 #include <QDesktopWidget>
+#include <QScreen>
 extern "C" {
 #include <libavformat/avformat.h>
 #include <libavdevice/avdevice.h>
@@ -132,11 +133,20 @@ CameraDevice* CameraDevice::open(QString devName, VideoMode mode)
         }
         else
         {
-            screen = QApplication::desktop()->screenGeometry().size();
+            QScreen* defaultScreen = QApplication::screens().at(0);
+            qreal pixRatio;
+
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 5, 0))
+            pixRatio = defaultScreen->devicePixelRatio();
+#else
+            pixRatio = 1.0;
+#endif
+
+            screen = defaultScreen->size();
             // Workaround https://trac.ffmpeg.org/ticket/4574 by choping 1 px bottom and right
             // Actually, let's chop two pixels, toxav hates odd resolutions (off by one stride)
-            screen.setWidth(screen.width()-2);
-            screen.setHeight(screen.height()-2);
+            screen.setWidth((screen.width() * pixRatio)-2);
+            screen.setHeight((screen.height() * pixRatio)-2);
         }
         av_dict_set(&options, "video_size", QString("%1x%2").arg(screen.width()).arg(screen.height()).toStdString().c_str(), 0);
         if (mode.FPS)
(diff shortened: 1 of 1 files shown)


GitCop:
- Commit: d3f3dca045628c75687e0a8bba3e5c573a13eb87
  - Invalid type. Valid types are **feat,  fix,  docs,  style,  refactor,  perf,  test,  chore,  revert**

Guidelines are available at

---

zetok:
Huh. Damn, looks like gitcop is misconfigured, or is failing to do what it is supposed to do. Looks like it can't be used after all. :/

tux3:
Maybe gitcop is confused, I noticed that by default gitcop is configured to allow a maximum of 72 chars in the subject AND in every line of the body, and this commit has a 75 character line.
So I upped the limit in the message body to 255, let's see if it helps.

initramfs:
@tux3 According to the [contributing guidelines](https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#commit-message-format), shouldn't the maximum commit line length be limited to 100 (and not 72)? Wouldn't a limit of 100 be sufficient?

zetok:
@initramfs right, sorry about that.

Related: https://github.com/tux3/qTox/issues/3089#issuecomment-204174033

GitCop:
- Commit: a54601f072c9dd277cb757573caa5dc99d4576bb
  - Invalid type. Valid types are **feat,  fix,  docs,  style,  refactor,  perf,  test,  chore,  revert**

Guidelines are available at

---

initramfs:
Updated commit message to fix spelling mistake.

tux3:
Welp, I'll just disable GitCop.

src/video/cameradevice.cpp
```diff
@@ -132,11 +133,20 @@ CameraDevice* CameraDevice::open(QString devName, VideoMode mode)
         }
         else
         {
-            screen = QApplication::desktop()->screenGeometry().size();
+            QScreen* defaultScreen = QApplication::screens().at(0);
```
sudden6: Will this also work in a multi-monitor environment?
Shouldn't https://doc.qt.io/qt-5/qguiapplication.html#primaryScreen-prop be used?

initramfs:
@sudden6 Sorry I didn't know about the existance of that function, I've added a commit to change the behaviour to use the aforementioned function, it should now explictily use the default screen. 

Regarding support in a multimonitor environment, before the latest commit, you are right in that it could accidentally select the second monitor if that monitor was somehow index 0. However, I believe a much better solution would be to give users a selector to show which monitor to display.

zetok:
> However, I believe a much better solution would be to give users a selector to show which monitor to display.

Yeah, until that will be possible though, selecting default screen is probably the best :)

zetok:
RE: screen selection:
Related: #1786

sudden6:
@initramfs if you are going to add a way to choose the screen please open a new PR.

PR #3099, opened 2016-04-03
fix(simple_make|ffmpeg-devel):Re-add ffmpeg-devel

linux-modder:


INSTALL.md
```diff
@@ -184,15 +184,28 @@ sudo apt-get install build-essential qt5-qmake qt5-default qttools5-dev-tools li
 
 <a name="fedora-other-deps" />
 #### Fedora:
+**Note ffmpeg-devel requires enabling / using rpmfusion-free repo (enable via  dnf config-manager)(!).**
```
sudden6: please remove the `(!)` to match the style of the debian section above.

INSTALL.md
```diff
@@ -184,15 +184,28 @@ sudo apt-get install build-essential qt5-qmake qt5-default qttools5-dev-tools li
...
+
+*Valid Keys* 
+
+F22:  97F4D1C1  | 5065 885A 371C 6200 3ED7  AC4F 81C9 B423 97F4 D1C1
+
+F23:  E051B67E  | 1E0D 69F0 77CA 4960 C32C  17E2 5B03 78C0 E051 B67E
+
 **Note that sqlcipher is not included in Fedora(!).**
```
sudden6: same as above

sudden6:
please also change the commit messages according to our guidelines

INSTALL.md
linux-modder: Fedora != Debian   honestly needs to stand out imo

sudden6: if you really think it's that important, use only `!` but not `(!).`

INSTALL.md
linux-modder: You do realize this one WAS there previously I did not add -- but has been changed as requested.

PR #3105, opened 2016-04-06

ovalseven8:


translations/de.ts
```diff
@@ -538,7 +538,9 @@ Viele Grüße,
     </message>
     <message>
         <source>/me offers friendship, &quot;%1&quot;</source>
-        <translation>/me macht eine Freundschaftsanfrage, &quot;%1&quot;</translation>
+        <translation>/me macht eine Freundschaftsanfrage:
+
+„%1“</translation>
```
ovalseven8, line 543: The reason for this change is that it looks much better in the client with a empty line for the German default friend request message.

Basically, it looks better in general. Perhaps this change should be done globally?

PR #3110, opened 2016-04-08
style: enhance and reformat issue template

ghost:
Add missing required information for new issues
Redo Markdown format for a cleaner raw display
Standarise requirements/terms with CONTRIBUTING.md

Template standard needed to finish #2983.

What do you think about moving `ISSUE_TEMPLATE.md` out of the root, into GH's custom `.github` dir?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
-#### Brief Description
+##### Brief Description
 
 OS: Windows / OS X / Linux (include version and/or distro)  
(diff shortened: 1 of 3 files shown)


zetok:
> What do you think about moving `ISSUE_TEMPLATE.md` out of the root, into GH's custom `.github` dir?

I don't see any reason to not do that.

As for the PR, I'm not sure if changing `**` to `####` is a good thing, since with this font will be overly large, while currently it's IMHO sufficiently visually marked. Perhaps use 5 `#`, which will give same size?

PR #3112, opened 2016-04-08
Fix regex because the string should contain `.`

ovalseven8:
For example `alice@example.com`


zetok:
Err, no, `bob@localhost` is perfectly valid.

Issue #3078, opened 2016-03-29
Detached chat windows remain open after disabling multi-window mode


smc2000:
Brief description: The title already does a good job at describing what happens.

Operating System: Linux Mint 17.2 XFCE 64bit
qTox version: v1.3.0-127-g https://github.com/tux3/qTox/commit/791d871737dc27371434386ca9b5c4def4d7a67a

Reproducible: Always

**Steps to reproduce:**  
1.  Start with qTox in single window mode.
2.  Enable multiple windows mode.
3.  Click a friend or group chat to open a detached chat window.
4.  Disable multiple windows mode.

**Observed Behavior:**  
The detached chat window remains open. Clicking the chat's entry in the friend list puts focus on the floating chat window.

**Expected Behavior:**  
Close all extra windows upon disabling multi windows mode.

**Additional info:**  
Messages written in the floating chat window correctly appear in the unified qTox window once I close the floating chat to carry on the conversation in single window mode.

zetok:
> Close all extra windows upon disabling multi windows mode.

In a "single window mode" you still can open detached windows, it's just not a "default" action.

So why would you expect them to be closed once option for just the behaviour gets changed?

smc2000:
When you right click a friend in single window mode to open a chat in a new window, that is a very direct action that is clearly described by the option on the right click menu. The resulting action is therefore expected from the user's perspective. On the other hand, if a user wants to return qTox to single window mode by disabling the multi window option, I think it makes most sense that the user expects all open qTox windows to merge together into one and that would mean closing the detached chats so that the conversation can continue in one window. After all, "not multiwindow == single window".

zetok:
> if a user wants to return qTox to single window mode by disabling the multi window option

Except that it's not a switch on/off, but a switch for a default behaviour, which user can "override" without touching the option.

I.e. option doesn't change qTox's capabilities, but merely its default behaviour.

Thus,

> After all, "not multiwindow == single window".

`no multiwindow by default ≠ single window` would be correct.

zetok:
Needed info not provided for over a month, thus closing.

Issue #3087, opened 2016-03-30
qTox crashed after I changed the language


ovalseven8:
**Brief description:**
I played with qTox and suddenly it crashed. I am not sure why, so I can't list steps to reproduce it. But below you can find a full backtrace.
Seems like it has something to do with changing the language?

**Update:** See _Steps to reproduce_ below!

Operating System: Lubuntu 15.10
qTox version: https://github.com/tux3/qTox/commit/af698d85372a5d2266c7f85ac35209b437cc5acb
Reproducible: Always

**Steps to reproduce:**
1. Start qTox
2. Write `invite` to groupbot so that you get a group invite
3. Go to the group invites window and click "Decline"
4. Go to the settings
5. Change language to e.g. German

Here you can see the log right before qTox crashed:

```
[17:49:13.493] src/core/core.cpp:132 : Debug: Core starting with IPv6 enabled
[New Thread 0x7fffcdffb700 (LWP 14002)]
[17:49:13.495] src/core/core.cpp:305 : Debug: Self avatar not found, will broadcast empty avatar to friends
[17:49:13.812] src/widget/systemtrayicon.cpp:93 : Debug: Using the Qt backend
[17:49:14.994] src/core/core.cpp:407 : Debug: "Connecting to 130.133.110.14:33445 (Manolis)"
[17:49:14.994] src/core/core.cpp:407 : Debug: "Connecting to 104.167.101.29:33445 (noisykeyboard)"
[17:49:19.494] src/core/core.cpp:407 : Debug: "Connecting to 205.185.116.116:33445 (Busindre)"
[17:49:19.495] src/core/core.cpp:407 : Debug: "Connecting to 198.98.51.198:443 (Busindre)"
[17:49:22.898] src/core/core.cpp:374 : Debug: Connected to the DHT
[17:49:22.898] src/persistence/profile.cpp:314 : Debug: Saving tox save to  "/home/test/.config/tox/Alice.tox"
[17:49:32.544] src/core/core.cpp:500 : Debug: "Text group invite by 0"
[Thread 0x7fffceffd700 (LWP 14000) exited]
[17:49:44.452] src/widget/translator.cpp:53 : Debug: Loaded translation "de"
[17:49:44.452] src/widget/translator.cpp:61 : Debug: System translation loaded "de"

Program received signal SIGSEGV, Segmentation fault.
[...]

PR #3116, opened 2016-04-10
Cleaning set of buttons after deleting objects and add translation

### LINKED ISSUE #3083: Langugage inconsistent in new group chat invites page
Brief description: The language in the new group chat invites page depends on which language you had set while you got the invite.

qTox version: https://github.com/tux3/qTox/commit/375c9657ceecc67a66efe05480958fdfd3122c15
Reproducible: Always

**Steps to reproduce:**  
1. Set language to "English"
2. Get group chat invite (for example with sending `invite` to group bot)
3. Change language
4. Get further group chat invite
5. Go to the group chat invites page
=> 6. Languages are inconsistent

**Observed Behavior:**
Languages are inconsistent

**Expected Behavior:**
Languages should be consistent

**Additional info:**  
Screenshot

Diadlo:
Fixed crash on language change after group invite #3087
Added translation to group invite message (Fix #3083)

### DIFF
--- a/src/widget/form/groupinviteform.cpp
+++ b/src/widget/form/groupinviteform.cpp
@@ -20,6 +20,7 @@
 #include "groupinviteform.h"
 
 #include <tox/tox.h>
+#include <QDebug>
 #include <QSignalMapper>
 #include <QPushButton>
 #include <QBoxLayout>
@@ -136,8 +137,7 @@ void GroupInviteForm::onGroupInviteAccepted()
     GroupInvite invite = groupInvites.at(index);
     groupInvites.removeAt(index);
 
-    groupWidget->deleteLater();
-    inviteLayout->removeWidget(groupWidget);
+    deleteInviteButtons(groupWidget);
 
     emit groupInviteAccepted(invite.friendId, invite.type, invite.invite);
 }
@@ -149,8 +149,27 @@ void GroupInviteForm::onGroupInviteRejected()
     int index = inviteLayout->indexOf(groupWidget);
     groupInvites.removeAt(index);
 
-    groupWidget->deleteLater();
-    inviteLayout->removeWidget(groupWidget);
+
+    deleteInviteButtons(groupWidget);
+}
+
+void GroupInviteForm::deleteInviteButtons(QWidget *widget)
+{
+    QList<QPushButton*> buttons = widget->findChildren<QPushButton*>();
+
+    if (acceptButtons.contains(buttons.at(0)))
+    {
+        acceptButtons.remove(buttons.at(0));
+        rejectButtons.remove(buttons.at(1));
+    }
+    else
+    {
+        acceptButtons.remove(buttons.at(1));
+        rejectButtons.remove(buttons.at(0));
+    }
+
+    widget->deleteLater();
+    inviteLayout->removeWidget(widget);
 }
 
 void GroupInviteForm::retranslateUi()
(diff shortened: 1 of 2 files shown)


zetok:
Yep, seems to fix #3078 for me.

src/widget/form/groupinviteform.cpp
```diff
@@ -149,8 +153,29 @@ void GroupInviteForm::onGroupInviteRejected()
...
-    groupWidget->deleteLater();
-    inviteLayout->removeWidget(groupWidget);
+    deleteInviteButtons(groupWidget);
+}
+
+void GroupInviteForm::deleteInviteButtons(QWidget *widget)
+{
+    QList<QPushButton*> buttons = widget->findChildren<QPushButton*>();
+
+    if (acceptButtons.contains(buttons.at(0)))
```
sudden6: This seems wrong to me, because you always delete the first or second buttons, what if the user has 5 friend requests?
Maybe take a look at addfriendfrom.cpp and see how it's handled there.

src/widget/form/groupinviteform.h
```diff
@@ -59,15 +61,18 @@ private slots:
 
 private:
     void retranslateUi();
-    void retranslateAcceptButton(QPushButton* acceptButton);
-    void retranslateRejectButton(QPushButton* rejectButton);
+    void retranslateAcceptButton(QPushButton *acceptButton);
```
sudden6: I think our Coding Style says `*` belong to the type.

zetok: Yep: https://github.com/tux3/qTox/wiki/Coding#coding-style

src/widget/form/groupinviteform.cpp
Diadlo: We have structure like that:
all requests -> request -> buttons
And "button->parentWidget()" return "request", which contains only 2 buttons

src/widget/form/groupinviteform.h
Diadlo: Sorry for my inattention to code style. Fixed

src/widget/form/groupinviteform.cpp
```diff
@@ -149,8 +153,29 @@ void GroupInviteForm::onGroupInviteRejected()
...
-    groupWidget->deleteLater();
-    inviteLayout->removeWidget(groupWidget);
+    deleteInviteButtons(groupWidget);
+}
+
+void GroupInviteForm::deleteInviteButtons(QWidget* widget)
+{
+    QList<QPushButton*> buttons = widget->findChildren<QPushButton*>();
+
+    if (acceptButtons.contains(buttons.at(0)))
```
sudden6: I still think it is better to replace this construct with something like:

```
acceptButtons.subtract(&QSet::fromList(buttons));
rejectButtons.subtract(&QSet::fromList(buttons));
```

(just an example, I don't know if it is correct)
See http://doc.qt.io/qt-5/qset.html#subtract and http://doc.qt.io/qt-5/qset.html#fromList for details.

As it's much cleaner and verbose.

Diadlo: You're right, it made the code much better, thank you

src/widget/form/groupinviteform.h
```diff
@@ -21,7 +21,9 @@
 #define GROUPINVITEFORM_H
 
 #include <QWidget>
+#include <QDateTime>
 #include <QSet>
+#include <src/widget/tool/croppinglabel.h>
```
sudden6: use `" "` for local includes

Diadlo: It's really strange. I never use <> for local include. Fixed

PR #3118, opened 2016-04-10
fix:restore links to local resources

PKEv:
Added support for links like "file://192.168.1.1/" and "smb://localnode/"

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -249,7 +249,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(?:\\b)(file:///)([\\S| ]*)" //link to a local file, valid until the end of the line
+                "|(?:\\b)((file|smb)://)([\\S| ]*)" //link to a local file, valid until the end of the line
                 "|(?:\\b)(tox:[a-zA-Z\\d]{76})" //link with full user address
                 "|(?:\\b)(mailto:\\S+@\\S+\\.\\S+)" //@mail link
                 "|(?:\\b)(tox:\\S+@\\S+)"); // starts with `tox` then : and only alpha-digits till the end


zetok:
Could you use https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#commit-message-format ?

vith:
It isn't necessary to escape `.` inside a regex character class, right?

zetok:
0e604af54021b69dcc593aaf3bb1058ddc13bd82: doesn't seem to work:
no highlight

Also, `[\.]` doesn't seem to be needed, since it's already ~covered by `\S`

```
\S  Matches a non-whitespace character.
```

?

Aside from "links" not working, is `file://` really a thing? I.e. I know that on ~unix `file:///` is a ~standard – is `file://` specified somewhere?

PKEv:
strange...
qtox_screenshot_2016-04-11 21-30-28 484

PKEv:
aboute "file:///...."
I can't check and was somewhat surprised as the links I have now are like: "file://192.168.1.8/Public" 

but look: http://stackoverflow.com/questions/18246053/how-can-i-create-a-link-to-a-local-file-on-a-locally-run-web-page/18246357

examples of different processing links:
qtox_screenshot_2016-04-11 22-02-38 351

it seems to me that we should support and format with two slashes...

zetok:
Hm, looks like I might have compiled/launched wrong version, after recompiling `smb://` seems to work:
smb works

zetok:
@PKEv working:

``` diff
diff --git a/src/chatlog/chatmessage.cpp b/src/chatlog/chatmessage.cpp
index 35327fd..017c34e 100644
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -249,7 +249,9 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
-                "|(?:\\b)((file:///)|(smb://))([\\S| |\\.]*)" //link to a local file, valid until the end of the line
+                // ↓ link to a file, or samba share
+                //   https://en.wikipedia.org/wiki/File_URI_scheme
+                "|(?:\\b)((file|smb)://)([\\S| ]*)"
                 "|(?:\\b)(tox:[a-zA-Z\\d]{76})" //link with full user address
                 "|(?:\\b)(mailto:\\S+@\\S+\\.\\S+)" //@mail link
                 "|(?:\\b)(tox:\\S+@\\S+)"); // starts with `tox` then : and only alpha-digits till the end
```

PKEv:
tested:
qtox_screenshot_2016-04-11 23-37-02 004
qtox_screenshot_2016-04-11 23-36-54 816

src/chatlog/chatmessage.cpp
```diff
@@ -249,7 +249,7 @@ QString ChatMessage::detectAnchors(const QString &str)
                 "(?:\\b)((www\\.)|(http[s]?|ftp)://)" // (protocol)://(printable - non-special character)
                 // http://ONEORMOREALHPA-DIGIT
                 "\\w+\\S+)" // any other character, lets domains and other
-                "|(?:\\b)(file:///)([\\S| ]*)" //link to a local file, valid until the end of the line
+                "|(?:\\b)((file|smb)://)([\\S| ]*)" //link to a local file, valid until the end of the line
```
zetok, line 252: could you correct the comment?

I.e. with `file://` or `smb://` it's no longer about only local files.

src/chatlog/chatmessage.cpp
```diff
@@ -1,4 +1,4 @@
-/*
+﻿/*
```
zetok:
``` diff
diff --git a/src/chatlog/chatmessage.cpp b/src/chatlog/chatmessage.cpp
index c2a563b..2bb986c 100644
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -1,4 +1,4 @@
-/*
+<U+FEFF>/
```

Could you fix that?

zetok:
Also, the commit message is kinda off – i.e. it's not really a fix, but introducing a new feature, where non-local `file://`s and `smb://` shares are now supported, so it should be:

```
feat(chatform): add support for non-local file and samba share links

E.g. links like "file://192.168.1.1/" and "smb://localnode/" are now
supported. Before only links to local files were supported, e.g.
"file:///local/file.txt".
```

src/chatlog/chatmessage.cpp
PKEv: i don' understand

PR #3134, opened 2016-04-12
Added checks pointers on existence

### LINKED ISSUE #3133: segfault when trying to reconnect when creating groupchats
##### Brief Description

When creating thousands of groupchats qTox noticeably slows down. When that happens, one has to click on `Reconnect` in settings to crash qTox.

OS: Gentoo GNU/Linux x86_64
qTox: 0a2f54194569e626ef8501c610647219887505d4
toxcore:  0.0.0
Qt:  5.5.1

Reproducible: Always
##### Steps to reproduce
1. Create a few thousands of groupchats (Note: after first 1k of groupchats qTox's UI has a tendency to slow down)
2. While creating groupchats, click on `Reconnect` in settings (Note: done in multiple windows mode)
3. Segfault
##### Observed Behavior

qTox segfaults.
##### Expected Behavior

qTox should reconnect.
##### Additional Info

backtrace: https://gist.github.com/zetok/f2b875f7ec047be0e5c1714b49b4aa5b

Note: to quickly create groupchats, exploit method from #3128 - i.e. hold down `Enter`.
If you're using xorg, you may want to speed that up further by e.g. `xset r rate 500 200`, which will make pressed key to start being repeated 200 times/s after first 500ms of key being pressed.

Diadlo:
Fix #3133.
- Added check, that core and coreAv exist, before use them
- Added check, that tox exit, before use them

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -978,6 +978,12 @@ ToxId Core::getGroupPeerToxId(int groupId, int peerId) const
 QList<QString> Core::getGroupPeerNames(int groupId) const
 {
     QList<QString> names;
+    if (!tox)
+    {
+        qWarning() << "Can't get group peer names, tox is null";
+        return names;
+    }
+
     int result = getGroupNumberPeers(groupId);
     if (result < 0)
     {
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1502,7 +1502,10 @@ void Widget::onGroupNamelistChanged(int groupnumber, int peernumber, uint8_t Cha
     {
         qDebug() << "onGroupNamelistChanged: Group "<<groupnumber<<" not found, creating it";
         g = createGroup(groupnumber);
+        if (!g)
+            return;
     }
+
 
     TOX_CHAT_CHANGE change = static_cast<TOX_CHAT_CHANGE>(Change);
     if (change == TOX_CHAT_CHANGE_PEER_ADD)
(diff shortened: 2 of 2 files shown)


src/widget/widget.cpp
```diff
@@ -1592,8 +1595,17 @@ Group *Widget::createGroup(int groupId)
 
     Core* core = Nexus::getCore();
 
+    if (!core)
+        return nullptr;
+
     QString groupName = QString("Groupchat #%1").arg(groupId);
```
PKEv, line 1604: Maybe can make "Groupchat" translatable?

zetok, line 1604: Why not. Though in a separate PR.

sudden6:
I'm not entirely sure all those checks are needed, but they certainly won't hurt.
Please add some error messages, because this way your checks just mask the problem and this could be hard to detect in the future.

Diadlo:
@sudden6 Not all. Check "if (!core)" add just for completely sure.

sudden6:
Ok, can you still add the error messages? just use the `qDebug() qWarning() qCritical() qFatal()` functions, whichever is appropriate.

Diadlo:
Add warning messages.

sudden6:
What about the two other checks? Is there a reason why you didn't add some messages there?

Also you don't necessarily have to include the function name into the debug output, because it already contains the line numbers.

Diadlo:
- core.cpp has many checks "if (!tox) return ...", and all without messages (should I add message to all?)
- widget.cpp check it is result of possible "return nullptr", which already have warning message

sudden6:
> core.cpp has many checks "if (!tox) return ...", and all without messages (should I add message to all?)

Just add it to your code for now, you can create another PR for the rest if you like. Just be careful not to spam the logfile.

> widget.cpp check it is result of possible "return nullptr", which already have warning message

ok, I didn't realize this, leave it as it is then.

src/core/core.cpp
```diff
@@ -978,6 +978,12 @@ ToxId Core::getGroupPeerToxId(int groupId, int peerId) const
 QList<QString> Core::getGroupPeerNames(int groupId) const
 {
     QList<QString> names;
+    if (!tox)
+    {
+        qWarning() << "Can't create get peer names. tox is null";
```
sudden6: I would rephrase this to: `Can't get group peer names, tox is null`.

zetok:
That travis failure should be fixed on 5b7002a846a27e33622cae7b1d13a46e88f417ff

Issue #3131, opened 2016-04-12
icon files do not change in systray


farseerfc:
##### Brief Description

OS: Arch Linux (rolling)
qTox version:  v1.3.0-170-g0a2f541
Commit hash:  0a2f54194569e626ef8501c610647219887505d4
toxcore:  0.0.0 (3747-1 in Arch Linux)
Qt:  5.6.0

This is the same issue as https://bugs.archlinux.org/task/47910
Keeping the icons as qtox.png / qtox.svg breaks the change of tray icon based on status.

Reproducible: Always
##### Steps to reproduce
1. Start qtox 
2. login
##### Observed Behavior

icon in systray is still a black lock, the default icon [blacklock](https://raw.githubusercontent.com/tux3/qTox/master/img/icons/24x24/qtox.png)
##### Expected Behavior

icon in systray should change to online icon [online](https://raw.githubusercontent.com/tux3/qTox/master/img/taskbar/dark/taskbar_online.svg)
##### Additional Info

See https://bugs.archlinux.org/task/47910

A workaround will be to rename icons from qtox.\* to qTox.*

zetok:
> icon in systray is still a black lock, the default icon blacklock

Related: https://github.com/tux3/qTox/issues/1820#issuecomment-133818693

After some update of Qt 5.5.1 it's fixed for me.

Could you perhaps test whether with different Qt version (e.g. 5.4) there is still problem with lower-case `qtox.*` icons?

farseerfc:
I will try to test on an older version of Qt but I need some time to setup the environment (because my main working DE is KDE5 and I can't risk to break it).

vith:
I get the bug on release builds of Qt from Ubuntu or Arch Linux packages but not debug builds from Qt git tags v5.5.0, v5.5.1, or v5.6.0. Haven't been able to properly debug the release builds to see what causes the difference but I created this workaround for now: https://github.com/tux3/qTox/pull/3151

zetok:
Since #3151 was merged, this should be fixed.
@farseerfc care to test with newest qTox?

farseerfc:
Wow tested ed9ccbe4716930ff27044201480589033c5d3b6f and it works!
Thank you @vith and @zetok .
I will close this issue.

PS:
@vith I added you as the co-maintainer of qtox-git on AUR. Feel free to update it if necessary. And let me know if this troubles you.

PPS:
[...]

Issue #1820, opened 2015-06-06
Tray Icon problems on Fedora 22 KDE x64 Linux (Qt 5.5 bug)


Getron:
The tray icon doesn't change on my Fedora 22 KDE x64 Linux.

The "Light icon" doesn't work.

And the icon not indicate my status to, doesn't work, it is just a black Tox icon, without any indications on it, like green midle dot, or so, nothing.

Can you please check this, and fix it please?

zetok:
@Getron if you could, please include more info: https://github.com/tux3/qTox/wiki/Useful-bug-reports

Getron:
@zetok EchoBot is functional, i'm online on qTox main Window, just the tray icon is not showing if i'm online, and not change the "Light icon" from options.

I think, this could be a problem with the new KDE 5.3, which is included in the new Fedora 22 KDE x64.

Many apps have this icon problems with the new KDE 5.3, but, this is one of the newest improvment of KDE.

I think you should try it on a virtual machine to, and you will see then the problem. On my machine, it is not virtual machine, it is on real hardware installed the system, but, i think, this would happend on virtual Fedora 22 KDE Spin x64 machine to.

On Fedora 20 and 21 KDE, i don't have this problems, but, the 20 and 21 have a KDE 4, and the 22 is ekiped with new KDE 5.3, which has other algorithms for icons, and many other things.

The best way is to check it yourself.

For example, Viber, Teamviewer, NoMachine and Skype has to this problem with tray icon on new KDE 5.3, it doesn't show at all the skype icon. And i'm not alone with this problems.

qTox showa the icon, but, is not functional (graphicaly)

tux3:
@Getron Hi, can you post the content of ~/.config/tox/qtox.log please?

Getron:
2015-06-03 21:00:53 file logger starting
[21:00:53.617] src/core/core.cpp:916 : Warning: The Tox configuration file  ""  was not found 

2015-06-03 21:07:43 file logger starting
[21:09:28.234] :0 : Warning: libpng warning: iCCP: known incorrect sRGB profile
[21:09:30.531] :0 : Warning: libpng warning: iCCP: known incorrect sRGB profile
[21:09:30.641] :0 : Warning: libpng warning: iCCP: known incorrect sRGB profile
[21:09:30.750] :0 : Warning: libpng warning: iCCP: known incorrect sRGB profile
[21:09:30.859] :0 : Warning: libpng warning: iCCP: known incorrect sRGB profile
[...]

PR #3151, opened 2016-04-17
fix(Widget::updateIcons): workaround QIcon fallback bug

vith:
Some builds of Qt appear to have a bug in icon loading: `QIcon::hasThemeIcon` is sometimes unaware that the icon returned from `QIcon::fromTheme` was a fallback icon, causing `hasThemeIcon` to incorrectly return `true`.

In qTox this leads to the tray and window icons using the static qTox logo icon instead of an icon based on the current presence status.

This workaround checks for an icon that definitely does not exist to determine if `hasThemeIcon` can be trusted.

On systems with the Qt bug, this workaround will always use our included icons but user themes will be unable to override them.

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -445,8 +445,35 @@ void Widget::updateIcons()
             status = QStringLiteral("offline");
     }
 
+    // Some builds of Qt appear to have a bug in icon loading:
+    // QIcon::hasThemeIcon is sometimes unaware that the icon returned
+    // from QIcon::fromTheme was a fallback icon, causing hasThemeIcon to
+    // incorrectly return true.
+    //
+    // In qTox this leads to the tray and window icons using the static qTox logo
+    // icon instead of an icon based on the current presence status.
+    //
+    // This workaround checks for an icon that definitely does not exist to
+    // determine if hasThemeIcon can be trusted.
+    //
+    // On systems with the Qt bug, this workaround will always use our included
+    // icons but user themes will be unable to override them.
+    static bool checkedHasThemeIcon = false;
+    static bool hasThemeIconBug = false;
+
+    if (!checkedHasThemeIcon)
+    {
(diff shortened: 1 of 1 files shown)


vith:
If anyone's not a fan of this mitigation, I believe another option would be renaming the static icon to `qtox-logo` or similar, so that there's no `qtox` to fallback to from `qtox-offline` etc.

Downstream worked around it by using a unique capitalization for the installed icon files.

vith:
Related: #3131 #1820

sudden6:
Is this only affecting certain versions of Qt? Because then I think it would be better to limit the workaround to these versions.

Using unique capitalization probably isn't a good idea, because it could be seen as an accident and be renamed in the future, which will break things again.

zetok:
> Is this only affecting certain versions of Qt?

Yep.

> Because then I think it would be better to limit the workaround to these versions.

No, it wouldn't be.
IIRC 5.3 are affected, 5.4 versions are not affected (kinda), 5.5.0 is affected, 5.5.1 is affected, 5.5.1 with some ~~path~~ _patch_ is not affected, 5.6 is affected…

So I'm kinda wondering how you distinguish between 5.5.1 that is affected, and 5.5.1 build that is not affected at compile time.

sudden6:
ok, I thought it would be an easier to distinguish

vith:
I tried to determine which Qt versions are affected, but any that I built myself with `./configure -developer-build` did not have the bug (v5.5.0, v5.5.1, v5.6.0 were the versions I tested). Earlier versions did not have the commit which added the fallback behavior: https://github.com/qtproject/qtbase/commit/a40d390a

I was unable to debug any Qt build without `-developer-build` as I always got an error that the `svg` module didn't exist when trying to build qTox with them. I did skim the patches that Arch Linux was using and it didn't look like they intentionally touched anything icon related.

sudden6:
Seems like your version is already the best solution, I'll merge it later.

PR #3162, opened 2016-04-19
Style fixes

Diadlo:
- Changed calls deprecated functions
- Replaced C-style cast to static_cast
- Constructors are made explicit
- Small style changes

### DIFF
--- a/src/chatlog/content/notificationicon.h
+++ b/src/chatlog/content/notificationicon.h
@@ -31,7 +31,7 @@ class NotificationIcon : public QObject,  public ChatLineContent
 {
     Q_OBJECT
 public:
-    NotificationIcon(QSize size);
+    explicit NotificationIcon(QSize size);
 
     virtual QRectF boundingRect() const override;
     virtual void paint(QPainter *painter, const QStyleOptionGraphicsItem *option, QWidget *widget) override;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -150,7 +150,11 @@ int main(int argc, char *argv[])
     IPC& ipc = IPC::getInstance();
 #endif
 
-    sodium_init(); // For the auto-updater
+    if (sodium_init() < 0) // For the auto-updater
+    {
+        qCritical() << "Can't init libsodium";
+        return EXIT_FAILURE;
+    }
 
 #ifdef LOG_TO_FILE
     QString logFileDir = Settings::getInstance().getAppCacheDirPath();
--- a/src/video/camerasource.cpp
+++ b/src/video/camerasource.cpp
@@ -65,12 +65,12 @@ void CameraSource::open()
     open(CameraDevice::getDefaultDeviceName());
 }
 
-void CameraSource::open(const QString deviceName)
+void CameraSource::open(const QString& deviceName)
 {
     open(deviceName, VideoMode{0,0,0,0});
 }
 
-void CameraSource::open(const QString DeviceName, VideoMode Mode)
+void CameraSource::open(const QString& DeviceName, VideoMode Mode)
 {
     streamBlocker = true;
     QMutexLocker l{&biglock};
--- a/src/video/camerasource.cpp
+++ b/src/video/camerasource.cpp
@@ -337,7 +337,7 @@ void CameraSource::stream()
         }
 
       // Free the packet that was allocated by av_read_frame
-      av_free_packet(&packet);
+      av_packet_unref(&packet);
     };
 
     forever {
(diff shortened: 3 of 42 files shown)


src/video/camerasource.cpp
```diff
@@ -337,7 +337,7 @@ void CameraSource::stream()
         }
 
       // Free the packet that was allocated by av_read_frame
-      av_free_packet(&packet);
+      av_packet_unref(&packet);
```
sudden6, line 340: are these new functions also present in older versions of ffmpeg? Because the oldest version we need to support is probably in Debian 8.

src/main.cpp
```diff
@@ -150,7 +150,11 @@ int main(int argc, char *argv[])
     IPC& ipc = IPC::getInstance();
 #endif
 
-    sodium_init(); // For the auto-updater
+    if (sodium_init() == -1) // For the auto-updater
```
sudden6: comparing for `== -1` is technically correct, but I think it is better to compare for `< 0` if there are error codes added in the future.

src/video/camerasource.cpp
```diff
@@ -65,12 +65,12 @@ void CameraSource::open()
     open(CameraDevice::getDefaultDeviceName());
 }
 
-void CameraSource::open(const QString deviceName)
+void CameraSource::open(const QString& deviceName)
```
sudden6, line 68: Is there a specific reason to change this?

src/widget/genericchatitemlayout.cpp
```diff
@@ -99,10 +99,10 @@ QLayout* GenericChatItemLayout::getLayout() const
 int GenericChatItemLayout::indexOfClosestSortedWidget(GenericChatItemWidget* widget) const
 {
     // Binary search: Deferred test of equality.
-    int min = 0, max = layout->count(), mid;
+    int min = 0, max = layout->count();
```
sudden6, line 102: I think performance an style wise the previous solution was better.

src/chatlog/content/notificationicon.h
```diff
@@ -31,7 +31,7 @@ class NotificationIcon : public QObject,  public ChatLineContent
 {
     Q_OBJECT
 public:
-    NotificationIcon(QSize size);
+    explicit NotificationIcon(QSize size);
```
sudden6, line 34: Can you give some reasons why an explicit constructor is better in this case?

src/video/camerasource.cpp
Diadlo, line 68: We have const. It means no difference between copy and reference. But send by reference more effectively then full copy of string

src/widget/genericchatitemlayout.cpp
Diadlo, line 102: Some C++ styles say: "Use narrowest scope as possible, and initialize variables in the declaration"

src/chatlog/content/notificationicon.h
Diadlo, line 34: It's just like a const. Used for inability of assignment QSize to NotificationIcon. I think it is logical

src/video/camerasource.cpp
Diadlo, line 340: I found stable ffmpeg v2.8 on Debian 8 [here](https://tracker.debian.org/pkg/ffmpeg)
In ffmpeg docs av_packet_unref exist in [v2.4](https://www.ffmpeg.org/doxygen/2.4/group__lavc__packet.html)

If something is wrong, please, tell me

src/main.cpp
Diadlo: Ok. Will be fixed

zetok:
@Diadlo could you rebase on newest master?

src/video/camerasource.cpp
sudden6, line 68: I did some reading in the meantime and now see your point, seems good.

src/chatlog/content/notificationicon.h
sudden6, line 34: I also get your point here now.

src/widget/genericchatitemlayout.cpp
sudden6, line 102: > I think ~~performance~~ an style wise the previous solution was better.

Forget what I said, I made a mistake.

src/video/camerasource.cpp
sudden6, line 340: Seems to be ok then.

sudden6:
@Diadlo as a general note, it would be nice to include reasons _why_ you changed something, not only what you changed.

Edit: ~~The PR seems to be ready for merge, I'll test it a little bit and then merge it.~~
Please fix https://github.com/tux3/qTox/pull/3162#discussion_r60319324 then it's ready for merge :)

Diadlo:
@sudden6 Ok. I will include reasons in future :)
[comment](https://github.com/tux3/qTox/pull/3162#discussion_r60319324) fixed

Issue #3125, opened 2016-04-11
Device name in config qTox is not correct....


borrougagnou:
OS: MAC OS X 10.9.5
qTox version: v1.3.0-38-ga48c422
…
Why the name of Audio Device is "weird"?
![capture d ecran 2016-04-11 a 22 30 05](https://cloud.githubusercontent.com/assets/10796546/14441629/07991f22-0036-11e6-8551-3963ced377c4.png)

Before, no name of device:
![capture d ecran 2016-04-11 a 22 46 15](https://cloud.githubusercontent.com/assets/10796546/14441881/53b7d0e6-0037-11e6-8365-09f64e7b884f.png)

but the really name in mac:
Legend: NAME --> TYPE
## Input:

Internal Microphone --> Built-in
Line In --> Audio line-in-port
--- OPTIONAL ---
Digital In --> Optical digital-in port
NAME OF DEVICE --> USB

---
## Output:

Internal Speakers --> Built-in
Line In --> Audio line-in-port
--- OPTIONAL ---
NAME OF DEVICE --> USB
NAME OF DEVICE --> HDMI
Apple TV --> AirPlay

Regards,
Boris.

zetok:
What is the problem?

borrougagnou:
what is this strange name in the latest version of qTox?
(name of device in the first picture)

The input / output (in the end of post) is a real name of device

zetok:
No idea. I'd assume that it's something from your system.

borrougagnou:
I search but I don't find the strange name in my computer x.X sorry
![capture d ecran 2016-04-12 a 00 24 59](https://cloud.githubusercontent.com/assets/10796546/14444371/49d17f6a-0045-11e6-85ba-e2111a1dda51.png)
![capture d ecran 2016-04-12 a 00 23 07](https://cloud.githubusercontent.com/assets/10796546/14444382/57349584-0045-11e6-8c2f-2e96a1d79983.png)
![capture d ecran 2016-04-12 a 00 27 52](https://cloud.githubusercontent.com/assets/10796546/14444388/693c3232-0045-11e6-95d2-9414999ed768.png)

Chiitoo:
It seems there are Swedish and French translations mingling together for some reason...

ghost:
But that's not part of the translations afaik. I get the string in English, always.

antis81:
[...]

PR #3175, opened 2016-04-21
fix and cleanup audio device lists

antis81:
- [x] cleanup audio device list creation and usage in qTox
- [x] fix selection of default audio device on initialization
- [x] fix initialization, when no device is selected (do not initialize)
- [x] actually disable an audio device without losing the name information
- [x] **WON'T FIX**: ~~fix unicode conversion on Windows (probably only affects pre Win10 versions ??)~~
  - String encoding in OpenAL is undocumented and can lead to incorrect device names on older platforms - especially Windows. With RtAudio, this problem is solved. IMO it is not worth the effort.

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -320,7 +320,7 @@ bool Audio::autoInitOutput()
 
 bool Audio::initInput(const QString& deviceName)
 {
-    if (deviceName.toLower() == QStringLiteral("none"))
+    if (!Settings::getInstance().getAudioInDevEnabled())
         return false;
 
     qDebug() << "Opening audio input" << deviceName;
@@ -363,7 +363,7 @@ bool Audio::initOutput(const QString& deviceName)
     outSources.clear();
 
     outputInitialized = false;
-    if (deviceName.toLower() == QStringLiteral("none"))
+    if (!Settings::getInstance().getAudioOutDevEnabled())
         return false;
 
     qDebug() << "Opening audio output" << deviceName;
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -50,6 +50,18 @@ public:
     {
     }
 
+    static const ALchar* inDeviceNames()
+    {
+        return alcGetString(NULL, ALC_CAPTURE_DEVICE_SPECIFIER);
+    }
+
+    static const ALchar* outDeviceNames()
+    {
+        return (alcIsExtensionPresent(NULL, "ALC_ENUMERATE_ALL_EXT") != AL_FALSE)
+                ? alcGetString(NULL, ALC_ALL_DEVICES_SPECIFIER)
+                : alcGetString(NULL, ALC_DEVICE_SPECIFIER);
+    }
+
     qreal inputGain() const
     {
         return gain;
@@ -323,7 +335,7 @@ bool Audio::initInput(QString inDevDescr)
     const ALCsizei bufSize = (frameDuration * sampleRate * 4) / 1000 * chnls;
     if (inDevDescr.isEmpty())
     {
-        const ALchar *pDeviceList = Audio::inDeviceNames();
+        const ALchar *pDeviceList = Private::inDeviceNames();
         if (pDeviceList)
             inDevDescr = QString::fromUtf8(pDeviceList, strlen(pDeviceList));
     }
@@ -366,7 +378,7 @@ bool Audio::initOutput(QString outDevDescr)
     if (outDevDescr.isEmpty())
     {
         // default to the first available audio device.
-        const ALchar *pDeviceList = Audio::outDeviceNames();
+        const ALchar *pDeviceList = Private::outDeviceNames();
         if (pDeviceList)
             outDevDescr = QString::fromUtf8(pDeviceList, strlen(pDeviceList));
     }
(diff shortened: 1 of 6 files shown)


src/audio/audio.cpp
```diff
@@ -524,20 +536,40 @@ bool Audio::isOutputReady()
...
-    return alcGetString(NULL, ALC_CAPTURE_DEVICE_SPECIFIER);
+    QStringList list;
+    const ALchar* pDeviceList = inDevNames();
+
+    if (pDeviceList)
+    {
+        while (*pDeviceList)
+        {
+            int len = strlen(pDeviceList);
+            list <<  QString::fromUtf8(pDeviceList, len);
```
TheNain38: There are two spaces, instead of one

src/widget/form/settings/avform.cpp
```diff
@@ -321,52 +321,30 @@ void AVForm::getVideoDevices()
 
 void AVForm::getAudioInDevices()
 {
-    QString settingsInDev = Settings::getInstance().getInDev();
-    int inDevIndex = 0;
+    QStringList deviceNames;
```
TheNain38: You don't use this variable, right? So, the "None" device is never added?

src/widget/form/settings/avform.cpp
```diff
@@ -321,52 +321,30 @@ void AVForm::getVideoDevices()
...
-        {
-            int len = strlen(pDeviceList);
-            QString inDev = QString::fromUtf8(pDeviceList, len);
-            bodyUI->inDevCombobox->addItem(inDev);
-            if (settingsInDev == inDev)
-                inDevIndex = bodyUI->inDevCombobox->count()-1;
-            pDeviceList += len+1;
-        }
-    }
+    bodyUI->inDevCombobox->addItems(Audio::inDeviceNames());
```
antis81: The `deviceNames` list should be used here and initialized with Audio::inDevNames. :) Sorry, will clean this up later today.

src/widget/form/settings/avform.cpp
antis81: Now it's used as it wasintended… :)

zetok:
@antis81 um, there is now used commit message format, which will allow to automatically generate changelog for the next, and following releases :)
https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#-git-commit-guidelines

antis81:
@zetok Thanks, I'll change the message.

zetok:
@antis81 um, you've said in #3125 that this PR might not be merged because it's still WIP – if the code is good for master, it should be merged, even if it doesn't fix all the bugs that are out there ;)

But due to "WIP" label this PR seemed like something that isn't finished yet, so e.g. I didn't even test it, and I'd guess that it goes in similar way for everyone else who would be interested in it getting merged.

Meanwhile, it got conflicts with master :(

Dunno, if it's ready for merge, even if it doesn't fix all the things that don't work too well, but if it improves stuff, or makes stuff ready for improvements, IMO "WIP" should be removed (after PR gets rebased on newest master).

antis81:
> @antis81 um, you've said in #3125 that this PR might not be merged because it's still WIP…

No, I said it probably will not be merged, because we plan to switch API. No problem. I have some local changes lingering around, which solve the conflicts and add some more value. I'll finish those changes and push, so it can actually be merged without conflicts. As always, job/life situation keeps changing in IT business and I have to prioritize.

antis81:
@zetok No more conflicts. I'll push some more changes, which correct enable/disable audio output without mixing it with device names. After that, we are ready to merge.

antis81:
@zetok Guess we're fine here. Some review would be nice, but not urgently required. FYI: As OpenAL's original text encoding concerning device names, the last topic cannot be fixed easily and requires deep investigation of OpenAL sources. This is the actual reason, I didn't merge this. But, as we decided to move from OpenAL and use a library more friendly to utf-8, I don't think it is worth the effort. If someone likes to take the time figuring out the encoding, feel free to fix it…

zetok:
@antis81 

> No, I said it probably will not be merged, because we plan to switch API. No problem. (…)

Looks like I've misunderstood >.<

# 

on 765eac9 I got some segfault: https://gist.github.com/zetok/f8e3ab215a355e0f1c46fdcb3b5c8fdd

Happened when I clicked on `Exit` in tray menu after
- starting audio conversation
- starting group audio call
- switching between devices multiple times in settings (also re-scanning devices after adding new one)
- ended audio conversation
- without ending group call tried to `Exit`, while qTox instance had `Audio/Video` settings tab open

backtrace looks a bit weird, could it be something in toxav ?

antis81:
@zetok Thanks for trying with a user's viewpoint. Are you sure the segfault only happens with the commit you stated above? I absolutely couldn't find anything related. Even the corrected boolean return value from `Audio::initInput` only returns correctly `false` now, when device is disabled - nothing bad happening here.

> backtrace looks a bit weird, could it be something in toxav ?

Probably. When looking at the backtrace, it seems to try to send audio buffer after (!) device was closed. This would be totally unrelated to the changes here though.

zetok:
~~I couldn't reproduce the segfault on master, while it's reproducible on this PR.~~

zetok:
Err, nvm, I've looked at the wrong terminal window, it's actually reproducible on master – PR is IMO good to go.

antis81:
> … PR is IMO good to go.

Then please merge it quick and don't ask for long… :smile:

zetok:
> Then please merge it quick and don't ask for long… :smile:

Um, thing is, I don't even know C++. Merging PRs without being able to evaluate what the code change does doesn't sound like something that one should do, so feel free to merge :)

antis81:
@zetok Ok, that's a sane decision then. Never meant to push you btw! Now I even understand better, why situation is as it is and merge process takes its time. You know - it's about responsibility and currently my personal situation is "suboptimal" (others might state it's disastrous actually :smile:) - IOW: My time on qTox is very limited and focused on pure development, code design & a bit review. Due to this, I am not able taking place in regular Saturday's online meetings. Anyway I'm convinced qTox will become the best (encrypted) messenger available and like to support the project. So… nvmd, I'll merge it myself.

zetok:
I see. It's really awesome that you find even a little bit of time for qTox even though things aren't going well well for you.

Hopefully things will get better for you soon :)

Btw, meetings lately haven't been really happening due to almost everyone not being able to participate, so don't worry about it :P

PR #3179, opened 2016-04-22
Added tr()

Diadlo:
Added call tr() function in few places


zetok:
> Added translation

um, the right description would be something like `make … translatable`, in this case `fix(l18n): make typing notification & groupchat name translatable`

Diadlo:
@zetok No problem

Issue #1401, opened 2015-03-15
display a date notification if a chat continues on another day


NoxNoctium:
Currently, qTox only displays date information in the chatlog for already saved messages, if you continue a chat with a person on another day this behaviour could lead to confusion (see pictures below)
![image](https://cloud.githubusercontent.com/assets/2098025/6658229/fa348fdc-cb62-11e4-879c-2bd33d3646dd.png)
in this case, the datechange is visible because of the overlapping times.
![image](https://cloud.githubusercontent.com/assets/2098025/6658264/caff0f0c-cb63-11e4-8f3b-35fddebf5c07.png)
didn't manage to capture a direct comparison of the problem here, originally the date information wasn't displayed, only after a reload of the application (this is how it should look in any case).

zetok:
Description:
When qTox is turned on across multiple days, there is no indicator in chatlog about new day, which is confusing.

"Notification" is visible if chat history is loaded upon fresh qTox start.

Reproducible: Always

Steps to reproduce:
1. Turn qTox on and message your contact
2. Wait for next day without turning qTox off
3. Message your contact again
4. there is no indicator of different day in chatlog

Edit: um, to be clear; "Notification" shouldn't appear on every new day, but rather only when there's a message on new day

TheLastProject:
Could this please get the bug label? Date changes not always being displayed is something I very much consider a bug because it really makes it confusing to tell if messages were sent/received on the same day or not.

zetok:
When #2013 gets merged this should be closable.

zetok:
Related: #687

zetok:
Done in #3181, thus closing.

PR #3181, opened 2016-04-23
fix(genericchatform):separate messages from different days

PKEv:
Before this new cut-off days is only shown when the load history, now added a feature which in the current chat checks the last post date and if the date does not coincide with current - output a system message about the new date.


zetok:
Fixes #1401 ?

PKEv:
yes

zetok:
Doesn't seem to work:
no date?

PKEv:
Need more info - can't repeat the wrong behavior

zetok:
Dunno, I've just send a message to the contact, left qTox on overnight, and in the morning I've sent the second message.

I'll try again with another instance and will see in the morning tomorrow.

PKEv:
ok, i tried and how you told and manual change date in system - no problem

zetok:
Hmm, it works now – weird I probably launched the wrong binary. :/

PR #3199, opened 2016-04-25
feat(importProfile): Add way to import profile

### LINKED ISSUE #1872: [Regression] No way to import profile
Since new login window was introduced, qTox doesn't allow importing profiles any longer.

I think that `Import profile` should be added as an option in login window, above `New profile`. Or perhaps below it?

When selected, it should:
- allow for importing profile through `drag & drop` - i.e. drag profile file into designated for that rectangle in login window to import profile
- allow for importing profile file through file picker

About `drag & drop` - something like img.bi has, but with `Drop profile here or just click` and `Import`? (and obviously colour scheme fitting qTox):
e.g. img.bi

PKEv:
close  #1872
change based on pr #2140 by @agilob

### DIFF
--- a/src/widget/loginscreen.cpp
+++ b/src/widget/loginscreen.cpp
@@ -27,6 +27,7 @@
 #include "src/widget/form/setpassworddialog.h"
 #include "src/widget/translator.h"
 #include "src/widget/style.h"
+#include "src/widget/tool/profileimporter.h"
 #include <QMessageBox>
 #include <QDebug>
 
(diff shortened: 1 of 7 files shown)


src/widget/tool/profileimporter.cpp
```diff
@@ -0,0 +1,85 @@
+/*
+    Copyright © 2015 by The qTox Project
```
zetok: Wouldn't that be `2016`?

Unless it wasn't modified since 2015?

~same goes for `profileimporter.h` file

zetok: err, sorry,  `2015-2016` rather than `2016`

src/widget/loginscreen.cpp
```diff
@@ -255,3 +257,12 @@ void LoginScreen::retranslateUi()
 {
     ui->retranslateUi(this);
 }
+
+void LoginScreen::onImportProfile()
+{
+    ProfileImporter *pi = new ProfileImporter(this);
```
sudden6, line 263: Matching delete needed to avoid memory leak

PR #3204, opened 2016-04-25
fix(cppcheck): fix functions type

PKEv:
The adjustment based on the results of cppcheck analyzer

### DIFF
--- a/src/chatlog/chatline.h
+++ b/src/chatlog/chatline.h
@@ -70,13 +70,13 @@ public:
     void replaceContent(int col, ChatLineContent* lineContent);
     void layout(qreal width, QPointF scenePos);
     void moveBy(qreal deltaY);
-    void removeFromScene();
-    void addToScene(QGraphicsScene* scene);
-    void setVisible(bool visible);
-    void selectionCleared();
-    void selectionFocusChanged(bool focusIn);
+    void removeFromScene() const;
+    void addToScene(QGraphicsScene* scene) const;
+    void setVisible(bool visible) const;
+    void selectionCleared() const;
+    void selectionFocusChanged(bool focusIn) const;
 
-    int getColumnCount();
+    int getColumnCount() const;
     int getRow() const;
 
     ChatLineContent* getContent(int col) const;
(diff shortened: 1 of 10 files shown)


sudden6:
What is the reason for these changes, i.e what will they improve?

initramfs:
@sudden6 Those changes merely enforce const-correctness (except for HistoryKeeper and EncryptedDb which changes member to static access) which is mostly for programmers to guard against doing stupid things. Though they can aid in compiler optimizations as well. As long as it compiles, everything should be fine (no functionality change).

zetok:
> `fix(cppcheck): fix functions type`

Shouldn't that be `refactor`, since what code does doesn't change?

sudden6:
@initramfs thanks for the explaination.
@zetok yes, I think `refactor` would be better.

PKEv:
@initramfs thanks

src/chatlog/chatline.h
```diff
@@ -70,13 +70,13 @@ class ChatLine
...
     void layout(qreal width, QPointF scenePos);
     void moveBy(qreal deltaY);
-    void removeFromScene();
-    void addToScene(QGraphicsScene* scene);
-    void setVisible(bool visible);
-    void selectionCleared();
-    void selectionFocusChanged(bool focusIn);
+    void removeFromScene() const;
+    void addToScene(QGraphicsScene* scene) const;
+    void setVisible(bool visible) const;
```
antis81, line 75: The above three methods are no const methods, because they alter the object or it's contents.

src/chatlog/chatmessage.cpp
```diff
@@ -167,14 +167,14 @@ void ChatMessage::setAsAction()
...
-void ChatMessage::hideSender()
+void ChatMessage::hideSender() const
 {
     ChatLineContent* c = getContent(0);
     if (c)
         c->hide();
 }
 
-void ChatMessage::hideDate()
+void ChatMessage::hideDate() const
```
antis81, line 177: …and here

src/chatlog/chatmessage.cpp
```diff
@@ -167,14 +167,14 @@ void ChatMessage::setAsAction()
     action = true;
 }
 
-void ChatMessage::hideSender()
+void ChatMessage::hideSender() const
```
antis81, line 170: same here…

Issue #3209, opened 2016-04-26
Corrupted .tox profile file disables the ability to remove profile


ghost:
I have recently found that if I create a Tox profile and then I go into the actual ".tox" file and edit some of the data such as the name, the name is changed but as qTox doesn't seem to recognise the file when I log in and press the "Delete" button it claims to have deleted it, but it hasn't because when I go back to the login screen I am able to load that profile again.
Though I understand that this is something a user is not meant to do, and you could just say to people not to do it like it's not an upstream bug if you edit the code and make a bug. However this shows that if a Tox profile got corrupted qTox may not allow its removal through the GUI so there really should be something in place that means I can delete it through the GUI even if the entire ".tox" file has been replaced with something else.

I am running Ubuntu GNOME 15.10 with GNOME 3.18 and qTox version `1.4.0-14-g5045c46-1`.

initramfs:
I've looked through the file delete code and it looks fine (paths check out too). Except for a missing delete directive for chat history, everything seems to check out. I would like to actually test this but I've been having a hard time "editing fields" in the .tox file (how are you editing it?) and getting it prompt an invalid profile. I'm happy to submit a PR that outlines some basic changes (check if the file was actually deleted, delete chat history, and so forth) but it'd be nice to fix this issue as well.

If possible, could you send a zipped dummy corrupted profile (.ini, .db and .tox) such that I can reliably get it to ask me to delete the profile?

ghost:
@initramfs, If you just create a new Tox profile called "Testing", then once you have done so (and make sure that you leave it as an unencrypted one) edit the file (I use `vim` but most text editors should do) and change the value "Testing" wherever it is (should be one near the top of the file) to "Corrupted", then save the file and load it with the Tox GUI. Then try and delete it using the "Delete" button on the profile page in the Tox GUI.
It shouldn't be too hard to do...

initramfs:
I've carried out the above procedure and found that:
1. Tox loads the new profile without issue (the name actually changes to "Corrupt" (or whatever value I changed it to)
2. The delete function works as intended, deleting the profile files as appropriate.
3. Tested on both v1.4.0 (19d3ca0) and the commit in question (5045c46)

Can you make sure that only one instance of tox is running and that tox has the appropriate write permissions to the file?

ghost:
[...]

PR #3213, opened 2016-04-27
Show warning on failure to delete profile files

initramfs:
This PR adds warnings if Qt was unable to successfully delete a profile file (.tox, .ini, .db, etc). Potentially allowing users to perform manual deletion (or, in the case of bug reports regarding profile deletion [e.g. issue #3209], allows us to verify correct operation).

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -361,7 +361,12 @@ void ProfileForm::onDeleteClicked()
                 tr("Are you sure you want to delete this profile?", "deletion confirmation text")))
     {
         Nexus& nexus = Nexus::getInstance();
-        nexus.getProfile()->remove();
+
+        if(!nexus.getProfile()->remove())
+        {
+            GUI::showError(tr("Files could not be deleted!"), tr("Some files could not be deleted, please manually remove them."));
+        }
+
         nexus.showLogin();
     }
 }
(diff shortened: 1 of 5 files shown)


PKEv:
Maybe add some message to the user - logs few looks?

SkyzohKey:
Maybe make the string translatable ?

initramfs:
@PKEv @SkyzohKey New error dialog box now appears upon failure to delete any file, prompting the users to manually delete files. The strings are made translatable.

src/widget/form/profileform.cpp
```diff
@@ -361,7 +361,12 @@ void ProfileForm::onDeleteClicked()
                 tr("Are you sure you want to delete this profile?", "deletion confirmation text")))
     {
         Nexus& nexus = Nexus::getInstance();
-        nexus.getProfile()->remove();
+
+        if(!nexus.getProfile()->remove())
+        {
+            GUI::showError(tr("Files could not be deleted!"), tr("Some files could not be deleted, please manually remove them."));
```
sudden6: How should the user know which files to delete?

initramfs: @sudden6 Well, they could check the log but I see your concern. Should the prompt show the names of the file that could not be deleted? Becauses chances are, unless the user manually tampered with the filesystem, either all will delete or none will.

sudden6: > Should the prompt show the names of the file that could not be deleted?

Maybe you could list all filenames that couldn't be deleted at once? Might produce a space problem in the error message tough...

initramfs:
@sudden6 Now shows the filenames that failed to delete in the prompt.

Example below:

Failed to delete dialog

Issue #3223, opened 2016-04-28
About some translating


ghost:
I think I can help with translating, but https://github.com/tux3/qTox/wiki/Translating is hard to read.
I'm using Windows(I use Linux only for servers and nothing more).

In other project which is not hosted on github, I contributed to them by these steps:
1. Ask them for a single file.
2. They give a "single file" to me(e.g., "en_US.xml" "lang_1033.dll.source")
Other project example (language files are single, and can be opened by Notepadd++)

```
goodapp.exe
|-lang
|-------en_US.xml
|-------pt_BR.xml
```

en_US.xml example:
    APP_NAME(tab)Good App
    APP_TEXT_HELLO01(tab)Guten morgen!
1. I edit the file, and send it back.

zetok:
Which language(s?) would you want to help with?

I've thought a bit about it, and I think that providing those files in an automated way would be possible (regenerating them on every push to master), provided that @tux3 would help.

verycrypt:
> Which language(s?) would you want to help with?

I learned some Japanese while working with some people.
Japanese care their privacy very seriously, and translate (q|)Tox into it may gather some attention from them.

So, if you make the translation step simple(single file), I'll do it.

verycrypt:
I also can translate Russian and German, but I'm not confident to translate them perfectly.

zetok:
`日本語` is correct name, right?

Also, it should be located between `Nederlands` and `Norsk Bokmål`, right? (at least wikipedia does that):
![wikipedia sorting](https://cloud.githubusercontent.com/assets/3148759/14910378/a0533d5e-0de4-11e6-8a97-32ab9b2c2433.png)

To translate: [ja.zip](https://github.com/tux3/qTox/files/241919/ja.zip)

And yeah, it kinda does suck to add languages according to instructions on wiki.
(I've edited it to point to opening an issue requesting file to translate for now)

ghost:
> 日本語 is correct name, right?

Yep.

> it should be located between Nederlands and Norsk Bokmål, right?

I don't think so; let's just call it "Japanese (Nihongo)" or "Japanese (日本語)".
And sort other languages in alphabetic order; I like a-z sorting.

e.g.,

Chinese ('add chinese word here')
Japanese (日本語)
Korean ('add korean language here')
Nederlands
Norsk Bokmal

And hey, got a ja.ts file.

> <message>
> <location filename="../src/widget/form/settings/avform.cpp" line="225"/>
[...]

PR #3228, opened 2016-04-29
fix(i18n): remove duplicated text with HTML tags from translations

zetok:
Starting from d121bd839b1ca36f201b79a9ce9cef3a78ce72e0 this string is
being set from .cpp, thus there's no need to keep a duplicate with HTML
tags that make it harder to translate in .ui file.

Duplicate pointed out in #3223 by @verycrypt.


zetok:
>   `fix:`

I wonder if that shouldn't be `refactor:`, since what code does doesn't really change from user perspective..?

sudden6:
I think refactor: is the better choice here.

PR #3229, opened 2016-04-29
fix(simple_make.sh): add a way to make compiling on Fedora work

zetok:
By installing RPM Fusion repo, which provides required ffmpeg.

Also add missing qt5-qtsvg-devel dependency.

Closes #2998

Someone with Fedora should check whether it works.

PR #3232, opened 2016-04-29
fix(bootstrap): Adding auto compile of Sqlcipher to bootstrap (Fedora)

linux-modder:
Will need a review of  DIR pinnings but  otherwise should be ready to ship.


zetok:
Needed:
- pass travis test
- make installation optional depending on whether the distro is Fedora

zetok:
Closing, since #3229 was merged.

PR #3249, opened 2016-05-03
Delete widget from traced widgets list after reading message

### LINKED ISSUE #2247: Green "unread messages" notification stays forever.
After qTox start and then reception of an offline message, green "unread messages" notification stays on the bottom-left side, even after the message is read. It should dissapear then...

Diadlo:
Added public method Wigdet::updateScroll, which needed to update "Unread message" bar.

Fix #2247.

### DIFF
--- a/src/widget/contentdialog.cpp
+++ b/src/widget/contentdialog.cpp
@@ -484,18 +484,18 @@ bool ContentDialog::event(QEvent* event)
                 updateTitle(activeChatroomWidget);
 
                 Friend* frnd = activeChatroomWidget->getFriend();
+                Group* group = activeChatroomWidget->getGroup();
+
+                GenericChatroomWidget *widget = nullptr;
 
                 if (frnd)
-                {
-                    frnd->getFriendWidget()->resetEventFlags();
-                    frnd->getFriendWidget()->updateStatusLight();
-                }
+                    widget = frnd->getFriendWidget();
                 else
-                {
-                    Group* g = activeChatroomWidget->getGroup();
-                    g->getGroupWidget()->resetEventFlags();
-                    g->getGroupWidget()->updateStatusLight();
-                }
+                    widget = group->getGroupWidget();
+
+                widget->resetEventFlags();
+                widget->updateStatusLight();
+
                 Widget::getInstance()->resetIcon();
             }
 
--- a/src/widget/contentdialog.cpp
+++ b/src/widget/contentdialog.cpp
@@ -496,6 +496,7 @@ bool ContentDialog::event(QEvent* event)
                 widget->resetEventFlags();
                 widget->updateStatusLight();
 
+                Widget::getInstance()->updateScroll(widget);
                 Widget::getInstance()->resetIcon();
             }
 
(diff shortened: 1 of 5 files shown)


src/widget/notificationscrollarea.cpp
```diff
@@ -66,12 +66,23 @@ void NotificationScrollArea::trackWidget(GenericChatroomWidget* widget)
     }
 }
 
-void NotificationScrollArea::updateTracking()
+/**
+ * @brief Delete notification bar to visible elements on scroll area
+ */
+void NotificationScrollArea::updateVisualTracking() {
```
sudden6, line 72: This method is only used in the two `connect()` above, would it be possible to adjust the `connect()` calls to directly call `updateTracking()`?

src/widget/notificationscrollarea.cpp
```diff
@@ -66,12 +66,23 @@ void NotificationScrollArea::trackWidget(GenericChatroomWidget* widget)
...
+    updateTracking(nullptr);
+}
+
+/**
+ * @brief Delete notification bar to visible elements and to widget on scroll area
```
sudden6: You mean `Delete notification bar from visible elements and widget on sroll area` here?

src/widget/contentdialog.cpp
```diff
@@ -484,19 +484,21 @@ bool ContentDialog::event(QEvent* event)
...
-                    g->getGroupWidget()->updateStatusLight();
-                }
+                    widget = group->getGroupWidget();
+
+                widget->resetEventFlags();
+                widget->updateStatusLight();
+
+                Widget::getInstance()->updateScroll(widget);
                 Widget::getInstance()->resetIcon();
+
```
sudden6: unnecessary empty line

src/widget/notificationscrollarea.cpp
Diadlo, line 72: Yes. I try set default value to first argument of `updateTracking`, but have a truble.
But in this case I must write something like that:

```
connect(verticalScrollBar(), &QAbstractSlider::rangeChanged, this, static_cast<void (NotificationScrollArea::*)(void)>(&NotificationScrollArea::updateTracking));
```

src/widget/notificationscrollarea.cpp
Diadlo: Yes :)

src/widget/contentdialog.cpp
Diadlo: Fixed

zetok:
on cc490f56ddd0ce1361e8e0e13f8e638fa0a5b931 #2247 is still not fixed – the other way to reproduce is to:
1. Have an offline contact that after coming online will be visible on the friendlist, but is not visible when offline
2. While contact is offline, minimize qTox to tray
3. While qTox is minimized in tray, receive message from the contact that came online
4. Restore from tray
5. Clicking on unread message notification doesn't make it disappear

other way
(using `\n`s as a status message might prove helpful in reproducing)

Diadlo:
@zetok can't reproduce. When I open window if contact is still online it visible (condition of point 1.). If contact goes offline and invisible I haven't `new message` bar

zetok:
webm with steps to reproduce: [qtox.zip](https://github.com/tux3/qTox/files/262231/qtox.zip)

Diadlo:
I think, that found the reason of this behaviour.
`QWidget::mapTo` work incorrect to not yet repainted widgets. It use old widget position.
Should I call `updateVisualTracking()` every time (for example on repaint main window)?

Diadlo:
@zetok Fixed

PR #3252, opened 2016-05-04
docs(README.md): add info about meetings

zetok:
Also change unicode `⦁` to `|`, since e.g. on default Debian 8 install
`⦁` is not displayed correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,15 @@
 qTox
 ======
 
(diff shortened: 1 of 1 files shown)


zetok:
background: https://lists.tox.chat/pipermail/general/2016-May/000065.html

PR #3253, opened 2016-05-05
Enable HiDPI support if Qt supports it

initramfs:
Previously HiDPI support is conditionally enabled based on compilation parameters as well as environmental variables which makes binary distribution difficult as it requires users to either manually recompile from source and/or set environmental variables/use command-line parameters to fully enable HiDPI.

This PR automatically uses Qt's runtime HiDPI scaling as introduced with Qt 5.6, allowing automatic scaling of the application as necessary.

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -108,11 +108,6 @@ contains(DISABLE_FILTER_AUDIO, NO) {
      DEFINES += QTOX_FILTER_AUDIO
 }
 
-contains(HIGH_DPI, YES) {
-    QT_DEVICE_PIXEL_RATIO= auto
-    DEFINES += HIGH_DPI
-}
-
 contains(JENKINS,YES) {
     INCLUDEPATH += ./libs/include/
 } else {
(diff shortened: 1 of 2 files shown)


nurupo:
Adding support for automatic HiDPI scaling for Qt 5.6 is good, but why did you remove a way to enable HiDPI for Qt <5.6?

initramfs:
@nurupo Well the two would seem conflicting, one requires you to enable qmake flags and the other would automatically be enabled >= Qt 5.6. It'd cause minor confusion at best.

Regarding actually "enabling" HiDPI, that was never done prior to this PR. That had to be set via environmental variables or command line options. The only thing this PR disables when compiling < Qt 5.6 is the `Qt::AA_UseHighDpiPixmaps` attribute. Which marks a small visual degradation in clarity. Besides, I don't think many packages ship with that option enabled, I only found out there was such an option a few days ago when snooping around. I can confirm that both Windows and Arch packages don't.

Regarding the `QT_DEVICE_PIXEL_RATIO=auto` thing, that attribute has since been deprecated. I could add that back but that would mean adding a "HiDPI" compliation flag that only works for Qt 5.4 - 5.5.1 which seems kinda ugly. Besides, I thought that `QT_DEVICE_PIXEL_RATIO` was supposed to be an environmental variable set on the user's machine, does including it as a enviornmental variable during compliation do anything?

PR #3257, opened 2016-05-05
feat(loginscreen): Added CapsLock indicator

### LINKED ISSUE #3245: Password fields should indicate if caps lock are on
Like with other applications where you are entering your password or setting a new one it would be useful if qTox indicated when the caps lock is on by perhaps putting a warning triangle with an explanatory tooltip in the password box on the right when the caps lock is on.

Diadlo:
Fix #3245.

### DIFF
--- a/src/platform/capslock.h
+++ b/src/platform/capslock.h
@@ -1,0 +1,33 @@
+/*
+    Copyright © 2016 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifdef QTOX_PLATFORM_EXT
+
+#ifndef PLATFORM_CAPSLOCK_H
+#define PLATFORM_CAPSLOCK_H
+
+
+namespace Platform
+{
+    bool capsLockEnabled();
+}
+
+#endif // PLATFORM_CAPSLOCK_H
+
+#endif // QTOX_PLATFORM_EXT
--- a/src/platform/capslock_osx.cpp
+++ b/src/platform/capslock_osx.cpp
@@ -18,26 +18,13 @@
 */
 
 #include <QtCore/qsystemdetection.h>
-#if defined(Q_OS_UNIX) && !defined(__APPLE__) && !defined(__MACH__)
-#include <X11/XKBlib.h>
+#if defined(__APPLE__) && defined(__MACH__)
 #include "src/platform/capslock.h"
-#undef KeyPress
-#undef KeyRelease
-#undef FocusIn
-#undef FocusOut
 
 bool Platform::capsLockEnabled()
 {
-    Display * d = XOpenDisplay((char*)0);
-    bool caps_state = false;
-    if (d)
-    {
-        unsigned n;
-        XkbGetIndicatorState(d, XkbUseCoreKbd, &n);
-        caps_state = (n & 0x01) == 1;
-    }
-    return caps_state;
+    // TODO: implement for osx
+    return false;
 }
 
-
-#endif  // defined(Q_OS_UNIX) && !defined(__APPLE__) && !defined(__MACH__)
+#endif  // defined(__APPLE__) && defined(__MACH__)
--- a/src/platform/capslock_win.cpp
+++ b/src/platform/capslock_win.cpp
@@ -22,7 +22,7 @@
 #include <windows.h>
 #include "src/platform/capslock.h"
 
-bool Platform::checkCapsLock()
+bool Platform::capsLockEnabled()
 {
     return GetKeyState(VK_CAPITAL) == 1;
 }
--- a/src/platform/capslock_win.cpp
+++ b/src/platform/capslock_win.cpp
@@ -1,0 +1,30 @@
+/*
+    Copyright © 2016 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <QtCore/qsystemdetection.h>
+#ifdef Q_OS_WIN32
+#include <windows.h>
+#include "src/platform/capslock.h"
+
+bool Platform::checkCapsLock()
+{
+    return GetKeyState(VK_CAPITAL) == 1;
+}
+
+#endif  // Q_OS_WIN32
--- a/src/platform/capslock_x11.cpp
+++ b/src/platform/capslock_x11.cpp
@@ -1,0 +1,43 @@
+/*
+    Copyright © 2016 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <QtCore/qsystemdetection.h>
+#if defined(Q_OS_UNIX) && !defined(__APPLE__) && !defined(__MACH__)
+#include <X11/XKBlib.h>
+#include "src/platform/capslock.h"
+#undef KeyPress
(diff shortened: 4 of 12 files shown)


src/widget/loginscreen.cpp
```diff
@@ -31,6 +31,16 @@
 #include <QMessageBox>
 #include <QDebug>
 
+#ifdef Q_OS_WIN32
+    #include <windows.h>
+#else
+    #include <X11/XKBlib.h>
```
sudden6: This will introduce a new dependency, which will have to be listed in `INSTALLING.md` and also `simple_make.sh` will have to be updated.
Also new dependencies have to be agreed by our new committee.

sudden6:
Additionally for ui changes, it would be nice to have a screenshot, so one can comment without having to build your branch.

Diadlo:
spectacle a18295
spectacle m18295

ProMcTagonist:
Can we do something like this instead with a tooltip on the icon?

qtox capslock

src/widget/loginscreen.cpp
zetok: Err, this needs to be fixed:
- it's broken on osx
- introduction of a hard dep on xorg is a no-go – there's a define for ~that: https://github.com/tux3/qTox/blob/4432ab6398aa4c9cdc222cf46170b820b9e33fa0/qtox.pro#L127

zetok: @sudden6 also, it would need a decision if there was a disagreement whether dep should be introduced.

There's little to no point in discussing/voting on something that ~everyone agrees on.
With that being said, I have ~no opinion on this one – provided that code works on supported platforms, there are docs, etc. Generally, what I want is "get it to work".

If there is something that you perceive to be an issue, just simply state it – if it's something that simply needs to be fixed, then it needs to be fixed. If it's something that depends on ~opinion, there might be someone disagreeing, in which case that would need to be discussed/voted on, or no one would, in which case just proceed with whatever needs to be done.

And if there is something that you want opinions of someone else before proceeding – IMO there's little to no point in waiting for an official meeting to get it – just simply get it whenever you have a chance. Now, if the only chance of getting opinion of e.g. @tux3 on some matter is the meeting, that's a separate problem…

sudden6: ok, I interpreted "dependency changes" in the literal sense, because it'll break most build scripts. Of course it shouldn't need a vote if ~everyone agrees, but I think it should at least get brought up before merging.

ghost:
I'm in agreement with @ProMcTagonist on this. Just having it as text below is a little ugly, and the standard way of doing it is to have a little icon in the actual box.

src/widget/loginscreen.cpp
Diadlo: We can avoid xorg dependence of will check CapsLock status after first button press in password field (if user enter letter in uppercase without `shift` or lowercase with it). What you think about it?

zetok: @Diadlo There's no real need to do it, since qTox already "depends" on a few xorg-only things – what needs to be done, is to properly `define` things.

Diadlo:
@ProMcTagonist can you send to me this icon is svg format?

ProMcTagonist:
@Diadlo the one in that picture is proprietary (grabbed off google images), and I'm not an icon maker, but I edited [this](https://commons.wikimedia.org/wiki/Category:Arrow_icons#/media/File:Octicons-arrow-up.svg) into [this](https://a.pomf.cat/jtvpwu.svg) if that's usable (MIT?)

zetok:
@ProMcTagonist it's great :)

Btw, now it's possible to attach .zip files to issues :)

ProMcTagonist:
@zetok thanks :)

[caps lock.svg.zip](https://github.com/tux3/qTox/files/256500/caps.lock.svg.zip)

Diadlo:
I have a problem. I try add image `:img/caps_lock.svg` , it is invisible. But if I use, for example `:img/add.svg` all works good. What I do wrong?

zetok:
You need to add it to `res.qrc`

src/platform/capslock.h
```diff
@@ -0,0 +1,33 @@
+/*
+    Copyright © 2014 by The qTox Project
```
zetok: `2016`?
`2014, 2016` if used parts of code from file from 2014?

src/platform/capslock_win.cpp
```diff
@@ -0,0 +1,30 @@
+/*
+    Copyright © 2014-2015 by The qTox Project
```
zetok: `2014-2016` ?

src/platform/capslock_x11.cpp
```diff
@@ -0,0 +1,43 @@
+/*
+    Copyright © 2014 by The qTox Project
```
zetok: `2014, 2016` ?

src/platform/capslock.h
Diadlo: Sorry. Just copy from other file :)
Fixed

Diadlo:
I have small problem. When I apply `padding-right: 20px;` style to QLineInput, it broke other paddings
spectacle g17596
spectacle g17597

zetok:
Could you perhaps add tooltip for the caps-lock icon with something like `CAPS-LOCK ENABLED`?

Diadlo:
@zetok Added :+1: 
spectacle q18297

Diadlo:
Do you have any idea?

Diadlo:
I think, it is complete

src/platform/capslock_osx.cpp
```diff
@@ -0,0 +1,29 @@
...
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <QtCore/qsystemdetection.h>
+#if defined(__APPLE__) && defined(__MACH__)
+#include "src/platform/capslock.h"
+
+bool Platform::capsLockEnabled()
+{
+    return false;
```
zetok, line 27: um, I would think that osx does support caps-lock detection – could you at least add `TODO: implement for osx` ?

Diadlo, line 27: @zetok Yes, sorry. Fixed

zetok:
Ok, now the only missing part is documentation about the new optional dependency.

Diadlo:
@zetok What and where I should add?

zetok:
The absolute minimum is https://github.com/tux3/qTox/blob/master/INSTALL.md#dependencies

Diadlo:
@zetok Added

zetok:
**New "optional" dependency**: b7677fc58f516f242b1e297797836d8137c96afb
cc affected package maintainers: @abbat, @farseerfc, @mpxc, @vith, @yurivict

Hopefully I didn't miss anyone – if I did, please say so.

sudden6:
@Diadlo I think it would make sense to make the capslock indicator an independent class, because then we could reuse it in `profileform` also.

Diadlo:
@sudden6 Good idea :+1: 
Completed

src/widget/capslockindicator.cpp
```diff
@@ -0,0 +1,42 @@
...
+    cleanInputStyle = parentWidget()->styleSheet();
+
+    QIcon icon = QIcon(":img/caps_lock.svg");
+    setIcon(icon);
+    QSize iconSize(inputSize.height(), inputSize.height());
+    setIconSize(iconSize);
+    setCursor(Qt::ArrowCursor);
+    move(inputSize.width() - inputSize.height(), 0);
+    setStyleSheet("border: none; padding: 0;");
+    setToolTip("CAPS-LOCK ENABLED");
```
sudden6: I think this should be translateable

Diadlo: @sudden6 Of course. Fixed.

sudden6:
spectacle z18979

I found a small styling issue, the text in the tooltip is unreadable when using a dark theme for the desktop. Tooltips in the rest of qTox and also the username and password are ok.

Reproduceable with https://opendesktop.org/content/show.php?content=124222 KDE colorscheme.

Also the tooltip text still seems not to be translateable.

Diadlo:
@sudden6 Fixed

src/widget/capslockindicator.cpp
```diff
@@ -0,0 +1,42 @@
+#include "capslockindicator.h"
+#ifdef QTOX_PLATFORM_EXT
+#include "src/platform/capslock.h"
+#endif
+
+CapsLockIndicator::CapsLockIndicator(QWidget *parent) : QToolButton(parent) {
```
sudden6: Coding Style: opening `{` braces belong to new line

Diadlo: Fixed.

sudden6:
@Diadlo while trying to merge I found a bug:

The icon doesn't appear for languages which shorten the input field, eg. German, Russian,...

spectacle d18565

I think it's because the Icon is moved to it's position only when creating it, but not on subsequent resize/move events.

Diadlo:
@sudden6 Ok I have a big problem. I can't get real size of QLineEdit. size() and frameSize() give me `QSize(597, 32)` (width bigger, than window width) In real it's size around `QSize(130, 23)`

sudden6:
@Diadlo maybe that's already the bug? Maybe the QLineEdit is already extending beyond the window, because of some unknown reason and this makes your code fail?

Diadlo:
@sudden6 timing :) Right now I found the reason of it. Qt calculate size only when widget visible.

Diadlo:
@sudden6 Fixed
spectacle h28222

src/widget/capslockindicator.h
```diff
@@ -9,6 +9,9 @@ class CapsLockIndicator : QToolButton
     CapsLockIndicator(QWidget *widget);
     void updateIndicator();
 
+public slots:
```
sudden6: Is there a reason to make this a slot?

src/widget/capslockindicator.cpp
```diff
@@ -2,20 +2,27 @@
...
-    QSize iconSize(inputSize.height(), inputSize.height());
-    setIconSize(iconSize);
     setCursor(Qt::ArrowCursor);
-    move(inputSize.width() - inputSize.height(), 0);
     setStyleSheet("border: none; padding: 0; color: white");
     setToolTip(tr("CAPS-LOCK ENABLED"));
+    updateSize();
+}
+
+void CapsLockIndicator::updateSize() {
```
sudden6: `}` should go on a new line

Diadlo: Argg.. Always forget about it. Fixed

sudden6:
There's still the question about the `public slot` instead of the `public` method.

src/widget/capslockindicator.h
Diadlo: Fixed

src/widget/capslockindicator.cpp
sudden6: If you are using qt-creator you can set the coding style to "Qt" which is what we are using.

sudden6:
Thanks, great work :+1:

PR #3266, opened 2016-05-08
feat(l10n): Delete tr() for toxme errors

Diadlo:
Deleted tr() call for errors which cannot be reproduced throw qtox client

### DIFF
--- a/src/net/toxme.cpp
+++ b/src/net/toxme.cpp
@@ -288,21 +288,65 @@ Toxme::ExecCode Toxme::deleteAddress(QString server, ToxId id)
     return extractError(response);
 }
 
+/**
+ * @brief Return string of the corresponding error code
+ * @param errorCode Code to get error message
+ * @return Source error message
+ */
 QString Toxme::getErrorMessage(int errorCode)
 {
     switch (errorCode) {
     case IncorrectResponse:
-        return QObject::tr("Incorrect response");
+        return "Incorrect response";
     case NoPassword:
-        return QObject::tr("No password in response");
+        return "No password in response";
+    case ServerError:
+        return "Server doesn't support Toxme";
+    case -1:
+        return "You must send POST requests to /api";
+    case -2:
+        return "Problem with HTTPS connection";
+    case -3:
+        return "I was unable to read your encrypted payload";
+    case -4:
+        return "You're making too many requests. Wait an hour and try again";
+    case -25:
+        return "This name is already in use";
+    case -26:
+        return "This Tox ID is already registered under another name";
+    case -27:
+        return "Please don't use a space in your name";
+    case -28:
+        return "Password incorrect";
+    case -29:
+        return "You can't use this name";
+    case -30:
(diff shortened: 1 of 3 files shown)


sudden6:
Why exactly do you want to remove these translations? They are shown to the user here: https://github.com/tux3/qTox/blob/master/src/widget/form/profileform.cpp#L522-L523 so I don't see any reason not to translate them?

Diadlo:
@sudden6 because it is error Toxme server. 
For example: `You must send POST requests to /api`. Request type controled by qtox. And if user have this error, he, or she can't do anything to change it

tux3:
We should probably log those errors in english in profileform.cpp with qWarning(), and show a generic translated error message to the user, that they can actually understand.
If you could do that in the PR, then it's reasonable to not translate those.

sudden6:
+1 for logging all errors, but I think stuff like "You're making too many requests. Wait an hour and try again" should also be shown to the user. Like @Diadlo said, everything the user can change should also get displayed in the GUI

Diadlo:
Fixed

src/widget/form/profileform.cpp
```diff
@@ -520,7 +520,10 @@ void ProfileForm::onRegisterButtonClicked()
             break;
         default:
             QString errorMessage = Toxme::getErrorMessage(code);
-            GUI::showWarning(tr("Toxme error"),  errorMessage);
+            qWarning() << errorMessage;
+            QString translated = Toxme::translateErrorMessage(code);
+            if (translated != QString())
```
sudden6: We should still show the default error message here, because else the user might be confused why it didn't work, but showed no message.

src/net/toxme.cpp
```diff
@@ -319,13 +363,9 @@ QString Toxme::getErrorMessage(int errorCode)
...
         return QObject::tr("Tox ID not sent");
-    case -41:
-        return QObject::tr("Lookup failed because the server replied with invalid data");
     case -42:
         return QObject::tr("That user does not exist");
-    case -43:
-        return QObject::tr("Internal lookup error. Please file a bug");
     default:
-        return QObject::tr("Unknown error (%1)").arg(errorCode);
+        return QString();
```
sudden6: I'd prefer if the default message returned something like "Internal ToxMe error" and not an empty string.

src/widget/form/profileform.cpp
Diadlo: Fixed

src/net/toxme.cpp
Diadlo: Fixed

sudden6:
The commit message of your second commit looks like a mistake, if it was, please change. The rest seems ready for merge.

Diadlo:
@sudden6 Sorry. Fixed :)

Diadlo:
I have small problem. When I apply `padding-right: 20px;` style to QLineInput, it broke other paddings
spectacle f17597
spectacle g17597

PR #3281, opened 2016-05-09
fix(simple_make.sh): add sqlite dependencies for Fedora

zetok:


linux-modder:
WFM on Fedora 23, GTG in my  book..
  for a  virgin updated install no less

PR #3282, opened 2016-05-09
feat(l10n): Update Polish translation from Weblate

zetok:


tux3:
Why the change to lt.ts?

zetok:
https://github.com/tux3/qTox/blob/master/tools/pull-translations.sh#L50 does it automatically – different versions of lupdate put there different version..

Want me to return it back to `2.1`?

tux3:
No, no, I was just wondering why it only did that to lt.ps in particular, but the difference shouldn't matter.

zetok:
Well, technically, all the .ts were affected when I first ran it on the .ts files.

Then there was a PR with translations to `lt.ts`, which changed it to `2.1`, and here we go again.. :D

tux3:
It'll do.

PR #3289, opened 2016-05-10
Make markdown strings translatable

agilob:

PR #3293, opened 2016-05-11
Translatable markdown settings & activity labels translated by locale

zetok:
Renamed & squashed commits from #3289


src/widget/friendlistwidget.cpp
```diff
@@ -286,6 +288,9 @@ void FriendListWidget::setMode(Mode mode)
         categoryNever->setName(tr("Unknown", "Category for sorting friends by activity"));
         activityLayout->addWidget(categoryNever);
 
+        delete ql;
+        ql = 0;
```
sudden6: I think this line should be `ql = nullptr;`

PR #3294, opened 2016-05-11
Translations update

zetok:
- [x] Depends on #3293

### DIFF
--- a/translations/ar.ts
+++ b/translations/ar.ts
@@ -312,49 +312,6 @@ which may lead to problems with video calls.</source>
     </message>
 </context>
 <context>
-    <name>Android</name>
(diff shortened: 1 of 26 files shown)


sudden6:
uhm, isn't this just #3293 plus polish translations update? Why keep the other PR open then?

zetok:
It additionally contained also commit with update of all translation files.

I've removed that commit, since other PR changed translation stuff, and possibly other PRs that will be merged before this one will also change stuff to translate – in this case I'd rebase this PR on top of master, and add a commit with update of all translation files again.

As for why separate PR for translations – that's the way @tux3 likes it.

Issue #3196, opened 2016-04-25
Add an option to disable logging.


ghost:
##### Brief Description

OS: Windows 7
qTox version: v1.4.0-2-g24a7f4d
Commit hash:  (isn't this hash value a privacy?)
toxcore:  0.0.0
Qt:  5.5.1
Hardware:  ?
…

Reproducible: Always
##### Steps to reproduce
1. Launch qTox.
##### Observed Behavior

qTox start writing many lines to "qTox.log".
##### Expected Behavior

qTox doesn't write anything.

Let the user disable logging by adding another setting to settings.
Advanced -> [X] Enable Logging (when turned off, never create/write logs)

---

zetok:
> Commit hash: (isn't this hash value a privacy?)

No.

> Add an option to disable logging.

What is the use case?

ghost:
> What is the use case?

Writing any log to the disk(not memory) is bad for security and privacy.
"Debug log" is useful only for developers, not for normal users.
If I need support/want to help, I will enable debug log to know what's going on.
There's no reason to enable logging as a default.
Am I clear?

> Commit hash: No.

So, what's your hash? (just to make sure you're using same as mine)

ghost:
Another example: if I moved my qTox to USB disk, I want qTox to write less.
(USB stick has limited read/write counts)

zetok:
> So, what's your hash? (just to make sure you're using same as mine)

Here you go: https://github.com/tux3/qTox/commits/master ← on the right

> Writing any log to the disk(not memory) is bad for security and privacy.

If you care about stuff like that, you wouldn't use windows in the first place, and on top of that you'd configure your system to allow writes only to tmpfs, with other partitions mounted RO. Furthermore, you'd use FDE. Just in case.

I'm asking about the actual reasons, not some keywords.

> Another example: if I moved my qTox to USB disk, I want qTox to write less.

And that would be a valid reason. Not really good enough, given that one could configure things to mount `~/.cache` / `$HOME` directory as tmpfs, which would result in 0 writes to your persistent memory, regardless of whether that would be from qTox, or other software. And it's not qTox's job to do that.

Also, given the great durability & ~low cost of pendrives these days, it's not a real issue even when there is a bit of writes from qTox.

# 

So far, 1 weak use case, which is solved by using other tools, available on any Linux distro. If your OS doesn't support that, I'd suggest to get some better OS.

And if there is a good reason, I would like to hear about it :)

# 

On a side note..

[...]

PR #3295, opened 2016-05-12
feat(widget): Disable sound notification for `busy` status

### LINKED ISSUE #2893: [Regression] notification sound isn't being suppressed when status is being set to `busy`
Description:
When user status is set to `busy`, qTox should suppress notification sound for the new message. Probably due to a fix to the #2328 it doesn't do that any longer.

qTox: 67620f96617312ccbb27c721c2c3090f036d52e3

Reproducible: Always

Steps to reproduce:
1. Set status to `Busy`
2. Make sure that `Play sound` is checked in settings
3. Receive a message when qTox is minimized to tray

What happens:
qTox plays notification sound for a new message.

What should happen:
qTox should should not play the notification sound on a new message.

Diadlo:
Fix #2893.

BREAKING CHANGE: Disabling notification sound for `busy` status

### DIFF
--- a/src/widget/form/settings/generalsettings.ui
+++ b/src/widget/form/settings/generalsettings.ui
@@ -39,8 +39,8 @@
        <rect>
         <x>0</x>
         <y>0</y>
-        <width>1278</width>
-        <height>1413</height>
+        <width>1270</width>
+        <height>1468</height>
        </rect>
       </property>
       <layout class="QVBoxLayout" name="verticalLayout_4" stretch="0,0,0,1">
@@ -327,6 +327,20 @@ instead of closing itself.</string>
                  <string>Play sound</string>
                 </property>
                </widget>
+              </item>
+              <item>
+               <layout class="QVBoxLayout" name="verticalLayout_13">
+                <property name="leftMargin">
+                 <number>40</number>
+                </property>
+                <item>
+                 <widget class="QCheckBox" name="busySound">
+                  <property name="text">
+                   <string>Play sound while Busy</string>
+                  </property>
+                 </widget>
+                </item>
+               </layout>
               </item>
               <item>
                <widget class="QCheckBox" name="showWindow">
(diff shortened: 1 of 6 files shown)


ovalseven8:
I think it would be cool if the user could set in the settings if he wants to have sound notification when status is "busy".

Diadlo:
I think more settings equals less usability. Not sure

Chiitoo:
I will probably never understand how exactly more options makes things less usable.  They should in my opinion do the exact opposite: make things even more usable.  That is of course unless every user /has/ to go through all the settings and options, but that isn't needed with defaults being OK.

They could be hidden behind a reverse Konami code or something, just so the “common user” doesn't ever see them unless they really want to.  ^^

That said, I'm in favour of disabling the notification sound while 'busy', or at least having an option to do so.

Diadlo:
Ok. If I add this in settings, where it should be?

Chiitoo:
If it were to be not hidden, how about something simple like this:

```
On new message:
    [ ] Play sound
        [ ] Play sound while busy
    [ ] Open window
        [ ] Focus window
```

Diadlo:
What do everybody think about adding new point in settings?

zetok:
> What do everybody think about adding new point in settings?

Basically https://github.com/tux3/qTox/issues/3196#issuecomment-214376759.

With that being said, @Chiitoo's proposal isn't bad – it's just that ~whole `Chat` part of `General` tab should be moved to a new settings tab.

src/widget/form/settings/generalsettings.ui
```diff
@@ -329,6 +329,20 @@ instead of closing itself.</string>
...
               </item>
               <item>
+               <layout class="QVBoxLayout" name="verticalLayout_13">
+                <property name="leftMargin">
+                 <number>40</number>
+                </property>
+                <item>
+                 <widget class="QCheckBox" name="busySound">
+                  <property name="text">
+                   <string>Play sound while busy</string>
```
zetok: `s/busy/Busy/` ?

Diadlo: Fixed

Diadlo:
@zetok rebased

PR #3297, opened 2016-05-12
[WIP] Added drag and drop remote file

### LINKED ISSUE #1276: Can't drag and drop online images into qTox
For example, dragging
exampleimage into the qTox chat gives a "wasn't able to open" error:
dragfile

Dragging it into the chat box just displays the URL:
dragfile2

Diadlo:
When drag'n'drop remote file in qTox chat it will be downloaded in temporary directory and sent from it.

Fix #1276.

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -64,6 +64,8 @@
 #include "src/video/camerasource.h"
 #include "src/nexus.h"
 #include "src/persistence/profile.h"
+#include <QMimeDatabase>
+#include <QMimeType>
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
 #include <QTemporaryDir>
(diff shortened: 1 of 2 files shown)


SkyzohKey:
Does this introduce ip leak if a proxy is set?

Diadlo:
@SkyzohKey It use proxy for file downloads, if it's what you mean

zetok:
on b0a0617e04351d1abe0a0e86ac61b294a0680001
- [x] Fails with

``` html
<html>
<body>
<img src="https://i.chzbgr.com/full/8764238592/hA2A477F4/"
    alt="Cat :3" title="Cat :3"></img>
</body>
</html>
```

zetok:
on b0a0617e04351d1abe0a0e86ac61b294a0680001
- [x] with things like https://pbs.twimg.com/media/CiP5ubQXIAAgv1t.jpg:large drag & drop can cause problems on windows, given that filename is something like `qTox-KUe9EFCiP5ubQXIAAgv1t.jpg:large` → https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#filesystem

Diadlo:
@zetok If I reset file name it will be normal

Diadlo:
@zetok If I reset file name, does it will be normal?

zetok:
If you reset file name, how will you know file extension? On windows it actually does mean something.

Diadlo:
@zetok. MIME type database

zetok:
- [x] segfault on 76a58d699eb3631afb79c1ec6e25364f76a82c5a: https://gist.github.com/zetok/e650971450dafa4c29c3d7f20ba6dbeb

zetok:
So, the general problem is that I don't think that it's really possible to make this functionality actually work reliably – out of images that I've tested with, most of file transfers resulted in `.html` files being transferred, rather than graphical ones.

Now, this is a problem with websites, but that doesn't matter, since it's the qTox that provides the functionality. From user PoV browser offers a `Save image as..` button, so they rightfully would expect from qTox to send an image which they have dragged & dropped. Instead, qTox sends some `.html` file…

Diadlo:
@zetok About problem with `.html`: users frequently use drag and drop in other applications. I think, that they accustomed to this behaviour

Diadlo:
@zetok Fixed

zetok:
- [x] still segfaults on dba938a9751d89d6a8827695e7a40bb070c573be: https://gist.github.com/zetok/8806fd6220d076a1a596cff6495d3cc0
- [ ] there's no indicator that qTox is doing anything when it is downloading big file before sending it to peer – if not for monitoring network speed, I'd suspect that functionality simply doesn't work

Diadlo:
@zetok I can't reproduce. But believe, that fixed now
What kind of indicator are you want to see? Change cursor on "waiting" will be normal?

zetok:
> What kind of indicator are you want to see?

Probably should be something like file transfer widget – dunno about colour, perhaps white?
It should say that $file is being downloaded to send it to your friend.

And since would be similar to file transfer widget, user should be able to see progress, cancel and pause download.

zetok:
on 8a9c666
- [ ] file doesn't stop downloading when user logs out & tries to exit

zetok:
on 8a9c666
- [ ] when download finishes "after" exit, there are leftover files in `/tmp`

ProMcTagonist:
Not going to be worked on anymore?

Diadlo:
@ProMcTagonist Not now. Maybe later

Issue #1167, opened 2015-02-15
non-alphabetic(?) nicks aren't highlighted


zetok:
from https://github.com/krepa098/qTox/issues/5 : 

![muh.. no highlight](https://cloud.githubusercontent.com/assets/3148759/6093592/73bf079c-aefd-11e4-86c5-282b314e475f.png)

dunno if chatform related

EDIT: though it worked with old chatform time ago

zetok:
Still reproducible on e797bf958279ea9eaa858bd377532b22de79bef7

ProMcTagonist:
If a nickname starts with $ and I type $ and hit tab, I get $$restofnick. Seems like nick handling in general doesn't work well with non alphabetic characters.

PR #3298, opened 2016-05-12
fix: correctly tab-complete nicks starting with `$`

zetok:
Without it, nicks starting with `$` would have doubled `$`.

Continuation of #1377

Pointed out by @ProMcTagonist in #1167


Diadlo:
> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

PR #3304, opened 2016-05-13
fix(bootstrap.sh):Re-add /usr/local/locallibs

linux-modder:
Somewhere in the  optimization of bootstrap.sh recently  the  following line as removed:

> ```
> echo '/usr/local/lib/' | sudo tee -a /etc/ld.so.conf.d/locallib.conf
> ```

This  commit  reverts that omission so  a  compile  works properly  after  a  sudo make install && qtox &

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -170,8 +170,8 @@ if [[ $INSTALL_TOX = "true" ]]; then
         make install
     else
         sudo make install
-    echo '/usr/local/lib/' | sudo tee -a /etc/ld.so.conf.d/locallib.conf
-	sudo ldconfig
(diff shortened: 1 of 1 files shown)


zetok:
> Somewhere in the optimization of bootstrap.sh recently the following line as removed:

Err, it was never included in the first place.

bootstrap.sh
```diff
@@ -170,7 +170,8 @@ if [[ $INSTALL_TOX = "true" ]]; then
         make install
     else
         sudo make install
-        sudo ldconfig
+    	echo '/usr/local/lib/' | sudo tee -a /etc/ld.so.conf.d/locallib.conf
+	sudo ldconfig
```
zetok, line 174: Could you fix indentation?

linux-modder, line 174: yeah sorry

PR #3305, opened 2016-05-14
fix(profile): change password buttons behaviour

### LINKED ISSUE #3300: [UI] Profile - show/hide "x password" button
##### Brief Description

OS: Windows 7 32
qTox version: v1.4.0-68-g8d225a8
Commit hash: 8d225a8
Qt: 5.5.1
…

Reproducible: Always
##### Steps to reproduce
1. In "profile" window, look at "profile" section.
##### Observed Behavior

"Remove Password" and "Change Password" is always there.
##### Expected Behavior

if tox-profile-isEncrypted === true then show ("Remove Password" and "Change Password")
else show ("Set new password")
hide else
fi

a68366:
When profile is encrypted:

Not encrypted:

Closes #3300

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -393,15 +393,15 @@ void ProfileForm::onLogoutClicked()
 
 void ProfileForm::setPasswordButtonsText()
 {
-    if (!Nexus::getProfile()->isEncrypted())
-    {
-        bodyUI->changePassButton->setText(tr("Set profile password", "button text"));
-        bodyUI->deletePassButton->setVisible(false);
-    }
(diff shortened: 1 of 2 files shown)


src/widget/form/profileform.cpp
```diff
@@ -389,6 +391,20 @@ void ProfileForm::onLogoutClicked()
     nexus.showLogin();
 }
 
+void ProfileForm::setPasswordButtonsText()
+{
+    if (!Nexus::getProfile()->isEncrypted())
```
zetok: `!` is unnecessary – simply change placement of statements.

PR #3308, opened 2016-05-15
fix(bootstrap.sh):Re-add /usr/local/locallibs--spacing should be fixed now)

linux-modder:


antis81:
@linux-modder You deleted the file and I doubt this is what you intended, right? Further, please cleanup and squash your branch into a single commit? In your case, you also need to "reword" ad3d035 with message of 3dc1699.

sudden6:
close since implemented in #3402

PR #3309, opened 2016-05-16
refactor(audio): remove obsolete signal "groupAudioPlayed"

antis81:
Removed the unconnected (broken) and now obsolete `Audio::groupAudioPlayed` signal, which got replaced by the single `Core::groupAudioPlaying` signal. This signal is emitted correctly from within `CoreAV::groupCallCallback`.

Following the code flow, I found the former implementation is pretty much bullshit and - for whatever reason - the "active (group & audio & video) peer" was assumed to be the "loudest" peer. This is completely wrong. However if anyone gets the intention behind this, I'd be glad to hear about what it should actually do and I'll do my very best to reimplement it.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3309)

PR #3311, opened 2016-05-16
feat(audio): add real gain control of the input device

### LINKED ISSUE #2756: qtox - 406-7f34290 - horrible echo on audio / video call
Hello,

I am doing some test calls between 2 laptops (in different rooms) and audio is not usable...
Echo is truly horrible, disabling one of microphones breaks the loop. Both clients use latest qtox with 'Filter audio' ON (which doesn't seem to work). Video is OK.

I see nothing special in log:

```
[14:39:16.096] src/core/coreav.cpp:142 : Debug: "answering call 0"
[14:39:16.107] src/widget/form/chatform.cpp:305 : Debug: onAvStart
[14:39:16.107] src/widget/form/chatform.cpp:511 : Debug: disableCallButtons
[14:39:16.112] src/widget/form/chatform.cpp:621 : Debug: creating netcam
[14:39:16.113] src/video/netcamview.cpp:91 : Debug: SIZER QSize(640, 480)
[14:39:18.707] src/audio/audio.cpp:550 : Debug: Audio output not closed -> there are pending subscriptions.
[14:39:23.437] src/video/camerasource.cpp:209 : Debug: Opening device  "/dev/video0"
[14:41:42.343] src/widget/form/chatform.cpp:420 : Debug: onHangupCallTriggered
[14:41:42.343] src/core/coreav.cpp:211 : Debug: "Cancelling call with 0"
[14:41:42.343] src/audio/audio.cpp:402 : Debug: Unsubscribed from audio input device [ 0 subscriptions left ]
[14:41:42.343] src/audio/audio.cpp:666 : Debug: stopping audio capture
[14:41:42.343] src/audio/audio.cpp:670 : Debug: Closing audio input
[14:41:42.344] src/audio/audio.cpp:768 : Debug: Audio source 2 deleted. Sources active: 0
[14:41:42.344] src/audio/audio.cpp:688 : Debug: Closing audio output
[14:41:42.346] src/widget/form/chatform.cpp:491 : Debug: enableCallButtons
[14:41:42.346] src/widget/form/chatform.cpp:511 : Debug: disableCallButtons
[14:41:42.347] src/widget/form/chatform.cpp:504 : Debug: timer started!!
[14:41:42.379] src/video/camerasource.cpp:282 : Debug: Closing device  "/dev/video0"
[14:41:43.922] src/widget/form/chatform.cpp:542 : Debug: onEnableCallButtons
```

Also, when calling there is a long delay after pressing answer until call is actually initiated, I can see this here:

```
[14:39:18.707] src/audio/audio.cpp:550 : Debug: Audio output not closed -> there are pending subscriptions.
[14:39:23.437] src/video/camerasource.cpp:209 : Debug: Opening device  "/dev/video0"
```

It's a 5 second gap for... I don't know what, there were no other tasks for qtox.

And both clients are on ubuntu one 14.04 and one 15.04

### LINKED ISSUE #3109: Allow higher Volume / Gain
Operating System: Windows 10
qTox version: 1.3.0
Sound-Card: ASUS Xonare Essence STX Audio Device
Microphone: König CMP-MIC8
Headphones: Sennheiser HD 380 Pro

The person who I am talking to is also using qTox and has set the volume and microphone gain to maximum.

If i turn the microphone level and Volume to maximum, it is still quite low compared to other programs which have the same global volume in Windows. If i tune up the volume in windows such that the volume of qTox is normal, other programs are really loud.

A workaround would be to increase just the qtox volume in windows, since it is possible to adjust the volume for a single program. One problem of this workaround is, that the incrementation of the qTox volume also increases the global volume and every program has the same volume as the global volume when it gets started -> so every newly started program (e.g. a media player) is loud.

In Skype the volume is fine using the same Hardware and OS.

I'd like to have greater range for the volume and / or gain such that i can adjust the qtox volume to a higher level as the current maximum.

antis81:
This PR adds a long awaited feature. We now have a controllable **Audio Gain**, no really! :smile:

Gain can be controlled in positive/negative dB values from -30~30dB and allows for amplification or mitigation of the input signal level, resulting in highly improved audio quality. Thanks to @initramfs for the technical advice and review.

~~The gain is calculated with the common `10*log10()` algorithm, which is normally used for dB calculation of the electrical power level - and it gives pretty good results. Values can be set via qTox's public Audio API as percentage values (between 0 and 1).~~

Note: This PR also brings back const-correctness to `Audio`; some minor cleanup with naming conventions also included.

~~Important: As this has quite an impact on qTox's Audio experience --> Anyone who is able to compile qTox is invited to try this PR from a users view point!~~

The following values should fit for most common microphones, producing splendid clear results. Helps to check vs. [Tox EchoBot](https://toxme.io/u/echobot)):
- System settings:
  - Master Volume: ~80%
  - Gain: ~50% (~10dB)
- qTox Audio Settings:
  - Master volume: ~80%
  - Gain: ~60-85%

**TODO:**
- [x] Add sane min/max gain boundaries (-30/+30 dB) in `Audio`.

**UI will be done in separate PR**
- Add tick marker in slider at "unity gain" level.
- Add clipping indicator.
- [x] ~~Change label text in A/V settings to "Input gain" or only "Gain".~~ --> Wrong German translation.

Closes the following issues:
- close #3109 
- close #2756 

Have fun talking with qTox!

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -27,6 +27,7 @@
 #include <QMutexLocker>
 #include <QPointer>
 #include <QThread>
+#include <QtMath>
 #include <QWaitCondition>
 
 #include <cassert>
@@ -34,6 +35,49 @@
 #ifdef QTOX_FILTER_AUDIO
 #include "audiofilterer.h"
 #endif
+
+/**
+@internal
+
+@class Audio::Private
+
+@brief Encapsulates private audio framework from public qTox Audio API.
+*/
+class Audio::Private
+{
+public:
+    Private()
+        : minInputGain{-30.0}
+        , maxInputGain{30.0}
+        , gain{0.0}
+        , gainFactor{0.0}
+    {
+    }
+
+    qreal inputGain() const
+    {
+        return gain;
+    }
+
+    qreal inputGainFactor() const
+    {
+        return gainFactor;
+    }
+
+    void setInputGain(qreal dB)
+    {
+        gain = qBound(minInputGain, dB, maxInputGain);
+        gainFactor = qPow(10.0, (gain / 20.0));
+    }
+
+public:
+    qreal   minInputGain;
+    qreal   maxInputGain;
+
+private:
+    qreal   gain;
+    qreal   gainFactor;
+};
 
 /**
 Returns the singleton instance.
@@ -45,15 +89,15 @@ Audio& Audio::getInstance()
 }
 
 Audio::Audio()
-    : audioThread(new QThread)
-    , alInDev(nullptr)
-    , inGain{1.f}
-    , inSubscriptions(0)
-    , alOutDev(nullptr)
-    , alOutContext(nullptr)
-    , alMainSource(0)
-    , alMainBuffer(0)
-    , outputInitialized(false)
+    : d{new Private}
+    , audioThread{new QThread}
+    , alInDev{nullptr}
+    , inSubscriptions{0}
+    , alOutDev{nullptr}
+    , alOutContext{nullptr}
+    , alMainSource{0}
+    , alMainBuffer{0}
+    , outputInitialized{false}
 {
     // initialize OpenAL error stack
     alGetError();
@@ -87,6 +131,7 @@ Audio::~Audio()
 #ifdef QTOX_FILTER_AUDIO
     filterer.closeFilter();
 #endif
+    delete d;
 }
 
 void Audio::checkAlError() noexcept
@@ -106,13 +151,14 @@ void Audio::checkAlcError(ALCdevice *device) noexcept
 /**
 Returns the current output volume (between 0 and 1)
 */
-ALfloat Audio::outputVolume()
+qreal Audio::outputVolume() const
 {
     QMutexLocker locker(&audioLock);
 
     ALfloat volume = 0.0;
 
-    if (alOutDev) {
+    if (alOutDev)
+    {
         alGetListenerf(AL_GAIN, &volume);
         checkAlError();
     }
@@ -125,28 +171,78 @@ Set the master output volume.
 
 @param[in] volume   the master volume (between 0 and 1)
 */
-void Audio::setOutputVolume(ALfloat volume)
+void Audio::setOutputVolume(qreal volume)
 {
     QMutexLocker locker(&audioLock);
 
-    volume = std::max(0.f, std::min(volume, 1.f));
+    volume = std::max(0.0, std::min(volume, 1.0));
 
-    alListenerf(AL_GAIN, volume);
+    alListenerf(AL_GAIN, static_cast<ALfloat>(volume));
     checkAlError();
 }
 
-ALfloat Audio::inputVolume()
+/**
+@brief The minimum gain value for an input device.
+
+@return minimum gain value in dB
+*/
+qreal Audio::minInputGain() const
 {
     QMutexLocker locker(&audioLock);
-
-    return inGain;
+    return d->minInputGain;
 }
 
-void Audio::setInputVolume(ALfloat volume)
+/**
+@brief Set the minimum allowed gain value in dB.
+
+@note Default is -30dB; usually you don't need to alter this value;
+*/
+void Audio::setMinInputGain(qreal dB)
 {
     QMutexLocker locker(&audioLock);
+    d->minInputGain = dB;
+}
 
-    inGain = std::max(0.f, std::min(volume, 1.f));
+/**
+@brief The maximum gain value for an input device.
+
+@return maximum gain value in dB
+*/
+qreal Audio::maxInputGain() const
+{
+    QMutexLocker locker(&audioLock);
+    return d->maxInputGain;
+}
+
+/**
+@brief Set the maximum allowed gain value in dB.
+
+@note Default is 30dB; usually you don't need to alter this value.
+*/
+void Audio::setMaxInputGain(qreal dB)
+{
+    QMutexLocker locker(&audioLock);
+    d->maxInputGain = dB;
+}
+
+/**
+@brief The dB gain value.
+
+@return the gain value in dB
+*/
+qreal Audio::inputGain() const
+{
+    QMutexLocker locker(&audioLock);
+    return d->inputGain();
+}
+
+/**
+Set the input gain dB level.
+*/
+void Audio::setInputGain(qreal dB)
+{
+    QMutexLocker locker(&audioLock);
+    d->setInputGain(dB);
 }
 
 void Audio::reinitInput(const QString& inDevDesc)
@@ -172,7 +268,8 @@ void Audio::subscribeInput()
 {
     QMutexLocker locker(&audioLock);
 
-    if (!autoInitInput()) {
+    if (!autoInitInput())
+    {
         qWarning("Failed to subscribe to audio input device.");
         return;
     }
@@ -235,7 +332,8 @@ bool Audio::initInput(QString inDevDescr)
     const uint16_t frameDuration = AUDIO_FRAME_DURATION;
     const uint32_t chnls = AUDIO_CHANNELS;
     const ALCsizei bufSize = (frameDuration * sampleRate * 4) / 1000 * chnls;
-    if (inDevDescr.isEmpty()) {
+    if (inDevDescr.isEmpty())
+    {
         const ALchar *pDeviceList = Audio::inDeviceNames();
         if (pDeviceList)
             inDevDescr = QString::fromUtf8(pDeviceList, strlen(pDeviceList));
@@ -246,10 +344,13 @@ bool Audio::initInput(QString inDevDescr)
                                        sampleRate, stereoFlag, bufSize);
 
     // Restart the capture if necessary
-    if (!alInDev) {
+    if (!alInDev)
+    {
         qWarning() << "Failed to initialize audio input device:" << inDevDescr;
         return false;
     }
+
+    d->setInputGain(Settings::getInstance().getAudioInGain());
 
     qDebug() << "Opened audio input" << inDevDescr;
     alcCaptureStart(alInDev);
@@ -273,7 +374,8 @@ bool Audio::initOutput(QString outDevDescr)
 
     assert(!alOutDev);
 
-    if (outDevDescr.isEmpty()) {
+    if (outDevDescr.isEmpty())
+    {
         // default to the first available audio device.
         const ALchar *pDeviceList = Audio::outDeviceNames();
         if (pDeviceList)
@@ -293,7 +395,8 @@ bool Audio::initOutput(QString outDevDescr)
     alOutContext = alcCreateContext(alOutDev, nullptr);
     checkAlcError(alOutDev);
 
-    if (!alcMakeContextCurrent(alOutContext)) {
+    if (!alcMakeContextCurrent(alOutContext))
+    {
         qWarning() << "Cannot create output audio context";
         return false;
     }
@@ -306,7 +409,8 @@ bool Audio::initOutput(QString outDevDescr)
     checkAlError();
 
     Core* core = Core::getInstance();
-    if (core) {
+    if (core)
+    {
         // reset each call's audio source
         core->getAv()->invalidateCallSources();
     }
@@ -421,7 +525,8 @@ void Audio::cleanupOutput()
 {
     outputInitialized = false;
 
-    if (alOutDev) {
+    if (alOutDev)
+    {
         alSourcei(alMainSource, AL_LOOPING, AL_FALSE);
         alSourceStop(alMainSource);
         alDeleteSources(1, &alMainSource);
@@ -486,8 +591,15 @@ void Audio::doCapture()
     }
 #endif
 
-    for (size_t i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
-        buf[i] *= inGain;
+    for (quint32 i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
+    {
+        // gain amplification with clipping to 16-bit boundaries
+        int ampPCM = qBound<int>(std::numeric_limits<int16_t>::min(),
+                                 qRound(buf[i] * d->inputGainFactor()),
+                                 std::numeric_limits<int16_t>::max());
+
+        buf[i] = static_cast<int16_t>(ampPCM);
+    }
 
     emit frameAvailable(buf, AUDIO_FRAME_SAMPLE_COUNT, AUDIO_CHANNELS, AUDIO_SAMPLE_RATE);
 }
@@ -495,7 +607,7 @@ void Audio::doCapture()
 /**
 Returns true if the input device is open and suscribed to
 */
-bool Audio::isInputReady()
+bool Audio::isInputReady() const
 {
     QMutexLocker locker(&audioLock);
     return alInDev && inSubscriptions;
@@ -504,7 +616,7 @@ bool Audio::isInputReady()
 /**
 Returns true if the output device is open
 */
-bool Audio::isOutputReady()
+bool Audio::isOutputReady() const
 {
     QMutexLocker locker(&audioLock);
     return alOutDev && outputInitialized;
@@ -530,12 +642,14 @@ void Audio::subscribeOutput(ALuint& sid)
 {
     QMutexLocker locker(&audioLock);
 
-    if (!autoInitOutput()) {
+    if (!autoInitOutput())
+    {
         qWarning("Failed to subscribe to audio output device.");
         return;
     }
 
-    if (!alcMakeContextCurrent(alOutContext)) {
+    if (!alcMakeContextCurrent(alOutContext))
+    {
         qWarning("Failed to activate output context.");
         return;
     }
@@ -554,8 +668,10 @@ void Audio::unsubscribeOutput(ALuint &sid)
 
     outSources.removeAll(sid);
 
-    if (sid) {
-        if (alIsSource(sid)) {
+    if (sid)
+    {
+        if (alIsSource(sid))
+        {
             alDeleteSources(1, &sid);
             qDebug() << "Audio source" << sid << "deleted. Sources active:"
                      << outSources.size();
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -75,6 +75,10 @@ AVForm::AVForm() :
                 .arg(audio.maxInputGain()));
     bodyUI->microphoneSlider->setMinimum(qRound(audio.minInputGain()) * 10);
     bodyUI->microphoneSlider->setMaximum(qRound(audio.maxInputGain()) * 10);
+    bodyUI->microphoneSlider->setTickPosition(QSlider::TicksBothSides);
+    bodyUI->microphoneSlider->setTickInterval(
+                (qAbs(bodyUI->microphoneSlider->minimum()) +
+                 bodyUI->microphoneSlider->maximum()) / 4);
     bodyUI->microphoneSlider->setTracking(false);
     bodyUI->microphoneSlider->installEventFilter(this);
     connect(bodyUI->microphoneSlider, &QSlider::valueChanged,
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -45,6 +45,8 @@ AVForm::AVForm() :
     bodyUI = new Ui::AVSettings;
     bodyUI->setupUi(this);
 
+    const Audio& audio = Audio::getInstance();
+
     bodyUI->btnPlayTestSound->setToolTip(
                 tr("Play a test sound while changing the output volume."));
 
@@ -65,6 +67,14 @@ AVForm::AVForm() :
     bodyUI->playbackSlider->installEventFilter(this);
     connect(bodyUI->playbackSlider, &QSlider::valueChanged,
             this, &AVForm::onPlaybackValueChanged);
+
+    bodyUI->microphoneSlider->setToolTip(
+                tr("Use slider to set the gain of your input device ranging"
+                   " from %1dB to %2dB.")
+                .arg(audio.minInputGain())
+                .arg(audio.maxInputGain()));
+    bodyUI->microphoneSlider->setMinimum(qRound(audio.minInputGain()) * 10);
+    bodyUI->microphoneSlider->setMaximum(qRound(audio.maxInputGain()) * 10);
     bodyUI->microphoneSlider->setTracking(false);
     bodyUI->microphoneSlider->installEventFilter(this);
     connect(bodyUI->microphoneSlider, &QSlider::valueChanged,
@@ -378,7 +388,7 @@ void AVForm::onInDevChanged(QString deviceDescriptor)
     Audio& audio = Audio::getInstance();
     audio.reinitInput(deviceDescriptor);
     bodyUI->microphoneSlider->setEnabled(bodyUI->inDevCombobox->currentIndex() != 0);
-    bodyUI->microphoneSlider->setSliderPosition(audio.inputVolume() * 100.f);
+    bodyUI->microphoneSlider->setSliderPosition(qRound(audio.inputGain() * 10.0));
 }
 
 void AVForm::onOutDevChanged(QString deviceDescriptor)
@@ -390,7 +400,7 @@ void AVForm::onOutDevChanged(QString deviceDescriptor)
     Audio& audio = Audio::getInstance();
     audio.reinitOutput(deviceDescriptor);
     bodyUI->playbackSlider->setEnabled(audio.isOutputReady());
-    bodyUI->playbackSlider->setSliderPosition(audio.outputVolume() * 100.f);
+    bodyUI->playbackSlider->setSliderPosition(qRound(audio.outputVolume() * 100.0));
 }
 
 void AVForm::onFilterAudioToggled(bool filterAudio)
@@ -414,10 +424,10 @@ void AVForm::onPlaybackValueChanged(int value)
 
 void AVForm::onMicrophoneValueChanged(int value)
 {
-    Settings::getInstance().setInVolume(value);
+    const qreal dB = value / 10.0;
 
-    const qreal percentage = value / 100.0;
-    Audio::getInstance().setInputVolume(percentage);
+    Settings::getInstance().setAudioInGain(dB);
+    Audio::getInstance().setInputGain(dB);
 }
 
 void AVForm::createVideoSurface()
--- a/src/audio/audio.h
+++ b/src/audio/audio.h
@@ -56,20 +56,26 @@ class Audio : public QObject
 {
     Q_OBJECT
 
+    class Private;
+
 public:
     static Audio& getInstance();
 
-    ALfloat outputVolume();
-    void setOutputVolume(ALfloat volume);
+    qreal outputVolume() const;
+    void setOutputVolume(qreal volume);
 
-    ALfloat inputVolume();
-    void setInputVolume(ALfloat volume);
+    qreal minInputGain() const;
+    void setMinInputGain(qreal dB);
+    qreal maxInputGain() const;
+    void setMaxInputGain(qreal dB);
+    qreal inputGain() const;
+    void setInputGain(qreal dB);
 
     void reinitInput(const QString& inDevDesc);
     bool reinitOutput(const QString& outDevDesc);
 
-    bool isInputReady();
-    bool isOutputReady();
+    bool isInputReady() const;
+    bool isOutputReady() const;
 
     static const char* outDeviceNames();
     static const char* inDeviceNames();
@@ -114,11 +120,13 @@ private:
 #endif
 
 private:
+    Private* d;
+
+private:
     QThread*            audioThread;
-    QMutex              audioLock;
+    mutable QMutex      audioLock;
 
     ALCdevice*          alInDev;
-    ALfloat             inGain;
     quint32             inSubscriptions;
     QTimer              captureTimer, playMono16Timer;
 
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -236,7 +236,7 @@ void Settings::loadGlobal()
     s.beginGroup("Audio");
         inDev = s.value("inDev", "").toString();
         outDev = s.value("outDev", "").toString();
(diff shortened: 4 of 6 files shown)


zetok:
cc @towlie ?

towlie:
For what am I needed here? Only compiling for windows?
I saw no issues while testing.

antis81:
> For what am I needed here? Only compiling for windows?

@towlie This and I'd like to get some feedback about how usable that is. Yep, thanks!

If someone could check it shortly on OSX, this would be nice. Other than that -> we're ready to merge.

src/audio/audio.cpp
```diff
@@ -486,16 +498,30 @@ void Audio::doCapture()
     }
 #endif
 
-    for (size_t i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
-        buf[i] *= inGain;
+    for (quint32 i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
+    {
+        int16_t ampPCM = static_cast<int16_t>(qRound(buf[i] * gain));
+        buf[i] = static_cast<int16_t>(ampPCM);
```
initramfs, line 601: Why is the `int16_t` being cast again to an `int16_t`?

src/audio/audio.cpp
```diff
@@ -486,16 +498,30 @@ void Audio::doCapture()
     }
 #endif
 
-    for (size_t i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
-        buf[i] *= inGain;
+    for (quint32 i = 0; i < AUDIO_FRAME_SAMPLE_COUNT * AUDIO_CHANNELS; ++i)
+    {
+        int16_t ampPCM = static_cast<int16_t>(qRound(buf[i] * gain));
```
initramfs: Given that the gain can be as high as 20 now, this line does not stop the samples from railing against `INT16_MAX` (or it's negative), insteading allowing it to roll over, creating horrible distortion instead of digital peaking.

initramfs:
This code dramatically changes the gain characteristics of the input source from having unity gain at 100% to having unity gain at 10^(1/10)% (~ 1.26%). Whilst this would work well for most microphone sources who's preamps supply signals way below the input line level, people who have microphones or input sources that supply at nominal line levels will have a very hard time controlling the gain since deviation in either direction (around ~1.26%) results in a very large change in the resulting gain.

If possible, I would prefer if unity was somewhere at say 30% of the way (or at worst 10% of the way) with a small marker indicating where it is (and perhaps the slider can snap to it when near).

src/audio/audio.cpp
```diff
@@ -486,16 +498,30 @@ void Audio::doCapture()
...
 
 /**
+@internal
+*/
+void Audio::privateSetGain(qreal percentage)
+{
+    percentage = std::max(0.0, std::min(percentage, 1.0));
+    gainPercentage = percentage;
+
+    gain = (percentage > 0.0) ? 10.0 * qAbs(log10(percentage * 100.0)) : 0.0;
```
initramfs: The use of `qAbs()` in this context causes incorrect behaviour when input gain percentage is < 1% (it causes a massive rise in gain since the log curve is reflected upwards).

src/audio/audio.cpp
antis81, line 601: ~~Indeed not required here. Will remove it.~~

Fixed!

src/audio/audio.cpp
antis81: Fixed!

src/audio/audio.cpp
antis81: Fixed!

antis81:
> If possible, I would prefer if unity was somewhere at say 30% of the way (or at worst 10% of the way) with a small marker indicating where it is (and perhaps the slider can snap to it when near).

Sure, the markers are fine. We don't want to snap, as it's less accurate and resolution is kinda rough by 1% per tick. So let's see how the markers work for users. I'd rather finish and re-animate the gain level meter providing a visual feedback.

@initramfs Thanks for the discussion. Really helpful. :)

src/audio/audio.cpp
initramfs: Removing the abs is no better since now gain can go negative (causing channel inversion) with the same effect (increasing negative gain). If this new gain curve is to be used, the log value has to be clipped at some lower bound (0 for example, such that it doesn't go negative).

antis81: Yeah, I just remembered why I put it there. Actually removing the abs changes "nothing" here as values are always within range 1..100 (percentage \* 100.0) -> I'll re-add the abs to be on the safe side. Btw. Percentage is already clipped above.

Still 1% results in a 0 gain factor…??

initramfs: Sorry, maybe I'm missing something but where does it indicate that percentages are always greater than 1%? If this is guaranteed by some other code perhaps add a comment somewhere (in the function documentation perhaps). You may need to check consistentcy elsewhere as well (particularly in `inputGainPercentage()` where the return documentation states percentage can be anything between 0 and 100%).

TheNain38:
@antis81 All of the functions and variables having `percentage` in their name in this code, aren't actual percentage, normally, this is a value between 0 and 100 (because it is a value to be divised by 100), so, I suggest you rename these to something else than `percentage`, because it is inducing that this should be a value between 0 and 100.
See [here](https://en.wikipedia.org/wiki/Percentage)

antis81:
@TheNain38 Thanks for the comment. So what? :)

Like written in the API description, the gain level describes a non-linear amplification on a logarithmic scale (0..20dB). Any linear representation (e.g. in percent) is by far not optimum. A gain level of 50% amplifies the input signal already by ~17dB, while 100% represents maximum 20dB amplification. The optimum would be probably to actually create a logarithmic slider, which we currently do not have. :)

TheNain38:
@antis81 I mean, in your code, for example, [here](https://github.com/tux3/qTox/pull/3311/files#diff-f23c32cfb93eaa31ca828e9e9f75b2a8R517), you called the argument `percentage`, when it actually is a value between 0 and 1, and not a percentage (which is between 0 and 100 by definition), so I was suggesting you just change the name of variables and functions, using `percentage` in their names (whithout it being a real percentage) :)

initramfs:
@TheNain38 To my knowledge a percentage essentially represents a fraction out of 100. The common notation of appending the % symbol after a valid indicates the value is to be taken (divided) out of 100. In the wikipedia article you linked (emphasis mine):

> For example, 45% (read as "forty-five percent") is equal to 45⁄100, or **0.45**.

In my experience I've seen percentages expressed (in a programmatic context) as both integers (0 ~ 100) and floating points (where 1.0f is 100%). As such, I don't see a problem with the representation antis81 choses to use within this PR.

I may have misrepresented your concern in terms of consistentcy. If percentages are represented elsewhere as `qreal`s bounded between 0.0 ~ 100.0 (especially within Qt's libraries) then I support your decision to change public functions to take inputs bounded between 0.0 ~ 100.0.

Just my two cents.

initramfs:
@antis81 You need to be careful with your terminology, this code **does not** provide 20 dB of gain, it instead provides around 26.02 dB of gain. Absolute gain value is **not** equal to decibel gain value. At 50% of the input "gain percentage" the actual gain is ~24.6 dB.

Upon further inspection of your code, it appears you may have the volume control backwards (i.e. you need to use exponentials instead of logarithms). Since human hearing scales exponentially (i.e. you need exponentially more power/voltage/signal for a linear increase in perceived volume).

Take a look at [this diagram](http://sound.westhost.com/pots-f4.gif) denoting graphs of potentiometers (passive electronic devices that are commonly used as volume knobs for audio appliances). The "log pot" is somewhat of a misnomer in this regard.

Furthermore, I think the audio code would benefit greatly if all percentage units are replaced with dB units since percentage units are extremely vague.

As it currently stands, when this PR gets merged existing percentage units saved in configuration files will now mean something completely different. If I had my input gain/volume (this PR makes this terminology vague) set to 30% this would translate to a 70% **reduction** in absolute signal level (or a 10.45 dB **reduction**). After this PR gets merged the same slider value of 30% will now translate to a 1377% **increase** in absolute signal level (or a 23.39 dB **increase**).

If the audio code could instead simply deal in linear units of dB (and calculate the gain based on that). It would make things much simpler (and allow my earlier concern regarding unity gain at 30% to be implemented more easily).

See [this page](http://www.sengpielaudio.com/calculator-amplification.htm) on calculating dB gain from absolute gain.

antis81:
@initramfs 
Glad talking to someone interested in the maths behind the gain topic. I'm just scratching the surface (with the objective to make things better for the qTox community). :)

**FYI:** With the algo and value range used in this PR we get **only increased** PCM values. The minimum factor `10*log(2) = 3.01". Value of 0 or 1 results in a muted signal, which is unwanted here of course -> we want the factor to be 1 (==no amplification) in this case. Thus, there's never a decrease happening as the factor doesn't drop below 1.

:question_mark: Does decreasing the input level make any sense in our use case?

After thinking and reading more about adding gain to PCM encoded signal, I think we are best off by providing several lookup tables with the pre-calculated factors in a defined resolution. Current resolution would be 100 values, but we can use 200 or 500 as well. The advantage of this is, that we can set the slider's maximum to `sizeof(arr)-1` and it always uses the correct resolution. Further we can actually map the distance to linear values, so that a slider in the middle represents actually "50%" of maximum amplification. For the algo I think it's best to let the user decide between logarithmic, exponential or unity gain - depending on what fits best for him/her. For me, the `10*log10` gain provides excellent results and I do not need anything else.

I'd like to state out one thing clearly:
I'm always talking about **input gain**, not "input volume", which absolutely makes no sense in this context. Maybe this came out a little confusing in the code due to "historical reasons", but gain is what I always meant in this context.

This being said, I don't think the actual dB does matter to qTox users and we're good to go without exposing any technical background.

> Furthermore, I think the audio code would benefit greatly if all percentage units are replaced with dB units since percentage units are extremely vague.

Well, we should use the right thing for the right purpose. So… percentage for output volume (OpenAL and Qt both use the 0..1 percentage terminology) and dB (I think it's best using a LUT with index) for the gain.

@initramfs Would you mind creating the LUT(s) with maybe 100 (or maybe better resolution) `float` factors?

initramfs:
@antis81 

> With the algo and value range used in this PR we get only increased PCM values.

Which, under the conditions of uniform gain application applied across all samples will translate into a output gain of the same factor.

> Does decreasing the input level make any sense in our use case?

Yes, because not everyone uses input devices that input at low levels. Some people may have devices that have a high average level (of say - 3 dBFS) and want to cut it down to match their friends for example. I understand the fact that lowering the gain is a feature that'll be less common for the masses of people using qTox but that's the current way that input is being handled (see below), merging this PR would essentially remove this feature (which I personally find useful anyway).

> After thinking and reading more about adding gain to PCM encoded signal, I think we are best off by providing several lookup tables with the pre-calculated factors in a defined resolution.

By using decibel measure directly, one can linearly interpolate without loss the appropriate gain value for a given percentage/slider position, eliminating the need for any lookup table (see below).

> For me, the `10*log10` gain provides excellent results and I do not need anything else.

This is probably due to the fact that the log curve rises very quickly initially, allowing good control of gain that somewhat seems reasonable. Whilst this is not a bad idea, I personally believe you should give a try to my proposal below.

<hr>

Just to clear things up, I'll state a new proposed algorithm for gain control (it's not really new, the same algorithm used in basically all gain control systems that I've dealt with [I'm somewhat of an audio engineer]).

The current qTox implementation of "gain" is best described like this:
master-gain

The y axis "Gain (absolute)" represents the actual gain multiplier applied to the PCM samples whereas the x axis "Gain (percentage)" represents the input percentage taken by `audio.cpp` as well as the slider position. In this (and all futher diagrams) the blue line represents the relation and the red line represents the position of unity gain.

In the current qTox implementation of input "gain", we have something that's essentially a linear cut in which 100% represents unity and 0% represents 0 absolute gain. In this system the highest gain is +0 dB (i.e. no amplification at all) whilst providing a linear cut to -∞ dB. It's clear the see why people will complain about low microphone levels since there is no way to digitally boost the levels beyond what the line in level from the audio source is.

Your PR changes the gain characteristics to this:

pr-gain

It's easy to see why this PR can satisfy the needs of most since it massively increases the digital gain to ~26.02 dB (instead of 0.0 dB from the current implementation). Not only that, the log curve fits nicely since it boosts gain very fast as you traverse up the slider which makes it sound like a linear travesal in slider position is indeed causing a proportional increase in perceived audio output. The shortfalls of this method are that it no longer provides a way to accurately control gain values below 0.0 dB. Another issue is that for people with truly quiet input devices that require 80% or more gain, the changes between 80, 90 and 100% sliders don't feel like they change audio output level at all (given the nature of the curve compounded with the nature of human hearing).

My proposed method of control is to treat the input not as percentages, but directly as decibel values. The slider can have something like -20 dB on one side and +20 dB on the other in which the slider will **linearly** traverse between the maxima and the minima. This is illustrated as:

proposal-20

Note how using decibel measure (and then converting it to absolute gain via exponents) allows a linear value (dB/slider position) to be transformed into gain values that scale with human hearing. Not only that, such a control allows both amplification and damping to be applied to the PCM samples, retaining existing functionality whilst adding the ability to boost the samples by large gain ratios. In the case that people see more value in amplfication than in damping, we can bias the 0 dB point more to the left, allowing more positive gain headroom for amplication (as shown below):

proposal-10-30

The choices of values (20 dB/-10~30 dB) is arbitary here, any range of values will work (I chose said values to best match the range that you've set out and had success with). No LUT is required since we can map any slider positon/value to a decibel value via simple linear algebra. Any precision can be specified by the slider since the dB -> gain multiplier algorithm is a simple exponential (more precisely `10^(dB / 20)`) which can be calculated as stored everytime the slider is moved (as your code currently does).

The user does not need knowledge of what decibels is, only that a linear change in the slider results in a perceived linear increase/decease in output level throughout the whole slider (e.g. moving from 50% of the slider to 60% of the slider feels like it gets louder by the same quantity as moving from 80% to 90%).

I urge you try this algorithm and see if it works for you (I really don't see why it won't).

antis81:
@initramfs Great to hear you are kind of involved in sound design - I'm kind of hobbyist musician :). I implemented your proposed algorithm. **Surprise:** With my lousy setup, the log algorithm works better. With this information, I think the optimum would be combining the linear algo for negative amp and the log for positive amp. With the 10^(dB/20) algo I found, that a range from -20dB~12dB (x10 => 320 slider ticks) fits best. This is wide range and values near or above 12dB easily result in heavy distortion and clipping effects. So the proposed algo is a good trade-off, but not optimum. I will add the tick marker at unity gain (see todo list in description). After this is done we're pretty much through with the PR and ready to merge.

initramfs:
@antis81

> This is wide range and values near or above 12dB easily result in heavy distortion and clipping effects.

That cannot be right, 12 dB corresponds to an absolute gain of 10^(12/20) = ~3.981. From your initial comments, you claim that you get good results with your log algorithm at the 50~65% range, this corresponds to an absolute gain of **at least** 16.89 or nearly **4.27** times larger than what you claim to cause distortion. Are you sure you've implemented the proposed algorithm correctly?

> I think the optimum would be combining the linear algo for negative amp and the log for positive amp

I would oppose changing algorithms between positive and negative gain since it essentially causes confusion as the sliders no longer change sound in any way proportional to the perceived audio change.

> I found, that a range from -20dB~12dB

Any reason to put more on the negative side? I thought you brought up the point that negative gain isn't very useful (which I somewhat agree with and thus proposed a bias towards positive gain).

antis81:
> Are you sure you've implemented the proposed algorithm correctly?

Yep. But maybe we are misunderstanding. Please have a look for yourself in `Audio::privateSetGain` at [audio.cpp:L518](https://github.com/tux3/qTox/pull/3311/commits/a558f03967ea4a77a4b5745cb5b28de2b8224c7a#diff-f23c32cfb93eaa31ca828e9e9f75b2a8R518).

initramfs:
@antis81 Yep, that implementation is correct.

If possible, can you print the associated gain dB vs gain absolute values for +12 dB versus the gain values you get for your log curve/algorithm at 50% percent position? Because my earlier concern still holds...

[audio.cpp:L262](https://github.com/tux3/qTox/pull/3311/files#diff-f23c32cfb93eaa31ca828e9e9f75b2a8R262) uses an incorrect multiplier to calculate dB gain values from settings.

src/audio/audio.cpp
```diff
@@ -251,6 +259,8 @@ bool Audio::initInput(QString inDevDescr)
         return false;
     }
 
+    privateSetGain(Settings::getInstance().getInVolume() * 0.01);
```
antis81: Of course multiplier no longer applies here -> will be removed. Thanks @initramfs.

src/audio/audio.cpp
```diff
@@ -486,16 +496,32 @@ void Audio::doCapture()
...
     emit frameAvailable(buf, AUDIO_FRAME_SAMPLE_COUNT, AUDIO_CHANNELS, AUDIO_SAMPLE_RATE);
 }
 
 /**
+@internal
+*/
+void Audio::privateSetGain(qreal dB)
+{
+    gainDecibel = dB;
+    gain = qPow(10.0, (dB * 0.2));
```
antis81: @initramfs Duh! Factor is wrong -> will be changed to **0.05**. Will add some graphs visualizing the (corrected) values. Stay tuned.

antis81:
@initramfs 

Here's the graph for my first logarithmic approach with positive (percentage) integer values ranging from 1~100:
logarithmic

And here's current (corrected) exponential approach using lower amp factors within ~~current~~ range -20~12dB:
exponential

I now widened the range from **-30~30dB**, which now also provides good results in both directions (lowest factor: ~0.03 ; highest factor: ~31.62).

src/audio/audio.cpp
initramfs: @antis81 Wow... I missed that too... Must have been real late when I read that.

src/audio/audio.cpp
```diff
@@ -251,6 +259,8 @@ bool Audio::initInput(QString inDevDescr)
         return false;
     }
 
+    privateSetGain(Settings::getInstance().getInVolume());
```
initramfs: Just to tidy up, you need to make sure the input values from the settings are sane (since users with old configurations may contain some very wacky values). I'd suggest defaulting to unity gain (0.0 dB) if values are out of range.

initramfs:
@antis81 Read the line comment I made earlier, but everything else looks good to me. A few small (but not entirely necessary) things that can be added are a marker to where unity gain is (halfway on the slider), a peak indicator (if the UI allows for it) to indicate when samples are railing against 0 dBFS.

src/audio/audio.cpp
antis81: Yes, that's totally ok. I guess it makes sense adding a (probably customizable) `minGain/maxGain` values to Audio API, defaulting to +-30dB.

antis81:
@initramfs We had a gain meter implemented (removed it, because it was unfinished) and I think it makes sense reviving the idea, only doing it right this time.

Both, the marker and a gain indicator or meter are totally doable with acceptable effort.

antis81:
@initramfs
Could you do a last complete code review please? I am going to implement the UI features in another PR as this introduces new widget(s) completely independent of qTox audio API.

Most relevant changes:
- added the min/max gain values; set to -30/+30dB; stored in `Audio::Private` object
- min/max gain corrects wrong user settings; **defaults** to 0dB unity gain, if settings value does not exist; settings value **bounds** to min/max gain in dB (e.g. previously 100% ==> +30dB ==> users will moan about great feedback for one last time :smile:).
- audio input device is initalized with gain from settings ("audio/inVolume") in `Audio::initInput`; same applies to slider in a/v settings form; **result:** Users will no longer think, that settings are not correctly saved, because slider always jumps to minimum/maximum. :)

**One important thing I noticed when implementing this:**
As the term "inVolume" is wrong in settings anyway, IMO it is best to introduce a new value "audio/inGain" and leave the old as a "dead" value. This would reset everything for users and initialize with the 0dB unity gain. Would also prevent people from moaning about the big feedback, before checking their audio configuration.

@tux3 @zetok
Whoever writes the documentation: We should give a hint to users in release notes, that audio settings changed and they need to re-adjust their in/out levels (setting sliders to out:~80/in:~70 "percent" is a good common setup).

antis81:
@TheNain38 Btw.: No percentage values were harmed during the making of this PR (**EDIT:** a.k.a no percentages are used any longer). :smile:

src/audio/audio.cpp
```diff
@@ -36,6 +37,37 @@
...
+        , maxInputGain{30.0}
+        , gain{0.0}
+        , gainDecibel{0.0}
+    {
+    }
+
+    void setInputGain(qreal dB)
+    {
+        gainDecibel = qBound(minInputGain, dB, maxInputGain);
+        gain = qPow(10.0, (dB * 0.05));
```
initramfs: The parameter `gain` is being calculated from the unbounded input argument `dB`, resulting in incorrect behaviour when out of bounds values are used (i.e. whilst `gainDecibel` will correctly clip to +-30 dB, `gain` will be arbitarily set based on the input dB value, which could potentially come from wrong/outdated settings).

Not an issue but, is there a particular reason to use `dB * 0.05` instead of `dB / 20.0`? I feel the latter is more clear.

src/audio/audio.cpp
```diff
@@ -36,6 +37,37 @@
...
+    }
+
+public:
```
initramfs, line 73: Keep these fields public somewhat defeats the above function. Any function can arbitarily modify the fields without the logic defined above. In particular, it allows for the consistentcy of the four parameters to be broken by mistakes that are no easily seen (since they can be anywhere in the code base that uses this code). I understand that mutation is required by the `Audio` class, see my line comments below.

src/audio/audio.cpp
```diff
@@ -125,28 +158,78 @@ Set the master output volume.
...
+    return d->maxInputGain;
+}
+
+/**
+@brief Set the maximum allowed gain value in dB.
 
-    inGain = std::max(0.f, std::min(volume, 1.f));
+@note Default is 30dB; usually you don't need to alter this value.
+*/
+void Audio::setMaxInputGain(qreal dB)
```
initramfs, line 222: What is the advantage of creating a whole new class to encapsulate mutable parameters that that are essentially only changed by this class? I feel like that makes the code less readable by the constant access to object `d`. Isn't holding the values as member fields of this class good enough? Comments and variable isolation can help with establishing that the values +-30 dB are defaults. You can also declare the defaults as `constexpr` and completely make them immutable, allowing modification only via code changes.

If the separate class approach is necessary in your opinion, please move the `setX` `getX` functions that relate to gain to that class, or else the `Audio::Private` class is merely a glorified `struct` whose functionality merely adds an additional layer of indirection to otherwise things that could have been member fields of `Audio`.

Another potential idea to reduce the size of the `Audio` class would be to declare a `AudioGainDevice` (or a name of similar sort) that defines an abstract/virtual class that provides functions for gain control with suitable defaults. The `Audio` class can then appropriately derive from it as a subclass. Access control can be done via private/protected fields in the base class.

src/widget/form/settings/avform.cpp
```diff
@@ -65,6 +67,12 @@ AVForm::AVForm() :
     bodyUI->playbackSlider->installEventFilter(this);
     connect(bodyUI->playbackSlider, &QSlider::valueChanged,
             this, &AVForm::onPlaybackValueChanged);
+
+    bodyUI->microphoneSlider->setToolTip(
+                tr("Use slider to set the gain of your input device ranging"
+                   " from -30dB to +30dB."));
```
initramfs: These numbers should be retrieved from the defined constants currently in `Audio::Private`

initramfs:
> As the term "inVolume" is wrong in settings anyway, IMO it is best to introduce a new value "audio/inGain" and leave the old as a "dead" value.

I fully support this idea, in fact I was thinking about it since I first saw what this PR did since it radically changes the way input is handled (it is no longer a linear volume, but an actual digital amplifier now). Moreover, we can derive input gain from the existing input volume field via the equation `20 * log10( input_volume)` such that users who eventually run a version of qTox with this PR merged will not hear any difference initially (they'll only notice that the slider is now in a different position [much lower than where they initially placed it before this PR, allowing for a lot more extra headroom]).

src/audio/audio.cpp
antis81: Thanks, will correct that.

As we have a constant value, there's no reason. When factor is variable, this avoids "division by zero" error.

src/audio/audio.cpp
antis81, line 73: > Keep these fields public somewhat defeats the above function. …

Yep, we can make them immutable by adding one or two more wrapper functions to completely control this throughout qTox Audio API.

src/audio/audio.cpp
antis81, line 222: > What is the advantage of creating a whole new class to encapsulate mutable parameters that that are essentially only changed by this class?

It is more than just that. Actually, I prepare to encapsulate the qTox Audio API completely and bring back the pimpl idiom. This is absolutely necessary, to keep the API controllable, making it "future proof" (had a hard time btw. collecting all the code fragments and clean up the mess to get to the current stable state). Further it actually eases handling of audio data within qTox by providing a sane interface, which cannot be mistaken. It has some other advantages as well. For example it allows an atomic replacement of underlying `OpenAL` with another library (e.g. RtAudio).

> You can also declare the defaults as constexpr and completely make them immutable, allowing modification only via code changes.

This also was my first thought, but it is just not the right thing to do here. Those values are really meant as defaults and thus mutable through public audio api.

> Another potential idea to reduce the size of the Audio class would be to declare a `AudioGainDevice`…

Yes, why not. I'd simply name that class `qTox::Audio::InputDevice` (including the namespaces). I think about splitting up audio devices for some time now, being able to actually use them when required and replace the "crappy" singleton and subscription mechanisms. OTOH, this PR is actually ready, making qTox "audio conferencing" finally controllable to users, which probably helps to extend the community (user base). I don't want to break that by bringing in new concepts too quickly. Hope you understand. Anyway I'd love to discuss future of qTox Audio API with a "professional attitude". ^^

src/widget/form/settings/avform.cpp
antis81: Yep, thanks. Will change that.

antis81:
> I fully support this idea, in fact I was thinking about it since I first saw what this PR…

Ok, I'll do that as well then and rename the settings value.

antis81:
@initramfs The discussed changes are ready. Merge?

src/audio/audio.cpp
```diff
@@ -36,6 +37,47 @@
...
+    }
+
+    qreal inputGainFactor() const {
+        return gainFactor;
+    }
+
+    void setInputGain(qreal dB)
+    {
+        gain = qBound(minInputGain, dB, maxInputGain);
+        gainFactor = qPow(10.0, (gain / 20.0));
```
antis81, line 70: @initramfs Hint: I exchanged meaning to `gain` (in decibel) and `gainFactor`, so it is more clear.

src/audio/audio.cpp
```diff
@@ -36,6 +37,47 @@
...
+    }
+
+public:
+    qreal   minInputGain;
+    qreal   maxInputGain;
+
+private:
```
antis81, line 77: Look, a private private -> neat. :) Now values can only be changed through `setGain()`, making it "fool proof".

antis81:
@initramfs A small "post implementation UI change": I added tickmarks to the gain slider, improving usability. Unfortunately QSlider does not provide setting tickmarks on an array of certain positions, but on a fixed interval only. This requires implementing our own slider widget (started with implementation, but it will take some time).

initramfs:
@antis81 

> It is more than just that. Actually, I prepare to encapsulate the qTox Audio API completely and bring back the pimpl idiom.
> 
> I think about splitting up audio devices for some time now, being able to actually use them when required and replace the "crappy" singleton and subscription mechanisms.

Well okay, then I guess I can see where this is going. Forgive my earlier comments on this matter then since I didn't know exactly where this was leading to. Funnily enough, I've been doing the same thing for qTox's video code (replacing singleton patterns and allowing multiple object/devices). In particular, my changes plan to finally allow for proper desktop sharing control.

Regarding the settings change from `inVolume` to `inGain`, I noticed you haven't added any conversion code to transform old `inVolume` settings to `inGain`, is this intentional? I mean, it's not really that much of a big deal, I just want to clarify.

Apart from that there is just a few stylistic changes that is required (some opening brackets need to be on new lines). Also, can the private instance be named something more descriptive than `d`?

antis81:
> I noticed you haven't added any conversion code to transform old inVolume settings to inGain, is this intentional?

Yes. Like discussed, the `inGain` value is "reset" to unity gain (this matches the before 100% `inVolume`). This will be least painful for users and exceeds the range seamlessly. Converting the old linear values to negative exponential dB values from -30~0dB doesn't pay off, don't you think? Btw. apparently qTox drops the old `inVolume` settings key - that's a good thing actually.

> Also, can the private instance be named something more descriptive than d?

The `d` nomenclature is a quite common practice and therefore actually very descriptive. It is even well documented in Qt's internal and public docs and used consequently throughout the library, where I adopted some quite useful programming techniques from. :) For an example, have a look at `QExplicitlySharedDataPointer` and `QSharedData`. This is actually what I have in mind for the future to split audio devices. I started a "test implementation" locally to see how much effort this will become. In short, this is unfinished and it will take several days to complete.

antis81:
> Apart from that there is just a few stylistic changes that is required…

Whoopsy, will be corrected. I'd like to merge afterwards keeping focused on what's most relevant here. Could you review #3309 and #3175 as well? The latter is another candidate for moving the abstraction layer to the `d` pointer. And now it starts to create a more complete picture… :)

initramfs:
> Converting the old linear values to negative exponential dB values from -30~0dB doesn't pay off, don't you think?

Yep, no problem, just wanted to verify correct behaviour.

> The `d` nomenclature is a quite common practice and therefore actually very descriptive.

Okay, I didn't know that. I've seen it before somewhere else but didn't think too much of it. Thanks for the info.

> Could you review #3309 and #3175 as well?

Sure, I'll do it later today when I have time.

antis81:
This PR is ready to merge. @zetok Would you do it? I can also merge myself. As this includes some important improvements to the audio code, it would help merging soon.

antis81:
@zetok Bump.

zetok:
@antis81 Sorry for no earlier answer.

If PR is ready, LGTM :)

Issue #2726, opened 2015-12-22
Messages are sent to me twice on some occasions


ghost:
I have recently noticed (after one of the last updates) that when somebody sends me a message, about 2 or 3 seconds after they have sent it they send me an exact duplicate, however this does not happen with all messages they send me, nor do I see it in relation to any of my messages. The other strange thing is that the other person cannot see the duplicate message and only sees the original.

Here is an example of a duplicate message I see:

![duplicate_messages_qtox](https://cloud.githubusercontent.com/assets/11575271/11957831/f210ee18-a8bb-11e5-81df-43c175d8a304.png)

ghost:
Happens to me also, running latest git version

ProMcTagonist:
Same.

ElLamparto:
If the ACK is lost for any reason, the message is resent. Not much to do about it.

ghost:
If a message needs to be resent though can't it send some information with it that tells the qTox client that it has been resent, and then if the qTox client registers that there is another exact duplicate message already being displayed, and this is an automatic resend, and not what the user has requested, then just not to display the duplicate message?

agilob:
You should discuss this in toxcore repo.

SeventhSonOfASeventhSon:
@ElLamparto, Sorry, but what exactly is "ACK"?

ghost:
I have filed a bug report on this [here](https://github.com/irungentoo/toxcore/issues/1486).

LittleVulpix:
People keep saying this is a toxcore issue and it is an extension thereof, but I believe qtox is "worse than others" at sending double messages. All of my friends that have qtox occassionally send me duplicate messages. Whereas I (using miranda NG) and some of my uTox friends NEVER send me duplicate messages.

There must be some kind of timeout in qtox that tries to resend a message if an ack(?) is not received within a given timeframe. I would recommend extending this timeout to at least 15-20 seconds. It is somewhat annoying to keep getting duplicate messages from qTox users.

ghost:
Though it may be something else, this may be completely failing here https://github.com/tux3/qTox/issues/3022.

zetok:
> (…) I believe qtox is "worse than others" at sending double messages. (…)
> 
> There must be some kind of timeout in qtox that tries to resend a message if an ack(?) is not received within a given timeframe. I would recommend extending this timeout to at least 15-20 seconds. It is somewhat annoying to keep getting duplicate messages from qTox users.

[...]

PR #3316, opened 2016-05-17
fix: increase timer for checking offline messages timeout (again)

zetok:
Should make duplicated messages even less common (#2726)
Increased ~proportionally to change in
76d8e19320d2f15aeb019f72ecaaac2d6a23feab

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -120,7 +120,8 @@ void Widget::init()
     timer = new QTimer();
     timer->start(1000);
     offlineMsgTimer = new QTimer();
-    offlineMsgTimer->start(15000);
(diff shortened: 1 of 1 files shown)


src/widget/widget.cpp
```diff
@@ -120,7 +120,7 @@ void Widget::init()
     timer = new QTimer();
     timer->start(1000);
     offlineMsgTimer = new QTimer();
-    offlineMsgTimer->start(15000);
+    offlineMsgTimer->start(120000);
```
sudden6: would you mind changing this to the form `2*60*1000` ? makes it easier to read.

zetok: Done.

Issue #3161, opened 2016-04-19
Can't see white icon in tray (add contrasting (black?) line around white icon)


meamuri:
##### Brief Description

OS: Windows 
qTox version: v1.3.0.
Commit hash: b69042ec407c30b27ce3e56bb6daf14300facc8e  
toxcore:  0.0.0.
Qt: 5.5.1. 

Reproducible: Always 
##### Steps to reproduce
1. Open settings
2. Set light icon
3. Open "show hide icons" in tray. 
##### Observed Behavior

Can't see white icon on default white background color
##### Expected Behavior

Make black border
##### Additional Info

![tox](https://cloud.githubusercontent.com/assets/15677472/14636524/a7bed45c-0634-11e6-8f32-0b95eabb4575.jpg)

sudden6:
Isn't that exactly why we have a dark icon?

zetok:
> Isn't that exactly why we have a dark icon?

Kinda. On the other hand, the reasoning here is that someone wants to have _white_ icon, but still wants it to be visible on white background.
Same (in reverse) could be applied to dark icon.

ProMcTagonist:
The cloud and flag icons are white but can still be seen. I don't think it's unreasonable to expect the same of the white tox icon.

agilob:
![screenshot_20160426_213418](https://cloud.githubusercontent.com/assets/2544251/14833334/a6a9c65e-0bf6-11e6-8a3b-2a22f5558d5f.png)

Change it in settings to dark icon?

Diadlo:
@agilob I think both icons (light and dark) should be possible to showing on any background

sudden6:
It seems this is not possible to do without making the icon look bad, see: https://github.com/qTox/qTox/pull/3336

Issue #2671, opened 2015-12-13
Windows: Icons in file properties displayed incorrectly and too less sizes


rugk:
Just right-click on `qtox.exe` or `qtox-updater.exe` and choose `Properties`:
![propertiesicons](https://cloud.githubusercontent.com/assets/11966684/11767100/7ad32cbe-a1a2-11e5-8d53-1c2cbcbbceef.PNG)

This grey part does not look nice. :smile: 

v1.2.1

rugk:
I also see what causes this issue: You only include the 64x64 px icon.
![qtox icon](https://cloud.githubusercontent.com/assets/11966684/11767107/b2a85808-a1a2-11e5-9dc2-f42578efe756.PNG)

This is not good and I'd recommend you to include all usual icons sizes\* from 16x16px to 256x256px.

If you want you could also add icons for different colour depths (8bit color and so on). As the tox icon does not have so many colours there should be no problem. :wink: 

*which means: 16x16, 24x24, 32x32, 48x48, 128x128 and 256x256

NoxNoctium:
so yeah uh, it's a try.
![pic](https://i.imgur.com/1grAng3.png)
combined it with the ideas from https://github.com/tux3/qTox/issues/2246 and added a seperate updater icon.
the outline doesn't scale too 'pretty' though, opinions?

ico files here: https://docs.google.com/uc?id=0B4hJgknPmitzdDBPLVV1cUZQZHM

rugk:
Okay, the downloader icon looks nice. :smiley:

rugk:
@tux3 BTW: Is this really only Windows? I mean I only tested it on Windows, but how are the icons for other OSes added?

NoxNoctium:
OSX uses it's own proprietary format, linux usually supports svg, so yeah.
Usually the icon is replaced by one with a status indicator, that behavior is currently broken in windows.
Included icon sizes and color depths are the same as in the Windows 10 explorer.exe, would also need feedback there.

agilob:
> OSX uses it's own proprietary format, linux usually supports svg, so yeah.

I love that "usually" :D I spent 5 hours experimenting with SVGs as tray icon :D

PR #3336, opened 2016-05-26
feat(icons): Icons was updated

Diadlo:
Icon border was added


zetok:
#3161

Diadlo:
Fix #3161.

Diadlo:
Screenshots. Left is new, right -- old

spectacle bz6510
spectacle mn6723
spectacle dt6723
spectacle lh6723

ProMcTagonist:
@Diadlo The loop of the lock looks very skinny now, and the point of the speech bubble is distorted as well. Could the outline be thinner or further away?

Diadlo:
@ProMcTagonist It is the most tiny outline, which seems normal. I think, that point of the speech bubble is normal. But yes, loop looks skinny. May be I should make circle and body of lock smaller?

NoxNoctium:
Proposals here somewhat related: #2671

sudden6:
The outline definitely makes the icon look smaller, is there a way to put the border on the outsides of the current icon?

Diadlo:
@sudden6 I don't know way. Because size of old icon was equals to size of picture. I must make icon smaller to add border

sudden6:
ok, makes sense, I'm unsure if that's what everybody wants.
@zetok @tux3 opinions?

zetok:
new icon looks kinda weird: weird look
and the white one looks outright broken white broken

PR #3351, opened 2016-06-04
remove filter_audio

### LINKED ISSUE #3194: filter_audio is unmaintained
hi there, i am very new to this project, so understand my naivety.

filter_audio aka libfilteraudio seems to be unmaintained and grayhatter means, 
that it requires your audio card delay be EXACTLY the same as irungentoo's

Is it still the default and only one way for qtox?

antis81:
The filter_audio API is not working for qTox/Tox and thus disabled by default via compiler switch. In current qTox UI, it is not configurable, even when this switch is enabled. According to issue #3194 the library is no longer maintained as well. I don't see any reason to keep it in qTox.

For reference, visit the libs repository: https://github.com/irungentoo/filter_audio

Closes #3194


sudden6:
If there's nobody who is still using I'm fine with this.

zetok:
There are instructions in `INSTALL.md` that shouldbe removed, and parts of scripts that aim to provide dependencies for qTox: `bootstrap.sh` and `windows/bootstrap.sh`.

There might be something that I forgot about.

antis81:
@zetok I now also removed all references in the scripts. There's no leftovers in the repo, yet there might be some in the wiki pages or other docs in general.

PR #3353, opened 2016-06-04
[WIP] Initial multi-device support

tux3:
This adds multi-device options in the Advanced settings tab, requires my (or @grayhatter's) toxcore on the multi-device branch to compile.
Currently lacking a UI to see blacklisted devices, to add a real name to a device, and it's not translatable yet.

Comments welcome.

snap_43

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3353)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -792,6 +792,42 @@ QPair<QByteArray, QByteArray> Core::getKeypair() const
     return keypair;
 }
 
+QVector<ToxDevice> Core::getDeviceList() const
+{
+    QVector<ToxDevice> devs;
+
+    if (!tox)
+        return devs;
+
+    size_t count = tox_self_get_device_count(tox);
+    for (size_t i=0; i<count; ++i)
+    {
+        char name[TOX_MAX_NAME_LENGTH+1];
+        QByteArray pk(TOX_PUBLIC_KEY_SIZE, 0);
+        TOX_DEVICE_STATUS status;
+        if (!tox_self_get_device(tox, i, (uint8_t*)name, &status, (uint8_t*)pk.data(), nullptr))
+            break;
+
+        devs.append({CString::toString((uint8_t*)name, strlen(name)), pk, (DeviceStatus)status});
+    }
+
+    return devs;
+}
+
+bool Core::addDevice(QString name, QByteArray pk)
+{
+    if (pk.size() != TOX_PUBLIC_KEY_SIZE)
+        return false;
+
+    CString cName(name);
+    return (bool)tox_self_add_device(tox, (uint8_t*)cName.data(), cName.size(), (uint8_t*)pk.data(), nullptr);
+}
+
+bool Core::removeDevice(QByteArray pk)
+{
+    return (bool)tox_self_delete_device(tox, (uint8_t*)pk.data(), nullptr);
+}
+
 QString Core::getStatusMessage() const
 {
     QString sname;
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -81,7 +81,7 @@ public:
 
     QVector<ToxDevice> getDeviceList() const; ///< Return a list of names and public keys for our devices
     bool addDevice(QString name, QByteArray pk); ///< Add a device we want to sync with
-    bool removeDevice(QByteArray pk); ///< Permanently remove a device we're synced with, by index in the list
+    bool removeDevice(QByteArray pk); ///< Permanently remove and blacklist a device in the list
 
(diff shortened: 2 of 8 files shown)


agilob:
Does it mean I can use the same toxID on +1 devices and keep messages on them in sync as long as they're all online?

Is it similar to carbon sync (xmpp)?

> XEP-0280: Message Carbons which automatically syncs the messages you send to your desktop client and thus allows you to switch seamlessly from your mobile client to your desktop client and back within one conversation.

or do I have to have different tox IDs on each device and provide it to all my friends who will save my IDs as `agilob pc` and `agilob mobile`?

GrayHatter:
Speaking not for this PR but for Tox MultiDevice in it's current state.

> Does it mean I can use the same toxID on +1 devices and keep messages on them in sync as long as they're all online?

If they're all online, yes. If one is off line, message history syncing is planned, but not yet implemented.

> Is it similar to carbon sync (xmpp)?

Don't think so.

> XEP-0280: Message Carbons which automatically syncs the messages you send to your desktop client and thus allows you to switch seamlessly from your mobile client to your desktop client and back within one conversation.

If they're both online, then yes. That should be the experience.

> or do I have to have different tox IDs on each device and provide it to all my friends who will save my IDs as agilob pc and agilob mobile?

No, Toxcore will handle the syncing step for you. You'll just have to have all of your connected devices sync together at some point. Then you'll have to give your friend an ID that is currently online.

I like these questions, because now I have a solution for that problem. Thanks @agilob

src/core/core.cpp
```diff
@@ -792,6 +792,42 @@ QPair<QByteArray, QByteArray> Core::getKeypair() const
...
+    if (pk.size() != TOX_PUBLIC_KEY_SIZE)
+        return false;
+
+    CString cName(name);
+    return (bool)tox_self_add_device(tox, (uint8_t*)cName.data(), cName.size(), (uint8_t*)pk.data(), nullptr);
+}
+
+bool Core::removeDevice(QByteArray pk)
+{
+    return (bool)tox_self_delete_device(tox, (uint8_t*)pk.data(), nullptr);
```
sudden6, line 828: Add some check, that `pk` is a valid PublicKey?

src/core/core.h
```diff
@@ -79,6 +79,10 @@ class Core : public QObject
     ToxId getSelfId() const; ///< Returns our Tox ID
     QPair<QByteArray, QByteArray> getKeypair() const; ///< Returns our public and private keys
 
+    QVector<ToxDevice> getDeviceList() const; ///< Return a list of names and public keys for our devices
+    bool addDevice(QString name, QByteArray pk); ///< Add a device we want to sync with
+    bool removeDevice(QByteArray pk); ///< Permanently remove a device we're synced with, by index in the list
```
sudden6: Comment says by index, but it seems to be by public key.

agilob:
Nice, that's a small game changer :)

src/core/core.cpp
tux3, line 828: The key can only come from the list of devices, so there's not really any way it could be invalid. Since this API will probably be used in exactly one place, I think we can assume this one precondition is going to hold up.

src/core/core.h
tux3: Thank you, my mistake.

src/core/core.cpp
```diff
@@ -792,6 +792,42 @@ QPair<QByteArray, QByteArray> Core::getKeypair() const
...
+    if (!tox)
+        return devs;
+
+    size_t count = tox_self_get_device_count(tox);
+    for (size_t i=0; i<count; ++i)
+    {
+        char name[TOX_MAX_NAME_LENGTH+1];
+        QByteArray pk(TOX_PUBLIC_KEY_SIZE, 0);
+        TOX_DEVICE_STATUS status;
+        if (!tox_self_get_device(tox, i, (uint8_t*)name, &status, (uint8_t*)pk.data(), nullptr))
```
Diadlo, line 808: What about using static_cast, instead C-style cast?

agilob:
Is that PR dead?

GrayHatter:
this one, or mdevice?

agilob:
The functionality in qTox that would provide multi-device support (both core and client).

GrayHatter:
yeah, it's probably close to dead

The mdevice branch in toxcore a few months away at the earliest. 

And I'm pretty sure I'm going to end up changing the API a bit too.

GrayHatter:
I'm going to need to rewrite the api a bit (a lot?) you can safely close this

Issue #1701, opened 2015-05-21
[Feature request] way to paste text as a quote


zetok:
There should be a way (and preferably a shortcut) to paste text as a quote in input field.

I.e. `>` ("arrow" followed by a space) should be put before every line of pasted text when option is used.
- Option should appear in **context menu when r-clicked** in input field:
  - `Paste quote`, located below option `Paste`
- Additionally there should be **shortcut**:
  - `CTRL + SHIFT + V`
# 

Potential problems:
- there is possibility of problem(s) of it not working too well when the will be a need to use text splitting with larger texts - i.e. there's a chance that `>` will be left at the end of one message, while text that was supposed to be treated as quote will be in "next" message, which will cause visual inconsistency.
  - to avoid it, perhaps splitting mechanism should be somehow made aware that `> text.. rawr` shouldn't be split, and if it has to be, then text that was split into next message should be again prepended with `>` before splitting it again? or somehow make splitting mechanism to be more generic, so that it could consider different kind of values when splitting messages? (dunno really)
# 

Probably related issue: #784

ProMcTagonist:
+1

ghost:
+1

SkyzohKey:
+1

sudden6:
Implemented in recent builds.

PR #3360, opened 2016-06-06
feat(genericchatform): add "Quote selected text" feature to chat window

fazotron:
This commit adds feature to quote selected text in chat window.
How-To:
1. Select text in chat window
2. Press RMB and click "Quote selected text" OR press "ALT+Q"
3. Selected text will be automatically inserted in input text area

Animated screenshot instead of a thousand words:

quote

### DIFF
--- a/doc/user_manual_en.md
+++ b/doc/user_manual_en.md
@@ -6,6 +6,7 @@
 * [Settings](#settings)
 * [Groupchats](#groupchats)
 * [Message Styling](#message-styling)
+* [Quotes](#quotes)
 * [Multi Window Mode](#multi-window-mode)
 * [Keyboard Shortcuts](#keyboard-shortcuts)
 
@@ -239,6 +240,14 @@ Additionally, qTox supports three modes of Markdown parsing:
 
 *Note that any change in Markdown preference will require a restart.*
 
+## Quotes
+
+qTox has feature to quote selected text in chat window:
+
+1. Select the text you want to quote.
+2. Right-click on the selected text and choose "Quote selected text" in the context menu. Also you can use `ALT` + `q` shortcut.
+3. Selected text will be automatically quoted into a message input area in a pretty formatting.
+
 ## Multi Window Mode
 
 In this mode, qTox will separate its main window into a single contact list and
@@ -260,3 +269,4 @@ The following shortcuts are currently supported:
 | `CTRL` + `Page Up` | Switch to the previous contact|
 | `CTRL` + `TAB` | Switch to the next contact |
 | `CTRL` + `SHIFT` + `TAB` | Switch to the previous contact|
+| `ALT` + `q` | Quote selected text |
--- a/doc/user_manual_en.md
+++ b/doc/user_manual_en.md
@@ -245,8 +245,10 @@ Additionally, qTox supports three modes of Markdown parsing:
 qTox has feature to quote selected text in chat window:
 
 1. Select the text you want to quote.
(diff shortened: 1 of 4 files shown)


sudden6:
Please add some documentation to https://github.com/tux3/qTox/blob/master/doc/user_manual_en.md

fazotron:
@sudden6 Done.

zetok:
Related #1701

src/widget/form/genericchatform.cpp
```diff
@@ -199,6 +203,7 @@ GenericChatForm::GenericChatForm(QWidget *parent)
             this, &GenericChatForm::onChatContextMenuRequested);
 
     new QShortcut(Qt::CTRL + Qt::SHIFT + Qt::Key_L, this, SLOT(clearChatArea()));
+    new QShortcut(Qt::ALT + Qt::Key_Q, this, SLOT(quoteSelectedText()));
```
zetok, line 206: Wouldn't `CTRL` + `SHIFT` + `v` be better?

fazotron, line 206: @zetok  `CTRL` + `SHIFT` + `v` is a common shortcut to paste unformatted text in different software. I think it is a bad way to use it for another purposes.

SkyzohKey, line 206: Maybe the system should works differently. Example : copy a quote with ctrl+shift+c or r-click, this would simply add the quote in the clipboard. Letting the user do a simple ctrl+v, this way feels more intuitive/natural to me.

fazotron:
@SkyzohKey Indeed. What about "Right-click selected text and choose "Quote selected text" in a context menu" ?

src/widget/form/genericchatform.cpp
fazotron, line 206: @SkyzohKey As for me, the way I implemented it is more intuitive. I used it in other old instant messengers.
Anyway, your suggestion may be implemented as radio buttons in qTox settings: "Insert quotation to message input area" (by default) or "Copy quotation to clipboard". On the other hand, it makes current simple implementation more complicated.

zetok, line 206: @fazotron

> `CTRL` + `SHIFT` + `v` is a common shortcut to paste unformatted text in different software. I think it is a bad way to use it for another purposes.

 I see. On the other hand, using `ALT` + `q` to paste quote feels a bit weird to me.

Good to know that some older IMs used that :)

And yeah, in future it would be nice to have keyboard shortcuts configurable in settings.

fazotron:
Previous commit was amended to make documentation more clear.

doc/user_manual_en.md
```diff
@@ -239,6 +240,14 @@ Additionally, qTox supports three modes of Markdown parsing:
 
 *Note that any change in Markdown preference will require a restart.*
 
+## Quotes
+
+qTox has feature to quote selected text in chat window:
+
+1. Select the text you want to quote.
+2. Right-click on the selected text and choose "Quote selected text" in the context menu. Also you can use `ALT` + `q` shortcut.
```
sudden6: I think "You also can use the `ALT` + `q` shortcut." sounds better, I'm no native speaker tough. Also please shorten to 80 chars per line.

doc/user_manual_en.md
```diff
@@ -239,6 +240,14 @@ Additionally, qTox supports three modes of Markdown parsing:
...
+3. Selected text will be automatically quoted into a message input area in a pretty formatting.
```
sudden6: "Selected text will be automatically quoted into the message input area in a pretty formatting." ?

fazotron:
Fix according to @sudden6 comments

Issue #1294, opened 2015-02-27
Redesign emoticon popup for Stickers and no more pagination


ProMcTagonist:
Mockup based on butchered version of the one by @krepa098 in #1027:

![stickar](https://cloud.githubusercontent.com/assets/4358079/6890403/5920c69e-d65f-11e4-88f2-694ea894806f.png)
- Stickers supersede emoticons (those benefits can be found in #402)
- No more pagination issues
- Nicer-looking

This is conceptual as ittner's work isn't merged in core yet https://github.com/irungentoo/toxcore/pull/1148 but the redesign needs its own issue for discussion.

Edit: updated mockup to properly show size of stickers.
Edit: the new API enables stickers which can be discussed [here](https://wiki.tox.chat/developers/feature_ideas/stickers), this issue remains for qTox's popup

redsteakraw:
How will this handle selecting variations such as the proposed skin tone modifier for the smilies  / people emoji?

ProMcTagonist:
I don't see how that would be different than selecting any other emoji from their page when you've selected the emoticon entry in the sidebar.

If it's a UX problem you're worried about, 
1. Emoticons are going to be used actively (clicking and sending) much less than stickers
2. When skin tones are actually added by the UC we can see how others implement it
3. The grouping proposal you've talked about before will probably aid with this too

We've got nothing to worry about concerning this.

All-in-all, my experience is that people use emoji where they **can't** use stickers. When they can use stickers, they prefer the freedom of expression they offer. The advantage of the tox networks's incompatibility with other networks is that we don't need to worry about supporting users of ~~inferior~~ other systems.

redsteakraw:
1, I only see stickers as an alternative input for emoji but only limited to emoji.(one I am currently neither for or against)
2. qTox has lead the way of emoji support being the first project to implement a color middle finger emoji.  As such we shouldn't just wait for others and play catchup being ahead of the curve is where we should remain, let Skype play catchup.
3.yes indeed the standard groupings will compliment the proposed ui.

As for the stickers, emoji are being expanded and by staying on top of the proposals and following unicode, stickers are a proposal for future emoji work I would be weary of implementing a possible incompatible standard and making things difficult later down the road.  Furthermore the advantage or emoji are that they have codepoints they are encoded and it makes preserving the content and archiving so much easier.  Looking back on old reddit posts and you see broken links after broken links and history is lost.  I would be cautious to learn from the mistakes of current systems.  
@ProMcTagonist

[...]

Issue #992, opened 2015-01-07
Emotion input issues


redsteakraw:
Emotion input with large sets such as the GoogleEmojiComplete shows only 15 at a time and the popup is large compared to the emotions displayed.  It is hard to navigate find the emotions with so many pages.

zetok:
There is a proposal improving it: https://wiki.tox.chat/developers/feature_ideas/stickers

ProMcTagonist:
https://github.com/tux3/qTox/issues/1027

https://github.com/tux3/qTox/issues/1294

Related. I think it's getting to be about time for stickers and that redesign, since FILE_KIND is done (I think? see https://github.com/tux3/qTox/issues/1804  and  https://github.com/irungentoo/toxcore/pull/1148)

zetok:
> (…) since FILE_KIND is done (I think? (…)

Well, using `FILE_KIND` constitutes adding to toxcore `TOX_FILE_KIND_NEWTYPE` - adding new types wouldn't result in API breakage, but clients would have to be updated in order to support new stuff.

ProMcTagonist:
> clients would have to be updated in order to support new stuff.

That would have to happen regardless. I think a bigger concern is requiring a core change for every new filetype. I thought the whole point was to avoid that, so clients could rapidly come up with and agree on new filetypes themselves.

jessica181920:
This is no longer an issue.

PR #3364, opened 2016-06-07
feat(smileys): add emojione emoji-pack and make it the default

### LINKED ISSUE #3315: [change icon suggestion] Turn :) from the red cheeks into a :grinning: closed mouth smile
Let's face it, the current : ) or : - ) smiley icon in qTox looks almost embarrassed with those red cheeks and closed eyes. 

Even the ^ _ ^ icon has that 'joyful in frosty weather' type look with all that red.

I'd prefer to see the : ) smiley be made into a "more standard" looking smileyface _without_ the red cheeks. Something like the :grinning: icon for example, but with a closed mouth smile. 

Thank you.

+1 if you concur. :beer:

sudden6:
Thanks to @ProMcTagonist who suggested to use emojis from emojione.com

Fixes #3315


zetok:
- [x] not available not available

sudden6:
@zetok should be fixed

zetok:
- [x] `®` and `©` don't work, even though they're "shown" in the list – i.e. instead of signs, there are 2 empty places in the list
  muh ©®

sudden6:
Also fixed, please note, that not every emoji in the pack is currently available in the emoji picker, because I didn't yet find a way to convert an emoji.json file to an emoticon.xml file.

zetok:
```
zetok@galaxor ~/git_zetok/tox/qTox/smileys $ ls Universe/ | wc -l
877
zetok@galaxor ~/git_zetok/tox/qTox/smileys $ ls emojione/ | wc -l
1836
```

IMO it would be good to add all of them.

sudden6:
I don't know if it would be a good idea, because our emoji picker isn't exactly useful when choosing between that many emoji and these are probably the most used ones.

zetok:
> I don't know if it would be a good idea, because our emoji picker isn't exactly useful when choosing between that many emoji.

Yep, it needs to be fixed :)
Issue:  #992
Proposal: #1294

sudden6:
So, display all when the new picker is done?

zetok:
I'd prefer to display all, and if that breaks current emoticon picker, fix the picker.

sudden6:
I'd prefer to use the current state of this PR, then fix the emoji picker (won't probably happen very soon), then upgrade to the full set.

zetok:
About using current set – I've been wondering whether including that in qTox repo is really the best choice.

I.e. given that emojine is being actively developed, and what's more, it's using git, I was wondering whether using it as a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) wouldn't be better.

sudden6:
I also thought about this, but I don't know if this is a good idea here. I'm not sure if our user will appreciate if the emojis change between releases.

Also there's the problem of generating an appropriate emoticons.xml but I think we could change the qTox code to  support the emoji.json format.

Issue #3144, opened 2016-04-14
"*** Error in `qtox': free(): invalid pointer: 0x000000000245b040 ***" on normal exit


ghost:
Since the most recent update of qTox (`1.3.0-177-g1d6f32c-1`) I have noticed that this is the output in Terminal:

> [[19:57:00.168] src/persistence/settings.cpp:121 : Debug: "Loading settings from /home/boghog/.config/tox/qtox.ini"
> [19:57:00.169] src/ipc.cpp:46 : Debug: Our global IPC ID is  3993560339317860218
> [19:57:00.169] src/ipc.cpp:64 : Debug: Attaching to the global shared memory
> [19:57:00.169] src/ipc.cpp:259 : Debug: Previous owner timed out, taking ownership 14014811563473618655 -> 3993560339317860218
> [19:57:00.169] src/main.cpp:196 : Debug: built on:  21:43:31 Apr 13 2016 ( 1460583550 )
> [19:57:00.169] src/main.cpp:197 : Debug: commit:  built without git 
> 
> [19:57:00.170] src/nexus.cpp:81 : Debug: Starting up
> [19:57:07.358] src/persistence/profile.cpp:86 : Debug: Loading tox save  "/home/boghog/.config/tox/遊牧民.tox"
> [19:57:07.581] src/persistence/settings.cpp:371 : Debug: "Saving global settings at /home/boghog/.config/tox/qtox.ini"
> [19:57:07.691] src/persistence/settings.cpp:296 : Debug: Loading personal settings from "/home/boghog/.config/tox/遊牧民.ini"
> [19:57:08.814] src/core/core.cpp:236 : Debug: Loading user profile
> [19:57:08.814] src/persistence/profile.cpp:249 : Debug: Loading tox save  "/home/boghog/.config/tox/遊牧民.tox"
> [19:57:08.815] src/widget/systemtrayicon.cpp:65 : Debug: Using GTK backend
> 
> (qtox:5982): GLib-GObject-CRITICAL *_: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
> [19:57:08.951] src/core/core.cpp:132 : Debug: Core starting with IPv6 enabled
> [19:57:12.208] src/core/core.cpp:407 : Debug: "Connecting to 178.62.250.138:33445 (Impyy)"
> [19:57:12.209] src/core/core.cpp:407 : Debug: "Connecting to 130.133.110.14:33445 (Manolis)"
[...]

PR #3365, opened 2016-06-08
fix(widget): memory issues

### LINKED ISSUE #2308: Memory leak after closing main window

PafXUS:
Small memory leaks and uninitialized memory usage fixes. Should fix #2308 at least for disabled systray unity backend. Shouldn't create #3144 problem as cbb7eec.

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -93,7 +93,7 @@ ProfileForm::ProfileForm(QWidget *parent) :
     bodyUI->qrLabel->setWordWrap(true);
 
     QRegExp re("[^@ ]+");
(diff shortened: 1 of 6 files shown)


sudden6:
It seems there got a merge commit in there by accident.

PafXUS:
Merge commit removed. Now only fix commits.

PR #3371, opened 2016-06-09
fix(video): fix video resolution setting

### LINKED ISSUE #1033: Video Resolution setting not retained
I'm setting Video Resolution to 160x120, but if I move out of the Audio/Video tab and back, the resolution is back to the highest. This is as of commit aef083934f8a9366a855ec474bd6dec7f3f88867

b4n:
Index the filtered modes we show to the user when selecting from the
combobox, not the full list of modes actually supported by the device.

Fixes #1033.

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -162,8 +162,8 @@ void AVForm::updateVideoModes(int curIndex)
         return;
     }
     QString devName = videoDeviceList[curIndex].first;
-    videoModes = CameraDevice::getVideoModes(devName);
-    std::sort(videoModes.begin(), videoModes.end(),
+    QVector<VideoMode> allVideoModes = CameraDevice::getVideoModes(devName);
+    std::sort(allVideoModes.begin(), allVideoModes.end(),
         [](const VideoMode& a, const VideoMode& b)
             {return a.width!=b.width ? a.width>b.width :
                     a.height!=b.height ? a.height>b.height :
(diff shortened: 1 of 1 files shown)


src/widget/form/settings/avform.cpp
```diff
@@ -162,8 +162,8 @@ void AVForm::updateVideoModes(int curIndex)
         return;
     }
     QString devName = videoDeviceList[curIndex].first;
-    videoModes = CameraDevice::getVideoModes(devName);
-    std::sort(videoModes.begin(), videoModes.end(),
+    QVector<VideoMode> allVideoModes = CameraDevice::getVideoModes(devName);
```
b4n, line 165: A large part of the diff relates to renaming this variable.

src/widget/form/settings/avform.cpp
```diff
@@ -227,12 +227,14 @@ void AVForm::updateVideoModes(int curIndex)
     QSize prefRes = Settings::getInstance().getCamVideoRes();
     unsigned short prefFPS = Settings::getInstance().getCamVideoFPS();
     // Iterate backwards to show higest resolution first.
```
b4n, line 229: The actually interesting changes start here

src/widget/form/settings/avform.cpp
```diff
@@ -227,12 +227,14 @@ void AVForm::updateVideoModes(int curIndex)
...
+    videoModes.clear();
     for(auto iter = bestModeInds.rbegin(); iter != bestModeInds.rend(); ++iter)
     {
         int i = iter->second;
-        VideoMode mode = videoModes[i];
+        VideoMode mode = allVideoModes[i];
+        videoModes.push_back(mode);
```
sudden6: Since we use qt it would be better to use `append(mode);` here.

b4n: I wondered for a second, grepped the source and found it was more or less random apparently, so I kept what I had, but I'll change it.

``` console
$ git grep -F .append | wc -l
42
$ git grep -F .push_back | wc -l
33
```

b4n:
@sudden6 I updated the patch.

PR #3376, opened 2016-06-11
fix(video): do not list the same mode twice

b4n:
Do not provide more than one entry in the settings for the same actual
mode, even if it was selected for different expected resolutions.

For example, do not list 640x480 twice for both 480p and 360p if the
device doesn't have a better 360p mode, and simply skip the 360p entry.

---

Another solution could be to select the second-best match if the first-best is already used.  For example, my webcam has 640x480 and 352x288, so no real 360p mode, but 352x288 still is reasonably close (although it's beyond the hard-coded threshold of 300, as `640-352+360-288=360`), and slightly better than 240p.
This would however be a little more work, and I'm not certain it's wanted.

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -232,6 +232,10 @@ void AVForm::updateVideoModes(int curIndex)
     {
         int i = iter->second;
         VideoMode mode = allVideoModes[i];
+
+        if (videoModes.contains(mode))
+            continue;
+
         videoModes.append(mode);
         if (mode.width==prefRes.width() && mode.height==prefRes.height() && mode.FPS == prefFPS && prefResIndex==-1)
             prefResIndex = videoModes.size() - 1;


b4n:
(re-pushed in the hope to trigger Travis build that seem to have been missing for some reason)

src/widget/form/settings/avform.cpp
```diff
@@ -228,9 +228,15 @@ void AVForm::updateVideoModes(int curIndex)
     unsigned short prefFPS = Settings::getInstance().getCamVideoFPS();
     // Iterate backwards to show higest resolution first.
     videoModes.clear();
+    QVector<int> usedInds;
     for(auto iter = bestModeInds.rbegin(); iter != bestModeInds.rend(); ++iter)
     {
         int i = iter->second;
+        // don't use the same mode twice
+        if (usedInds.contains(i))
```
sudden6: uhm, why introduce a new variable here and not just do `videoMode.contains(mode)` ?

b4n: Hum, good remark,I don't know why I didn't do that.  I'll fix that.

b4n:
@sudden6 fixed, thanks

src/widget/form/settings/avform.cpp
```diff
@@ -232,6 +232,10 @@ void AVForm::updateVideoModes(int curIndex)
     {
         int i = iter->second;
         VideoMode mode = allVideoModes[i];
+        // don't use the same mode twice
```
antis81: Just a small style fix. It's totally clear what the code does. No comment needed here. Can you just replace with empty line?  Thanks.

b4n: Sure, done.

PR #3380, opened 2016-06-12
fix serious type conversion warnings

antis81:
The used var types in the qTox Settings class require some cleanup. For some weeks (months?) now, LLVM's static code analysis is available in Qt Creator and helps a bit. It moans about lots of missing type casts, which is also an indicator, that those types do not match well and need to be changed.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3380)

### DIFF
--- a/src/core/indexedlist.h
+++ b/src/core/indexedlist.h
@@ -11,24 +11,37 @@ public:
     explicit IndexedList() = default;
 
     // Qt
-    bool isEmpty()
+    inline bool isEmpty()
     {
         return v.empty();
     }
 
-    bool contains(int i)
+    template <typename cmp_type>
+    bool contains(cmp_type i)
     {
-        return std::find_if(begin(), end(), [i](T& t){return (int)t == i;}) != end();
+        return std::find_if(begin(), end(), [i](T& t)
+        {
+            return static_cast<cmp_type>(t) == i;
+        }) != end();
     }
 
-    void remove(int i)
+    template <typename cmp_type>
+    void remove(cmp_type i)
     {
-        v.erase(std::remove_if(begin(), end(), [i](T& t){return (int)t == i;}), end());
+        v.erase(std::remove_if(begin(), end(), [i](T& t)
+        {
+            return static_cast<cmp_type>(t) == i;
+        }), end());
     }
 
-    T &operator[](int i)
+    template <typename cmp_type>
+    T& operator[](cmp_type i)
     {
-        iterator it = std::find_if(begin(), end(), [i](T& t){return (int)t == i;});
+        iterator it = std::find_if(begin(), end(), [i](T& t)
+        {
+            return static_cast<cmp_type>(t) == i;
+        });
+
         if (it == end())
             it = insert({});
 
@@ -44,34 +57,32 @@ public:
     {
         return v.begin();
     }
-    inline const_iterator begin() const
-    {
-        return v.begin();
-    }
+
     inline const_iterator cbegin() const
     {
         return v.cbegin();
     }
+
     inline iterator end()
     {
         return v.end();
     }
-    inline const_iterator end() const
-    {
-        return v.end();
-    }
+
     inline const_iterator cend() const
     {
         return v.cend();
     }
+
     inline iterator erase(iterator pos)
     {
         return v.erase(pos);
     }
+
     inline iterator erase(iterator first, iterator last)
     {
         return v.erase(first, last);
     }
+
     inline iterator insert(T&& value)
     {
         v.push_back(std::move(value));
--- a/src/persistence/serialize.cpp
+++ b/src/persistence/serialize.cpp
@@ -67,18 +67,19 @@ float dataToFloat(QByteArray data)
 }
 
 // Converts a string into PNet string data
-QByteArray stringToData(QString str)
+QByteArray stringToData(const QString& str)
 {
     QByteArray data(4,0);
     // Write the size in a Uint of variable lenght (8-32 bits)
     int i=0;
-    uint num1 = (uint)str.toUtf8().size();
+    int num1 = str.toUtf8().size();
     while (num1 >= 0x80)
     {
-        data[i] = (unsigned char)(num1 | 0x80); i++;
+        data[i] = static_cast<char>(num1 | 0x80);
+        i++;
         num1 = num1 >> 7;
     }
-    data[i]=num1;
+    data[i] = static_cast<char>(num1);
     data.resize(i+1);
     data+=str.toUtf8();
     return data;
@@ -86,24 +87,24 @@ QByteArray stringToData(QString str)
 
 QString dataToString(QByteArray data)
 {
-    // Variable UInt32
-    unsigned char num3;
-    int num = 0;
+    char num3;
+    int strlen = 0;
     int num2 = 0;
     int i=0;
     do
     {
-        num3 = data[i]; i++;
-        num |= (num3 & 0x7f) << num2;
+        num3 = data[i++];
+        strlen |= (num3 & 0x7f) << num2;
         num2 += 7;
     } while ((num3 & 0x80) != 0);
-    unsigned int strlen = (uint) num;
 
-    if (!strlen)
+    if (strlen <= 0)
         return QString();
 
-    data = data.right(data.size()-i); // Remove the strlen
+    // Remove the strlen
+    data.remove(0, i - 1);
     data.truncate(strlen);
+
     return QString(data);
 }
 
@@ -113,35 +114,35 @@ float dataToRangedSingle(float min, float max, int numberOfBits, QByteArray data
     uint value=0;
     if (numberOfBits <= 8)
     {
-        endvalue = (uchar)data[0];
+        endvalue = static_cast<uchar>(data[0]);
         goto done;
     }
-    value = (uchar)data[0];
+    value = static_cast<uchar>(data[0]);
     numberOfBits -= 8;
     if (numberOfBits <= 8)
     {
-        endvalue = (value | ((uint) ((uchar)data[1]) << 8));
+        endvalue = value | (static_cast<uint>(data[1]) << 8);
         goto done;
     }
-    value |= (uint) (((uchar)data[1]) << 8);
+    value |= static_cast<uint>(data[1]) << 8;
     numberOfBits -= 8;
     if (numberOfBits <= 8)
     {
-        uint num2 = (uint) (((uchar)data[2]) << 0x10);
+        uint num2 = static_cast<uint>(data[2]) << 0x10;
         endvalue = (value | num2);
         goto done;
     }
-    value |= (uint) (((uchar)data[2]) << 0x10);
+    value |= static_cast<uint>(data[2]) << 0x10;
     numberOfBits -= 8;
-    endvalue =  (value | ((uint) (((uchar)data[3]) << 0x18)));
+    endvalue = value | (static_cast<uint>(data[3]) << 0x18);
     goto done;
 
     done:
 
     float num = max - min;
-    int num2 = (((int) 1) << numberOfBits) - 1;
+    int num2 = (static_cast<int>(1) << numberOfBits) - 1;
     float num3 = endvalue;
-    float num4 = num3 / ((float) num2);
+    float num4 = num3 / num2;
     return (min + (num4 * num));
 }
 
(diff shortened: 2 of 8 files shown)


antis81:
@tux3 Here's another chance to get rid of some dead code :smile:. I don't wanna simply merge without review. Anyone with time slot available for review? Though not directly related, this PR is mandatory for further improvements on the audio api as well.

Diadlo:
src/nexus.cpp:195 (code format)
src/core/corefile.cpp (Filename -> filename)
src/core/corefile.h (Filename -> filename)
ctrl + shift + f => `//qDebug()` => delete them all :)

antis81:
@Diadlo Actually most of the stuff you state is unrelated and has to be done separately. Anyway, did some minor further corrections and one major in IndexList :smile:. **This requires review to avoid unexpected crashes!**

Diadlo:
@antis81 Deleted. Sorry. I misunderstand you)

src/core/indexedlist.h
```diff
@@ -13,30 +13,91 @@ class IndexedList
     explicit IndexedList() = default;
```
sudden6, line 11: Why do we have our own IndexedList and don't use QVector or QList?

antis81, line 11: Thanks for taking a look inside. No idea why this exists. I just didn't want to change "too much". Though it looks like this IndexedList is used solely for management of ToxCall objects. Btw.: a QMap/QHash would be the appropriate thing to replace it - it is actually about key-value mapping.

src/persistence/serialize.cpp
```diff
@@ -62,27 +62,27 @@ float dataToFloat(QByteArray data)
...
 
 // Converts a string into PNet string data
-QByteArray stringToData(QString str)
+QByteArray stringToData(const QString& str)
 {
     QByteArray data(4,0);
     // Write the size in a Uint of variable lenght (8-32 bits)
     int i=0;
-    uint num1 = (uint)str.toUtf8().size();
+    uint num1 = static_cast<uint>(str.toUtf8().size());
```
sudden6: num1 is always used as `char` why not make it one?

src/persistence/serialize.cpp
```diff
@@ -62,27 +62,27 @@ float dataToFloat(QByteArray data)
...
 // Converts a string into PNet string data
-QByteArray stringToData(QString str)
+QByteArray stringToData(const QString& str)
 {
     QByteArray data(4,0);
     // Write the size in a Uint of variable lenght (8-32 bits)
     int i=0;
-    uint num1 = (uint)str.toUtf8().size();
+    int num1 = str.toUtf8().size();
     while (num1 >= 0x80)
```
antis81: @sudden6 Cast to char is not possible because of this line (biggest positive value is 0x7f == 127!). Would need to be uchar, ergo, nothing won here. However no cast at all is valid :smile:.

antis81:
@tux3 Most functions in `serialize.h/cpp` seem to be dead code. Is it safe to remove them?

src/persistence/serialize.cpp
sudden6: Ah, didn't think of that.

antis81: Fortunately the static analyzer warns in this case.

Diadlo:
What about to make methods that do not change the object, constant?

antis81:
@sudden6 @zetok We could replace the IndexedList with QMap or QHash and serialization functions with QVariant. But - I have very limited time for qTox and it's fine, if we do that later in the process. If one of you wants to adopt this PR, feel free to cherry-pick and close it. Other than that -> it's reviewed and ready to merge.

antis81:
@tux3 I removed all unused conversion methods from `persistence/serialize.h / .cpp`. It would probably be the best to convert and use QVariant compliant types. In this case, our own conversion methods might simply "fade away" - makeing life "easier" for contributors. :)

src/persistence/serialize.h
```diff
@@ -28,23 +28,12 @@
...
-QByteArray stringToData(QString str);
 QString dataToString(QByteArray data);
-float dataToRangedSingle(float min, float max, int numberOfBits, QByteArray data);
-QByteArray rangedSingleToData(float value, float min, float max, int numberOfBits);
-uint8_t dataToUint8(QByteArray data);
-uint16_t dataToUint16(QByteArray data);
-uint32_t dataToUint32(QByteArray data);
-uint64_t dataToUint64(QByteArray data);
+uint64_t dataToUint64(const QByteArray& data);
+int dataToVInt(const QByteArray& data);
```
tux3, line 30: This should say "vuint" instead of "vint", the encoding assumes unsignedness.

src/core/indexedlist.h
tux3, line 11: IndexedList could be replaced by a map or a hash table, the only difference is the performance and memory usage, for our usage it easily beats the hash table or map on both points.
That said performance isn't really crucial here, so I'm fine with removing it if people find it too complicated to deal with.

tux3:
The stuff in serialize.(h|cpp) was shamelessly copypasted from a side project (hence the unused functions), and the implementation is absolutely awful, only very efficient in terms of density of the serialized format.
It's fine to replace all of this with Qt's built-in serialization. The current format is several times smaller than what Qt does if I remember correctly, but it shouldn't really matter since the settings save file isn't very big. And the crypto we use on it is decent enough to not care if half the file is known plaintext.

src/persistence/settingsserializer.cpp
```diff
@@ -510,38 +518,37 @@ void SettingsSerializer::readIni()
...
-                //qDebug() << "Found key"<<v.key<<"at index"<<groupArrayIndex<<"of array"<<a.name;
+                v.arrayIndex = static_cast<int>(groupArrayIndex);
+                a.values.append(static_cast<quint64>(vi));
             }
         }
     }
 
     // Clean up spurious array element groups
-    sort(begin(groupsToKill), end(groupsToKill), std::greater_equal<int>());
+    sort(std::begin(groupsToKill), std::end(groupsToKill),
```
tux3: What sort is this calling if not std::sort :) ?
I'm not super fond of gratuitous nested-name specifiers when there's no ambiguity, but if you're going to add them, why leave an unqualified name in the middle of it!?

src/core/indexedlist.h
sudden6, line 11: I still don't understand why `IndexedList` isn't equal to `QVector`, as far as I can tell `IndexedList` is basically a wrapper around `std::vector`, but `QVector` and `std::vector` are both a dynamically resized array.

I would be happy if somebody could explain this to me :)

src/persistence/settingsserializer.cpp
```diff
@@ -180,9 +188,12 @@ const SettingsSerializer::Value* SettingsSerializer::findValue(const QString& ke
...
                 continue;
-            for (int vi : a.values)
-                if (values[vi].arrayIndex == arrayIndex && values[vi].key == key)
-                    return &values[vi];
+            for (int vi = 0; vi < a.values.size(); vi++)
+            {
+                const Value& v = values[vi];
+                if (v.arrayIndex == arrayIndex && v.key == key)
+                    return &v;
+            }
```
tux3, line 226: This completely changes the meaning of the loop! You're now iterating over all values, instead of the values of the array. And you're doing it while using the array's number of values as a stop condition. 
What was wrong with the simple range-for loop?

src/persistence/settingsserializer.cpp
```diff
@@ -251,10 +262,10 @@ void SettingsSerializer::save()
             writeStream(stream, a.name.toUtf8());
             writeStream(stream, vuintToData(a.size));
 
-            for (uint64_t vi : a.values)
+            for (int vi = 0; vi < a.values.size(); vi++)
```
tux3: This change seems wrong again, it's not the same meaning.
And why bother changing from a modern range-for to a for(;;) anyway? 
It's longer and I'm not convinced it's more readable.

src/core/indexedlist.h
tux3, line 11: IndexedList is really used as a map<int, ToxCall>, that's the easiest way to see it.
Each ToxCall has a unique int ID that toxcore gives it, we use that as the int key of the map. But since that ID is unique and it's already in the ToxCall structure, we don't really need to make a map that stores the keys, we can just make a vector and ask the ToxCall what its key (it's ID) is.
Since there are rarely more than a couple ToxCalls going on at the same time, the vector never has more than a couple elements, and just iterating over it is super fast (cache locality, predictors friendliness and all that). It also uses less memory than a map of hash table would, since it only has to store an array of the values instead of a a more complicated tree or hash table.

So I though it was simple enough to bother with, but after having to explain it I can see how it's confusing, we can absolutely remove it if it makes the code easier to understand. A map<int, ToxCall> will work just the same :)

src/persistence/serialize.h
antis81, line 30: Just look the two following code lines below. If you're confused now, I'm totally with you… :) Btw. What does the "V" stand for? At least It is not a vector/pointer.

sudden6:
I think "V" stands for variable, because I thing this is some sort of code to store a variable byte count together with the number of bits, so you can get the length and the data afterwards.

src/persistence/settingsserializer.cpp
antis81, line 226: > What was wrong with the simple range-for loop?

The `int` type is simply wrong. Would be a `Value` struct. Now how do you get the index, which is used to retrieve the value from the `values` array for comparison. Note: I'm not fixing/changing code architecture, just the type cast warnings!

And, uhm… how does that change meaning? Maybe I overlooked something, but it's the same, just done right. If however there is a modern way to retreive the container index, I'll gladly adopt it.

antis81, line 226: @tux3 Can you check this again please. Should be good now! Sorry, I got confused by the array index and the contained index value. Thanks for spotting this :innocent:. However, I now corrected the types in header, so most of the casts completely vanished.

src/persistence/settingsserializer.cpp
antis81: @tux3 Will change this as well. Same problem with the index.

antis81:
It's ready to merge. @tux3 Can you give feedback please about the for loops. If you have the time feel free to review.

src/persistence/serialize.cpp
```diff
@@ -97,13 +97,16 @@ QString dataToString(QByteArray data)
...
 
-    if (!strlen)
+    int strlen = num;
+
+    if (strlen <= 0)
         return QString();
 
-    data = data.right(data.size()-i); // Remove the strlen
+    // Remove the strlen
+    data = data.right(data.size() - i);
```
BiTOk: data = data.mid(i) looks simpler and works faster 
or
data.remove(0, i)

src/core/indexedlist.h
```diff
@@ -44,34 +57,42 @@ class IndexedList
...
+
     inline iterator end()
     {
         return v.end();
     }
+
     inline const_iterator end() const
     {
-        return v.end();
+        return v.cend();
```
BiTOk: Why end() returns const iterator?

src/persistence/serialize.cpp
```diff
@@ -97,128 +37,56 @@ QString dataToString(QByteArray data)
...
-{
-    return ((uint64_t)(uint8_t)data[0])
-            +(((uint64_t)(uint8_t)data[1])<<8)
-            +(((uint64_t)(uint8_t)data[2])<<16)
-            +(((uint64_t)(uint8_t)data[3])<<24)
-            +(((uint64_t)(uint8_t)data[4])<<32)
-            +(((uint64_t)(uint8_t)data[5])<<40)
-            +(((uint64_t)(uint8_t)data[6])<<48)
-            +(((uint64_t)(uint8_t)data[7])<<56);
+        num3 = data[i]; i++;
```
BiTOk: ++i faster

src/persistence/serialize.cpp
```diff
@@ -227,52 +95,29 @@ size_t dataToVUint(const QByteArray& data)
...
-    data[5] = (uint8_t)((num>>40) & 0xFF);
-    data[6] = (uint8_t)((num>>48) & 0xFF);
-    data[7] = (uint8_t)((num>>56) & 0xFF);
+    QByteArray data(sizeof(int), 0);
+    // Write the size in a Uint of variable lenght (8-32 bits)
+    int i=0;
+    while (num >= 0x80)
+    {
+        data[i] = static_cast<char>(num | 0x80);
+        i++;
```
BiTOk, line 114: ++i faster

src/core/indexedlist.h
Diadlo: Because it's const context

src/persistence/serialize.cpp
sudden6, line 114: I highly doubt that.

BiTOk, line 114: @sudden6 For some cases for some compillers it can have matter.

src/core/indexedlist.h
antis81: Not exactly, but the return value is wrong (it's a typedef). Will change to non-const iterator.

src/core/indexedlist.h
```diff
@@ -44,34 +57,32 @@ class IndexedList
     {
         return v.begin();
     }
-    inline const_iterator begin() const
-    {
-        return v.begin();
-    }
```
antis81, line 50: @BiTOk @sudden6 @Diadlo FYI

Removed the overload completely: `cbegin` is a c++11 feature, which removes ambiguity. As we already set C++11 as a minimum requirement, we don't need old method overloads, which are just being kept for compatibility.

src/persistence/serialize.cpp
```diff
@@ -25,200 +25,66 @@ Most of functions in this file are unsafe unless otherwise specified.
...
     } while ((num3 & 0x80) != 0);
-    unsigned int strlen = (uint) num;
 
-    if (!strlen)
+    if (strlen <= 0)
         return QString();
 
-    data = data.right(data.size()-i); // Remove the strlen
+    // Remove the strlen
+    data.remove(0, i - 1);
```
antis81, line 45: @BiTOk Better? :)

src/persistence/serialize.cpp
sudden6, line 114: Yep, for some cases and some compilers it may matter, but we are writing for many compilers. Despite that, premature optimization kills the program.

sudden6:
@antis81 ahem, what does "refactor(settings): correct use of std namespace" have to do with std namespace?

else, LGTM

antis81:
@sudden6 merged the evil commit as "fixup" (making it vanish), as it didn't contain much changes anyway and fits now nicely into history as well (see 6be8893).

src/persistence/serialize.cpp
BiTOk, line 114: > premature optimization

? Ok, let's write equally complex but more slowly code, it's path to success :-)
For int type prefix and postfix form of increment almost always has same speed. But few programmers using different increment type in that case for different values types. For iterator types prefix and postfix forms has big difference in speed. Speed measurement: http://www.viva64.com/media/images/content/b/0093_Iterators/image1.png
The same advice can be found in the book of Scott Meyers: Effective C++.

antis81, line 114: @BiTOk Calm down, this is nitpicking and completely irrelevant here (and anywhere else!!). Fact is: Behaviour is neither standardized nor consistent throughout compilers or even their versions respectively. The ONLY difference between prefix and postfix is simply, that prefix has an immediate effect on the variable, while postfix alters the variable at "end-of-command". I could also write: Use a compiler, that does things right and leave MSC behind. :)

antis81:
Guess we're ready to merge are we?

sudden6:
ok, ready to merge then

src/persistence/serialize.cpp
nurupo: According to @tux3, this `data.remove(0, i - 1);` change has broken deserialization and by extension the auto-update feature in qTox since v1.5.0 (related issue https://github.com/qTox/qTox/issues/3910)

antis81: Thanks for spotting that!

Issue #3378, opened 2016-06-12
[Feature request] pw field when adding / changing password should indicate whether CAPS-LOCK is on


zetok:
##### Brief Description

Currently only login page shows indicator of CAPS-LOCK being on when typing password – it should be available also when changing / adding a password on profile page.

Reproducible: Always
##### Steps to reproduce
1. Turn on CAPS-LOCK
2. Click on `Add` / `Change password` on profile page
##### Observed Behavior

Password filed doesn't have caps-lock indicator
##### Expected Behavior

CAPS-LOCK indicator should be shown
##### Additional Info

Related: #3245

b4n:
Same applies to the "load profile" login screen field.  Only the "New profile" fields show a capslock warning.

b4n:
I've got a fix for that, but I made it on top of #3383 because it changes API, so I'd rather wait for that one to be dealt with first.

However, it's not as pretty as I'd wish it would be.  Ideally, I guess we should have a custom QLineEdit widget for passwords that does it itself, but:
1. I don't know if it'd be possible to use to from _.ui_ definitions?
2. I failed to find an encapsulated way to get notified when capslock changes.  Current way (from #3257) is on the `::event()` virtual method on the top level (window), so that's not very encapsulated.
   
   Encapsulating would be interesting even without 1 above, because it would avoid the need for a caller to manually call `::updateIndicator()`.

But as I'm totally new to Qt, maybe someone more experienced would know.

PR #3383, opened 2016-06-12
fix(capslockindicator): fix altering the line edit height

### LINKED ISSUE #3379: Password fields with CapsLock warning shrink vertically
##### Brief Description

OS: Linux (Debian Sid, up-to-date as of today), under MATE desktop
qTox version: Both Nightly package from this morning (1.4.0-151-g7ca9809-1) and custom build of Git's master (645e9504eccae3da811237a1ab1c20e1709df048)
Commit hash: 7ca98093a23ee8fbdba83eb5b2e9b2a4b7dac994 and 645e9504eccae3da811237a1ab1c20e1709df048
Qt:  5.5.1

Reproducible: Always
##### Steps to reproduce
1. Launch qTox
2. Go to the New Profile screen (if not already the case)
3. Hit CapsLock
##### Observed Behavior

The password fields shrink and its content is mostly invisible:
qtox_new_profile_2capslock

A slight variant is when CapsLock is enabled before the screen shows up: same but with small icon instead of overflowing one.
qtox_new_profile_3capslock
##### Expected Behavior

No shrinking, just a CapsLock warning.  Here is a screenshot of the screen when CapsLock is off (and everything displays fine):
qtox_new_profile_1
##### Additional Info

This might be related to be running on MATE and Qt trying to look GTK, not sure.

b4n:
Fix altering the Line Edit widget height when the indicator is on (e.g.
when caps lock is active).

To avoid messing with the style and all the complex issues that arise
from it, use an action to display the icon inside the line edit, as it
is a built-in feature that work just fine.

Fixes #3379.

---

CapsLock off:
qtox_new_profile_capslock_off
CapsLock on:
qtox_new_profile_capslock_on

As this approach doesn't alter any style and just relies on _QLineEdit_ support for _QAction_s, I'm fairly confident it should work everywhere, but it might be better for someone with another setup than mine to test to avoid similar issues to the one this fixes.

Also, be warned that I'm very new to Qt, so it might be prudent to carefully review my use of it (even though I checked out the docs a fair bit and tested it fairly correctly I believe, but still).


b4n:
I don't think https://github.com/tux3/qTox/pull/3257#issuecomment-219309291 is relevant, but it might be worth checking it out.

anoadragon453:
Can confirm this fixes the issue on my end. :+1: 

Now I want to know why there isn't a capslock indicator on the Load Profile password field! :)

b4n:
@anoadragon453 thanks for testing!  and for the other field, see #3378

Diadlo:
Maybe disable clickability?

b4n:
I can `setEnabled(false)`, but then the cursor doesn't switch back to arrow on it.  But either way's preferred, and there might be a way to do that that I don't know about.

zetok:
on 653e0b5
- [x] caps-lock indicator disappears once password is long enough to get to its position
- [x] after ↑ there's stuff printed to stdout whenever something is typed in password field:
  
  ```
  [Thread 0x7fffcb8c5700 (LWP 5506) exited]
  Maximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reachedMaximum number of clients reached
  ```

On a plus side, I wasn't able to reproduce issue present on master, where changing to caps-lock would cause password dots to become invisible.

b4n:
> caps-lock indicator disappears once password is long enough to get to its position

I don't see this.  Though yes, it does disappear when the "Maximum number of clients reached" message starts to get spewed.

> after ↑ there's stuff printed to stdout whenever something is typed in password field: […]

That's also reproducible on current master.  You just need to perform enough events on the line edit (i.e. holding `a` for a while, then backspace, or anything like this).
I have no clue from where it comes, but I would guess some Qt internal maybe?  Nor where to look to fix this.

Edit: how, looks like https://stackoverflow.com/questions/1409886/help-maximum-number-of-clients-reached-segmentation-fault, I'll take a look

b4n:
@zetok it indeed was the XDisplay not being properly closed, I now added a patch fixing this.

b4n:
I did some further refactoring on the `CapsLockIndicator` to improve encapsulation, and later event handling: https://github.com/b4n/qTox/commits/widget/capslockindicator/more-encapsulated
Should I add these here or wait for this to be dealt with and then open a new PR?  And, are these wanted?  (I believe it'll become interesting when spreading use of this to other password fields not to duplicate some code)

sudden6:
@b4n I think it's best to directly include your other work into this PR, because it's all about improving the capslockindicator. A few comments on the changes of your branch:
- https://github.com/b4n/qTox/commit/283a98828113fc08693b81379c361d9d56e60f3b isn't a good idea in my eyes, because if the compiler does it's job, the code is optimized away anyways. Also to implement caps-lock support for OSX this would have to be reverted.
- I don't know if installing a custom eventhandler will impact the performance in a noticeable way. Mainly my concern is, that the eventhandler is properly unregistered if no password field is visible. Would be neat if you or someone else could elaborate a bit more on this.

b4n:
> b4n@283a988 isn't a good idea in my eyes, because if the compiler does it's job, the code is optimized away anyways.

I don't see how the compiler could know that calling `setVisible(false)` isn't needed when an even comes up, there could be other code having changed that, so I doubt the compiler can optimize it away.

> Also to implement caps-lock support for OSX this would have to be reverted.

Yes, but if I trust the comment, OSX already shows an indicator?  I can't check, but that's what the comment _It doesn't needed for OSX, because it shows indicator by default_ seemed to suggest.

> I don't know if installing a custom eventhandler will impact the performance in a noticeable way. 

I doubt it'd be a real concern, but I might be wrong.  But I would try and avoid having several of these handlers for no reason, just in case; hence b4n@00a4a629ffc12b3a03eb48b74113cf5278c02096

> Mainly my concern is, that the eventhandler is properly unregistered if no password field is visible. Would be neat if you or someone else could elaborate a bit more on this.

In my branch the event handler is remove when there are no more `CapsLockIndicator`s around.  Unfortunately this means when the widgets get destroyed, and that just doesn't happen: even the loginscreen seem to live for the whole application's lifetime.

I didn't try yet, but we probably could relatively easily only have one when the field is visible by using `showEvent()`/`hideEvent()` or `setVisible()`, but unfortunately I couldn't find a way to listen to such events on another widget.  So that would be easily doable if we had a custom `PasswordEdit : QLineEdit`, but I don't know how easy it'd be to use such a custom widget in the current code (e.g. plug it in the generated UI).
If anyone has knowledge on that, I'd be happy to try and implement such a widget, that would be even better API that whatever we can get right now.

b4n:
I added all the commits, but I can still remove b4n@00a4a62 if you want.

sudden6:
> >    Also to implement caps-lock support for OSX this would have to be reverted.
> 
> Yes, but if I trust the comment, OSX already shows an indicator? I can't check, but that's what the comment It doesn't needed for OSX, because it shows indicator by default seemed to suggest.

@Diadlo can you explain this?

I thought, it just isn't implemented for OSX

> >    I don't know if installing a custom eventhandler will impact the performance in a noticeable way.
> 
> I doubt it'd be a real concern, but I might be wrong. But I would try and avoid having several of these handlers for no reason, just in case; hence b4n@00a4a62

After some further reading, I don't think this will have a negative impact on performance.

Indeed the best way would be to make a custom widget.

Seems like it isn't actually that difficult, for our usecase we could probably use "Promote to" described here: http://doc.qt.io/qt-4.8/designer-using-custom-widgets.html

Diadlo:
@sudden6 As I know OSX don't need caps lock indicator for password fields. But in theory caps lock checker may have other use case

b4n:
@sudden6 I have pushed new commits that introduce a new custom widget that wraps and hide it all, and implements only listening to all events when a password field is actually visible.  That was actually fairly easy to do after searching a little in Qt's docs :)

Also, I added b4n@145b4f0 here to fix #3378 and generalize the use of that new `PasswordEdit` widget, so now all password fields should be covered -- and it's easy to make new fields use it from inside QtCreator's designer.

b4n:
OK, builds fails because of a typo in a commit message, and apparently it's a new check (the commit was here before).  I'll fix this.

b4n:
(fixed, build in progress but passed this step)

sudden6:
@b4n very well done, looks pretty good at first glance, but I'll review and test it some more before merge.

EDIT: what about https://github.com/tux3/qTox/pull/3383#issuecomment-226199526 do they still occur?

b4n:
@sudden6 cool :)  I just added a commit not adding the action at all if the indicator is disabled, as it'd be always hidden anyway.  This might be candidate to get squashed into the `PasswordEdit` commit, but I'll let you tell me if/what you want squashed.

b4n:
> EDIT: what about https://github.com/tux3/qTox/pull/3383#issuecomment-226199526 do they still occur?

No, it's fixed by f919073

PR #3385, opened 2016-06-13
feat(l10n): update Italian translation

ghost:


translations/it.ts
```diff
@@ -1282,11 +1282,11 @@ will be sent to them when they appear online to you.</source>
     </message>
     <message>
         <source>Remove chat from this window</source>
-        <translation type="unfinished"></translation>
+        <translation>Rimuovi la chat da questa finestra</translation>
     </message>
     <message>
-        <source>1 user in chat</source>
-        <translation type="unfinished"></translation>
+        <source>%1 user in chat</source>
```
zetok: this too needs to be fixed – there's no `%` there

translations/it.ts
```diff
@@ -1229,9 +1229,9 @@ will be sent to them when they appear online to you.</source>
         <translation>Avvia chiamata</translation>
     </message>
     <message>
-        <source>1 user in chat</source>
+        <source>%1 user in chat</source>
```
zetok: same here, remove `%`

zetok:
`<source>` strings are to not be edited at all – they're there for translation. `lupdate` is used to update source strings.

Issue #3154, opened 2016-04-17
"(qtox:8687): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed" error on startup


ghost:
I have noticed that since one of the recent updates I am getting this error when starting qTox:

```
(qtox:8687): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```

This is the full log:

> [21:57:07.914] src/persistence/settings.cpp:121 : Debug: "Loading settings from /home/boghog/.config/tox/qtox.ini"
> [21:57:07.915] src/ipc.cpp:46 : Debug: Our global IPC ID is  15503824813809099664
> [21:57:07.915] src/main.cpp:196 : Debug: built on:  13:53:32 Apr 16 2016 ( 1460814569 )
> [21:57:07.915] src/main.cpp:197 : Debug: commit:  built without git 
> 
> [21:57:07.916] src/nexus.cpp:81 : Debug: Starting up
> [21:57:14.683] src/persistence/profile.cpp:86 : Debug: Loading tox save  "/home/boghog/.config/tox/遊牧民.tox"
> [21:57:14.931] src/persistence/settings.cpp:378 : Debug: "Saving global settings at /home/boghog/.config/tox/qtox.ini"
> [21:57:15.050] src/persistence/settings.cpp:296 : Debug: Loading personal settings from "/home/boghog/.config/tox/遊牧民.ini"
> [21:57:15.934] src/core/core.cpp:236 : Debug: Loading user profile
> [21:57:15.934] src/persistence/profile.cpp:249 : Debug: Loading tox save  "/home/boghog/.config/tox/遊牧民.tox"
> [21:57:16.051] src/core/core.cpp:132 : Debug: Core starting with IPv6 enabled
> [21:57:17.595] src/widget/systemtrayicon.cpp:65 : Debug: Using GTK backend
> 
> (qtox:8687): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
> [21:57:19.098] src/core/core.cpp:407 : Debug: "Connecting to 194.249.212.109:33445 (fluke571)"
> [21:57:19.099] src/core/core.cpp:407 : Debug: "Connecting to 185.25.116.107:33445 (MAH69K)"
> [21:57:23.598] src/core/core.cpp:407 : Debug: "Connecting to 144.76.60.215:33445 (sonOfRa)"
[...]

PR #3390, opened 2016-06-13
Fix SystemTrayIcon GTK implementation critical warning

b4n:
See #3154.  This fixes the runtime warning under Xfce/GNOME/MATE and Cinnamon desktops:

```
[01:06:45.205] src/widget/systemtrayicon.cpp:65 : Debug: Using GTK backend

(qtox:28186): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```

While at it, refactor some code to avoid 4 time code duplication.

### DIFF
--- a/src/widget/systemtrayicon.cpp
+++ b/src/widget/systemtrayicon.cpp
@@ -116,6 +116,26 @@ QString SystemTrayIcon::extractIconToFile(QIcon icon, QString name)
     return iconPath;
 }
 
+#if defined(ENABLE_SYSTRAY_GTK_BACKEND) || defined(ENABLE_SYSTRAY_STATUSNOTIFIER_BACKEND)
+GdkPixbuf* SystemTrayIcon::convertQIconToPixbuf(const QIcon &icon)
+{
+    void (*callbackFreeImage)(guchar*, gpointer) =
+        [](guchar* image_bytes, gpointer)
+    {
+        delete[] image_bytes;
+    };
+    QImage image = icon.pixmap(64, 64).toImage();
+    if (image.format() != QImage::Format_RGBA8888_Premultiplied)
+        image = image.convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+    guchar* image_bytes = new guchar[image.byteCount()];
+    memcpy(image_bytes, image.bits(), image.byteCount());
+
+    return gdk_pixbuf_new_from_data(image_bytes, GDK_COLORSPACE_RGB, image.hasAlphaChannel(),
+                                    8, image.width(), image.height(), image.bytesPerLine(),
+                                    callbackFreeImage, NULL);
+}
+#endif
+
 void SystemTrayIcon::setContextMenu(QMenu* menu)
 {
     if (false);
@@ -145,21 +165,7 @@ void SystemTrayIcon::setContextMenu(QMenu* menu)
                 item = gtk_menu_item_new_with_label(aText.toStdString().c_str());
             else
             {
-                void (*callbackFreeImage)(guchar*, gpointer) =
-                    [](guchar*, gpointer image_bytes)
-                {
-                    free(reinterpret_cast<guchar*>(image_bytes));
-                };
-                QImage image = a->icon().pixmap(64, 64).toImage();
-                if (image.format() != QImage::Format_RGBA8888_Premultiplied)
-                    image = image.convertToFormat(QImage::Format_RGBA8888_Premultiplied);
-                guchar* image_bytes = (guchar*)malloc(image.byteCount());
-                memcpy(image_bytes, image.bits(), image.byteCount());
-
-                GdkPixbuf* pixbuf = gdk_pixbuf_new_from_data(image_bytes, GDK_COLORSPACE_RGB, image.hasAlphaChannel(),
-                                        8, image.width(), image.height(), image.bytesPerLine(),
-                                        callbackFreeImage, image_bytes);
-
+                GdkPixbuf* pixbuf = convertQIconToPixbuf(a->icon());
                 item = gtk_image_menu_item_new_with_label(aText.toStdString().c_str());
                 gtk_image_menu_item_set_image(GTK_IMAGE_MENU_ITEM(item), gtk_image_new_from_pixbuf(pixbuf));
                 gtk_image_menu_item_set_always_show_image(GTK_IMAGE_MENU_ITEM(item),TRUE);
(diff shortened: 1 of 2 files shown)


src/widget/systemtrayicon.cpp
```diff
@@ -116,6 +113,26 @@ QString SystemTrayIcon::extractIconToFile(QIcon icon, QString name)
...
+{
+    void (*callbackFreeImage)(guchar*, gpointer) =
+        [](guchar*, gpointer image_bytes)
+    {
+        free(reinterpret_cast<guchar*>(image_bytes));
+    };
+    QImage image = icon.pixmap(64, 64).toImage();
+    if (image.format() != QImage::Format_RGBA8888_Premultiplied)
+        image = image.convertToFormat(QImage::Format_RGBA8888_Premultiplied);
+    guchar* image_bytes = (guchar*)malloc(image.byteCount());
```
antis81: This is C syntax from copied code, right? Would you mind converting to C++?

``` c++
guchar* image_bytes = new guchar[image.byteCount()];

// counterpart in callback above:
delete[] image_bytes;
```

src/widget/systemtrayicon.cpp
```diff
@@ -116,6 +113,26 @@ QString SystemTrayIcon::extractIconToFile(QIcon icon, QString name)
     return iconPath;
 }
 
+#if defined(ENABLE_SYSTRAY_GTK_BACKEND) || defined(ENABLE_SYSTRAY_STATUSNOTIFIER_BACKEND)
+GdkPixbuf* SystemTrayIcon::convertQIconToPixbuf(const QIcon &icon)
+{
+    void (*callbackFreeImage)(guchar*, gpointer) =
+        [](guchar*, gpointer image_bytes)
+    {
+        free(reinterpret_cast<guchar*>(image_bytes));
```
antis81: What does the `guchar*` pointer give? That cast smells ugly as hell…

b4n: Agreed, but I didn't want to start argue with the code that was there before me (I just moved it around).  But if we're at it, let's improve this a little.

src/widget/systemtrayicon.cpp
b4n: OK.  Another solution could be keeping the QImage around.  It would keep a little more memory around, but would avoid the copy.

antis81: That's totally fine. We're talking about the application icon here, which is usually a single instance.

b4n: What do you mean, keeping the QImage or the current copy?

antis81: The QImage.

b4n: (I pushed a change for your version, but I can add/amend it)

src/widget/systemtrayicon.cpp
```diff
@@ -116,6 +113,26 @@ QString SystemTrayIcon::extractIconToFile(QIcon icon, QString name)
     return iconPath;
 }
 
+#if defined(ENABLE_SYSTRAY_GTK_BACKEND) || defined(ENABLE_SYSTRAY_STATUSNOTIFIER_BACKEND)
+GdkPixbuf* SystemTrayIcon::convertQIconToPixbuf(const QIcon &icon)
+{
+    void (*callbackFreeImage)(guchar*, gpointer) =
+        [](guchar* image_bytes, gpointer)
+    {
+        delete[] image_bytes;
```
antis81, line 122: Wait a moment. If this works, keeping the copy is for the better. Whatever you prefer.

antis81:
Can you squash into a single commit, please?

src/widget/systemtrayicon.cpp
b4n, line 122: This works, GdkPixbuf passes the bytes to free as first argument, the other is for any possible additional data required. (see https://developer.gnome.org/gdk-pixbuf/unstable/gdk-pixbuf-Image-Data-in-Memory.html#gdk-pixbuf-new-from-data)

And if you like it this way, I'll keep it, because for now I'm not quite sure how to keep the image around, as I'm a C++ noob and `QPixmap::toImage()` doesn't return a pointer, so I'd somewhat either need to trick the C++ code to understand the C callback took a reference, or copy the QImage, which, unless it implements clever buffer sharing, is likely to involve the same copy.

b4n:
@antis81 done (I assumed you only wanted the allocator commit to be squashed into the unifying one.  If you want all together I can also do it, but I think it makes less sense)

src/widget/systemtrayicon.cpp
antis81, line 122: Well, here's how to keep the image around:

``` c++
GdkPixbuf* SystemTrayIcon::convertQIconToPixbuf(const QIcon &icon)
{
    // create the image before the callback/lambda assignment. 
    QImage image;

    // copy the image and carry to the callback (the [=image]; not [&image], which means reference scope -> becomes dangling reference here)
    // note, that the copy c'tor of QImage works with copy-on-write -> no bytes are  copied.
    void (*callbackFreeImage)(guchar*, gpointer) = [=image](guchar* image_bytes, gpointer)
    {
        // nothing to be done here.
    }

…
}
```

b4n, line 122: Oh cool, real closure :)

Though… how does C++ know when the callback is gone for good?  C++ garbage collector magic even if it's passed to a C library?

antis81, line 122: No garbage collection magic. Qt relies on clever smart pointers and copy-on-write with most of it's objects. Not sure about the closure though. Might leave a mem-leak on the function pointer. Hm… if you have info about the life-span of lambda assignments I'd be interested in that as well.

**EDIT** Yep, looks like we found a mem leak. Have a look here:
http://stackoverflow.com/questions/7941562/what-is-the-lifetime-of-a-c-lambda-expression

b4n, line 122: Yeah so as I expected the lambda will live on, and so will its state.  So yeah, I'll keep the manual copy :)

b4n, line 122: hum… though, would that mean that any lambda use leads to creating a new instance?  that would make them a lot more costly than a normal function or even method :S
But yeah, I don't see how it could do otherwise if the user doesn't destroy the object, so if it works as a standard function pointer.

antis81:
@sudden6 Do you have system to test this? I'd like to merge soon.

sudden6:
@antis81 tested on Fedora 23

antis81:
@sudden6 Thanks! Ok, I'm going to merge it then.

zetok:
@antis81 um, why did you push unsigned commits on master?

zetok:
https://github.com/tux3/qTox/blob/master/MAINTAINING.md#pull-requests eh.

zetok:
@antis81 "fixed", but please don't push unsigned by yourself stuff on master.

antis81:
@zetok Sorry, thanks for reminder. Btw. can't we reject those on GH via commit hook?

zetok:
> Btw. can't we reject those on GH via commit hook?

I don't know if anything like that already exists – do you have something specific in mind?

antis81:
@zetok Signed off commits do append a special text line in the commit. If this line doesn't exist (plus PGP key), the ~~commit~~ push should be rejected. We can even check, if it was one of the maintainers who pushed that commit (e.g. by nick name or registered GPG-Signatures).

sudden6:
would be a really nice safeguard

zetok:
@antis81 yes, but AFAIK there's nothing providing service like that. Unless you mean that there should be custom hook made & ran on server?

Also, it wouldn't require only checking whether given commit is signed, but whether merge commit is signed, regardless of whether commit in merged branch is signed + checking whether given commit is signed itself when not being pushed as a part of merged branch.

antis81:
> Unless you mean that there should be custom hook made & ran on server?

Exactly, this is what I mean. Probably Github provides a template for this already - not sure. But it would help to guide people.

zetok:
> Probably Github provides a template for this already - not sure. But it would help to guide people.

There are docs: https://developer.github.com/guides/building-a-ci-server/

I'm not interested though in writing something that is useful only with a proprietary service.

With that being said, when finally GH will go evil and it'll be time to move on to gitlab, I could try to write something that would integrate with it.

antis81:
> I'm not interested though in writing something that is useful only with a proprietary service.

Totally with you. We can do the same within Jenkins - up to now it's just a vague idea from my side to "bullet proof" our workflow. :)

> With that being said, when finally GH will go evil and it'll be time to move on to gitlab, I could try to write something that would integrate with it.

Don't forget about BitBucket, which looks like quite a good deal since it completely supports Git repos. :)

sudden6:
Sometime ago I already suggested to create a third travis instance which only does the syntax check for our commits, maybe we could also do the "is signed" check there and reject commits that fail?

zetok:
> maybe we could also do the "is signed" check there and reject commits that fail?

AFAIK can't be done.

PR #3402, opened 2016-06-16
chore(simple_make): Fix build for Fedora

sudden6:
Add some missing dependencies and the fix from #3308

Tested on a fresh Fedora 23 install

### DIFF
--- a/simple_make.sh
+++ b/simple_make.sh
@@ -13,7 +13,8 @@ elif which pacman; then
         git base-devel qt5 openal libxss qrencode opus libvpx libsodium
 elif which dnf; then
     sudo dnf group install \
-        "Development Tools"
+        "Development Tools" \
+        "C Development Tools and Libraries"
     # pure Fedora doesn't have what it takes to compile qTox (ffmpeg)
     sudo dnf install \
         http://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm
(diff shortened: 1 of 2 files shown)


simple_make.sh
```diff
@@ -13,15 +13,19 @@ elif which pacman; then
...
         http://download1.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm
     sudo dnf install \
         git qt-devel qt-doc qt-creator qt5-qtsvg qt5-qtsvg-devel \
         openal-soft-devel qt5-qttools-devel libXScrnSaver-devel \
         qrencode-devel opus-devel libvpx-devel glib2-devel gdk-pixbuf2-devel \
-        gtk2-devel libsodium-devel ffmpeg-devel sqlite sqlite-devel
+        gtk2-devel libsodium-devel ffmpeg-devel sqlite sqlite-devel libtool \
+        openssl-devel
+    # Fedora by default doesn't include libs in /usr/local/lib so add it
+    echo '/usr/local/lib/' | sudo tee -a /etc/ld.so.conf.d/locallib.conf
```
zetok:
``` bash
is_locallib() {
    grep -q '^/usr/local/lib/$' "$@"
}

readonly LOCALLIB="/etc/ld.so.conf.d/locallib.conf"

is_locallib "$LOCALLIB" \
    || echo '/usr/local/lib/' \
        | sudo tee -a "$LOCALLIB"

```

or something..?

sudden6: could you please explain what your code does?

zetok: `is_locallib()` checks whether line `/usr/local/lib/` exists in supplied file.

Only if line doesn't exist (non-zero exit status of `grep`) it is being appended to the file.

I didn't really write it well – confusing function & variable names, etc, just wanted to express the idea in which direction it should go.

`/usr/local/lib/` also should be a variable, for that though I'd need to test whether I'm escaping correctly `grep` regexp, so I just left it as is.

zetok: oh, right, a good resource: http://www.kfirlavi.com/blog/2012/11/14/defensive-bash-programming/

PR #3411, opened 2016-06-18
fix(persistence): pass by reference to const

PafXUS:
This commit changes passing by value behavior for password occurrences to pass by reference to const.
I've noticed great inconsistency in codebase about how "in" parameters of large (with sizeof(T) > 8 bytes) non-trivial types are passed around. For example QString, QByteArray, QList, std::function, IPCEventHandler, ToxFile sometimes are passed around by value, sometimes, by reference to const. I think all of them should be passed by reference to const by default. This advice can be found in [CoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const). This must be applied to Qt types with implicit sharing as well (you can see how Qt library itself pass it's types). Exceptions to this can be intentional moving around types, [copy-and-swap idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Copy-and-swap), or copy for modification.
   If everybody agree on this, i can help with fixing ~400 occurrences of this issue in codebase. Please discuss.

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -492,7 +492,7 @@ void Settings::savePersonal(Profile* profile)
     savePersonal(profile->getName(), profile->getPassword());
 }
 
-void Settings::savePersonal(QString profileName, QString password)
+void Settings::savePersonal(QString profileName, const QString &password)
 {
     if (QThread::currentThread() != settingsThread)
         return (void) QMetaObject::invokeMethod(&getInstance(), "savePersonal",
@@ -924,7 +924,7 @@ QString Settings::getToxmePass() const
     return toxmePass;
 }
 
-void Settings::setToxmePass(QString pass)
+void Settings::setToxmePass(const QString &pass)
 {
     QMutexLocker locker{&bigLock};
     toxmePass = pass;
--- a/src/persistence/db/rawdatabase.cpp
+++ b/src/persistence/db/rawdatabase.cpp
@@ -312,7 +312,7 @@ bool RawDatabase::remove()
 }
 
 
-QString RawDatabase::deriveKey(QString password)
+QString RawDatabase::deriveKey(const QString &password)
 {
     if (password.isEmpty())
         return {};
--- a/src/persistence/db/rawdatabase.h
+++ b/src/persistence/db/rawdatabase.h
@@ -89,7 +89,7 @@ protected slots:
 
 protected:
     /// Derives a 256bit key from the password and returns it hex-encoded
-    static QString deriveKey(QString password);
+    static QString deriveKey(const QString &password);
     /// Extracts a variant from one column of a result row depending on the column type
     static QVariant extractData(sqlite3_stmt* stmt, int col);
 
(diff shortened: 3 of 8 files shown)


sudden6:
Personally I think this is a great idea, not only because of possible performance gains, but also because a consistent style makes modifying the code easier.

@antis81 @tux3 what do you think about this?

Diadlo:
What about to make methods that do not change the object, constant?

PafXUS:
Making methods constant can be next step in our journey.

tux3:
@PafXUS It's not a bad idea, but we have to be **very** careful. If you just replace every large parameter with a const& without checking, you'll introduce several use-after-frees in the process.
I'm wary of mass changes, it's easy to make a mistake when those 400 occurences need to be checked very thoroughly.

src/persistence/settings.cpp
```diff
@@ -492,7 +492,7 @@ void Settings::savePersonal(Profile* profile)
     savePersonal(profile->getName(), profile->getPassword());
 }
 
-void Settings::savePersonal(QString profileName, QString password)
+void Settings::savePersonal(QString profileName, const QString &password)
```
tux3, line 495: And that looks like it introduces a use-after-free.
`QMetaObject::invokeMethod` is asynchronous here, it will send a message to the Settings thread, passing it the `const& password`, which might have been deleted by the time the Settings thread wakes up and processes the event.

PafXUS, line 495: @tux3 I agree that we should be very careful. But, i don't agree that const QString & will introduce use-after-free problem. Why so? Because - all "in" arguments by default are copy-constructed on caller side and their lifetime is connected to event not to original object from call. You can find in [Qt docs](https://doc.qt.io/qt-5.7/qmetaobject.html#invokeMethod): 

> With asynchronous method invocations, the **parameters** must be of types that are known to Qt's meta-object system, because **Qt needs to copy the arguments to store them** in an event behind the scenes.

And there were at least 2 separate discussions about this behavior [about signal-slots in general](http://www.embeddeduse.com/2013/06/29/copied-or-not-copied-arguments-signals-slots/) and about [invokeMethod ](http://www.qtcentre.org/threads/46934-Using-QMetaObject-invokeMethod-on-a-SLOT-that-has-a-QString-Reference-argument?s=680ce9ced4a4b0ef87d82615926d0071) 
 So in short: no, there is no any harm by doing const QString & instead of  QString in that case. Or can you please bring some small illustration in case of some misunderstanding on my part, of what exactly will happen.

tux3, line 495: You're righht! It's okay in this case, then.

PafXUS, line 495: So, how it's better to set up workflow? Should i add new commits as part of this pr, or this can be merged, and new one go as separate pr?

sudden6, line 495: append commits to this PR please

PR #3413, opened 2016-06-18
refactor(.qrc): split emoticon resource files into separate ones

zetok:
cc @Diadlo 

background: https://github.com/qTox/qtox-irc-logs/blob/master/2016/06/%23qtox_20160617.log.txt#L70,L87


Diadlo:
LGTM

sudden6:
I wonder if this is still a good idea, since 279225cd2d875a78a12774fd353070b52f9908bc proved to be much more effective.

PR #3414, opened 2016-06-18
chore(test-pr.sh): add test-pr.sh script as requested

zetok:
Also refactored & split `merge-pr.sh` script, since most of required
code for `test-pr.sh` already existed there.

Differences from `merge-pr.sh`:
- if there's no `upstream` remote, https version is used instead of
  git
- merge commit _is not_ GPG-signed (less hassle)
- merge branch is named e.g. `test1234` instead of `merge1234`
- instructions for pushing test branch onto upstream master are not
  printed

### DIFF
--- a/test-pr.sh
+++ b/test-pr.sh
@@ -15,10 +15,9 @@
 #    You should have received a copy of the GNU General Public License
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-# Script for merging pull requests. Works only when there are no merge
+# Script for testing pull requests. Works only when there are no merge
 # conflicts. Assumes that working dir is a qTox git repo.
 #
-# Requires SSH key that github accepts and GPG set to sign merge commits.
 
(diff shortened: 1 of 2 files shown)


test-pr.sh
```diff
@@ -0,0 +1,61 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU General Public License for more details.
+#
+#    You should have received a copy of the GNU General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+# Script for merging pull requests. Works only when there are no merge
```
sudden6: The description should probably say that it's only for testing.

zetok: Fixed.

PR #3418, opened 2016-06-19
fix(passwordedit.cpp): Fix build issue with Qt 5.3

### LINKED ISSUE #3416: qTox does not build on Debian 8 Jessie because of QVector<T>::removeOne() usage
##### Brief Description

OS: Debian 8 KDE
qTox version: HEAD
toxcore:  HEAD
Qt:  5.3.2, stock Debian 8
Hardware:  VirtualBox

Reproducible: Always
##### Steps to reproduce

Build qTox using stock Qt on Debian 8
##### Observed Behavior

```
src/widget/passwordedit.cpp: In member function ‘void PasswordEdit::unregisterHandler()’:
src/widget/passwordedit.cpp:47:31: error: ‘class QVector<QAction*>’ has no member named ‘removeOne’
         eventHandler->actions.removeOne(action);
                               ^
```

As removeOne() was added in [Qt 5.4](https://doc.qt.io/qt-5/qvector.html#removeOne)
##### Expected Behavior

Should build on all supported systems, mentioned on INSTALL.md.

I believe it should be stressed in contribution guidelines to use lowest-common-denominator Qt version. This is not the first time I have reported such incompatibility error.

Talkless:
QVector<T>::removeOne() was added in Qt 5.4, and this method usage
broke build for Debian 8 whitch has Qt 5.3.2.

Add alternative implementation and comment for the future to use
removeOne() when this becomes possible.

Closes #3416

### DIFF
--- a/src/widget/passwordedit.cpp
+++ b/src/widget/passwordedit.cpp
@@ -44,7 +44,8 @@ void PasswordEdit::unregisterHandler()
 #ifdef ENABLE_CAPSLOCK_INDICATOR
     if (eventHandler && eventHandler->actions.contains(action))
     {
-        eventHandler->actions.removeOne(action);
+        //TODO: future: use removeOne() when Qt 5.3 (Debian 8) support ends.
+        eventHandler->actions.remove(eventHandler->actions.indexOf(action));
         if (eventHandler->actions.isEmpty())
         {
             delete eventHandler;


Diadlo:
I think, better will be used construction, like this

```
#if (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
...
#else
...
#endif
```

Talkless:
Yes, but even with #ifdef I would like to add //TODO to remember to cleanup that #else when time comes . Maintenance is needed in both cases. Personally I think this is too small issue to start ifdefing actually, but if you still think it is better, I can change as suggested of course.

sudden6:
@Diadlo I agree with @Talkless this is a very tiny issue, not worth the hassle of an ifdef.

Diadlo:
Oh, it's just syntactic sugar, ifdef is excessively. You absolutely right. Sorry

Talkless:
Thanks!

b4n:
Oops, sorry.

Maybe a better fix would be to do an optimisation I discarded because it was useless, e.g. walking the `QVector` only once:

``` diff
diff --git a/src/widget/passwordedit.cpp b/src/widget/passwordedit.cpp
index e6ef11f..bbcc42b 100644
--- a/src/widget/passwordedit.cpp
+++ b/src/widget/passwordedit.cpp
@@ -42,10 +42,11 @@ void PasswordEdit::registerHandler()
 void PasswordEdit::unregisterHandler()
 {
 #ifdef ENABLE_CAPSLOCK_INDICATOR
-    if (eventHandler && eventHandler->actions.contains(action))
+    int idx;
+
+    if (eventHandler && (idx = eventHandler->actions.indexOf(action)) >= 0)
     {
-        //TODO: future: use removeOne() when Qt 5.3 (Debian 8) support ends.
-        eventHandler->actions.remove(eventHandler->actions.indexOf(action));
+        eventHandler->actions.remove(idx);
         if (eventHandler->actions.isEmpty())
         {
             delete eventHandler;
```

Would both support older versions as with this patch, and also avoid double lookup.

sudden6:
@b4n even better, mind a PR?

b4n:
sure

Talkless:
Yeah I had this idea, but that list contains only two items... so I didn't bother to overcomplicate.

b4n:
@Talkless that's why I didn't do that in the initial patchset, but it seems better to me to gain something out of the added complexity than making it a temporary workaround.  My take is, if we're gonna complicate the code and use `indexOf()` anyway, use it all the way so it's not a workaround but an actual improvement there's no point to remove later on.

PR #3424, opened 2016-06-21
refactor(passwordedit): optimize event handler unregistration

b4n:
Only search the vector once, and using old enough Qt API.

Follow up to #3418.

### DIFF
--- a/src/widget/passwordedit.cpp
+++ b/src/widget/passwordedit.cpp
@@ -42,10 +42,11 @@ void PasswordEdit::registerHandler()
 void PasswordEdit::unregisterHandler()
 {
 #ifdef ENABLE_CAPSLOCK_INDICATOR
-    if (eventHandler && eventHandler->actions.contains(action))
+    int idx;
+
+    if (eventHandler && (idx = eventHandler->actions.indexOf(action)) >= 0)
     {
-        //TODO: future: use removeOne() when Qt 5.3 (Debian 8) support ends.
-        eventHandler->actions.remove(eventHandler->actions.indexOf(action));
+        eventHandler->actions.remove(idx);
         if (eventHandler->actions.isEmpty())
         {
             delete eventHandler;


src/widget/passwordedit.cpp
```diff
@@ -42,10 +42,11 @@ void PasswordEdit::registerHandler()
 void PasswordEdit::unregisterHandler()
 {
 #ifdef ENABLE_CAPSLOCK_INDICATOR
-    if (eventHandler && eventHandler->actions.contains(action))
+    int idx;
+
+    if (eventHandler && (idx = eventHandler->actions.indexOf(action)) >= 0)
```
sudden6, line 47: I think it's more readable if you move the `= eventHandler->actions.indexOf(action)` part outside the if.

b4n, line 47: If you want, but that means we need 2 separate `if` clauses as we need to check `eventHandler` before accessing it.

b4n, line 47: FWIW, it would look like this:

``` diff
diff --git a/src/widget/passwordedit.cpp b/src/widget/passwordedit.cpp
index bbcc42b..9757082 100644
--- a/src/widget/passwordedit.cpp
+++ b/src/widget/passwordedit.cpp
@@ -42,15 +42,18 @@ void PasswordEdit::registerHandler()
 void PasswordEdit::unregisterHandler()
 {
 #ifdef ENABLE_CAPSLOCK_INDICATOR
-    int idx;
-
-    if (eventHandler && (idx = eventHandler->actions.indexOf(action)) >= 0)
+    if (eventHandler)
     {
-        eventHandler->actions.remove(idx);
-        if (eventHandler->actions.isEmpty())
+        int idx = eventHandler->actions.indexOf(action);
+
+        if (idx >= 0)
         {
-            delete eventHandler;
-            eventHandler = nullptr;
+            eventHandler->actions.remove(idx);
+            if (eventHandler->actions.isEmpty())
+            {
+                delete eventHandler;
+                eventHandler = nullptr;
+            }
         }
     }
 #endif
```

sudden6, line 47: oh, right, then the initial variant is probably the best

PR #3431, opened 2016-06-23
docs(README.md): add links to GPG signatures for windows installer

### LINKED ISSUE #3427: Add PGP signatures and authenticode signing
When obtaining qTox setup, we currently rely on HTTPS for the download from the tox webpage. We have no guarantee that the binary is exactly the binary that the developer uploaded - it can be modified by, for example, a different disgruntled developer who has access to the tox webpage.

The solution to this problem is to include PGP signatures for the setup files, portable package, and source code tarballs. This makes it possible for us to verify the files come from the developer. This solution does not cost any money to implement.

On Windows, the setup files and binaries should additionally be authenticode signed. This provides an additional benefit of Windows SmartScreen filter not interfering with installation or execution of qTox. On older Windows systems, such as Windows 7, it also removes the red/yellow warning screen when installing, instead showing a pleasant blue information screen instead. This solution requires purchase of a signing certificate, which can cost ~70$ per year or less if you look around, and there are also some ways of obtaining it for free if the software is classified as FOSS.

This is important for security focused tools, and to security minded people.

zetok:
Closes #3427.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ running on all major platforms.**
 ----
 
 * **Windows**:
-  - [**64 bit installer**](https://build.tox.chat/view/qtox/job/qTox_pkg_windows_x86-64_stable_release/lastSuccessfulBuild/artifact/setup-qtox.exe)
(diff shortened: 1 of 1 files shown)


tux3:
LGTM

PR #3432, opened 2016-06-23

zetok:
@qqq3 not sure if I got name in settings right..?


qqq3:
@zetok sorry I don't understand you. What do you mean?

zetok:
I was wondering whether that's correct: https://github.com/tux3/qTox/pull/3432/files#diff-cade401ac6b89223354ef2a70f84fe81R71

Issue #2847, opened 2016-01-23
Drag and Drop should work with text input area


LittleVulpix:
Almost everyone who uses qTox initially tries to drag and drop files into the text input area, because it's considered to be the input area for any kind of input. I agree and when I first tried qTox to understand what's going on, I realized drag and drop only works if you drop atop of the chat history, which doesn't make much sense.

tl;dr - drag and drop should work even if you drop the file atop of the text input area. Currently it gives you the "browser" file path, e.g.

file:///C:/Users/poorperson/Desktop/somefile.txt

appears in the chat instead of the transfer being initiated.

zetok:
Kinda a duplicate of #1947.

LittleVulpix:
Interesting, yeah I looked over issues but didn't stumble upon this one. Why does the "paste path" behavior exist at all? You only ever drag and drop a file to someone when you intend to send it.

sudden6:
The behaviour of drag and drop seems to have been changed. Now the whole chat + text input area act the same. Thus I will close this issue.

zetok:
Apparently that works only on Linux – windows still has old behaviour, thus reopening.

```
<LittleVulpix> zetok: Also why the hell does it still only paste the file path when you drag and drop a file into the input box x3 Everyone who ever sends me a file says it is not workin and then I have to explain that you need to drop the file into the "history" i.e. the area where you normally cannot input anything :D
<LittleVulpix> It doesn't work on windows... neither in 1.3 nor in 1.4
```

LittleVulpix:
Example of different behavior depending upon where you drop the file.
![Imgur](http://i.imgur.com/et0sjtp.gif)

Talkless:
I've tested this behaviour on Kubuntu 14.04 (my OS), Debian 8 GNOME, Debian 8 KDE, Kubuntu 15.10, Kubuntu 16.04, Xubuntu 16.10, Ubuntu 16.10  and Windows 10.

This works only on Kubuntu 15.10 and Kubuntu 16.10 for some reason.

> Now the whole chat + text input area act the same

Maybe event does not fall through to that parent widget somehow.

Is anyone working on a fix? Or should I try that?

zetok:
> This works only on Kubuntu 15.10 and Kubuntu 16.10 for some reason.

Most likely Qt version – works for me on 5.5.1

Talkless:
[...]

Issue #3372, opened 2016-06-10
Update qTox Windows version building manual


DX37:
But CMake adding exists in **Setting up PATH**.
https://github.com/tux3/qTox/blob/master/INSTALL.md#setting-up-path

zetok:
> Select mingw-developer-toolkit, mingw32-base, mingw32-gcc-g++, msys-base and mingw32-pthreads-w32

It's not provided with one of those? Or Qt toolchain that is supposed to be optional?

DX37:
> It's not provided with one of those? Or Qt toolchain that is supposed to be optional?

It's not exists in MinGW or Qt basic installation. But I deselect MinGW 4.9.2 in Tools, as described in docs.

DX37:
Well... OpenAL got updated from 1.16.0 to 1.17.2 and FFmpeg from 2.7 to 3.0.2.
Maybe it's time to update Windows building manual? And I notice, what after [Getting dependencies](https://github.com/tux3/qTox/blob/master/INSTALL.md#getting-dependencies) nothing listed what to do after it. Exactly what you should run to build qTox.

DX37:
Found gcc in _Qt for Desktop 5.6 (MinGW 4.9.2)_, WTF.

DX37:
Well... It's getting errors when I try to get and compile dependencies with original manual. Now I try to install MSYS2 and MinGW64.

zetok:
> Well... OpenAL got updated from 1.16.0 to 1.17.2 and FFmpeg from 2.7 to 3.0.2.
> Maybe it's time to update Windows building manual?

Sure. Instructions in there are supposed to be for something that worked. With that in mind, there has to be someone who would write & test newer instructions – note that out of qTox maintainers currently only @tux3 uses windows to test qTox. Dunno though if he even compiles qTox on windows..

If instructions for windows need to be updated, PRs would be very welcome.

DX37:
> If instructions for windows need to be updated, PRs would be very welcome.

Maybe I will do this...
At this moment I downloading Qt 5.6.0 in MSYS2. It's already have compiled openal and ffmpeg in it's repos. I will test that.

DX37:
I give up for now. I've got fully generated dependencies from manual, but I can't compile qTox. `qmake` behavior is strange, `make -j4` fails with error: 

> src/persistence/db/rawdatabase.cpp:15:31: fatal error: sqlcipher/sqlite3.h: No such file or directory

[...]

PR #3433, opened 2016-06-24
fix(chattextedit.cpp): fix drag-and-drop to be consistent across systems

Talkless:
Add explicit setAcceptDrops(false) to allow parent widget to handle
drag-and-drops consistently across various desktop environments.

Closes: #2847 

Before:
before

After:
after

Tested on KDE/Gnome/XFCE/Unity, now they behave the same same (previously, only later KDE accepted drag-and-drops).

Cannot build and test on Windows because of #3372


PKEv:
This is certainly cool but need to add button modifiers allow you to insert a link to the file

PR #3437, opened 2016-06-25
fix(loginscreen.cpp): fix password input focus after mouse click

Talkless:
Add setFocus() call to make password input focus work after user
clicks Load button (hence losing focus) with invalid password.
Without this patch, focus only works when using Enter key to login.


linux-modder:
As an aside to this we might as well also do some auto-clear of  pass on bad pass for completeness

zetok:
> As an aside to this we might as well also do some auto-clear of pass on bad pass for completeness

Would be ~redundant, possibly counter-intuitive and annoying e.g. in case where one just didn't type one last character.

Issue #3404, opened 2016-06-16
change marks for text formatting for more "natural" & familiar ones


zetok:
##### Brief Description

Current text formatting (markdown) is far from being familiar / natural to people who would want to use it.

Thus, it would be good to change it to something that lots of people already are familiar with, and if they're not, they could organically learn about it.

Pointed out & proposed in #3373, to use same marks that Ricin uses.

Thanks to @SkyzohKey:
![what ricin uses](https://i.imgur.com/RxfliqE.png)

cc interested parties from #3373:
@ddobrev, @DX37, @ProMcTagonist, @Chiitoo

Does anyone think that there are good reasons to not switch over?

Chiitoo:
Looks pretty good!

I'm not too keen on ~tilde~ making the strikethrough, but I can likely live with that.  :]

I don't really have a suggestion for any better alternative for that either.  Only things that come in mind are <s> and [s], but those don't quite fit in with the rest.

zetok:
> I'm not too keen on ~tilde~ making the strikethrough, but I can likely live with that.

Yeah, I'm not too happy about it myself, so I was wondering whether it would be acceptable to have only double tilde (i.e. `~~strikethrough~~`) as strikethrough.

Chiitoo:
Methinks I for one would be OK with double tilde.

yurivict:
Please add the markup explanation in a tooltip over the edit area.

ddobrev:
The last suggestion is not a good idea, this tool-tip will get in the way each time someone wants to write a message.

yurivict:
Code is also commonly marked up with triple-backquotes.

sudden6:
I think ~strikethrough~ would be still better than ~~strikethrough~~, since it fit's the style of the rest in a better way.

DX37:
Syntax proposed by @SkyzohKey seems more intuitive than Markdown. I think maybe even doesn't make sense to add an alternative form of syntax. Or vice versa is best to use an alternative syntax instead of the original. But it will be a bit cumbersome.

anoadragon453:
I wrote the code for the markdown interpretation in qTox. I agree it doesn't exactly line up with other applications and have heard complaints about it before. I'm open to changing the syntax as changes are quite easy to make, though I would like a formal discussion on it where most people are in agreement before we make the change. 

[...]

PR #3441, opened 2016-06-26
refactor(markdown): Change markdown syntax to be more intuitive

anoadragon453:
As suggested by #3404, the current markdown syntax is unintuitive, and thus we have agreed on switching to @SkyzohKey's implementation.

### DIFF
--- a/doc/user_manual_en.md
+++ b/doc/user_manual_en.md
@@ -220,15 +220,15 @@ Videochats and file transfers are currently unsupported in groupchats.
 ## Message Styling
 
 Similar to other messaging applications, qTox supports stylized text formatting.
-Formatting follows
-[Markdown syntax](https://daringfireball.net/projects/markdown/syntax), thus:
 
-* For **Bold**, surround text in double asterisks or underscores: `**text**`
-  or `__text__`
-* For **Italics**, surround text in single asterisks or underscores: `*text*`
-  or `_text_`
-* For **Strikethrough**, surround text in single tilde's: `~text~`
-* For **Underline**, surround text in single dashes: `-text-`
+* For **Bold**, surround text in single or double asterisks: `*text*`
+ or `**text**`
+* For **Italics**, surround text in single or double forward slashes: `/text/`
+ or `//text//`
+* For **Strikethrough**, surround text in single or double tilde's: `~text~`
+ or `~~text~~`
+* For **Underline**, surround text in single or double underscores: `_text_`
+ or `__text__`
 * For **Code**, surround your code in in single backticks: `` `text` ``
 
 Additionally, qTox supports three modes of Markdown parsing:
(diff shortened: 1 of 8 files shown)


zetok:
`Closes #3404` could be added to commit message, to automatically close the issue when PR gets merged

anoadragon453:
Huh, did not know that. I'll add it in if it's not merged by the time I get back from work.

anoadragon453:
Fixed errors and removed qDebug stuff.

zetok:
~~Would be good to also update [user manual](https://github.com/tux3/qTox/blob/4f493c5df73dffef25e4f34984fdb736f8da626c/doc/user_manual_en.md#message-styling)~~

Edit: nvm the brainfart

zetok:
@anoadragon453 btw, could you add a [`BREAKING CHANGE:`](https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#body) part to commit message?

doc/user_manual_en.md
```diff
@@ -223,12 +223,14 @@ Similar to other messaging applications, qTox supports stylized text formatting.
 Formatting follows
 [Markdown syntax](https://daringfireball.net/projects/markdown/syntax), thus:
```
zetok: Since it's no longer markdown, those 2 lines should be removed

src/chatlog/chatmessage.cpp
```diff
@@ -186,14 +183,16 @@ QString ChatMessage::detectMarkdown(const QString &str)
     QString out = str;
 
     // Create regex for certain markdown syntax
```
zetok: not a markdown anymore

zetok:
PR could also make some use of removing `Markdown` in settings, and in other places in code.

Though, if rename were to happen ~now, it would conflict with #3449, since it touched strings with `Markdown`.

anoadragon453:
So it seems all translations have left 'Markdown' in English. Search and replacing Markdown with something like Style might break those translations.

anoadragon453:
Added Breaking Change, and all instances of Markdown except in translations have been removed. See earlier comment for reasoning.

zetok:
> `refactor(markdown)`

Not a `refactor`, since it changes what users interact with. More a `feat`.

anoadragon453:
@zetok updated.

zetok:
@anoadragon453 um, now PR doesn't compile?

anoadragon453:
Alright! Fixed: 
- Bug in that _not_ showing formatting characters actually works
- Multi-line code formatting doesn't insert random newlines
- Settings shows 'Show Formatting Characters' by default

Rebased and yes, it compiles :)

anoadragon453:
@zetok Merge?

zetok:

anoadragon453:
@zetok Rebased.

PR #3444, opened 2016-06-26
fix(chatform): Fixed memory leak

Diadlo:
Memory for ScreenshotGrabber was allocated, but don't deallocated

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -67,7 +67,8 @@
 
 ChatForm::ChatForm(Friend* chatFriend)
     : f(chatFriend)
-    , isTyping{false}
+    , screenshotGrabber(nullptr)
+    , isTyping(false)
 {
     Core* core = Core::getInstance();
     coreav = core->getAv();
@@ -208,7 +209,7 @@ void ChatForm::onAttachClicked()
             file.close();
             continue;
         }
-        long long filesize = file.size();
+        qint64 filesize = file.size();
         file.close();
         QFileInfo fi(path);
 
@@ -783,9 +784,14 @@ void ChatForm::onScreenshotClicked()
 
 void ChatForm::doScreenshot()
 {
-    ScreenshotGrabber* screenshotGrabber = new ScreenshotGrabber(this);
-    connect(screenshotGrabber, &ScreenshotGrabber::screenshotTaken, this, &ChatForm::onScreenshotTaken);
+    if (!screenshotGrabber)
+        screenshotGrabber = new ScreenshotGrabber(this);
+
+    connect(screenshotGrabber, &ScreenshotGrabber::screenshotTaken,
+            this, &ChatForm::onScreenshotTaken);
+
     screenshotGrabber->showGrabber();
+
     // Create dir for screenshots
     QDir(Settings::getInstance().getAppDataDirPath()).mkpath("screenshots");
 }
@@ -807,16 +813,21 @@ void ChatForm::onScreenshotTaken(const QPixmap &pixmap) {
         QMessageBox::warning(this,
                              tr("Failed to open temporary file", "Temporary file for screenshot"),
                              tr("qTox wasn't able to save the screenshot"));
+
+        delete screenshotGrabber;
+        screenshotGrabber = nullptr;
         return;
     }
 
     pixmap.save(&file, "PNG");
 
-    long long filesize = file.size();
+    qint64 filesize = file.size();
     file.close();
     QFileInfo fi(file);
 
     emit sendFile(f->getFriendID(), fi.fileName(), fi.filePath(), filesize);
+    delete screenshotGrabber;
+    screenshotGrabber = nullptr;
 }
 
 void ChatForm::onLoadHistory()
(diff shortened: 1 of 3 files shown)


src/widget/form/chatform.cpp
```diff
@@ -783,9 +784,11 @@ void ChatForm::onScreenshotClicked()
 
 void ChatForm::doScreenshot()
 {
-    ScreenshotGrabber* screenshotGrabber = new ScreenshotGrabber(this);
-    connect(screenshotGrabber, &ScreenshotGrabber::screenshotTaken, this, &ChatForm::onScreenshotTaken);
+    screenshotGrabber = new ScreenshotGrabber(this);
+    connect(screenshotGrabber, &ScreenshotGrabber::screenshotTaken,
+            this, &ChatForm::onScreenshotTaken, Qt::QueuedConnection);
```
antis81: Queued connection is automagically taken - it does the right thing for you.

Diadlo: It seems you don't right. Because I tried it with `AutoConnection` and had error:
After deleting screenshotGrabber object, `ScreenshotGrabber::acceptRegion` call `restoreHiddenWindows()` to already deleted object.

antis81: Btw. by using a lambda you can avoid the member. ^^ Other than that - totally perfect.

Diadlo: Should I replace on lambda?

Diadlo: @antis81 I tried with lambda and without QueuedConnection. It didn't work

antis81: Yeah, I think I know what's going on here. I remember that grabber code and it's greatly f***ed up… :smile:

Anyway, strange things happen, because the grabber `window` member executes code after (!) the object got deleted -> memory corruption :boom:. By forcing a queued connection, the slot (event) is executed after the window is closed and leaving the function - but it's a pure coincidence and even f***s up the code more. Would you mind cleaning this up, by copying the rect to another var and move the "emit" to the end inside `ScreenshotGrabber::acceptRegion`?

Diadlo: Yes. It's good idea)

src/widget/form/chatform.cpp
```diff
@@ -784,39 +784,44 @@ void ChatForm::onScreenshotClicked()
...
+        pixmap.save(&file, "PNG");
 
-    if (!file.open(QFile::ReadWrite))
-    {
-        QMessageBox::warning(this,
-                             tr("Failed to open temporary file", "Temporary file for screenshot"),
-                             tr("qTox wasn't able to save the screenshot"));
-        return;
-    }
+        long long filesize = file.size();
```
antis81: What about using `qint64` type as supposed to be… :)

src/widget/form/chatform.cpp
```diff
@@ -784,39 +784,44 @@ void ChatForm::onScreenshotClicked()
...
+        file.close();
+        QFileInfo fi(file);
 
-    pixmap.save(&file, "PNG");
+        emit sendFile(f->getFriendID(), fi.fileName(), fi.filePath(), filesize);
+        delete screenshotGrabber;
```
antis81, line 817: In this case it helps to also set to `nullptr` after the delete.

Diadlo, line 817: Why? screenshotGrabber now is local variable of `ChatForm::doScreenshot`

src/widget/form/chatform.cpp
Diadlo: Fixed

src/widget/form/chatform.cpp
antis81, line 817: Yes, but it's added as child object (otherwise could become a dangling pointer). Thus, exiting qTox from within screenshot grabber actually leads to SIGSEGV. ^^

src/widget/form/chatform.cpp
antis81: Obviously not. Something wrong with commit?

src/widget/form/chatform.cpp
Diadlo, line 817: Ok. Fixed

src/widget/form/chatform.cpp
antis81: Ah now I get it, it's another occurence. Could you please change type here, too?

Diadlo: I fixed it in wrong place :) 
Now fixed both

antis81:
If it compiles -> ready from my side! LGTM!

sudden6:
@Diadlo Segfaults for me

https://gist.github.com/sudden6/e6729d19dae625ddf2562507fa6d445b

Steps to reproduce:
- try sending screenshot to currently offline contact (not sure if other contact beeing offline is relevant)
- as soon as [Enter] is pressed -> Segfault

Additionally this is the first time I compiled qTox with GCC 6.1, I will test further if that's relevant.

Diadlo:
@sudden6 Fixed

PR #3445, opened 2016-06-26
fix(main): Closing file before removing

### LINKED ISSUE #3440: Log file error "Critical: Unable to move logs"
OS: Windows 7
qTox version: (version numbers in menu `Settings → About`)  
Commit hash:  8d225a8c97a36838092b325ca830c14546f3a0c6 and newer
toxcore:  
Qt:  5.4.2
Hardware:  
…
https://gist.github.com/towlie/9419eea500e99bb95a4a9804d89dd880
Logging itself still works.

Diadlo:
Fix #3440

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -87,22 +87,19 @@ void logMessageHandler(QtMsgType type, const QMessageLogContext& ctxt, const QSt
     if (!logFilePtr)
     {
         logBufferMutex->lock();
-        if(logBuffer)
-        {
+        if (logBuffer)
             logBuffer->append(LogMsgBytes);
-        }
+
         logBufferMutex->unlock();
     }
     else
     {
         logBufferMutex->lock();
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -167,9 +167,13 @@ int main(int argc, char *argv[])
     if (QFileInfo(logfile).size() > 1000000)
     {
         qDebug() << "Log file over 1MB, rotating...";
-		
+
+        // close old logfile (need for windows)
+        if(mainLogFilePtr)
```
zetok: space between `if` and `(` ?

Diadlo: It's just copy-paste. Should I replace all `if(` in this function?

zetok: You could :)

Diadlo: Fixed

PR #3446, opened 2016-06-27
fix(addfriendform): Fixed problem with reading friend request

### LINKED ISSUE #3421: Notification about new friend request doesn't disappear after second friend request
##### Brief Description

When getting 2nd friend request during the same qTox "session", friend request notification doesn't disappear when clicked on it.

OS: Gentoo Linux amd64
qTox version: v1.4.0-159-ga851a5b
Commit hash: a851a5b18da8c295fba441c817920cf65db2cdbb
toxcore: 0.0.0
Qt: 5.5.1

Reproducible: Always
##### Steps to reproduce
1. Receive and accept friend request, clicking on green FR notification 
2. Repeat step 1.
3. Notification about 2nd friend request doesn't disappear
##### Observed Behavior

Green notification about friend request doesn't disappear, even though there's no FR.
##### Expected Behavior

Notification about new FR should disappear when clicked on it.
##### Additional Info

tfw no fr

(I thought that there was an issue for it already.. guess not.)

Diadlo:
Fix #3421

### DIFF
--- a/src/widget/form/addfriendform.cpp
+++ b/src/widget/form/addfriendform.cpp
@@ -113,6 +113,11 @@ void AddFriendForm::show(ContentLayout* contentLayout)
     head->show();
     setIdFromClipboard();
     toxId.setFocus();
+
+    // Fix #3421
+    // Needed to update tab after opening window
+    int index = tabWidget->currentIndex();
+    onCurrentChanged(index);
 }
 
 QString AddFriendForm::getMessage() const


src/widget/form/addfriendform.cpp
```diff
@@ -113,6 +113,8 @@ void AddFriendForm::show(ContentLayout* contentLayout)
     head->show();
     setIdFromClipboard();
     toxId.setFocus();
+    int index = tabWidget->currentIndex();
```
zetok, line 119: perhaps comment could be added why that part is needed?

zetok, line 119: I.e. that it fixes $bug

Diadlo, line 119: Added :)

PR #3449, opened 2016-06-28
feat(chat): add settings to alter the chat view's base font

### LINKED ISSUE #672: Add GUI option to change chat font size

antis81:
Closes #672
- [x] Introduce `Chat/chatMessageFont` settings value
  -  Defaults to GUI style font according to previous behavior.
- [x] Add Settings UI
- [x] Cleanup (chat) settings UI
- [x] Set the font censequently as the chat's base font.
  - Author, message and timestamp columns are based on a single font now. I can separate those to set a font on each column, but actually I think it is totally fine as is (See screenshots below).
# Settings UI:

qtox-chat-base-font-settings-ui

**Note:** @zetok I moved both text format rows to the top of the group box and shorten the markdown label to something like "Markdown format" (the gap makes the dialog look like a mixed fruit salad :smile:)
# Settings UI after cleanup:

qtox-chat-settings-ui
# Example 1 (from WIP interims-state) for chat view with "Comic Sans MS" font (size: 18px):

qtox-chat-base-font
# Example 2 (after) -> consistent appearance with same base font!

qtox-chat-base-font-complete

### DIFF
--- a/src/chatlog/content/text.cpp
+++ b/src/chatlog/content/text.cpp
@@ -31,10 +31,13 @@
 #include <QDesktopServices>
 #include <QTextFragment>
 
-Text::Text(const QString& txt, QFont font, bool enableElide, const QString &rwText, const QColor c)
+Text::Text(const QString& txt, const QFont& font, bool enableElide, const QString &rwText, const QColor c)
     : rawText(rwText)
     , elide(enableElide)
     , defFont(font)
+    , defStyleSheet(QString::fromUtf8("body{font: '%1' %2px %3;}")
+                    .arg(font.family()).arg(font.pixelSize())
+                    .arg(font.bold() ? "bold" : QString()))
     , color(c)
 {
     setText(txt);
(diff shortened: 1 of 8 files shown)


src/widget/form/settings/generalform.cpp
```diff
@@ -109,6 +110,8 @@ GeneralForm::GeneralForm(SettingsWidget *myParent) :
     bodyUI = new Ui::GeneralSettings;
     bodyUI->setupUi(this);
 
+    Settings& s = Settings::getInstance();
```
sudden6, line 113: would be neat to make the choice between `s` or `Settings::getInstance()` consistent throughout the method.

src/chatlog/content/text.cpp
```diff
@@ -247,9 +250,12 @@ void Text::regenerate()
         doc->setDefaultFont(defFont);
 
         if (!elide)
+        {
+            doc->setDefaultStyleSheet(defStyleSheet);
             doc->setHtml(text);
-        else
+        } else {
```
zetok: style should be 

``` c++
}
else
{
```

src/widget/form/settings/generalform.cpp
antis81, line 113: Done.

antis81:
@zetok It's ready!

PR #3456, opened 2016-07-01
Code cleanup

Diadlo:


src/core/coreav.cpp
```diff
@@ -113,7 +113,8 @@ bool CoreAV::anyActiveCalls()
 
 bool CoreAV::isCallVideoEnabled(uint32_t friendNum)
 {
-    assert(calls.contains(friendNum));
+    bool contains = calls.contains(friendNum);
+    assert(contains);
     return calls[friendNum].videoEnabled;
```
tux3, line 117: Why create a variable that is never reused?
Additionally, this makes the contains() call run in NDEBUG builds for no reason.

src/widget/loginscreen.cpp
```diff
@@ -40,7 +40,7 @@ LoginScreen::LoginScreen(QWidget *parent) :
     ui->setupUi(this);
 
     // permanently disables maximize button https://github.com/tux3/qTox/issues/1973
-    this->setWindowFlags(windowFlags() &! Qt::WindowMaximizeButtonHint);
+    this->setWindowFlags(windowFlags() ^ Qt::WindowMaximizeButtonHint);
```
tux3: ... uh, it looks like the previous code was wrong for using logical not instead of bitwise not, and the new code seems wrong because it will toggle the flag instead of disabling it.
What's the intent here? Shouldn't this be `& ~`?

src/core/coreav.cpp
Diadlo, line 117: `contains` didn't make `const` method. What if someone change realization, which will change object?

tux3, line 117: The solution is to make `contains` const then, not create variables. Semantically, a predicate called `contains` MUST be const or the code is insane.

src/widget/loginscreen.cpp
Diadlo: `permanently disables maximize button`
It should be disabled :)

Diadlo: You are right

Diadlo: Fixed

src/core/coreav.cpp
Diadlo, line 117: Fixed

src/audio/audio.cpp
```diff
@@ -558,7 +558,7 @@ void Audio::doCapture()
 {
     QMutexLocker lock(&audioLock);
 
-    if (!alInDev || !inSubscriptions)
+    if (!isInputReady())
```
antis81: Caution: This will cause a deadlock! There's a reason I privatized some things. Anyway, you could write an `Audio::Private::isInputReady` method.

src/core/toxcall.h
```diff
@@ -24,9 +24,9 @@ struct ToxCall
      ToxCall(const ToxCall& other) = delete;
      ToxCall(ToxCall&& other) noexcept;
 
-     inline operator int() {return callId;}
-     const ToxCall& operator=(const ToxCall& other) = delete;
-     const ToxCall& operator=(ToxCall&& other) noexcept;
+     inline operator int() const {return callId;}
```
antis81: Please keep style and expand it to the full 4 lines.

src/core/toxid.h
```diff
@@ -41,9 +41,9 @@ class ToxId
 public:
     ToxId(); ///< The default constructor. Creates an empty Tox ID.
     ToxId(const ToxId& other); ///< The copy constructor.
-    ToxId(const QString& id); ///< Create a Tox ID from QString.
-                              /// If the given id is not a valid Tox ID, then:
-                              /// publicKey == id and noSpam == "" == checkSum.
+    explicit ToxId(const QString& id); ///< Create a Tox ID from QString.
+                                       /// If the given id is not a valid Tox ID, then:
+                                       /// publicKey == id and noSpam == "" == checkSum.
```
antis81: Would you move that old indented comments **above** the function declaration, please? Optimum would be, to place it above function definition in .cpp file.

antis81: Btw.: Range selection is your friend -> `Shift+Alt+Mouse-Drag` :)

src/persistence/settings.h
```diff
@@ -308,10 +309,12 @@ public slots:
     void setAutoLogin(bool state);
 
     int getCircleCount() const;
+    void setCircleName(int id, const QString &name);
+
+    QString getCircleName(int id) const;
```
antis81: Getter/Setter alignment doesn't match.

src/persistence/settingsserializer.cpp
```diff
@@ -412,11 +411,9 @@ void SettingsSerializer::readIni()
         // Add all keys
         if (!s.group().isEmpty())
             beginGroup(s.group());
+
         for (QString k : s.childKeys())
-        {
             setValue(k, s.value(k));
-            //qDebug() << "Read key "<<k<<" in group "<<group<<":\""<<groups[group]<<"\"";
-        }
```
antis81, line 419: Keep those brackets. Removing them is potentially "dangerous".

src/platform/install_osx.cpp
```diff
@@ -98,7 +98,7 @@ void osx::migrateProfiles()
     if (checkDir.exists() && checkDir.isDir())
     {
         qDebug() << "OS X: Old settings directory detected migrating to default";
-        if( !dir.rename(oldPath, newPath) )
+        if ( !dir.rename(oldPath, newPath) )
```
antis81: Remove spaces.

src/widget/form/genericchatform.cpp
```diff
@@ -466,21 +464,17 @@ void GenericChatForm::onSelectAllClicked()
...
+QString GenericChatForm::resolveToxId(const ToxId &id) const
 {
     Friend *f = FriendList::findFriend(id);
     if (f)
-    {
         return f->getDisplayedName();
-    }
-    else
+
+    for (auto it : GroupList::getAllGroups())
```
antis81: Probably should become `for (const auto& group : GroupList::getAllGroups())`. Btw. `auto` is actually a copy of a group, not an iterator as the variable name suggests.

**EDIT:** Single ampersand (&) should be enough.

src/widget/friendwidget.h
```diff
@@ -47,7 +47,7 @@ class FriendWidget : public GenericChatroomWidget
 public slots:
     void onAvatarChange(int FriendId, const QPixmap& pic);
     void onAvatarRemoved(int FriendId);
-    void setAlias(const QString& alias);
+    void setAlias(const QString& alias) const;
```
antis81: Setters are never const!

src/widget/genericchatroomwidget.h
```diff
@@ -43,12 +43,12 @@ class GenericChatroomWidget : public GenericChatItemWidget
     virtual void setChatForm(ContentLayout* contentLayout) = 0;
     virtual void resetEventFlags() = 0;
     virtual QString getStatusString() const = 0;
-    virtual Friend* getFriend() const{return nullptr;}
-    virtual Group* getGroup() const{return nullptr;}
+    virtual Friend* getFriend() const {return nullptr;}
+    virtual Group* getGroup() const {return nullptr;}
```
antis81: Please expand methods to lines.

src/widget/groupwidget.cpp
```diff
@@ -211,7 +211,7 @@ Group* GroupWidget::getGroup() const
 
 bool GroupWidget::chatFormIsSet(bool focus) const
 {
-    (void)focus;
+    Q_UNUSED(focus);
```
antis81: :+1:

src/widget/loginscreen.cpp
```diff
@@ -40,7 +40,7 @@ LoginScreen::LoginScreen(QWidget *parent) :
     ui->setupUi(this);
 
     // permanently disables maximize button https://github.com/tux3/qTox/issues/1973
-    this->setWindowFlags(windowFlags() &! Qt::WindowMaximizeButtonHint);
+    this->setWindowFlags(windowFlags() &~ Qt::WindowMaximizeButtonHint);
```
antis81: This was actually right before. An XOR would enable the `Qt::WindowMaximizeButtonHint` flag and thus declines the description above. Please revert

antis81:
@Diadlo Thanks for helping with code cleanup! Doing such things is just calling for conflicts and I tend to break it up into several "topic related" PR's and usually avoid producing a complex code mesh. This is one reason for people to leave the project in the past, as this can quickly become very time consuming!

AFAIK such a situation can only be solved (or even completely avoided) by declaring a "feature freeze" date and use it for cleanup. Do you think this makes sense? ( CC: @zetok @tux3 )

src/core/toxid.h
Diadlo: What about change all `///<` comments on blocks?

```
/**
 * @bref
 */
```

src/persistence/settingsserializer.cpp
Diadlo, line 419: Why? Our coding style speaks, that `if / while / for / switch` with one line should be without brackets (https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#coding-style)

src/core/toxid.h
antis81: Would be great! If you're at it, would you also move the blocks to CPP? In general, there shouldn't be comments in header files. This helps to create a quite usable documentation (doxygen/javadoc)! This in turn helps newcomers to better understand API and the program itself.

src/widget/form/genericchatform.cpp
Diadlo: `it` is pointer (`getAllGroups()` return list of pointers). Why we should make it reference?

src/core/toxid.h
antis81: Ah, almost forget. You'd want to leave out the "leading stars" -> they can cause trouble and it is consistent to leave them out. This is btw. how Qt's very documentation is written!

src/widget/friendwidget.h
Diadlo: This setter real don't change nothing in the class

Diadlo:
@antis81 Should I break up into several PR? Should I break up by type of change? Or by directories?

src/persistence/settingsserializer.cpp
antis81, line 419: Yeah, and especially on loops, this easily leads to irritations and error prone code. I prioritize avoidance of confusion rather than following a dumb style guide! Please @tux3, @zetok: Can we change (actually correct) the "bracket rules" for loops please and make brackets mandatory there? It's fine with "if" statements, yet I tend to put brackets there as well. The problem is, after you followed lots of (different) style guides, you finally learn, that you don't need one. Fortunately that "tab against white space" war is finally over! :)

Btw. @Diadlo A single line `switch` statement doesn't make any sense. So brackets go there "automagically"… :)

src/widget/form/genericchatform.cpp
antis81: Ah sorry. It's fine then!

src/widget/loginscreen.cpp
Diadlo: But now it not XOR. Before was used logical NOT, instead of bitwise

src/widget/friendwidget.h
antis81: Ok, then the question is: If it has nothing to do with the surrounding class context, why does this slot even exist? :)

src/persistence/settingsserializer.cpp
Diadlo, line 419: I agree, that brackets should be everywhere (for `if / while / for` with one line too). I will happy if maintainers change this rule :)

src/widget/loginscreen.cpp
antis81: Well, no! Writing it a bit differently, it probably becomes more clear, that actually the `operator!` is used - which is very important here. Yet both the NOT and XOR operators relate to the flag and it probably has the same effect. :).

Just write it different:

``` c++
this->setWindowFlags(windowFlags() & !Qt::WindowMaximizeButtonHint);
```

All clear? :P

src/widget/friendwidget.h
Diadlo: It use `friendId` of this class :)

src/widget/loginscreen.cpp
Diadlo: You use operator! to number. It's wrong

src/core/toxid.h
Diadlo: I think, comments should be in .h file, because if somebody will use this method he or she shouldn't open implementation to know what this method does. Or I'm wrong?
Why "leading stars" can cause trouble?

antis81:
> Should I break up into several PR? Should I break up by type of change? Or by directories?

First, it's a good idea to split commits by topic (e.g. "audio, video, core, style, …"). Second it's a (very) good idea to differ between "API" and "UI" and put widget related changes into separate commits. Now the last (optional) step is, to split topics up to separate branches. Anyway, I'm not always loyal to my own rules, if they are "oversized" and things work as they are. But it's a rather strong "rule of thumb", which helps to work with the team and actually creates an essential basis to handle conflicts, by avoiding them most of the time.

src/widget/loginscreen.cpp
Diadlo: Are you agree?

antis81: Ah, damn, got me. You're right! So yes, I agree in this case, as the `!` "converts" to `bool`, while the `~` returns the bitwise negation.

antis81: Anyway, would you change writing -> it's just more clear!

zetok:
@antis81 

> AFAIK such a situation can only be solved (or even completely avoided) by declaring a "feature freeze" date and use it for cleanup. Do you think this makes sense?

There are projects that successfully use this model, so it might make sense to use it.

On the other hand, there's e.g. Rust, which is also massive and fast-paced, but AFAICS it [doesn't have stuff like freezes](https://internals.rust-lang.org/t/release-channels-git-branching-and-the-release-process/1940).

Both of those models work.

And I think that both of them wouldn't work for qTox right now, since that's not the actual problem.
I.e. either of those models would require lots of work on the maintainers' side, something which cannot be right now provided by qTox for people sending PRs.

> This is one reason for people to leave the project in the past, as this can quickly become very time consuming!

If someone doesn't really know for what they're signing up for when making a PR, they can feel disenchanted, and possibly think that their work is not being appreciated when their PR is clearly(!)[1] ready, is not being merged, but instead lays waiting, until there are conflicts caused by other PRs.

This can be fixed by informing people upfront what sending PR will constitute of, and the they will be required to rebase their PR if they don't manage to fix code in PR before there are conflicts.

qTox has problems in this area – there's not enough of maintainers active enough to be able to review & merge in timely manner PRs that are ready for merge, which leads to a situation where only a subset of PRs are merged, causing conflicts for other, waiting for a long time PRs, situation that could be avoided if there was enough of review done in timely manner, so that PR submitter would know whether there are needed some changes to PR needed. PRs also should be reviewed by anyone who knows how to code and is interested in qTox, but that requires advertising & encouraging people to do so.

With review & merge done in timely manner, and clear expectation[2] from PR submitters that they might need to rebase if they don't manage to fix stuff in timely manner for merge without conflicts there's only very little friction.

I kinda have tried to push things in this direction → flags `PR-changes-needed` and `PR-rebase-required`, which are supposed to tell PR submitter whether there is stuff that needs to be changed in PR in order for it to be merged. I wonder if that works though.

I.e. people seemed to be rather reluctant in using those labels to mark which PRs are in need of review/merge/changes. PRs without labels are in need of review/merge, while ones with label require something that PR submitter is required to do.

It would seem to me that "freeze" model encourages lots of breaking changes at the same time, which are simply bad. And no, that part is not avoidable, given that "freeze" constitutes changing lots of stuff in short amount of time, which is going to lead to problems.

On the other hand, "rust model" is more in spirit of small, incremental changes that need to work in order to be accepted, and seems to have smaller potential for breaking things.

In any case, it would be good to pick something that works, but the general attitude of people when it comes to adopting any (new?) workflow at all seems to be 'nah'. 

Does pilling up lots of changes + 'nah' attitude sound like something that will work well? ← I really wonder about that.

# 

> First, it's a good idea to split commits by topic (e.g. "audio, video, core, style, …"). Second it's a (very) good idea to differ between "API" and "UI" and put widget related changes into separate commits. Now the last (optional) step is, to split topics up to separate branches. Anyway, I'm not always loyal to my own rules, if they are "oversized" and things work as they are. But it's a rather strong "rule of thumb", which helps to work with the team and actually creates an essential basis to handle conflicts, by avoiding them most of the time. 

And that sounds like something that should be added to CONTRIBUTING.md ;)

---

[1] from their PoV ;)
[2] that is, it needs to be mentioned _everywhere_, so that people would actually read it at some point _before_ they make a huge PR, won't be able to fix it in timely manner, and end up with huge amount of code that they need to rebase, since other merged PRs caused conflicts

Diadlo:
I don't fully understand.
I should break up this PR on 3-4 PR or more (`Audio style fixes`, `Core style fixes`)?
Or just break up commits in this PR?

antis81:
@zetok 

> I kinda have tried to push things in this direction → flags PR-changes-needed and PR-rebase-required, which are supposed to tell PR submitter whether there is stuff that needs to be changed in PR in order for it to be merged. I wonder if that works though.

It works and is a good direction! Setting that flag informs people via mail: "You won the conflict of the week award! Please clean up." Yet the problem is, people "forget" to remove it by themselves, after the conflict(s) were solved. And to my knowledge, this is not possilble due to Github's permssions, right? So… maybe they can get informed by another (better) mechanism?

> And that sounds like something that should be added to CONTRIBUTING.md ;)

Agree'd! Moment… we should probably shortly discuss that on qTox better than here… just firing it up… :)

antis81:
@Diadlo Ok, I'll make it short.

> Or just break up commits in this PR?

Step 1: Break up into smaller, topic related commits (what you already partly did!)
Step 2 (optional): Break up topic related commits into PR's

Btw. this strategy is specific and only required when working with stuff, that spans several (or all) topics, like this PR! Usually a topic is a small unit.

sudden6:
@Diadlo since you submitted the smaller PRs can we close this one?

Diadlo:
Yes

PR #3462, opened 2016-07-03
docs(CONTRIBUTING.md): Added a new commit message formatting type for translations.

ghost:
It provides better simplicity and avoids confusions over commits related to translations of the GUI.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,6 +91,7 @@ Must be one of the following:
 * **test**: Adding missing tests
 * **chore**: Changes to the build process or auxiliary tools and libraries such
   as documentation generation
+* **tran**: Changes to the translation of the GUI
 
 ##### Revert
 If the commit reverts a previous commit, it should begin with `revert: `,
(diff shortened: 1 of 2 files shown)


zetok:
Err, simple way is to use `feat(l10n): {update,add} translation xx`. Not sure what you mean by "confusions" ?

Aside from that, PR doesn't add custom section for clog: https://github.com/clog-tool/clog-cli#custom-sections

ghost:
Well feat is described as "A new feature" which in my opinion is more related to functionality. I think, that it would be better if we separate some things more in order to have a clearer understanding of what a commit is supposed to do. feat(l10n) works too, but isn't it easier just to separate the translations as a new type.

zetok:
Translations are functionality. For people who don't know English or don't feel comfortable using it, without translations software is close to being useless.

With that being said, separate section might do.

Wouldn't `trans` be better than `tran`? Or `transl`?

ghost:
I just thought that `trans` sounded a bit weird and that `transl` is a bit long. Yes the software is useless without text output and it is part of the "functionality", but I just think it would make things a bit simpler. 

> PR doesn't add custom section

Ops...

PR #3464, opened 2016-07-03
feat(chatform): Disable call buttons if friend is offline

### LINKED ISSUE #556: Button should turn gray when unusable
Currently, when a friend is offline, all button in qTox remain green, making it hard to spot the user is not actually online.

anoadragon453:
Call buttons would be enabled but non-functional if you were looking at a friend who was offline. We now check if the friend is offline and disable the call buttons if so. 

Closes #556

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -114,6 +114,7 @@ ChatForm::ChatForm(Friend* chatFriend)
     connect(msgEdit, &ChatTextEdit::enterPressed, this, &ChatForm::onSendTriggered);
     connect(msgEdit, &ChatTextEdit::textChanged, this, &ChatForm::onTextEditChanged);
     connect(core, &Core::fileSendFailed, this, &ChatForm::onFileSendFailed);
+    connect(core, &Core::friendStatusChanged, this, &ChatForm::onFriendStatusChanged);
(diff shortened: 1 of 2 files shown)


tux3:
That's a good idea, but currently if a friend comes online the buttons would stay disabled until the chat form was reloaded. 
Do you think you could wire up the friend online/offline events to the enable/disable button slots :) ?

anoadragon453:
Yes I was noticing that bug as well. Would it be better to add it
specifically for that functionality or create an onFriendStatusChange()
method?

On 07/03/2016 03:01 AM, Tux3 wrote:

> That's a good idea, but currently if a friend comes online the button
> would stay disabled until the chat form was reloaded.
> Do you think you could wire up the friend online/offline events to the
> enable/disable button slots :) ?
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> https://github.com/tux3/qTox/pull/3464#issuecomment-230145331, or mute
> the thread
> https://github.com/notifications/unsubscribe/ABR7mKXqvpCGXtZNUBZuUktVkFvovBVGks5qR4h4gaJpZM4JD05O.

## 

Andrew

PGP Fingerprint:
C5D9 804C 6FA0 AA57 7441 0641 174B EAB0 09FD 176D

tux3:
onFriendStatusChange makes more sense to me, but either could work

anoadragon453:
@tux3 onFriendStatusChanged method added, functionality works as expected where user status is checked on changed.

PR #3465, opened 2016-07-03
feat(l10n): update Bulgarian translation from weblate

ghost:


zetok:
This needs to be cleaned up using `lupdate` (with proper options, look at wiki for instructions) and all updates should be pulled from weblate in oder to keep things working well.

(other translations from weblate also need to be cleaned up after pulling them in)

ghost:
Done.

zetok:
And other translations?

PR #3475, opened 2016-07-04
fix(groupinviteform): consider dateTime format in group invites

### LINKED ISSUE #3058: Group invites page doesn't consider date format settings
**Observed Behavior:**  

The date is always formatted like this in the group invites page:
2016-03-26-102429_1366x768_scrot

It doesn't consider the general settings in your Tox profile:
2016-03-26-112106_1366x768_scrot

**Expected Behavior:**

The date is formatted like specified in your settings

a68366:
Now:
after

Closes #3058

### DIFF
--- a/src/widget/form/groupinviteform.cpp
+++ b/src/widget/form/groupinviteform.cpp
@@ -29,6 +29,7 @@
 #include <QLabel>
 #include <QWindow>
 #include "ui_mainwindow.h"
+#include "src/persistence/settings.h"
 #include "src/widget/tool/croppinglabel.h"
 #include "src/widget/translator.h"
 #include "src/nexus.h"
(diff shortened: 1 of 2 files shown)


sudden6:
This will probably not update the invite message if the date format is changed.

a68366:
@sudden6 Yes, it won't. Only new invites will have the new format.
Actually, I don't know how to make it updating.
Probably call retranslateUi() somehow. But there is currently no way to know when the format changes.

zetok:
> This will probably not update the invite message if the date format is changed.

Isn't that the same for other stuff that has configurable date/time format?

There's lots of settings that don't really fully take effect until restart.

sudden6:
@a68366 since the groupinviteform is already hooked up to the translator, it would be sufficient to call `Translator::translate()` from the `onDateFormatSelected()` in the generalsettingsform.

No idea if that will help with all date/time issues.

a68366:
@sudden6 Yeah, it will help with group invites log (retranslation should be called in `onTimestampSelected()` too).
Other date/time indicators(e.g. in chatlog, couldn't find them anywhere else) will remain unchanged.
BTW, full client retranslation to just change date format doesn't look good to me.

sudden6:
@a68366 probably yes, I just did a quick skim over the code.

It would still improve things a little bit. Why do you think it's bad? We could probably expand the translator, but I don't think it's worth the work.

PR #3477, opened 2016-07-05
fix(qtox.pro): don't depend on GTK in order to build on Linux

zetok:


.travis/build-ubuntu_14_04.sh
```diff
@@ -117,6 +117,8 @@ cd ..
 
 $CC --version
 $CXX --version
+
+export PKG_CONFIG_PATH="$PWD/libs/lib/pkgconfig"
```
antis81, line 122: ~~This should be the script dir shouldn't it?~~

**edit:** Nevermind, I was on the wrong path. All fine.

zetok:
cc @abbat , @tux3 

This might be a change that can break some build(s), if you build against ffmpeg libs located in `libs/` – to "fix" it, just export [`PKG_CONFIG_PATH`](https://github.com/tux3/qTox/pull/3477/files#diff-e132a75f99cd5d83bdfb317eb6fe0ef8R121)

PR #2771, opened 2016-01-03
Desktop Notifications Again

albel727:
This is the same as PR #2082, but rebased to latest master branch with appropriate adaptations. Original commits by @TheSpiritXIII are squashed to ease reviewing, as there's a lot of intermediate work that got redone later. But a split into two commits
1. Adding notification framework and GUI settings
2. Adding snorenotify backend
can be considered. As well as adding simple backends not relying on snorenotify.

I wish to avoid the same bikeshed that plagued the original PR. Notification popups is a very useful, if not to say critical, functionality, that in my opinion should've been introduced much earlier. Let's focus on critical bugs and regressions. Feature requests like "a separate option for group notifications" and non-critical bugs like "notifications not always/for everything shown" can be addressed later, and should not obstruct merging this PR, as seeing the previous attempt I'm quite unenthusiastic about promoting it as it is.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/2771)

PR #3485, opened 2016-07-06
chore(bootstrap.sh): account for spaces in paths

zetok:
@ddobrev mentioned in #2771 that because of bootstrap not handling
spaces in paths he wasn't able to build qTox, hopefully this fixes it.
# 

@ddobrev you might want to test whether that would work for you with `./test-pr.sh <pr number>`

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3485)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -29,13 +29,13 @@ fi
 
 ################ parameters ################
 # directory where the script is located
-SCRIPT_DIR=$( cd $(dirname $0); pwd -P)
+SCRIPT_DIR="$( cd $(dirname $0); pwd -P)"
 
 # directory where dependencies will be installed
 INSTALL_DIR=libs
 
 # just for convenience
-BASE_DIR=${SCRIPT_DIR}/${INSTALL_DIR}
+BASE_DIR="${SCRIPT_DIR}/${INSTALL_DIR}"
 
 # directory names of cloned repositories
 TOX_CORE_DIR=libtoxcore-latest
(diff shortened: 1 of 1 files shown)


sudden6:
still fails with spaces in build directory path

https://gist.github.com/sudden6/327d5435609ee251298f6afb9b100e9b

line 5 is interesting because I think this indicates bad whitespace handling in libtoxcore Makefile. I'm not sure if this is fixable on our side at all.

zetok:
> I'm not sure if this is fixable on our side at all.

It's not.

bootstrap.sh
```diff
@@ -29,13 +29,13 @@ fi
 
 ################ parameters ################
 # directory where the script is located
-SCRIPT_DIR=$( cd $(dirname $0); pwd -P)
+SCRIPT_DIR="$( cd $(dirname $0); pwd -P)"
```
antis81, line 32: The quotes have no effect here. They are parsed by the script interpreter and will not wrap the resulting path in quotes. Try with:

``` bash
SCRIPT_DIR="\"$( cd $(dirname $0); pwd -P)\""
```

zetok, line 32: but path shouldn't need to be wrapped in quotes?

I.e. from what I understand, quotes are parsed away, and `SCRIPT_DIR` gets all ths stuff from `$()`, and then later on all the contents of `SCRIPT_DIR` are being passed as a single argument to other stuff (since `SCRIPT_DIR` is enclosed in quotes whenever it is used.

zetok:
Stale, closing.

PR #3488, opened 2016-07-07
style(toxid, widgets): Made constructor explicit

Diadlo:


src/core/toxid.h
```diff
@@ -41,9 +41,9 @@ class ToxId
 public:
     ToxId(); ///< The default constructor. Creates an empty Tox ID.
     ToxId(const ToxId& other); ///< The copy constructor.
-    ToxId(const QString& id); ///< Create a Tox ID from QString.
-                              /// If the given id is not a valid Tox ID, then:
-                              /// publicKey == id and noSpam == "" == checkSum.
+    explicit ToxId(const QString& id); ///< Create a Tox ID from QString.
+                                       /// If the given id is not a valid Tox ID, then:
+                                       /// publicKey == id and noSpam == "" == checkSum.
```
antis81, line 46: Please move docs to definition in cpp.

Diadlo, line 46: Why? I think, docs should be in .h file

antis81, line 46: Nah, it shouldn't. Should be at the right place in cpp file. Up to now qTox gives a sh*\* about docs, but if you want to help improve that situation, please move that to cpp. Not only, but especially in this case, Qt shows how it is done right! They will even reject a commit, if undocumented. We didn't reach a point yet where we need to be this restrictive, but I personally strive for the best. :) Btw. didn't you want to put it in `/**…*/` blocks?

src/widget/form/tabcompleter.h
```diff
@@ -45,7 +45,7 @@ public slots:
 
 private:
     struct SortableString {
-        inline SortableString(const QString &n) : contents{n} {}
+        inline explicit SortableString(const QString &n) : contents{n} {}
```
antis81: inline constructor?

src/core/toxid.h
Diadlo, line 46: Yes. I want to change docs. But think, that it should be part of another PR (do you agree?).
Why docs should be in cpp? If I use "Go to definition", I want see docs. But I can't

antis81, line 46: Because docs can be placed always in cpp, but not in header. And thus, changing docs only happens in cpp files and keep lean code changes. And that's only one reason.

And no, you don't want to read the docs in the function (header), but on an API documentation. :)

antis81, line 46: But, if you want to do it in another PR, that's fine. Please leave it then.

Diadlo, line 46: In real, I want (because it's faster)
Until qtox hasn't any API docs moving docs to cpp made work horrible IMHO.
If you still think, that it should be moved to cpp, I do it in next PR, Ok?

antis81, line 46: Yeah that's fine. Thanks!

Diadlo, line 46: > Qt shows how it is done right

C++ std lib show `how it is done right` (they write docs comments in .h file)

antis81, line 46: Do you know what the "Template" in STL stands for? Probably STL consist almost solely on header files… :)

Diadlo, line 46: Why docs should be in cpp?

sudden6, line 46: If it would be a public API for a lib comments in the header would make sense, but these are all internal files, so if I change some functionality I find it more convenient if the doc what it should do is right above the function.

Just my 2 cents.

antis81, line 46: Btw. through the discussion I learned, that Qt also ships the STL and other 3rd party (offline) docs. Didn't know that. You might want to have a look over to https://wiki.qt.io/Qt_Creator_Documentation_Gallery

There's several advantages to keeping the documentation near the function definition:
1. The biggest reason is, that every header can be documented in cpp, while this is not true vice versa. This allows to keep documentation in one place, keeping it more clear.
2. Changes to documentation can never change the header, which in turn doesn't have to be re-shipped.
3. Internal API documentation cannot become public by accident. Even if there's not really an internal documentation existing in qTox, this makes it future proof.
4. One can quickly check, if the function actually does, what the docs say, which comes in handy on a code review.
5. …

@Diadlo If you didn't know already, in Qt Creator you can jump quickly to the function declaration by the F2 or Shift+F2 key. So it's actually easy to find documentation.

Diadlo, line 46: With such a reasoned response I agree :)
What about commets to signals?

antis81, line 46: > What about commets to signals?

Fits nicely with a "top" block. Let's say there's a class with several signals and an enum. Would become:

``` c++
/*…license info…*/
#include "blabla.h"

/**
I can be briefly described as well!
*/
static void method() 
{ 
    // I'm just a little example.
}

/**
@class OurClass
@brief Exists for some reason…!?

Longer description
*/

/**
@enum OurClass::OurEnum
@brief The brief description line.

@var EnumValue1
means something

@var EnumValue2
means something else

Optional long description
*/

/**
@fn OurClass::somethingHappened(const QString &happened)
@param[in] happened    tells what has happened…
@brief This signal is emitted when something has happened in the class.

Here's an optional longer description of what the signal additionally does.
*/
```

antis81, line 46: P.s.: Enum values can also be written like this:

``` c++
/**
@enum OurClass::OurEnum

@value EnumValue1   means something
@value EnumValue2   means something else

…
*/
```

Ok?

Diadlo, line 46: Ok. And should be comment have indent or "leading stars"?

antis81, line 46: I'd write it like here without indent or stars. Qt Creator however puts stars in front by default. Would be glad if you know a method to change that (in project settings maybe).

tux3:
The code looks good to me, for the documentation I'd agree with @antis81, but I don't think it's an urgent problem, since we don't publish our internal APIs.

Diadlo:
@zetok What changes are required? Docs will be changed in another PR https://github.com/tux3/qTox/pull/3488#discussion_r70526447

zetok:
From my understanding, comments were supposed to be moved to .cpp files, dunno if in another PR..

@antis81 ↑ ?

antis81:
I reviewed this again and…

@Diadlo …no changes required.

Anyway there's a design bug lingering in ToxId QString constructor (parameter must become either `const QByteArray&` or `const char* const`, as QString might destroy the tox id. It doesn't affect this PR though and so I'm rather asking to setup another one containing the required changes. @tux3 Not sure, if this is affects security. At least it weakens reliability keeping the tox-id's integrity when copied.

@zetok Although it is not an urgent requirement, we agreed to move docs to cpp and do it in a separate PR. So this one is ready to merge, enabling us to move on.

Diadlo:
@antis81 Can you describe, how QString might destroy the tox id?

antis81:
> Can you describe, how QString might destroy the tox id?

Yes, no problem. This is due to different string encodings used on different platforms. When using fixed ASCII encoding (by convention), which includes all hex characters (0-9 and A-F), this behaves correctly on all platforms and can be used with a fixed length (75 bytes??) QByteArray or even a plain C char array - because actually no en-/decoding takes place. A QString OTOH interprets that data internally to some uncicode encoding, depending on the used encoding. Especially with variable length encodings like UTF-8 or UTF-16 or the Windows-1252 codepage, this can change (== destroy) the bytes within the array and in turn create an invalid tox-id. In Qt, we can also introduce strict string cast behaviour and prevent auto-casting from a char array, that is always assumed to be Latin-1 encoded (see [Converting Between 8-Bit Strings and Unicode Strings](http://doc.qt.io/qt-5/qstring.html)). I used that in another project, when dealing with Git Id's.

zetok:
@antis81 if ready, merge?

antis81:
@zetok Yes, please.

PR #3489, opened 2016-07-07
Make methods const

Diadlo:


antis81:
@Diadlo Thanks! That's better

zetok:
Needs to be rebased, since merge of #3411 caused slight conflicts.

src/widget/emoticonswidget.cpp
```diff
@@ -165,29 +165,23 @@ void EmoticonsWidget::wheelEvent (QWheelEvent *e)
...
+            stack.setCurrentIndex(stack.currentIndex() + 1);
+        else
+            stack.setCurrentIndex(stack.currentIndex() - 1);
 
+        emit pageButtonsUpdate();
     }
 }
 
-void EmoticonsWidget::PageButtonsUpdate()
+void EmoticonsWidget::pageButtonsUpdate() const
```
sudden6: I could be wrong, but doesn't this change the object and should therefore not be const?

src/widget/form/addfriendform.cpp
```diff
@@ -336,12 +336,12 @@ void AddFriendForm::removeFriendRequestWidget(QWidget* friendWidget)
     friendWidget->deleteLater();
 }
 
-void AddFriendForm::retranslateAcceptButton(QPushButton *acceptButton)
+void AddFriendForm::retranslateAcceptButton(QPushButton *acceptButton) const
```
sudden6: same as above

src/widget/form/addfriendform.cpp
```diff
@@ -336,12 +336,12 @@ void AddFriendForm::removeFriendRequestWidget(QWidget* friendWidget)
...
 {
     acceptButton->setText(tr("Accept"));
 }
 
-void AddFriendForm::retranslateRejectButton(QPushButton *rejectButton)
+void AddFriendForm::retranslateRejectButton(QPushButton *rejectButton) const
```
sudden6: same

src/widget/emoticonswidget.cpp
Diadlo: This method change children, not the object

Diadlo: This method change children, not the object

Diadlo: This method change children, not the object

Diadlo: This method change children, not the object

Diadlo: This method change children, not the object

Diadlo: This method change children, not the object

src/widget/form/addfriendform.cpp
Diadlo: This method can be static, because modify only parameter

src/widget/form/addfriendform.cpp
Diadlo: Same :)

src/widget/form/addfriendform.h
```diff
@@ -55,9 +55,9 @@ class AddFriendForm : public QObject
     bool addFriendRequest(const QString& friendAddress, const QString& message);
 
 signals:
-    void friendRequested(const QString& friendAddress, const QString& message);
-    void friendRequestAccepted(const QString& friendAddress);
-    void friendRequestsSeen();
+    void friendRequested(const QString& friendAddress, const QString& message) const;
+    void friendRequestAccepted(const QString& friendAddress) const;
+    void friendRequestsSeen() const;
```
antis81: Making signals const doesn't make any sense.

Diadlo: Unfortunately, it make. We can't emit non-const signal from const context (don't know why)

src/widget/emoticonswidget.h
```diff
@@ -31,12 +31,12 @@ class EmoticonsWidget : public QMenu
...
 signals:
-    void insertEmoticon(QString str);
+    void insertEmoticon(QString str) const;
 
 private slots:
-    void onSmileyClicked();
+    void onSmileyClicked() const;
     void onPageButtonClicked();
-    void PageButtonsUpdate();
+    void pageButtonsUpdate() const;
```
antis81: Const slots don't make sense either, because they are literally meant to react to some event. If they don't change anything, their existence is questionable.

Diadlo: Look at conversation in cpp file

src/widget/emoticonswidget.cpp
antis81: > This method change children, not the object

If an object changes its children, the object is altered as well. As a rule of thumb, any method called something like add, udate, set, insert, etc. is not const. It doesn't matter, if that method incidentally doesn't change anything in its parent class.

src/widget/form/addfriendform.cpp
```diff
@@ -228,7 +228,7 @@ void AddFriendForm::setIdFromClipboard()
     }
 }
 
-void AddFriendForm::deleteFriendRequest(const QString& toxId)
+void AddFriendForm::deleteFriendRequest(const QString& toxId) const
```
antis81: Ah forgot the "delete" in the list of words above :)

Diadlo: Ok, if I will make this method `static` it will be better?

src/widget/form/addfriendform.h
antis81: No, it doesn't! A const method is meant to not alter anything within it's context, right? So any method or signal called inside a const method is enforced to be const as well. If a const method emits a signal, something's wrong! Why, because nothing changed actually! Breaking this so-called const-correctness by declaring non-const methods const will result in a lot of `const_cast<type>()` casts as a consequence!

src/widget/form/addfriendform.cpp
antis81: Yeah, that's a static method then.

src/widget/form/addfriendform.cpp
antis81: Yeah, will be better.

antis81: Actually code should be moved to `Settings` and this should change to:

``` c++
static inline void AddFriendForm::deleteFriendRequest(const QString& toxId)
{
    Settings::getInstance().removeFriendRequest(toxId);
}
```

As the `AddFriendForm` will only be a single instance (when adding a friend), it actually makes no difference, if it is static or not.

src/widget/form/addfriendform.h
Diadlo: For example `onCurrentChanged` don't change anything in object, but change settings and emit `friendRequestsSeen`. This method shouldn't be const?

antis81: Basically yes. Btw. found a nice article here: https://isocpp.org/wiki/faq/const-correctness#const-cast-and-optimization

src/chatlog/chatlog.h
```diff
@@ -60,26 +60,26 @@ class ChatLog : public QGraphicsView
...
 
-    void layout(int start, int end, qreal width);
+    void layout(int start, int end, qreal width) const;
     bool isOverSelection(QPointF scenePos) const;
     bool stickToBottom() const;
 
     qreal useableWidth() const;
 
-    void reposition(int start, int end, qreal deltaY);
+    void reposition(int start, int end, qreal deltaY) const;
```
tux3, line 82: Semantically, I'm not sure if it makes sense to have "layout" and "reposition" be const.
If I pass a const Chatlog& around, do I really want it to be possible to reposition and relayout things? Probably not.

src/core/core.h
```diff
@@ -126,70 +126,69 @@ public slots:
...
+    void friendshipChanged(uint32_t friendId) const;
+
+    void friendStatusChanged(uint32_t friendId, Status status) const;
+    void friendStatusMessageChanged(uint32_t friendId, const QString& message) const;
+    void friendUsernameChanged(uint32_t friendId, const QString& username) const;
+    void friendTypingChanged(uint32_t friendId, bool isTyping) const;
+    void friendAvatarChanged(uint32_t friendId, const QPixmap& pic) const;
+    void friendAvatarRemoved(uint32_t friendId) const;
+
+    void friendRemoved(uint32_t friendId) const;
```
tux3, line 146: If a signal is const, it means that it's allowed to call it from a const function.
Signals like friendRemoved shouldn't be const! If a friend was removed, then we _definitely_ modified the state, trying to do that from a const function ought to be a compiler error.

src/core/cstring.h
```diff
@@ -34,7 +34,7 @@ class CString
     explicit CString(const CString& cstr);
     ~CString();
 
-    uint8_t* data();
+    uint8_t* data() const;
```
tux3, line 37: If the getter is const, then the returned pointer needs to be const too.
Otherwise if the caller can modify the string data, the function can't be const!

src/ipc.cpp
```diff
@@ -221,20 +221,18 @@ IPC::IPCEvent *IPC::fetchEvent()
...
-                                  Q_RETURN_ARG(bool, result),
-                                  Q_ARG(IPCEventHandler, handler),
-                                  Q_ARG(const QByteArray&, arg));
-        return result;
-    }
-    else
+    if (QThread::currentThread() == qApp->thread())
     {
         result = handler(arg);
         return result;
```
tux3, line 227: If you're breaking the if/else, then you might as well turn the if into a one-liner `return handler(arg)`
The result variable is only needed because QMetaObject needs a reference.

src/persistence/profile.cpp
```diff
@@ -438,12 +439,12 @@ void Profile::removeAvatar()
...
 }
 
-bool Profile::isHistoryEnabled()
+bool Profile::isHistoryEnabled() const
 {
     return Settings::getInstance().getEnableLogging() && history;
 }
 
-History *Profile::getHistory()
+History *Profile::getHistory() const
```
tux3, line 447: If we can modify the profile's history, is it really const-correct?

src/core/cstring.h
Diadlo, line 37: May be I don't understand something. But **this** method don't change object. This is a reason, why I make it as `const`

src/persistence/settings.cpp
```diff
@@ -1786,6 +1786,16 @@ Settings::Request Settings::getFriendRequest(int index) const
     return friendRequests.at(index);
 }
 
+void Settings::deleteFriendRequest(const QString &stringId)
```
tux3, line 1789: This commit is a bit hidden in the middle of a const-correctness pull-request. It's okay since it's at least in the commit name, but please try to mention it in the description of the pull request if you do some other things.

src/widget/about/aboutuser.h
```diff
@@ -23,7 +23,7 @@ class AboutUser : public QDialog
     ToxId toxId;
 
 private slots:
-    void onAcceptedClicked();
+    void onAcceptedClicked() const;
```
tux3, line 26: Why is only this slot const and not the other ones?

src/widget/form/addfriendform.h
```diff
@@ -73,9 +73,9 @@ private slots:
     void retranslateUi();
     void addFriendRequestWidget(const QString& friendAddress, const QString& message);
     void removeFriendRequestWidget(QWidget *friendWidget);
-    void retranslateAcceptButton(QPushButton* acceptButton);
-    void retranslateRejectButton(QPushButton* rejectButton);
-    void deleteFriendRequest(const QString &toxId);
+    static void retranslateAcceptButton(QPushButton* acceptButton);
+    static void retranslateRejectButton(QPushButton* rejectButton);
+    static void deleteFriendRequest(const QString &toxId);
```
tux3, line 78: If you add statics here, there are a LOT of other places where you could have added static :)

src/widget/friendlistlayout.cpp
```diff
@@ -72,10 +72,11 @@ int FriendListLayout::indexOfFriendWidget(GenericChatItemWidget* widget, bool on
 {
     if (online)
         return friendOnlineLayout.indexOfSortedWidget(widget);
+
     return friendOfflineLayout.indexOfSortedWidget(widget);
 }
 
-void FriendListLayout::moveFriendWidgets(FriendListWidget* listWidget)
+void FriendListLayout::moveFriendWidgets(FriendListWidget* listWidget) const
```
tux3, line 79: The whole purpose of this function is to move elements of the layout around, it's not const at all!
Semantically, it makes no sense. The goal of the function is to modify its object, it's not a const function.

src/persistence/profile.cpp
Diadlo: Same with https://github.com/tux3/qTox/pull/3489#discussion_r71072335

src/widget/friendwidget.cpp
```diff
@@ -420,7 +420,7 @@ void FriendWidget::mouseMoveEvent(QMouseEvent *ev)
     }
 }
 
-void FriendWidget::setAlias(const QString& _alias)
+void FriendWidget::setAlias(const QString& _alias) const
```
tux3, line 423: Please don't mark setter functions const, either!
If I have a const FriendWidget, I shouldn't be able to modify the alias through it.

src/widget/gui.cpp
```diff
@@ -195,46 +163,38 @@ QString GUI::itemInputDialog(QWidget * parent, const QString & title,
...
 // Private implementations
 
-void GUI::_clearContacts()
+void GUI::_clearContacts() const
 {
     Nexus::getDesktopGUI()->clearContactsList();
 }
 
-void GUI::_setEnabled(bool state)
+void GUI::_setEnabled(bool state) const
```
tux3, line 197: This and the functions around modify the GUI, they can't be const either, even if they don't directly touch private variables.

src/widget/about/aboutuser.h
Diadlo, line 26: Because, they change object or sent non-const pointer on `this`.

src/core/cstring.h
tux3, line 37: If the method allows you to change the object, then it can't be const, otherwise passing a const Object& reference to people would still allow them to modify the objet.

If you look at [std::vector](http://www.cplusplus.com/reference/vector/vector/data/) you'll see it has both a `value_type* data();` and a `const value_type* data() const;` function.
The const function only allows you to read, the non-const function allows you to modify

If the method is const and it returns a pointer that belongs to the object, then the pointer it returns needs to be const. Const is a safety feature.

src/persistence/profile.cpp
tux3, line 447: I have replied in https://github.com/tux3/qTox/pull/3489#discussion_r71072335

src/widget/about/aboutuser.h
tux3, line 26: I think this function changes the state too; even though it doesn't show a message box like the others, and so doesn't use the non-const `this` pointer, it does something similar to the other functions.
You shouldn't mark const everywhere there isn't a compiler error, only when it makes sense in the interface.

onAcceptClicked means somebody interracted with the widget and asked to save the new settings, so that counts as modifying things for me.

Diadlo:
Ok. I think I should temporary close this PR and rewrite most part of it.

PR #3498, opened 2016-07-08
fix(chat): cleanup chat css base style

### LINKED ISSUE #3496: [regression] non-markdown text formatting is broken
##### Brief Description

`/me action` should be blue
`> quote` should be green

qTox version: qTox version v1.4.0-246-g2d06b99
Commit hash:  2d06b9960c60f1fc886ae04c595bd9a921cff316
toxcore:  0.0.0
Qt:  5.5.1

Reproducible: Always
##### Steps to reproduce
1. type `/me action` or `> quote`
##### Observed Behavior

Text is not in color.
##### Expected Behavior

Text should be colored.
##### Additional Info

no color

what was before:
quote
quote and action

antis81:
Fixes #3496

### DIFF
--- a/ui/chatArea/innerStyle.css
+++ b/ui/chatArea/innerStyle.css
@@ -1,42 +1,35 @@
+body {
+    font: @baseFont;
+}
+
 p {
-  white-space: pre-wrap;
+    white-space: pre-wrap;
 }
 
-p.msg {
-  color: @black;
-  font: @big;
+.action {
+    color: #1818FF;
+    font-style: italic;
 }
 
-p.action {
-  color: #1818FF;
-  font: @big;
-  font-style: italic;
+.typing {
+    color: @mediumGreyLight;
 }
 
-p.typing {
-  color: @mediumGreyLight;
-  font: @big;
+.quote {
+    color: #279419;
 }
 
-span.quote {
-  color: #279419;
+.alert {
+    margin-left: 0px;
+    margin-right: 0px;
+    background-color: @orange;
 }
 
-p.alert {
-  margin-left: 0px;
-  margin-right: 0px;
-  color: @black;
-  background-color: @orange;
-  font: @big;
-}
-
-p.alert_name {
-  color: @black;
-  background-color: @orange;
-  font: @bigBold;
+.alert_name {
+    background-color: @orange;
+    font: @bigBold;
 }
 
 a {
     color: blue;
 }
-
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -61,6 +61,9 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
 
     switch(type)
     {
+    case NORMAL:
+        text = wrapDiv(text, "msg");
+        break;
     case ACTION:
         senderText = "*";
         text = wrapDiv(QString("%1 %2").arg(sender.toHtmlEscaped(), text), "action");
@@ -69,8 +72,6 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     case ALERT:
         text = wrapDiv(text, "alert");
         break;
-    default:
-        text = wrapDiv(text, "msg");
     }
 
     // Note: Eliding cannot be enabled for RichText items. (QTBUG-17207)
--- a/src/chatlog/content/text.cpp
+++ b/src/chatlog/content/text.cpp
@@ -31,13 +31,13 @@
 #include <QDesktopServices>
 #include <QTextFragment>
 
+#include "src/widget/style.h"
+
 Text::Text(const QString& txt, const QFont& font, bool enableElide, const QString &rwText, const QColor c)
     : rawText(rwText)
     , elide(enableElide)
     , defFont(font)
-    , defStyleSheet(QString::fromUtf8("body{font: '%1' %2px %3;}")
-                    .arg(font.family()).arg(font.pixelSize())
-                    .arg(font.bold() ? "bold" : QString()))
+    , defStyleSheet(Style::getStylesheet(QStringLiteral(":/ui/chatArea/innerStyle.css"), font))
     , color(c)
 {
     setText(txt);
--- a/src/chatlog/customtextdocument.cpp
+++ b/src/chatlog/customtextdocument.cpp
@@ -29,9 +29,6 @@
 CustomTextDocument::CustomTextDocument(QObject *parent)
     : QTextDocument(parent)
 {
-    static QString css = Style::getStylesheet(":ui/chatArea/innerStyle.css");
-
-    setDefaultStyleSheet(css);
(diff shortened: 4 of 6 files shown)


antis81:
@zetok Couldn't find the colors. So I chose for some "place holders". Guess the "msg" HTML class is regular text??

zetok:
https://github.com/tux3/qTox/blob/464daebdd7f912fb48e63e9f2a25b719b764a9e7/ui/chatArea/innerStyle.css colors.

antis81:
@zetok Thanks -> It's ready.

zetok:
Shouldn't it use things from the CSS file?

antis81:
> Shouldn't it use things from the CSS file?

It does. :)

antis81:
@zetok Wrote a greatly improved version of the fix!

Chat message style sheet is now completely based on `innerStyle.css` -> introducing a "base font" and cleaned up that crap in `documentcache.cpp`.

ui/chatArea/innerStyle.css
```diff
@@ -1,42 +1,9 @@
...
-    color: blue;
-}
+body {font: @baseFont;}
+p {white-space: pre-wrap;}
+.action {color: #1818FF; font-style: italic;}
+.typing {color: @mediumGreyLight;}
+.quote {color: #279419;}
+.alert {margin-left: 0px; margin-right: 0px; background-color: @orange;}
+.alert_name {background-color: @orange; font: @bigBold;}
+a {color: blue;}
```
zetok: Why the change? Previous one was easier to read.

antis81: Why not? ^^
1. It's still easy to read (one rule per line)…
2. …plus it compresses text for the interpreter(s), which parse it over and over for each message.
3. It increases the hurdle to change the style for "inexperienced" people, which comes to another advantage. :)

Diadlo: > It's still easy to read (one rule per line)…

Nope

zetok: > one rule per line

What. It **was** one rule per line, but you've changed it.

> plus it compresses text for the interpreter(s), which parse it over and over for each message.

Err? Sounds like something that compiler/whatever should be able to optimize? Also, since you're claiming some negligible performance gain, mind providing some hard data on that?

> It increases the hurdle to change the style for "inexperienced" people, which comes to another advantage. :)

Yep, that's pretty much what your change does.

zetok: > > It increases the hurdle to change the style for "inexperienced" people, which comes to another advantage. :)
> 
> Yep, that's pretty much what your change does.

Sorry, I didn't make it really clear: making it harder for people to contribute for no good reason is **wrong**. Please don't do that shit.

antis81:
> Err? Sounds like something that compiler/whatever should be able to optimize? Also, since you're claiming some negligible performance gain, mind providing some hard data on that?

Contrary to what you probably think, this is not compiled in either way! Style is applied to a real HTML document (exactly one per message) and works exactly like any inlined `<style>…</style>` rule set in this context. As you might know, HTML is an interpreted text language and simply not build to be kept in an evergrowing large list of chat messages. You can experience a real world problem e.g. on Github, when opening a large diff set (this is pure HTML) - caution: computer may lockup! It is already written in Qt's docs, that HTML text comes at the price of performance and to avoid that wherever possible (e.g. read [QtQuick.Text.textFormat](http://doc.qt.io/qt-5/qml-qtquick-text.html)). For this reason Qt supports a "reduced HTML subset" and requires strict XHTML tags. The same set is also used in [QTextDocument::setHtml](http://doc.qt.io/qt-5/qtextdocument.html#setHtml)!

Btw. in our use case, this can be greatly improved, by implementing a model, that only contains the raw text data (string) instead of a QTextDocument. This allows a (presumed) "ChatView" to apply rendering only to the visible and changed QGraphicsItem, which is definitely much faster than current approach. Plus, it consumes less memory.

So: Why do I state, that CSS one rule per line is easy to read?
CSS rules can be seen similar to a function call -> they cannot be nested. If they could be nested, you'd be totally right! But that is not the case. Btw. a css rule is `element { <attribute-set> }` (e.g. `body {background-color: blue; border-color: green;}`). 

> Please don't do that shit.

It is no shit. It's actually a sane thing to do here. It is shit, to split up ~10 css rules into ~40 lines, which actually makes the style harder to follow, because you have to scroll up and down, to compare if a rule goes conform with it's "basis rule".

Of course anyone can and is invited to contribute. But… contributions will be reviewed and it is a process of learning. By asking people to change their (logical) code design and improve themselves, I always will try to make them understand.

zetok:
@antis81 Even with stuff being dynamically interpreted, is really amount of whitespaces and newlines really affecting performance to a degree that can be measured?

Since you claim that it is, I really would want to see data on it in qTox. Note that even QtQuick stuff you've linked to has examples of CSS that look ~exactly like the CSS before change..

> CSS rules can be seen similar to a function call -> they cannot be nested. If they could be nested, you'd be totally right! But that is not the case.

Yes, that's irrelevant.

> It is no shit. It's actually a sane thing to do here. It is shit, to split up ~10 css rules into ~40 lines, which actually makes the style harder to follow, because you have to scroll up and down, to compare if a rule goes conform with it's "basis rule".

Looking from different angle, one could argue that it was one "rule" per line, and then it was squeezed. And it didn't become more readable because of it, quite opposite. Before understanding what given line does was easier, with the change it's harder.

> But… contributions will be reviewed and it is a process of learning. By asking people to change their (logical) code design and improve themselves, I always will try to make them understand.

Change makes learning harder. Also, you helping people doesn't scale. On the other hand, clear CSS that was before was easy to understand even for someone completely unfamiliar to CSS, and that's available to anyone able to see the sauce and read it – it scales. Not so much with your change.

On a side note, it's usually better to have code that is easier to read, even if it's not as "optimized" as code that is hard to read, but is only very slightly faster, and can be understood only by a small subset of people that would understood the clear code. And that applies to all of the code except for the performance critical one. And as you've highlighted, this isn't the part that should be changed if we need to actually get some real performance gain.

ui/chatArea/innerStyle.css
sudden6: I agree with zetok and Diadlo, the previous one was easier to read. If performance is a concern we could probably generate minimized css files.

antis81:
I don't really give a shit about the styling of such a short css document and discussing ~40 white-spaces back and forth isn't worth it to drop a valuable PR  - and if a harder to handle long document is what people like more, which probably changes once in a decade - I happily provide that… :)

antis81:
Btw. get surprised of what happens, if you auto-indent a css document in Qt Creator by hitting `Ctrl+A` and then `Ctrl+I` - isn't life just great… :smile:

zetok:
@antis81 why unsigned commit was pushed on master with no signed merge?

antis81:
@zetok That's interesting. I sign every of my commits via config. Don't know what went wrong exactly…??

PR #3502, opened 2016-07-11
refactor: cleanup chat code

antis81:
- [x] fix elided text to return the correct plain text without the wrapping div-tags
  - Appears like dead code though…
- [x] fix type conversion warnings
- [x] cleanup singleton code in DocumentCache


sudden6:
Didn't test, but code changes LGTM.

Issue #518, opened 2014-10-24
qtox vanishes when I change monitors


ProMcTagonist:
Run qtox. 

Unplug VGA cable. 

Plug it into another monitor. 

qTox is gone.

I thought i could get around it by hiding it in the tray during the unplugging and replugging, but it vanishes all the same. Debian Jessie Openbox.

```
QBackingStore::flush() called for  QWidgetWindow(0x848ce78, name = "QSystemTrayIconSysWindow")  which does not have a handle.
requestSystemTrayWindowDock failed.
```

dubslow:
@agilob ? This sounds like a fun one to fix. :)

dubslow:
Also sort of related, when I change workspace qTox gets minimized to tray... that's a bit annoying.

agilob:
Possibly another(?) Qt5 bug?
https://bugs.launchpad.net/qpdfview/+bug/1334682
https://bugreports.qt-project.org/browse/QTBUG-40168

No idea, I don't have second screen, not even place to put on one my desk ;/

suhr:
Is this still an issue with fresh Qt5?

antis81:
I have a second monitor available and can at least help in reproducing this (tomorrow). Looks not like a Qt5 bug - we'll see.

agilob:
I couldn't reproduce it on Arch with qt5.4 but a friend I asked for helped reproduced it on fedora. I'm thinking this might be also desktop environment dependent.

suhr:
@agilob Does your friend have the same Qt5 version?

ProMcTagonist:
I can't re-test with multiple monitors right now but I'll get on it within the next couple of days.

antis81:
I tried to reproduce with Lubuntu 15.04. Short answer: Cannot reproduce.

Long answer:
I plugged in a single second VGA monitor. Checked all positions with qTox window ...
- ... on left (primary) monitor
- ... on right (secondary) monitor
- ... centred between both monitors

Then on each setup I did:
1. Unplug monitor 
2. Re-plug monitor

Result: Nothing happened -> qTox and every other application stays in place. 

@ProMcTagonist Question: Can you please provide more info about the steps to reproduce and system info (distro, qtox version, qt version)? Did you call XRandR (actively trigger the monitor detection) after re-plugging the second monitor?

---

Btw.: Where do I find exact qTox version info (with current build number)?

ProMcTagonist:
[...]

PR #3505, opened 2016-07-12
fix(x11grab): try and use the current display

### LINKED ISSUE #3500: Desktop capture not work with NVIDIA discrete video card
##### Brief Description

OS: Linux Debian 8
qTox version: Nightly
Commit hash:  -
toxcore:   0.0.0
Qt:  5.6.1
Hardware:  Nvidia Quadro k600

Reproducible: Always
##### Steps to reproduce
1. Run qtox.
2. Go to Audio/Video.
3. Set video device : Desktop
##### Observed Behavior

Don't see desktop, only avatar in preview window
and in videocall.
##### Expected Behavior

Want to see desktop in preview window
and in videocall.
##### Additional Info

In debug mode (running from terminal) when
I set desktop as video device, see next information:
[12:11:25.724] src/video/camerasource.cpp:209 : Debug: Opening device  "x11grab#:0"
Invalid MIT-MAGIC-COOKIE-1 key[x11grab @ 0x1c86f80] Cannot open display :0, error 1.
In log file see next information:
[12:04:00.560] src/video/camerasource.cpp:209 : Debug: Opening device  "x11grab#:0"
[12:04:00.562] src/video/camerasource.cpp:225 : Warning: Failed to open device!

my $DISPLAY:
$ echo $DISPLAY
:1

On other computer with integrated video card 
$ echo $DISPLAY
:0.0
And all works.

b4n:
Don't hard-code ":0" as the display and try and use the DISPLAY environment variable instead.

This fixes screen grabbing under X11 when the user is not on screen 0.

Fix #3500.

### DIFF
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -317,7 +317,17 @@ QVector<QPair<QString, QString>> CameraDevice::getDeviceList()
     if (idesktopFormat)
     {
         if (idesktopFormat->name == QString("x11grab"))
-            devices.push_back(QPair<QString,QString>{"x11grab#:0", QObject::tr("Desktop", "Desktop as a camera input for screen sharing")});
+        {
+            QString dev = "x11grab#";
+            QByteArray display = qgetenv("DISPLAY");
+
+            if (display.isNull())
+                dev += ":0";
+            else
+                dev += display.constData();
+
+            devices.push_back(QPair<QString,QString>{dev, QObject::tr("Desktop", "Desktop as a camera input for screen sharing")});
+        }
         if (idesktopFormat->name == QString("gdigrab"))
             devices.push_back(QPair<QString,QString>{"gdigrab#desktop", QObject::tr("Desktop", "Desktop as a camera input for screen sharing")});
     }


b4n:
(BTW, it's not only a theoretical fix for #3500, I had the same problem because my session runs as `:1` and then the x11grab source wouldn't work.  This patches fixes it)

src/video/cameradevice.cpp
```diff
@@ -317,7 +317,17 @@ QVector<QPair<QString, QString>> CameraDevice::getDeviceList()
...
         if (idesktopFormat->name == QString("x11grab"))
-            devices.push_back(QPair<QString,QString>{"x11grab#:0", QObject::tr("Desktop", "Desktop as a camera input for screen sharing")});
+        {
+            QString dev = "x11grab#";
+            QByteArray display = qgetenv("DISPLAY");
+
+            if (display.isNull())
+                dev += ":0";
+            else
+                dev += display.constData();
```
antis81, line 327: Aren't you missing a colon here?

b4n, line 327: unless `$DISPLAY` is malformed, no.

``` console
$ echo $DISPLAY
:1.0
```

As with what we want here, it's `[host]:display[.screen]`.  See https://tronche.com/gui/x/xlib/display/opening.html

antis81, line 327: Ok then. Nevermind.

agilob:
https://github.com/tux3/qTox/issues/518

b4n:
@agilob ?  I'm not too knowledgeable, but I don't think it's related, as AFAIK the DISPLAY cannot change inside an X session at all.
Screen too AFAIK, as it's _not_ a monitor but here talks about X screens (groups of devices).  Which virtually nobody uses nowadays and unfortunately desktop start to drop support for.

It's also unrelated to the new setting "Screen" in the preferences, that IIUC just selects the area of the Nth monitor, so it should work fine.  But admittedly I can't test that feature as I don't have a second monitor to plug right now.

PR #3509, opened 2016-07-14

zetok:
cc @minj ?

### DIFF
--- a/translations/ar.ts
+++ b/translations/ar.ts
@@ -12,11 +12,19 @@
         <translation>صوت/فيديو</translation>
     </message>
     <message>
-        <source>%1p</source>
-        <translation>%1p</translation>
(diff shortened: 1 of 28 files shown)


minj:
@zetok most updates are OK, but some are questionable at first sight. No time for a thorough review atm but I recommend accepting them. I'll make any necessary changes later

Do you figure I should be using weblate instead?

zetok:
> Do you figure I should be using weblate instead?

Weblate provides a nice UI for people who aren't familiar with git/github/Qt tools, or for whom using all that stuff is too bothersome when they can use weblate and focus on just the translating, without "unnecessary" overhead.

Dunno – use what works best for you, be it weblate or manually making translations/PRs :)

PR #3512, opened 2016-07-14
fix(groupinviteform): escape HTML

TheNain38:
Please test it, it's been a while since last time I did some code...

### DIFF
--- a/src/widget/form/groupinviteform.cpp
+++ b/src/widget/form/groupinviteform.cpp
@@ -111,7 +111,7 @@ void GroupInviteForm::addGroupInvite(int32_t friendId, uint8_t type, QByteArray 
     QDateTime currentDateTime = QDateTime::currentDateTime();
     QString date = currentDateTime.toString(Settings::getInstance().getDateFormat());
     QString time = currentDateTime.toString(Settings::getInstance().getTimestampFormat());
-    groupLabel->setText(tr("Invited by <b>%1</b> on %2 at %3.").arg(name, date, time));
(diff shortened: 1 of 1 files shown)


antis81:
@TheNain38 Code will work.

@zetok I don't like the idea of supporting HTML in the qTox name at all. Plain text should be used. If at all, RichText might be allowed (and not escaped in this case). But the problem is: Escaping HTML results in ugly appearance like e.g.:

`Invited by <b>&lt;font size="36"&gt;Some Body&lt;/font&gt;</b>`

Preventing use of HTML in names resolves all of those problems in one go. Well, users who made use HTML need to adjust their names, but that's it actually!

Agree'd?

zetok:
> Preventing use of HTML in names resolves all of those problems in one go.

No. Users should be allowed to put whatever text they want as their name/whatever. but the text shouldn't be parsed as HTML. Instead it should be displayed exactly the way it is, that is, as a plaintext.

_edited for clarification_

antis81:
> …Instead it should be displayed exactly the way it is, that is, as a plaintext.

That's exactly what I said. :) So the fix is alright by escaping any `<, >, & and "` characters and setting the result as "reduced HTML subset" to the label.

zetok:
> That's exactly what I said. :)

Looks like I've misunderstood.

> Preventing use of HTML in names resolves all of those problems in one go.

In the past some people wanted to actively limit what is allowed as username / etc, which is not going to work well, and only cause problems for users. So I guess I'm a bit touchy about things that sound exactly like those ideas, sorry about that.

PR #3519, opened 2016-07-16

zetok:


zetok:
cc @RowenStipe osx build fails

PR #3526, opened 2016-07-18
docs(INSTALL.md): add RtAudio to list of dependencies

zetok:
Added only for Debian, Ubuntu and Fedora, since Arch and openSUSE don't
have it in repos.

Also add dependency to Linux travis and `simple_make.sh`.

Minimal required version might be wrong – that's the lowest version that
is available on distro (Fedora) that is new enough to compile qTox.

Someone else will have to figure out osx travis.

cc @RowenStipe ?

### DIFF
--- a/.travis/build-ubuntu_14_04.sh
+++ b/.travis/build-ubuntu_14_04.sh
@@ -35,6 +35,7 @@ sudo apt-get install -y \
     libopenal-dev \
     libopus-dev \
     libqrencode-dev \
+    librtaudio-dev \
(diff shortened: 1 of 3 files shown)


antis81:
LGTM so far. Don't know what to do with the missing distros. Should we compile and package by ourselves on those?

zetok:
> Should we compile and package by ourselves on those?

That's up to people who package qTox for those distros :)

antis81:
@zetok Merged.

Issue #3527, opened 2016-07-18
[regression] segfault during a video call


zetok:
##### Brief Description

OS: Gentoo Linux amd64
qTox version: 1.4
Commit hash: f970bb5eaca45c5500b1723ff82116e5ffb0a878
toxcore:  0.0.0
Qt:  5.5.1

Reproducible: Always
##### Steps to reproduce
1. Start video call
2. crash
##### Observed Behavior

segfault
##### Expected Behavior

video call should start
##### Additional Info

(links, images, etc go here)

bt:

``` backtrace
[16:53:22.071] src/widget/form/chatform.cpp:324 : Debug: onAvStart
[16:53:22.073] src/widget/form/chatform.cpp:648 : Debug: creating netcam
[16:53:22.075] src/video/netcamview.cpp:88 : Debug: SIZER QRect(1625,0 24x20)
[swscaler @ 0x5555581c3660] Warning: data is not aligned! This can lead to a speedloss

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff4676542 in ff_hscale8to15_8_ssse3 () from /usr/lib64/libswscale.so.3
(gdb) bt
#0  0x00007ffff4676542 in ff_hscale8to15_8_ssse3 () from /usr/lib64/libswscale.so.3
#1  0x00007ffff462993e in lum_h_scale (c=0x5555581c3660, desc=0x5555581e29e0, sliceY=<optimized out>, sliceH=14) at /var/tmp/portage/media-video/ffmpeg-2.8.6/work/ffmpeg-2.8.6/libswscale/hscale.c:41
#2  0x00007ffff465b9e0 in swscale (c=0x5555581c3660, src=<optimized out>, srcStride=<optimized out>, srcSliceY=<optimized out>, srcSliceH=<optimized out>, dst=0x7fffffffb200, dstStride=0x7fffffffb1d0)
    at /var/tmp/portage/media-video/ffmpeg-2.8.6/work/ffmpeg-2.8.6/libswscale/swscale.c:589
#3  0x00007ffff465ce07 in sws_scale (c=<optimized out>, srcSlice=<optimized out>, srcStride=<optimized out>, srcSliceY=<optimized out>, srcSliceH=<optimized out>, dst=<optimized out>, dstStride=0x555558100f40)
    at /var/tmp/portage/media-video/ffmpeg-2.8.6/work/ffmpeg-2.8.6/libswscale/swscale.c:1272
[...]

PR #3531, opened 2016-07-19
fix(videoframe): Added buffer alignment between lines

Diadlo:
Fixed warning from #3527. Not sure, that fix issue.

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -1430,6 +1430,7 @@ QString Settings::getVideoDev() const
 void Settings::setVideoDev(const QString& deviceSpecifier)
 {
     QMutexLocker locker{&bigLock};
+    qDebug() << "Set video dev" << deviceSpecifier;
     videoDev = deviceSpecifier;
 }
 
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -327,6 +327,7 @@ QVector<QPair<QString, QString>> CameraDevice::getDeviceList()
             else
                 dev += display.constData();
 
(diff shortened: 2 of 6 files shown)


zetok:
on 824dd56d443c9c0d8b25e3b64017a642a16968f3
- [ ] fixes the crash, but introduces regression; on the receiving side:
  
  dat image
- [ ] even though log says
  
  ```
  [20:15:25.373] src/widget/form/chatform.cpp:324 : Debug: onAvStart
  [20:15:25.374] src/widget/form/chatform.cpp:648 : Debug: creating netcam
  [20:15:25.376] src/video/netcamview.cpp:88 : Debug: SIZER QRect(0,0 0x0)
  Lost packet
  Lost packet
  [20:15:38.078] src/video/camerasource.cpp:219 : Debug: Opening device  "x11grab#:0.0"
  [swscaler @ 0x5555580fcfe0] Warning: data is not aligned! This can lead to a speedloss
  ```
  
  whole screen is being streamed

Diadlo:
@zetok Are you sure, that it's fix crash? It seems that bug reproduce only when exist problem with opening (random)

zetok:
> It seems that bug reproduce only when exist problem with opening (random)

I could reliably reproduce, and now I can't, so it looks fixed to me.

Diadlo:
@zetok Can you test it now?

zetok:
on 075762c011d18b076113534f768ac75c3797ce10 still the same

Diadlo:
Warning fixed

zetok:
on bed1d66 started crashing again

Diadlo:
Some misunderstanding of problem. Closed

PR #3534, opened 2016-07-21
Fix desktop stream crash

Diadlo:


sudden6:
If I read this correctly the qDebug() statements will be called for every frame and create massive logspam. Is this PR intended only for testing?

Diadlo:
This PR for tests, now

sudden6:
ok, please put this information in the comment in the future.

zetok:
03c50764d6d0cc8acd8e319715ccd2d423e0bf2c fixes #3527 for me.

Add to commit message info "why", i.e. that it fixes an issue where stuff crashed because $reasons

Generally, commits should explain what their purpose is, not only what is done, but also why.

\+ `Fixes #3527`

PR #3540, opened 2016-07-23
[WIP] Moved loading video device in another thread.

### LINKED ISSUE #2058: Switching to 'Audio/Video' settings tab is blocked on camera initialisation

Diadlo:
Fix #2058.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3540)

### DIFF
--- a/src/video/videosurface.cpp
+++ b/src/video/videosurface.cpp
@@ -26,6 +26,7 @@
 #include "src/core/core.h"
 #include "src/widget/style.h"
 
+#include <QtConcurrent/QtConcurrentRun>
 #include <QPainter>
 #include <QLabel>
 #include <QDebug>
@@ -60,6 +61,7 @@ VideoSurface::VideoSurface(const QPixmap& avatar, VideoSource *source, QWidget* 
 
 VideoSurface::~VideoSurface()
 {
+    QMutexLocker l(&bigLock);
     unsubscribe();
 }
 
@@ -80,9 +82,14 @@ void VideoSurface::setSource(VideoSource *src)
     if (source == src)
         return;
 
-    unsubscribe();
-    source = src;
-    subscribe();
+    // To don't block GUI thread.
+    QtConcurrent::run([this, src]()
+    {
+        QMutexLocker l(&bigLock);
+        unsubscribe();
+        source = src;
+        subscribe();
+    });
 }
 
 QRect VideoSurface::getBoundingRect() const
@@ -110,6 +117,7 @@ QPixmap VideoSurface::getAvatar() const
 
 void VideoSurface::subscribe()
 {
+    QMutexLocker l(&subscribeLock);
     if (source && hasSubscribed++ == 0)
     {
         source->subscribe();
@@ -120,6 +128,7 @@ void VideoSurface::subscribe()
 
 void VideoSurface::unsubscribe()
 {
+    QMutexLocker l(&subscribeLock);
     if (!source || hasSubscribed == 0)
         return;
 
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -29,6 +29,7 @@
 #include "src/core/core.h"
 #include "src/core/coreav.h"
 
+#include <QtConcurrent/QtConcurrent>
 #include <QDebug>
 #include <QShowEvent>
 #include <map>
@@ -113,8 +114,8 @@ void AVForm::showEvent(QShowEvent* event)
 {
     getAudioOutDevices();
     getAudioInDevices();
-    createVideoSurface();
     getVideoDevices();
+    createVideoSurface();
 
     if (!subscribedToAudioIn) {
         // TODO: this should not be done in show/hide events
@@ -533,6 +534,7 @@ void AVForm::createVideoSurface()
 {
     if (camVideoSurface)
         return;
+
     camVideoSurface = new VideoSurface(QPixmap(), CamFrame);
     camVideoSurface->setObjectName(QStringLiteral("CamVideoSurface"));
     camVideoSurface->setMinimumSize(QSize(160, 120));
@@ -544,12 +546,16 @@ void AVForm::killVideoSurface()
 {
     if (!camVideoSurface)
         return;
+
     QLayoutItem *child;
     while ((child = gridLayout->takeAt(0)) != 0)
         delete child;
 
     camVideoSurface->close();
-    delete camVideoSurface;
+    QtConcurrent::run([this]()
+    {
+        delete camVideoSurface;
+    });
     camVideoSurface = nullptr;
 }
 
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -198,8 +198,8 @@ void AVForm::selectBestModes(QVector<VideoMode> &allVideoModes)
     for (int i = 0; i < allVideoModes.size(); ++i)
     {
         VideoMode mode = allVideoModes[i];
-        QString pixelFormat = CameraDevice::getPixelFormatString(mode.pixel_format);
-        qDebug("width: %d, height: %d, FPS: %f, pixel format: %s", mode.width, mode.height, mode.FPS, pixelFormat.toStdString().c_str());
+//        QString pixelFormat = CameraDevice::getPixelFormatString(mode.pixel_format);
+//        qDebug("width: %d, height: %d, FPS: %f, pixel format: %s", mode.width, mode.height, mode.FPS, pixelFormat.toStdString().c_str());
 
         // PS3-Cam protection, everything above 60fps makes no sense
         if (mode.FPS > 60)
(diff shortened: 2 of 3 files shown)


src/widget/form/settings/avform.cpp
```diff
@@ -116,6 +117,12 @@ void AVForm::showEvent(QShowEvent* event)
     createVideoSurface();
     getVideoDevices();
 
+    // To don't block GUI thread.
+    QtConcurrent::run([this]()
+    {
+        camVideoSurface->setSource(&camera);
```
BiTOk: Is this really can be called from separate thread?

Diadlo: I don't find any reason, why not

BiTOk: Straight call qwidget from another thread usually is bad practice. Are you sure that setSource is thread-safe?

antis81: @diadlo The actual problem here is: You're wrapping one problem in another one. However you can move the `QtConcurrent` call inside the setter and make that completely "self-managed", like this:

``` c++
void VideoSurface::setSource(VideoSource *src)
{
    if (source == src)
        return;

    QtConcurrent::run([this]()
    {
        // note:
        // The actual problem is inside the (un-)subscribe methods,
        // which are long running operations.
        // It is a good idea to turn "source" into a 
        // `QExplicitlySharedDataPointer`, doing the
        // subscription in c'tor/d'tor of a shared object
        // (inherit `VideoSource` from `QSharedData`)
        unsubscribe();
        source = src;
        subscribe();
    }
}
```

In theory, calling a QWidget's method, that doesn't update the GUI, is safe. Creating a QWidget itself within thread context, other than the main thread (aka "GUI thread") leads to an assertion: `ASSERT failure in QWidget: "Widgets must be created in the GUI thread."` Why? Because rendering is not thread-safe and would cause a race, when updating the widget. Thus, it is not only bad practice, but it doesn't work at all.

Diadlo: But this functions don't render anything

antis81: This is exactly what I said. Just move that code and it will be fine. :)

Diadlo: @antis81 Interesting. After move `QtConcurrent::run` into `VideoSurface::setSource`, returned old behavior.

TheNain38: @Diadlo I don't know how `QtConcurrent::run` works, but maybe the function can't exit, when there's still some code inside a `QtConcurrent::run` executing?

sudden6:
@TheNain38 that would completely defeat the purpose of another thread :)

src/widget/form/settings/avform.cpp
sudden6: @TheNain38 that would completely defeat the purpose of another thread :)

TheNain38: @sudden6 Yes, but the fact that the old behavior comes back... is weird, so, I thought of something weird ^^

Diadlo:
Fixed. But switching many times on/out av-tab creates queue and small freeze. But I think is not a problem for real user-case

src/video/videosurface.cpp
```diff
@@ -68,9 +70,14 @@ void VideoSurface::setSource(VideoSource *src)
     if (source == src)
         return;
 
-    unsubscribe();
-    source = src;
-    subscribe();
+    // To don't block GUI thread.
+    QtConcurrent::run([this, src]()
```
antis81, line 86: You don't require "this" here.

Diadlo, line 86: `bigLock`, `(un)subscribe`, `source` non-static.
They can't be called/setted without object.

src/video/videosurface.cpp
```diff
@@ -68,9 +70,14 @@ void VideoSurface::setSource(VideoSource *src)
     if (source == src)
         return;
 
-    unsubscribe();
-    source = src;
-    subscribe();
+    // To don't block GUI thread.
+    QtConcurrent::run([this, src]()
+    {
+        QMutexLocker l(&bigLock);
```
antis81, line 88: This is wrong. QtConcurrent is designed to be used without external mutexes. Following the code, you will find `bigLock` is relocked, which should cause a race to happen.

Btw. Why do you want to wrap an already running threaded operation in another thread. That totally makes no sense at all. Looks more like something is wrong with the underlying threads, so that the GUI thread is blocked.

Diadlo, line 88: I want to wrap `subscribe` in another thread because it call `CameraSource::openDevice()` for camera, which can load camera very long (and if had problem -- forever)

Diadlo, line 88: > QtConcurrent is designed to be used without external mutexes.

@antis81 Do you have any proof?

antis81, line 88: > Do you have any proof?

First, here's what the docs state:

``` text
The QtConcurrent namespace provides high-level APIs that make it possible to write multi-threaded programs without using low-level threading primitives.
…
```

That means, it is designed to not use any mutexes or semaphores! So if you feel the need to lock a mutex here, something's wrong… :)

Second, when using a mutex, you need to know exactly when to lock and unlock it to avoid very hard to fix races (deadlock)! In this specific case I wonder why it even works, because if you follow the flow into `CameraSource::subscribe` (and unsubscribe respectively), you'll find, that it should actually relock `bigLock`, causing a deadlock:

``` c++
bool CameraSource::subscribe()
{
    QMutexLocker l{&biglock};

    if (!_isOpen)
    {
        ++subscriptions;
        return true;
    }

    if (openDevice())
    {
        ++subscriptions;
        return true;
    }
    else
    {
        while (device && !device->close()) {}
        device = nullptr;
        cctx = cctxOrig = nullptr;
        videoStreamIndex = -1;
        // Memfence so the stream thread sees a nullptr device
        std::atomic_thread_fence(std::memory_order_release);
        return false;
    }

}
```

This already runs in a non-blocking context! So the question here is: What blocks the main thread?

Anyway, I found another QtConcurrent::run in another place in code. Maybe you can take some inspriation from there:

``` c++
ToxFriendCall::~ToxFriendCall()
{
    if (timeoutTimer)
        delete timeoutTimer;

    if (videoEnabled)
    {
        // This destructor could be running in a toxav callback while holding toxav locks.
        // If the CameraSource thread calls toxav *_send_frame, we might deadlock the toxav and CameraSource locks,
        // so we unsuscribe asynchronously, it's fine if the webcam takes a couple milliseconds more to poweroff.
        QtConcurrent::run([](){CameraSource::getInstance().unsubscribe();});
        …
    }
}
```

Probably it helps to move the `QtConcurrent::run` inside `Videsurface::subscribe(source)` and call it like so:

``` c++
VideoSurface::setSource(src) {
    if (src != source) {
         unsubscribe();

         QtConcurrent::run([src]() {
             subscribe(src);
         }
         // note: you can also think of running subscribe asynchronously in general and move the `QtConcurrent::run` inside.
    }
}

Hope that helps.


```

src/video/videosurface.cpp
antis81, line 86: See comment below… :)

antis81:
@Diadlo By @initramfs recent internal refactor/redesign activities concerning video resource management, it now doesn't hang the UI anymore an shows "instantly", which makes this PR obsolete. Please try it for yourself and give a short feedback, cause I like to close this one.

Diadlo:
@antis81 For me it is don't fixed. And now (on notebook) I have infinite freeze on go away from av tab (I think, depend on #3575)

initramfs:
@Diadlo The commit hash referenced in issue #3575 refers to a qTox revision before my videoframe refactor was merged. Does your deadlock issue still present on the current `HEAD` of master (63db185 as of writing)? If so, could you update the information in issue #3575 to indicate the new line numbers (blocking points)?

@antis81 Prior to my videoframe refactor the video loading time was: camerasource lock acquisition + video device loading time. Where the lock acquisition could block for a very long time. My PR mostly got rid of that. However, the actual video device loading time has not been eliminated (this is mostly a function of ffmpeg). I personally still experience a loading time of several seconds when selecting a desktop screen source (I have a 4K panel) which I think somewhat justifies the intent of this PR.

---

Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3540#-:-KPD9IefqzC8xPtNAETT:baw6xwy)_

Diadlo:
@initramfs Yes, still deadlock. Information was updated

PR #3543, opened 2016-07-24
docs(CONTRIBUTING.md): make more explicit what commit message needs

zetok:
By clearly stating what commit message needs to have.
Should help with commit messages that aren't really clear.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,8 @@ Qt:
 * Provide **screenshots** for UI related changes.
 * Keep your git commit history **clean** and **precise**. Commits like
   `xxx fixup` should not appear.
+* Commit message should state not only what has been changed, but also why a
+  change is needed.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -39,6 +39,8 @@ Qt:
 * Provide **screenshots** for UI related changes.
 * Keep your git commit history **clean** and **precise**. Commits like
   `xxx fixup` should not appear.
+* Commit message should state not only what has been changed, but also why
```
sudden6: I would write "why a change is needed". Else LGTM.

zetok: Well, there's nothing else out there that would say that commit message should say what is being changed – above there is a sentence about PR message, but not really about commits..

sudden6: eh, I'm just referring to the last part of the sentence. "what has been changed, but also why a change is needed" sounds better.

PR #3551, opened 2016-07-24
fix(translator): Added layout direction reset on translation.

### LINKED ISSUE #3542: Arabic language mirror window horizontal
##### Brief Description

OS: Arch Linux
qTox version: v1.4.0-2-g24a7f4d.
Commit hash: 24a7f4d51960a73cb7234cec9a4a2811e9175c6c
Qt: 5.7.0

Reproducible: Always
##### Steps to reproduce
1. Change language on Arabic
2. Change language on English
##### Observed Behavior

Window won't reset state
##### Expected Behavior

Resetting window state
##### Additional Info

Some languages reset state.
spectacle m15389

Diadlo:
Fix #3542.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3551)

### DIFF
--- a/src/widget/translator.cpp
+++ b/src/widget/translator.cpp
@@ -55,7 +55,8 @@ void Translator::translate()
             // system menu translation
             QTranslator *qtTranslator = new QTranslator();
             QString s_locale = "qt_"+locale;
-            if (qtTranslator->load(s_locale, QLibraryInfo::location(QLibraryInfo::TranslationsPath)))
+            QString location = QLibraryInfo::location(QLibraryInfo::TranslationsPath);
+            if (qtTranslator->load(s_locale, location))
             {
                 QApplication::installTranslator(qtTranslator);
                 qDebug() << "System translation loaded" << locale;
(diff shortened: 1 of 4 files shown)


src/widget/translator.cpp
```diff
@@ -72,6 +73,11 @@ void Translator::translate()
         QCoreApplication::installTranslator(translator);
     }
 
+    // Added to fix #3542
```
zetok: rather than mentioning issue number, it's better to describe what it actually fixes

src/widget/translator.cpp
```diff
@@ -72,6 +73,11 @@ void Translator::translate()
...
+    QSet<QString> rtlLanguages = {"ar", "he"};
```
zetok: hardcoding languages doesn't seem like a good idea, especially given that every time a rtl language will be added, this will need to be updated

Diadlo: May be add this?

```
QApplication::tr("QT_LAYOUT_DIRECTION",
                         "Translate this string to the string 'LTR' in left-to-right"
                         " languages or to 'RTL' in right-to-left languages (such as Hebrew"
                         " and Arabic) to get proper widget layout.");
```

zetok: If it will work, why not. Don't depend on this value being correct though – i.e. check should be for `RTL`, otherwise assume `LTR`.

`\"Translate\" this to 'LTR' in left-to-right` ?
`s/such as/for example/`, or `e.g.` ?

zetok: Actually, the comment should say that only if language is RTL it needs to be translated to `RTL`, otherwise they don't need to "translate" it.

Also, among qTox langs Uighur appears to be RLT too.

Diadlo:
Hm... Don't work for me. May be I did something wrong?

src/widget/translator.cpp
```diff
@@ -72,6 +73,15 @@ void Translator::translate()
         QCoreApplication::installTranslator(translator);
     }
 
+    // After change language from rtl to ltr not always restored layout direction
```
sudden6: `After the language is changed from RTL to LTR, the layout direction isn't always restored` ?

PR #3563, opened 2016-07-30
Fix clear log

### LINKED ISSUE #3275: Do not clear in-progress download/upload status message
##### Brief Description

OS: Windows 7 64
qTox version: v1.4.0-29-g08263e9
Qt:  5.5.1
…

Reproducible: Always
##### Steps to reproduce
1. I am Alice.
2. I clicked Bob, and send some messages.
3. I select "e.docx", and drag it from explorer to qTox's chat window.
4. Bob accept downloading(by auto-accept).
5. Bob right-click chat window and click "Clear displayed msg" while downloading(e.g., 48%).
##### Observed Behavior

Bob: All logs are gone(expected), but downloading-in-progress files are also gone from chat window(!).
Alice: Transfer is still going on.
##### Expected Behavior

When Bob clear his chat window, uploading-in-progress OR downloading-in-progress files
should not disappear from chat window.

Diadlo:
Fix #3275.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3563)

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -20,6 +20,8 @@
 #include "chatlog.h"
 #include "chatmessage.h"
 #include "chatlinecontent.h"
+#include "chatlinecontentproxy.h"
+#include "content/filetransferwidget.h"
 #include "src/widget/translator.h"
 
 #include <QDebug>
@@ -560,11 +562,20 @@ void ChatLog::clear()
 {
     clearSelection();
 
+    QVector<ChatLine::Ptr> savedLines;
+
     for (ChatLine::Ptr l : lines)
-        l->removeFromScene();
+    {
+        if (isActiveFileTransfer(l))
+            savedLines.push_back(l);
+        else
+            l->removeFromScene();
+    }
 
     lines.clear();
     visibleLines.clear();
+    for (ChatLine::Ptr l : savedLines)
+        insertChatlineAtBottom(l);
 
     updateSceneRect();
 }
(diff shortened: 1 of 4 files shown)


src/chatlog/chatlog.cpp
```diff
@@ -560,11 +562,19 @@ void ChatLog::clear()
 {
     clearSelection();
 
+    QVector<ChatLine::Ptr> savedLines;
+
     for (ChatLine::Ptr l : lines)
-        l->removeFromScene();
+        if (!isActiveFileTransfer(l))
```
sudden6: It would make more sense to remove the `!` and switch if and else parts. Removing the "not" removes one step of thinking -> easier :)

Diadlo: Oh.. of course :)

zetok:
>   feat(filetransferwidget): Added ability to check activity

This commit message is rather unclear & confusing.

I.e. no actual _feature_ is being added, there is only some code ~refactoring done in order to make a fix possible.

Since commits marked with `feat|fix` are being used in changelog, it probably ought to be preferable to limit their usage as much as possible, so that they would be used only for changes that users face. What do you think about it?

zetok:
on a4884487a59b5572f7fbebcb1da41aae6467db83:
- [x] segfault after I've clicked "clear messages" in sending client: https://gist.github.com/zetok/26667bd05d6c8b85fbb503f31446f927

Diadlo:
@zetok Some fixes. I believe, that fix your error

PR #3457, opened 2016-07-02
Setting rethink

Diadlo:
Fix #1069.

Tasks:
- [x] Remove `Reset to default settings` from `Porable`
- [X] Add implemenation `Reset to default settings` // provided that user will be asked whether they want to restart in order to reset settings
- [X] Move `Show contacts`, `Faux Offline Messaging`, `Play sound` and `Play sound while Busy` in `General`

**General tab**
spectacle i11800

**UI tab**
spectacle w11800

**Advanced tab**
spectacle e11800

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3457)

PR #3564, opened 2016-07-30
[WIP] create forms in the requested context

antis81:
This PR improves general resource management within the UI. While not the main purpose, It also has a positive effect on qTox's memory footprint. The goal is to mainly improve resource management, where camera & audio resources are requested/free'd.

So, instead of creating all widgets at program start, they are now created within the context and free'd, when "closed". That means, a widget is not in memory until the user requests it. To make it complete for settings-ui however, we need to introduce a widget factory, that creates the widget for a specific "settings tab".

TODO:
- [ ] Destroy widgets in "embedded window" mode, when they are replaced. (**big change**, but totally worth it) 
- [ ] Implement a widget factory for "settings tab" widgets, so that only the "visible" widget is in memory.

@Diadlo, @initramfs, @sudden6 FYI

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3564)

### DIFF
--- a/src/core/recursivesignalblocker.cpp
+++ b/src/core/recursivesignalblocker.cpp
@@ -1,0 +1,64 @@
+/*
+    Copyright © 2016 by The qTox Project
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "recursivesignalblocker.h"
+
+#include <QObject>
+#include <QSignalBlocker>
+
+namespace qTox {
+
+/**
+@class  RecursiveSignalBlocker
+@brief  Recursively blocks all signals from an object and its children.
+@note   All children must be created before the blocker is used.
+
+Wraps a QSignalBlocker on each object. Signals will be unblocked when the
+blocker gets destroyed. According to QSignalBlocker, we are also exception safe.
+*/
+
+/**
+@brief      Creates a QSignalBlocker recursively on the object and child objects.
+@param[in]  object  the object, which signals should be blocked
+*/
+RecursiveSignalBlocker::RecursiveSignalBlocker(QObject* object)
+{
+    recursiveBlock(object);
+}
+
+RecursiveSignalBlocker::~RecursiveSignalBlocker()
+{
+    qDeleteAll(mBlockers);
+}
+
+/**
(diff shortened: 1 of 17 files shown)


zetok:
on 0d8d46457c9ef6fb270ae1cb97f7fb2088740cd7
- [ ] lot of spam in stdout that looks like that
  
  ```
  [15:58:38.417 UTC] kernel/qobject.cpp:4657 : Warning: QObject::connect: invalid null parameter
  [15:58:38.472 UTC] kernel/qobject.cpp:4657 : Warning: QObject::connect: invalid null parameter
  [15:58:38.504 UTC] kernel/qobject.cpp:4657 : Warning: QObject::connect: invalid null parameter
  [15:58:38.536 UTC] kernel/qobject.cpp:4657 : Warning: QObject::connect: invalid null parameter
  ```
- [ ] there's no top settings icon when first clicked on settings for the first time, switching tabs brings them back
  no top bar
  with bar

I didn't really test whether changing settings did not regress.

antis81:
Thanks for running  this shortly.

> lot of spam in stdout that looks like that

Thanks -> solved! The magic happens in `Settings`, which now correctly emits the signals when the value changes. This solves quite some "connection knots" by centralizing the settings values. That means, instead of connecting slots to gui component (e.g. a QCheckBox), we connect to the settings value "directly". If that changes (and only then!), the value updates and notifies safely. This also works fine to connect independent widgets on the same settings value! Anyway, I did not consequently implement it for all the values (just for the ones affected by this PR), because it might interfere with Diadlo's PR (read below), that cleans up the settings.

> there's no top settings icon when first clicked on settings for the first time, switching tabs brings them back

I reverted a change and will solve that. Won't that title bar vanish soon anyway (see #3457)?

sudden6:
@antis81 I think you referenced a wrong issue

antis81:
@sudden6 Argh damn, once again, thanks. Will correct that of course.

src/core/recursivesignalblocker.cpp
```diff
@@ -0,0 +1,64 @@
+/*
+    Copyright © 2016-2017 by The qTox Project
```
initramfs: According to FSF & GPL, you can't "pre-license" a year in which the software hasn't been released for (2017). 

> The copyright notice should include the year in which you finished preparing the release (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and some were finished in 1999.

So I suggest you change the line to just:

```
Copyright © 2016 by The qTox Project
```

[Source link](https://www.gnu.org/licenses/gpl-howto.html)

src/core/recursivesignalblocker.h
```diff
@@ -0,0 +1,44 @@
+/*
+    Copyright © 2016-2017 by The qTox Project
```
initramfs: Same copyright issue as the other file.

zetok:
@antis81 

> Won't that title bar vanish soon anyway (see #3457)?

Yes, but this here is a bug, and provides only inconsistent behavior.

antis81:
@zetok Initially, I just wanted to create/destroy settings widget, but it totally does not work, because this crap `ContentLayout` comes in the way. It doesn't make sense to run it right now. Please stay tuned while I'm rewriting the layout stuff (~90% complete). It will also fix layout, when the friends list is near the right or left edge of the screen.

Issue #3572, opened 2016-07-31
Allow profile pictures directory to be customised


ghost:
When changing the profile picture it would be nice if the default directory for that could be customised within qtox.ini. That would allow me to have a directory with default profile images and the user could then just select what they want, rather than needing to navigate the home directory tree.

PR #3579, opened 2016-08-01
Add avatarsPath GUI option

ghost:
Allows you to set the path where avatar images exist. For example:

```
[GUI]
avatarsPath=/home/myusername/Pictures/Avatars
```

So when clicking on the profile picture it takes you straight to that directory.
#3572

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3579)


sudden6:
- Please follow: https://github.com/tux3/qTox/blob/master/CONTRIBUTING.md#commit-message-format
- Documentation for this option is needed in /doc/

I'm also not sure if having a config file only option is good or not. Other opinions on this?

ghost:
The documentation in /doc doesn't appear to cover the ini file settings.

sudden6:
Yeah, as they in general are only set by the GUI, but you have a special case here. Maybe it's time to document them? @zetok

zetok:
> I'm also not sure if having a config file only option is good or not. Other opinions on this?

I doubt whether this option is needed at all.

zetok:
Hi,

As you might be aware, repo has been moved under `qTox` org on github.

What does it have to do with your PR, you might ask.

Well, it doesn't change all that much, except that workflow has been slightly improved – now there are required checks that PR needs to pass in order to be merged.

Those checks are:
- travis (checks whether PR builds & commit message format)
- reviewable (requirement for PR to be reviewed & accepted)

Since osx travis was broken for a bit, you will need to rebase your PR on newest master to get it to pass travis check. Sorry for inconvenience.

Reminder: as always, PR needs to be possible to merge without conflicts. Make sure it does ;)

src/persistence/settings.cpp
```diff
@@ -223,6 +223,7 @@ void Settings::loadGlobal()
             else
                 style = "None";
         }
+		avatarsPath = s.value("avatarsPath",QDir::homePath()).toString();
```
agilob: This will always save QDir::homePath in settings.ini... I think that's not what you want, otherwise this PR is useless?

ghost: Right. I needed an extra setvalue.

agilob:
I don't understand point of this PR.

This doesn't allow to to set the path in GUI. It always sets QDir:home path as default, and even if you change it manually in settings.ini it will be overridden by QDir:home as soon as you change any other setting. 

It never **writes** `avatarPath`? There is no statement where it writes any value of `avatarPath` to settings...

so it also doesn't not handle empty string, invalid paths (non-existing, no-permissions, remote drives). 

[discussion] I'm not convinced about usage of writeGlobal, shouldn't avatar path be profile-specific from now?

Unfortunately, I'm strongly against merging this in provided state.

src/persistence/settings.cpp
```diff
@@ -235,7 +235,7 @@ void Settings::loadGlobal()
             else
                 style = "None";
         }
-		avatarsPath = s.value("avatarsPath",QDir::homePath()).toString();
+        avatarsPath = s.value("avatarsPath",QDir::homePath()).toString();
```
agilob: I think here you should use `setAvatarPath` function method which would validate that the path exists and is writeable, I see big potential for breaking stuff here, as this path will be used by user and qtox and toxcore.

ghost:
To explain the context. I'm making mesh images with qTox pre-installed. qtox.ini is part of the setup and so exists before the user does anything. When the user opens qTox and clicks on their profile image to change it they're taken straight to the correct folder and can then select their avatar.

Note that because the system is not connected to the internet the user can't just download some image from some internet site. It needs to already exist within the default disk image.

This makes things maximally simple for the user, and they don't need to explore the directory structure hunting for images. The fact that the path isn't settable within the GUI doesn't matter in this case, although potentially it could be.

If there's a nice image selection dialog within Qt then I might add another patch to use that and improve the user experience a little more, instead of the file selection dialog.

src/persistence/settings.cpp
ghost: Ah, I just followed the convention in the lines above it.

agilob:
> This makes things maximally simple for the user, and they don't need to explore the directory structure hunting for images. The fact that the path isn't settable within the GUI doesn't matter in this case, although potentially it could be.

I'm against adding new filed in GUI to set the path. I agree with you, this should not be settable by user manually. 

I just noticed one thing, not sure what will happen when:
1. First install of qTox
2. `avatarsPath` is not in default qtox.ini file
3. It's not set by settings reader, so will it be null or empty?

I think you need to add that validation thing there, add `avatarPath=` in default ini file we have in rsc folder and handle empty value in setter? Sorry, I'm not trying to be picky or against you, I think this is important change that can break stuff. Let's wait for review from @antis81 and testing from someone.

ghost:
Is there an existing path validation function? I suppose I could just check that the path exists before changing it.

ghost:
There is a settings.ini within the res directory, but if I was to set a default avatarsPath there I don't know what I'd set it to, because the home directory may be somewhat unique to each install.

agilob:
@bashrc that's why you need to handle null value and empty string. I would just set it to nothing `avatarPath=` and handle it in code.

zetok:

Diadlo:

PR #3587, opened 2016-08-04
fix(gui): don't quit application before last event

antis81:
Quitting the application kills the event loop. So any event past that will not be handled. ~~Probably causes the "BAD!" entries in log.~~

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3587)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -583,8 +583,8 @@ void Widget::closeEvent(QCloseEvent *event)
         }
         saveWindowGeometry();
         saveSplitterGeometry();
-        qApp->exit(0);
         QWidget::closeEvent(event);
+        qApp->quit();
     }
 }
 


initramfs:
Not that it really ratters to this PR, but I see get the "BAD!" entries with this PR applied. Upon investigation they come from code in the v4l2 `getPixelFormatString()` function, which seems to be called by `CameraDevice`'s `open()` function...

antis81:
@initramfs Good to know, thanks! Just wondered why they appear after `Debug: Clean exit with status 0` message. I'll remove that assumption from description - nevermind.

initramfs:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3587#-:-KO_tC5At3xFxl9sfTRv:bnfp4nl)_

antis81:
Thanks! Will merge.

PR #3577, opened 2016-08-01
Docs refactor

Diadlo:
Fix #3559.

---

This change is [Reviewable](https://reviewable.io/reviews/tux3/qtox/3577)

Issue #3559, opened 2016-07-29
[Discussion] Issues with new code commenting style


initramfs:
Apologies for posting this here rather than in the [comment style change PR](https://github.com/tux3/qTox/pull/3536), I didn't see it until a few hours ago when it was already merged.

I'm in the processing of rebasing my code to the style introduced in PR #3536 when I found a few issues with the new style. First of all I would like to state that evaluating pros/cons, I agree with putting comments in the source files for **most cases**, but not all.

In particular, I would like to make the argument against putting comments for class descriptions (e.g. comments tagged with `@class`) and comments for variable descriptions (e.g. comments tagged with `@var`) in source files.

First of all, class comments basically declare the outline of a class, the functionality and the intended purpose or role for the class. These concepts usually do not change very often making header-triggered recompilation times minimal compared to other changes. Furthermore, it is very likely that a documentation change for the class changes at least some part of the header as well (whether this is a new function, refactoring of existing functions or the change in the parameters of certain functions) which would result in header-trigged recompliation regardless of where the documentation is placed.

By the change in PR #3536 class documentations just kinda float around the top of source files, making getting to them harder (functions are fine since QtCreator has a "Follow Symbol Under Cursor" option that takes you between the source and header for the function, allowing easy preview). Additionally, in the case of nested classes or multiple classes in one header, things get weird (do you put both class definitions at the top of the file? Or do you put the second/nested class documentation at the start of it's function implementions/constructors).

I've mentioned nested classes specifically because in one of my PRs I have a case of a nested class which is completely delared in the header (no functions declared in source). Do I isolate the class documention from the rest of the class and put it in the source file? That seems kinda strange.

Moving on to variable documentation, I believe that member variable documention must be kept right at the point of declaration _because_ of it's volatility. The fact that members are separated from their documentation, in my opinion, will only lead to changes in name/usage/type/role of the member variable that becomes inconsistent with the documentation due to failure to update documentation (people might not notice there is documentation for the particular variable for example). The same idea applies to nested/multiple classes, where do those variable documentations go?

Again, I made this point when I was rebasing my PR to the changes in master, having to cross-reference repeatedly between the header and source whether a particular variable was changed, renamed or deleted (within my PR).

A third but minor point regards the style of commenting used. From the codebase I've seen:

```
/**
 * Comment
 */
```

and

```
/// Comment
```

styles being used, in which the former is what QtCreator defaults to (at least on my system). The style of comment implemented by the PR is:

```
/**
Comment
[...]

PR #3590, opened 2016-08-04
add settings change notification

antis81:
This PR turns our settings class into a "fully reflected" database. That means, we can now connect to it and get application wide notification signals, when a value changes. You can now stop to connect slots to a button or a checkbox within `SettingsWidget` and rely on `Settings` instead! ~~The PR itself does not change or correct any existing behaviour despite from few absolutely required changes.~~ **As some topics got corrected during the review process, it is just consequent to re-connect the corrected signals and get the values directly from settings changes.**

@initramfs Could you review please? I know it's huge, but we really need it - and it helps you as well with the video settings! :smile:

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3590)

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -364,7 +364,7 @@ bool Audio::initInput(const QString& deviceName)
         return false;
     }
 
-    d->setInputGain(Settings::getInstance().getAudioInGain());
+    d->setInputGain(Settings::getInstance().getAudioInGainDecibel());
 
     qDebug() << "Opened audio input" << deviceName;
     alcCaptureStart(alInDev);
--- a/src/chatlog/chatlog.h
+++ b/src/chatlog/chatlog.h
@@ -52,7 +52,6 @@ public:
     void setTypingNotificationVisible(bool visible);
     void scrollToLine(ChatLine::Ptr line);
     void selectAll();
-    void forceRelayout();
 
     QString getSelectedText() const;
 
@@ -66,6 +65,13 @@ public:
 
 signals:
     void selectionChanged();
+
+public slots:
+    void forceRelayout();
+
+private slots:
+    void onSelectionTimerTimeout();
+    void onWorkerTimeout();
 
 protected:
     QRectF calculateSceneRect() const;
@@ -99,10 +105,6 @@ protected:
     void updateBusyNotification();
 
     ChatLine::Ptr findLineByPosY(qreal yPos) const;
-
-private slots:
-    void onSelectionTimerTimeout();
-    void onWorkerTimeout();
 
 private:
     void retranslateUi();
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -56,7 +56,7 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString &sender, const QSt
     text = detectQuotes(detectAnchors(text), type);
 
     //text styling
-    if (Settings::getInstance().getStylePreference() != NONE)
+    if (Settings::getInstance().getStylePreference() != Settings::StyleType::NONE)
         text = detectStyle(text);
 
     switch(type)
@@ -229,7 +229,7 @@ QString ChatMessage::detectStyle(const QString &str)
         {
             int mul = 0; // Determines how many characters to strip from text
             // Set mul depending on styleownPreference
-            if (Settings::getInstance().getStylePreference() == WITHOUT_CHARS)
+            if (Settings::getInstance().getStylePreference() == Settings::StyleType::WITHOUT_CHARS)
                 mul = 2;
 
             // Match captured string to corresponding style format
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -122,8 +122,8 @@ void Core::makeTox(QByteArray savedata)
     // IPv6 needed for LAN discovery, but can crash some weird routers. On by default, can be disabled in options.
     bool enableIPv6 = Settings::getInstance().getEnableIPv6();
     bool forceTCP = Settings::getInstance().getForceTCP();
-    ProxyType proxyType = Settings::getInstance().getProxyType();
-    int proxyPort = Settings::getInstance().getProxyPort();
+    Settings::ProxyType proxyType = Settings::getInstance().getProxyType();
+    quint16 proxyPort = Settings::getInstance().getProxyPort();
     QString proxyAddr = Settings::getInstance().getProxyAddr();
     QByteArray proxyAddrData = proxyAddr.toUtf8();
 
@@ -147,7 +147,7 @@ void Core::makeTox(QByteArray savedata)
     toxOptions.savedata_data = (uint8_t*)savedata.data();
     toxOptions.savedata_length = savedata.size();
 
-    if (proxyType != ProxyType::ptNone)
+    if (proxyType != Settings::ProxyType::ptNone)
     {
         if (proxyAddr.length() > 255)
         {
@@ -157,9 +157,9 @@ void Core::makeTox(QByteArray savedata)
         {
             qDebug() << "using proxy" << proxyAddr << ":" << proxyPort;
             // protection against changings in TOX_PROXY_TYPE enum
-            if (proxyType == ProxyType::ptSOCKS5)
+            if (proxyType == Settings::ProxyType::ptSOCKS5)
                 toxOptions.proxy_type = TOX_PROXY_TYPE_SOCKS5;
-            else if (proxyType == ProxyType::ptHTTP)
+            else if (proxyType == Settings::ProxyType::ptHTTP)
                 toxOptions.proxy_type = TOX_PROXY_TYPE_HTTP;
 
             toxOptions.proxy_host = proxyAddrData.data();
@@ -592,7 +592,7 @@ void Core::requestFriendship(const QString& friendAddress, const QString& messag
         {
             qDebug() << "Requested friendship of "<<friendId;
             // Update our friendAddresses
-            Settings::getInstance().updateFriendAdress(friendAddress);
+            Settings::getInstance().updateFriendAddress(friendAddress);
             QString inviteStr = tr("/me offers friendship.");
             if (message.length())
                 inviteStr = tr("/me offers friendship, \"%1\"").arg(message);
@@ -1166,7 +1166,7 @@ bool Core::hasFriendWithPublicKey(const QString &pubkey) const
 QString Core::getFriendAddress(uint32_t friendNumber) const
 {
     QString id = getFriendPublicKey(friendNumber);
-    QString addr = Settings::getInstance().getFriendAdress(id);
+    QString addr = Settings::getInstance().getFriendAddress(id);
     if (addr.size() > id.size())
         return addr;
 
--- a/src/core/corestructs.cpp
+++ b/src/core/corestructs.cpp
@@ -7,17 +7,40 @@
 #define TOX_HEX_ID_LENGTH 2*TOX_ADDRESS_SIZE
 
 /**
-@file corestructs.h
-@brief Some headers use Core structs but don't need to include all of core.h
-
-They should include this file directly instead to reduce compilation times
-*/
+ * @file corestructs.h
+ * @brief Some headers use Core structs but don't need to include all of core.h
+ *
+ * They should include this file directly instead to reduce compilation times
+ *
+ *
+ * @var ToxFile::fileKind   Data file (default) or avatar
+ */
 
 /**
-@var uint8_t ToxFile::fileKind
-@brief Data file (default) or avatar
-*/
+ * @brief   Compare equal operator
+ * @param   other   the compared instance
+ * @return  true, if equal; false otherwise
+ */
+bool DhtServer::operator==(const DhtServer& other) const
+{
+    return this == &other ||
+            (port == other.port && address == other.address &&
+             userId == other.userId && name == other.name);
+}
 
+/**
+ * @brief   Compare not equal operator
+ * @param   other   the compared instance
+ * @return  true, if not equal; false otherwise
+ */
+bool DhtServer::operator!=(const DhtServer& other) const
+{
+    return !(*this == other);
+}
+
+/**
+ * @brief ToxFile constructor
+ */
 ToxFile::ToxFile(uint32_t fileNum, uint32_t friendId, QByteArray filename, QString filePath, FileDirection Direction)
     : fileKind{TOX_FILE_KIND_DATA}, fileNum(fileNum), friendId(friendId), fileName{filename},
       filePath{filePath}, file{new QFile(filePath)}, bytesSent{0}, filesize{0},
--- a/src/core/corestructs.h
+++ b/src/core/corestructs.h
@@ -15,6 +15,9 @@ struct DhtServer
     QString userId;
     QString address;
     quint16 port;
+
+    bool operator==(const DhtServer& other) const;
+    bool operator!=(const DhtServer& other) const;
 };
 
 struct ToxFile
--- a/src/persistence/db/plaindb.h
+++ b/src/persistence/db/plaindb.h
@@ -25,7 +25,7 @@
 #include <QSqlDatabase>
 
 namespace Db {
-    enum class syncType : int {stOff = 0, stNormal = 1, stFull = 2};
+    enum class syncType {stOff = 0, stNormal = 1, stFull = 2};
 }
 
 class PlainDb : public GenericDdInterface
--- a/src/persistence/historykeeper.h
+++ b/src/persistence/historykeeper.h
@@ -29,7 +29,10 @@
 
 class Profile;
 class GenericDdInterface;
-namespace Db { enum class syncType; }
+
+namespace Db {
+enum class syncType;
+}
 
 class HistoryKeeper
 {
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -171,7 +171,7 @@ void Settings::loadGlobal()
         autostartInTray = s.value("autostartInTray", false).toBool();
         closeToTray = s.value("closeToTray", false).toBool();
         forceTCP = s.value("forceTCP", false).toBool();
-        setProxyType(s.value("proxyType", static_cast<int>(ProxyType::ptNone)).toInt());
+        proxyType = static_cast<ProxyType>(s.value("proxyType", 0).toInt());
         proxyAddr = s.value("proxyAddr", "").toString();
         proxyPort = static_cast<quint16>(s.value("proxyPort", 0).toUInt());
         if (currentProfile.isEmpty())
@@ -198,7 +198,7 @@ void Settings::loadGlobal()
     s.endGroup();
 
     s.beginGroup("Advanced");
-        int sType = s.value("dbSyncType", static_cast<int>(Db::syncType::stFull)).toInt();
+        Db::syncType sType = static_cast<Db::syncType>(s.value("dbSyncType", 2).toInt());
         setDbSyncType(sType);
     s.endGroup();
 
@@ -701,8 +701,12 @@ const QList<DhtServer>& Settings::getDhtServerList() const
 void Settings::setDhtServerList(const QList<DhtServer>& newDhtServerList)
 {
     QMutexLocker locker{&bigLock};
-    dhtServerList = newDhtServerList;
-    emit dhtServerListChanged();
+
+    if (newDhtServerList != dhtServerList)
+    {
+        dhtServerList = newDhtServerList;
+        emit dhtServerListChanged(dhtServerList);
+    }
 }
 
 bool Settings::getEnableIPv6() const
@@ -714,7 +718,12 @@ bool Settings::getEnableIPv6() const
 void Settings::setEnableIPv6(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    enableIPv6 = newValue;
+
+    if (newValue != enableIPv6)
+    {
+        enableIPv6 = newValue;
+        emit enableIPv6Changed(enableIPv6);
+    }
 }
 
 bool Settings::getMakeToxPortable() const
@@ -726,15 +735,22 @@ bool Settings::getMakeToxPortable() const
 void Settings::setMakeToxPortable(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    QFile(getSettingsDirPath()+globalSettingsFile).remove();
-    makeToxPortable = newValue;
-    saveGlobal();
+
+    if (newValue != makeToxPortable)
+    {
+        QFile(getSettingsDirPath() + globalSettingsFile).remove();
+        makeToxPortable = newValue;
+        saveGlobal();
+
+        emit makeToxPortableChanged(makeToxPortable);
+    }
 }
 
 bool Settings::getAutorun() const
 {
-#ifdef QTOX_PLATFORM_EXT
     QMutexLocker locker{&bigLock};
+
+#ifdef QTOX_PLATFORM_EXT
     return Platform::getAutorun();
 #else
     return false;
@@ -745,7 +761,14 @@ void Settings::setAutorun(bool newValue)
 {
 #ifdef QTOX_PLATFORM_EXT
     QMutexLocker locker{&bigLock};
-    Platform::setAutorun(newValue);
+
+    bool autorun = Platform::getAutorun();
+
+    if (newValue != autorun)
+    {
+        Platform::setAutorun(newValue);
+        emit autorunChanged(autorun);
+    }
 #else
     Q_UNUSED(newValue);
 #endif
@@ -766,7 +789,12 @@ QString Settings::getStyle() const
 void Settings::setStyle(const QString& newStyle)
 {
     QMutexLocker locker{&bigLock};
-    style = newStyle;
+
+    if (newStyle != style)
+    {
+        style = newStyle;
+        emit styleChanged(style);
+    }
 }
 
 bool Settings::getShowSystemTray() const
@@ -775,16 +803,26 @@ bool Settings::getShowSystemTray() const
     return showSystemTray;
 }
 
-void Settings::setShowSystemTray(const bool& newValue)
+void Settings::setShowSystemTray(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    showSystemTray = newValue;
+
+    if (newValue != showSystemTray)
+    {
+        showSystemTray = newValue;
+        emit showSystemTrayChanged(newValue);
+    }
 }
 
 void Settings::setUseEmoticons(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    useEmoticons = newValue;
+
+    if (newValue != useEmoticons)
+    {
+        useEmoticons = newValue;
+        emit useEmoticonsChanged(useEmoticons);
+    }
 }
 
 bool Settings::getUseEmoticons() const
@@ -796,7 +834,12 @@ bool Settings::getUseEmoticons() const
 void Settings::setAutoSaveEnabled(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    autoSaveEnabled = newValue;
+
+    if (newValue != autoSaveEnabled)
+    {
+        autoSaveEnabled = newValue;
+        emit autoSaveEnabledChanged(autoSaveEnabled);
+    }
 }
 
 bool Settings::getAutoSaveEnabled() const
@@ -808,7 +851,12 @@ bool Settings::getAutoSaveEnabled() const
 void Settings::setAutostartInTray(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    autostartInTray = newValue;
+
+    if (newValue != autostartInTray)
+    {
+        autostartInTray = newValue;
+        emit autostartInTrayChanged(autostartInTray);
+    }
 }
 
 bool Settings::getCloseToTray() const
@@ -820,7 +868,12 @@ bool Settings::getCloseToTray() const
 void Settings::setCloseToTray(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    closeToTray = newValue;
+
+    if (newValue != closeToTray)
+    {
+        closeToTray = newValue;
+        emit closeToTrayChanged(newValue);
+    }
 }
 
 bool Settings::getMinimizeToTray() const
@@ -832,7 +885,12 @@ bool Settings::getMinimizeToTray() const
 void Settings::setMinimizeToTray(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    minimizeToTray = newValue;
+
+    if (newValue != minimizeToTray)
+    {
+        minimizeToTray = newValue;
+        emit minimizeToTrayChanged(minimizeToTray);
+    }
 }
 
 bool Settings::getLightTrayIcon() const
@@ -844,7 +902,12 @@ bool Settings::getLightTrayIcon() const
 void Settings::setLightTrayIcon(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    lightTrayIcon = newValue;
+
+    if (newValue != lightTrayIcon)
+    {
+        lightTrayIcon = newValue;
+        emit lightTrayIconChanged(lightTrayIcon);
+    }
 }
 
 bool Settings::getStatusChangeNotificationEnabled() const
@@ -856,7 +919,12 @@ bool Settings::getStatusChangeNotificationEnabled() const
 void Settings::setStatusChangeNotificationEnabled(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    statusChangeNotificationEnabled = newValue;
+
+    if (newValue != statusChangeNotificationEnabled)
+    {
+        statusChangeNotificationEnabled = newValue;
+        emit statusChangeNotificationEnabledChanged(statusChangeNotificationEnabled);
+    }
 }
 
 bool Settings::getShowInFront() const
@@ -868,7 +936,12 @@ bool Settings::getShowInFront() const
 void Settings::setShowInFront(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    showInFront = newValue;
+
+    if (newValue != showInFront)
+    {
+        showInFront = newValue;
+        emit showInFrontChanged(showInFront);
+    }
 }
 
 bool Settings::getNotifySound() const
@@ -880,7 +953,12 @@ bool Settings::getNotifySound() const
 void Settings::setNotifySound(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    notifySound = newValue;
+
+    if (newValue != notifySound)
+    {
+        notifySound = newValue;
+        emit notifySoundChanged(notifySound);
+    }
 }
 
 bool Settings::getBusySound() const
@@ -892,7 +970,12 @@ bool Settings::getBusySound() const
 void Settings::setBusySound(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    busySound = newValue;
+
+    if (newValue != busySound)
+    {
+        busySound = newValue;
+        emit busySoundChanged(busySound);
+    }
 }
 
 bool Settings::getGroupAlwaysNotify() const
@@ -904,7 +987,12 @@ bool Settings::getGroupAlwaysNotify() const
 void Settings::setGroupAlwaysNotify(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    groupAlwaysNotify = newValue;
+
+    if (newValue != groupAlwaysNotify)
+    {
+        groupAlwaysNotify = newValue;
+        emit groupAlwaysNotifyChanged(groupAlwaysNotify);
+    }
 }
 
 QString Settings::getTranslation() const
@@ -913,10 +1001,15 @@ QString Settings::getTranslation() const
     return translation;
 }
 
-void Settings::setTranslation(QString newValue)
+void Settings::setTranslation(const QString& newValue)
 {
     QMutexLocker locker{&bigLock};
-    translation = newValue;
+
+    if (newValue != translation)
+    {
+        translation = newValue;
+        emit translationChanged(translation);
+    }
 }
 
 void Settings::deleteToxme()
@@ -942,11 +1035,22 @@ QString Settings::getToxmeInfo() const
     return toxmeInfo;
 }
 
-void Settings::setToxmeInfo(QString info)
+void Settings::setToxmeInfo(const QString& info)
 {
     QMutexLocker locker{&bigLock};
-    if (info.split("@").size() == 2)
-        toxmeInfo = info;
+
+    if (info != toxmeInfo)
+    {
+        if (info.split("@").size() == 2)
+        {
+            toxmeInfo = info;
+            emit toxmeInfoChanged(toxmeInfo);
+        }
+        else
+        {
+            qWarning() << info << "is not a valid toxme string -> value ignored.";
+        }
+    }
 }
 
 QString Settings::getToxmeBio() const
@@ -955,10 +1059,15 @@ QString Settings::getToxmeBio() const
     return toxmeBio;
 }
 
-void Settings::setToxmeBio(QString bio)
+void Settings::setToxmeBio(const QString& bio)
 {
     QMutexLocker locker{&bigLock};
-    toxmeBio = bio;
+
+    if (bio != toxmeBio)
+    {
+        toxmeBio = bio;
+        emit toxmeBioChanged(toxmeBio);
+    }
 }
 
 bool Settings::getToxmePriv() const
@@ -970,7 +1079,12 @@ bool Settings::getToxmePriv() const
 void Settings::setToxmePriv(bool priv)
 {
     QMutexLocker locker{&bigLock};
-    toxmePriv = priv;
+
+    if (priv != toxmePriv)
+    {
+        toxmePriv = priv;
+        emit toxmePrivChanged(toxmePriv);
+    }
 }
 
 QString Settings::getToxmePass() const
@@ -979,10 +1093,17 @@ QString Settings::getToxmePass() const
     return toxmePass;
 }
 
-void Settings::setToxmePass(const QString &pass)
+void Settings::setToxmePass(const QString& pass)
 {
     QMutexLocker locker{&bigLock};
-    toxmePass = pass;
+
+    if (pass != toxmePass)
+    {
+        toxmePass = pass;
+
+        // password is not exposed for security reasons
+        emit toxmePassChanged();
+    }
 }
 
 bool Settings::getForceTCP() const
@@ -994,7 +1115,12 @@ bool Settings::getForceTCP() const
 void Settings::setForceTCP(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    forceTCP = newValue;
+
+    if (newValue != forceTCP)
+    {
+        forceTCP = newValue;
+        emit forceTCPChanged(forceTCP);
+    }
 }
 
 QNetworkProxy Settings::getProxy() const
@@ -1022,19 +1148,21 @@ QNetworkProxy Settings::getProxy() const
     return proxy;
 }
 
-ProxyType Settings::getProxyType() const
+Settings::ProxyType Settings::getProxyType() const
 {
     QMutexLocker locker{&bigLock};
     return proxyType;
 }
 
-void Settings::setProxyType(int newValue)
+void Settings::setProxyType(ProxyType newValue)
 {
     QMutexLocker locker{&bigLock};
-    if (newValue >= 0 && newValue <= 2)
-        proxyType = static_cast<ProxyType>(newValue);
-    else
-        proxyType = ProxyType::ptNone;
+
+    if (newValue != proxyType)
+    {
+        proxyType = newValue;
+        emit proxyTypeChanged(proxyType);
+    }
 }
 
 QString Settings::getProxyAddr() const
@@ -1046,7 +1174,12 @@ QString Settings::getProxyAddr() const
 void Settings::setProxyAddr(const QString& newValue)
 {
     QMutexLocker locker{&bigLock};
-    proxyAddr = newValue;
+
+    if (newValue != proxyAddr)
+    {
+        proxyAddr = newValue;
+        emit proxyAddressChanged(proxyAddr);
+    }
 }
 
 quint16 Settings::getProxyPort() const
@@ -1058,7 +1191,12 @@ quint16 Settings::getProxyPort() const
 void Settings::setProxyPort(quint16 newValue)
 {
     QMutexLocker locker{&bigLock};
-    proxyPort = newValue;
+
+    if (newValue != proxyPort)
+    {
+        proxyPort = newValue;
+        emit proxyPortChanged(proxyPort);
+    }
 }
 
 QString Settings::getCurrentProfile() const
@@ -1073,11 +1211,17 @@ uint32_t Settings::getCurrentProfileId() const
     return currentProfileId;
 }
 
-void Settings::setCurrentProfile(QString profile)
+void Settings::setCurrentProfile(const QString& profile)
 {
     QMutexLocker locker{&bigLock};
-    currentProfile = profile;
-    currentProfileId = makeProfileId(currentProfile);
+
+    if (profile != currentProfile)
+    {
+        currentProfile = profile;
+        currentProfileId = makeProfileId(currentProfile);
+        emit currentProfileChanged(currentProfile);
+        emit currentProfileIdChanged(currentProfileId);
+    }
 }
 
 bool Settings::getEnableLogging() const
@@ -1089,7 +1233,12 @@ bool Settings::getEnableLogging() const
 void Settings::setEnableLogging(bool newValue)
 {
     QMutexLocker locker{&bigLock};
-    enableLogging = newValue;
+
+    if (newValue != enableLogging)
+    {
+        enableLogging = newValue;
+        emit enableLoggingChanged(enableLogging);
+    }
 }
 
 Db::syncType Settings::getDbSyncType() const
@@ -1098,13 +1247,15 @@ Db::syncType Settings::getDbSyncType() const
     return dbSyncType;
 }
 
-void Settings::setDbSyncType(int newValue)
+void Settings::setDbSyncType(Db::syncType newValue)
 {
     QMutexLocker locker{&bigLock};
-    if (newValue >= 0 && newValue <= 2)
-        dbSyncType = static_cast<Db::syncType>(newValue);
-    else
-        dbSyncType = Db::syncType::stFull;
+
+    if (newValue != dbSyncType)
+    {
+        dbSyncType = newValue;
+        emit dbSyncTypeChanged(dbSyncType);
+    }
 }
 
 int Settings::getAutoAwayTime() const
@@ -1113,13 +1264,23 @@ int Settings::getAutoAwayTime() const
     return autoAwayTime;
 }
 
+/**
+ * @brief Sets how long the user may stay idle, before online status is set to "away".
+ * @param[in] newValue  the user idle duration in minutes
+ * @note Values < 0 default to 10 minutes.
+ */
 void Settings::setAutoAwayTime(int newValue)
 {
     QMutexLocker locker{&bigLock};
+
     if (newValue < 0)
         newValue = 10;
 
-    autoAwayTime = newValue;
+    if (newValue != autoAwayTime)
+    {
+        autoAwayTime = newValue;
+        emit autoAwayTimeChanged(autoAwayTime);
+    }
 }
 
 QString Settings::getAutoAcceptDir(const ToxId& id) const
@@ -1146,7 +1307,7 @@ void Settings::setAutoAcceptDir(const ToxId &id, const QString& dir)
     }
     else
     {
-        updateFriendAdress(id.toString());
+        updateFriendAddress(id.toString());
         setAutoAcceptDir(id, dir);
     }
 }
@@ -1174,7 +1335,7 @@ void Settings::setContactNote(const ToxId &id, const QString& note)
     }
     else
     {
-        updateFriendAdress(id.toString());
+        updateFriendAddress(id.toString());
         setContactNote(id, note);
     }
 }
@@ -1188,7 +1349,12 @@ QString Settings::getGlobalAutoAcceptDir() const
 void Settings::setGlobalAutoAcceptDir(const QString& newValue)
 {
     QMutexLocker locker{&bigLock};
-    globalAutoAcceptDir = newValue;
+
+    if (newValue != globalAutoAcceptDir)
+    {
+        globalAutoAcceptDir = newValue;
+        emit globalAutoAcceptDirChanged(globalAutoAcceptDir);
+    }
 }
 
 const QFont& Settings::getChatMessageFont() const
@@ -1200,13 +1366,24 @@ const QFont& Settings::getChatMessageFont() const
 void Settings::setChatMessageFont(const QFont& font)
 {
     QMutexLocker locker(&bigLock);
-    chatMessageFont = font;
+
+    if (font != chatMessageFont)
+    {
+        chatMessageFont = font;
+        emit chatMessageFontChanged(chatMessageFont);
+    }
 }
 
 void Settings::setWidgetData(const QString& uniqueName, const QByteArray& data)
 {
     QMutexLocker locker{&bigLock};
-    widgetSettings[uniqueName] = data;
+
+    if (!widgetSettings.contains(uniqueName) ||
+        widgetSettings[uniqueName] != data)
+    {
+        widgetSettings[uniqueName] = data;
+        emit widgetDataChanged(uniqueName);
+    }
 }
 
 QByteArray Settings::getWidgetData(const QString& uniqueName) const
@@ -1221,11 +1398,15 @@ QString Settings::getSmileyPack() const
     return smileyPack;
 }
 
-void Settings::setSmileyPack(const QString &value)
+void Settings::setSmileyPack(const QString& value)
 {
     QMutexLocker locker{&bigLock};
-    smileyPack = value;
-    emit smileyPackChanged();
+
+    if (value != smileyPack)
+    {
+        smileyPack = value;
+        emit smileyPackChanged(smileyPack);
+    }
 }
 
 int Settings::getEmojiFontPointSize() const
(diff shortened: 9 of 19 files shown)


initramfs:
@antis81 No problem. I'll do it sometime later today, it's getting late over here.

initramfs:
I've highlighted some formatting stuff that doesn't directly relate to your PR, but since you're already modifying the code around it, it'd be helpful to tweak the formatting as well.

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 26 unresolved discussions.

---

_[src/core/corestructs.h, line 19 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOP-PmkZl4otyV70UaC:-KOP-PmlrpAcOLBhyr7i:b-fim78b) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/core/corestructs.h#L19)):_

> ``` objc
>     quint16 port;
> 
>     bool operator==(const DhtServer& other) const
> ```

I think these should be implemented in the source cpp file.

---

_[src/core/corestructs.h, line 26 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOP-Y7o5EkIqdIF59Yz:-KOP-Y7o5EkIqdIF59Z-:b8eka3b) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/core/corestructs.h#L26)):_

> ``` objc
>     }
> 
>     bool operator!=(const DhtServer& other) const
> ```

This too.

---

_[src/persistence/settings.cpp, line 806 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOc5sdq7mYKFJcK1cb:-KOOc5selqv7RZkp29oF:b-ntdtyf) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L806)):_

> ``` cpp
> }
> 
> void Settings::setShowSystemTray(const bool& newValue)
> ```

Is there a reason why this function takes `const bool&` instead of just `bool`? (Actually, its probably the otherway around..)

---

_[src/persistence/settings.cpp, line 1004 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOd6fHfHYp-IDpIL7j:-KOOd6fHfHYp-IDpIL7k:b-q4w8co) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1004)):_

> ``` cpp
> }
> 
> void Settings::setTranslation(QString newValue)
> ```

Obtain via `const QString&` since you're not modifying it.

---

_[src/persistence/settings.cpp, line 1038 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOdQOwWewznu0V4BnQ:-KOOdQOxt5q1cgsClvS-:b2nreh6) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1038)):_

> ``` cpp
> }
> 
> void Settings::setToxmeInfo(QString info)
> ```

Same as above.

---

_[src/persistence/settings.cpp, line 1042 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOddVxVsu9x02RlkTW:-KOOddVxVsu9x02RlkTX:b-6p5fih) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1042)):_

> ``` cpp
>     QMutexLocker locker{&bigLock};
> 
>     if (info != toxmeInfo && info.split("@").size() == 2)
> ```

This function potentially discards a malformed toxme ID without informing the caller that it has done so. Whilst the caller can indeed, connect to the signal to see if a `toxmeInfoChanged` signal was given, this is not equivilent (and is quite a large hassle). I'd suggest either return `true`/`false` if the toxme ID was accepted or use a qWarning log line for malformed IDs.

Even better if you validate the given parameter before obtaining `biglock`, saves acquiring an unneccessary lock.

---

_[src/persistence/settings.cpp, line 1055 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOeY2uHKoM1jgrQN2u:-KOOeY2uHKoM1jgrQN2v:b-bhg6kf) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1055)):_

> ``` cpp
> }
> 
> void Settings::setToxmeBio(QString bio)
> ```

As same `const QString&` above.

---

_[src/persistence/settings.cpp, line 1089 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOegh5DbNf21z8CbsD:-KOOegh5DbNf21z8CbsE:b-3q5mz7) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1089)):_

> ``` cpp
> }
> 
> void Settings::setToxmePass(const QString &pass)
> ```

Coding style dictates `const QString& pass` rather than the one here.

---

_[src/persistence/settings.cpp, line 1151 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOfTOSXU8WvnP-I4lU:-KOOfTOSXU8WvnP-I4lV:b-9f0hmd) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1151)):_

> ``` cpp
>     if (newValue != proxyType)
>     {
>         proxyType = (newValue >= 0 && newValue <= 2)
> ```

This code potentially causes a `proxyTypeChanged()` signal to be emitted even when no change has occured. You perform bounds checking only after you checked it against the old value. For example:

Assume `proxyType = ProxyType::ptNone`
Call `setProxyType(4)`

The code will pass the if-statement to check whether it's the same as the old value, but then get assigned to the  `ProxyType::ptNone` (the old value) since it's out of range.

Perform the range check before the new-old value comparison. Perhaps add a warning to indicate an out of range value.

---

_[src/persistence/settings.cpp, line 1204 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOsaEfzS4Q-kYwqvXx:-KOOsaEg2kYq6lfQD_kB:b-9ftuw8) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1204)):_

> ``` cpp
> }
> 
> void Settings::setCurrentProfile(QString profile)
> ```

Same issue as `const String&`.

---

_[src/persistence/settings.cpp, line 1244 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOsn6VF4JqqBfzxxAa:-KOOsn6VF4JqqBfzxxAb:b9o075v) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1244)):_

> ``` cpp
>     QMutexLocker locker{&bigLock};
> 
>     if (static_cast<Db::syncType>(newValue) != dbSyncType)
> ```

Same as above, potential for emission of a `dbSyncTypeChanged()` signal even when there is no change if an out of range value is given.

---

_[src/persistence/settings.cpp, line 1265 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOtNyWsp8xHDOd9xZX:-KOOtNyWsp8xHDOd9xZY:b-e1afrd) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1265)):_

> ``` cpp
>     if (newValue != autoAwayTime)
>     {
>         autoAwayTime = newValue < 0 ? 10 : newValue;
> ```

Same as above, out of bounds value can cause needless signal.

---

_[src/persistence/settings.cpp, line 1385 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOuyldhcAE9x1VrBM6:-KOOuyldhcAE9x1VrBM7:b-v9x3ao) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1385)):_

> ``` cpp
> }
> 
> void Settings::setSmileyPack(const QString &value)
> ```

Formatting: `const QString& value`.

---

_[src/persistence/settings.cpp, line 1436 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvIUasOqhnuObq2hu:-KOOvIUasOqhnuObq2hv:b-pljm9r) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1436)):_

> ``` cpp
> }
> 
> void Settings::setSecondColumnHandlePosFromRight(int pos)
> ```

Any reason why this int is not `const` (or for that matter, why the int above (in `setFirstColumnHandlePos) _is_?)

---

_[src/persistence/settings.cpp, line 1453 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvWgiBH2lGCov0Mfk:-KOOvWgiBH2lGCov0Mfl:b-ez78q4) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1453)):_

> ``` cpp
> }
> 
> void Settings::setTimestampFormat(const QString &format)
> ```

Formatting: `const QString& format`.

---

_[src/persistence/settings.cpp, line 1470 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvamiTimSZRIsrjxk:-KOOvamiTimSZRIsrjxl:b-ez78q4) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1470)):_

> ``` cpp
> }
> 
> void Settings::setDateFormat(const QString &format)
> ```

Formatting: `const QString& format`.

---

_[src/persistence/settings.cpp, line 1504 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOveJ9Xgm3Xbx7ktb2:-KOOveJAGsnK38vrxDHE:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1504)):_

> ``` cpp
> }
> 
> void Settings::setWindowGeometry(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1521 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvk6DaiYb3zE7GeTx:-KOOvk6DaiYb3zE7GeTy:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1521)):_

> ``` cpp
> }
> 
> void Settings::setWindowState(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1572 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvr4m7JG67mfey2Lt:-KOOvr4m7JG67mfey2Lu:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1572)):_

> ``` cpp
> }
> 
> void Settings::setSplitterState(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1589 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvz8CvehU_KjONbj6:-KOOvz8CvehU_KjONbj7:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1589)):_

> ``` cpp
> }
> 
> void Settings::setDialogGeometry(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1606 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOw1Poky_EwcgwA_wf:-KOOw1Poky_EwcgwA_wg:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1606)):_

> ``` cpp
> }
> 
> void Settings::setDialogSplitterState(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1623 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOw2k1bUDnF9v79VGV:-KOOw2k2ml2WYJqjKZC2:bm2n0jc) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1623)):_

> ``` cpp
> }
> 
> void Settings::setDialogSettingsGeometry(const QByteArray &value)
> ```

Formatting: `const QByteArray& value`.

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KOOwQP6AWqhv3-3WzX-:b-uwtq33) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_

> ``` cpp
>     QMutexLocker locker{&bigLock};
> 
>     if (dB < audioInGainDecibel || dB > audioInGainDecibel)
> ```

Any reason why this is not just `dB != audioInGainDecibel`?

---

_[src/persistence/settings.cpp, line 1793 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwdcasBKCMJMOnmV6:-KOOwdcbgfgxGxaDpJqL:b17ub31) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1793)):_

> ``` cpp
> }
> 
> void Settings::setScreenRegion(const QRect &value)
> ```

Formatting: `const QRect& value`.

---

_[src/persistence/settings.cpp, line 1855 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOx-574DlTIdKkU6rD:-KOOx-58JK1G-1cqcK87:b-u2n8bv) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1855)):_

> ``` cpp
> }
> 
> QString Settings::getFriendAdress(const QString &publicKey) const
> ```

Formatting: `const QString& publicKey`.

---

_[src/persistence/settings.h, line 106 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOzUF3W9pPmHQlF911:-KOOzUF49Wv3Qt85zHml:bbrra0l) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.h#L106)):_

> ``` objc
>                WRITE setTypingNotification NOTIFY typingNotificationChanged
>                FINAL)
>     //Q_PROPERTY(Db::syncType dbSyncType READ WRITE NOTIFY FINAL);
> ```

Is this deliberately commented out? Doesn't look like it's done...

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590#-:-KOOziNnRdfvhv4SwvES:b31gue8)_

antis81:
Review status: 6 of 17 files reviewed at latest revision, 26 unresolved discussions.

---

_[src/core/corestructs.h, line 19 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOP-PmkZl4otyV70UaC:-KOPkUr4KxGizJM8MVJ4:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/core/corestructs.h#L19)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> I think these should be implemented in the source cpp file.
> </details>
> Done.

---

_[src/core/corestructs.h, line 26 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOP-Y7o5EkIqdIF59Yz:-KOPkViXRkrBn66_cijq:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/core/corestructs.h#L26)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> This too.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 806 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOc5sdq7mYKFJcK1cb:-KOPkXQit22V7PSNcGi5:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L806)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Is there a reason why this function takes `const bool&` instead of just `bool`? (Actually, its probably the otherway around..)
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1004 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOd6fHfHYp-IDpIL7j:-KOPkYMOyWG0DU7KjtAS:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1004)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Obtain via `const QString&` since you're not modifying it.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1038 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOdQOwWewznu0V4BnQ:-KOPkZpKP4HY-o4SqF0S:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1038)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same as above.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1042 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOddVxVsu9x02RlkTW:-KOPk_ewvLYRDGDfeq59:bbrtmah) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1042)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> This function potentially discards a malformed toxme ID without informing the caller that it has done so. Whilst the caller can indeed, connect to the signal to see if a `toxmeInfoChanged` signal was given, this is not equivilent (and is quite a large hassle). I'd suggest either return `true`/`false` if the toxme ID was accepted or use a qWarning log line for malformed IDs.
> 
> Even better if you validate the given parameter before obtaining `biglock`, saves acquiring an unneccessary lock.
> </details>
> I changed those types to "enum class", making the check obsolete and enforce type usage (or cast) in code. There could be a validation method for reading/writing to settings, but this is just as error prone and what is it good for? If the enum is invalid, it should actually kick one in the ass and tell to do it right in code. Btw. there's nothing in the world, that saves one from exchanging the int value meanings. So those checks are good for nothing actually. :) 

---

_[src/persistence/settings.cpp, line 1055 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOeY2uHKoM1jgrQN2u:-KOPlwQcEFp4MCrkp18k:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1055)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> As same `const QString&` above.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1089 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOegh5DbNf21z8CbsD:-KOPlyKh8Do9Rc1qzdbC:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1089)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Coding style dictates `const QString& pass` rather than the one here.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1151 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOfTOSXU8WvnP-I4lU:-KOPm-ZOcHp7DMB2HrOg:b-ioso83) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1151)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> This code potentially causes a `proxyTypeChanged()` signal to be emitted even when no change has occured. You perform bounds checking only after you checked it against the old value. For example:
> 
> Assume `proxyType = ProxyType::ptNone`
> Call `setProxyType(4)`
> 
> The code will pass the if-statement to check whether it's the same as the old value, but then get assigned to the  `ProxyType::ptNone` (the old value) since it's out of range.
> 
> Perform the range check before the new-old value comparison. Perhaps add a warning to indicate an out of range value.
> </details>
> Done.

(see previous comment on "enum class")

---

_[src/persistence/settings.cpp, line 1204 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOsaEfzS4Q-kYwqvXx:-KOPmAgf4ZgpGK-hNOk1:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1204)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same issue as `const String&`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1244 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOsn6VF4JqqBfzxxAa:-KOPmCgCsggVQP1dzS6m:b-sc6zot) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1244)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same as above, potential for emission of a `dbSyncTypeChanged()` signal even when there is no change if an out of range value is given.
> </details>
> Done.

(see previous comment about "enum class")

---

_[src/persistence/settings.cpp, line 1265 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOtNyWsp8xHDOd9xZX:-KOPmIo2DSWTFQK4OotP:b-sc6zot) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1265)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same as above, out of bounds value can cause needless signal.
> </details>
> Done.

(see previous comment about "enum class")

---

_[src/persistence/settings.cpp, line 1385 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOuyldhcAE9x1VrBM6:-KOPmPRaNv4v5ZbTTFEO:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1385)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QString& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1436 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvIUasOqhnuObq2hu:-KOPjASsTdIZKGgdoXbh:b-ameb6b) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1436)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Any reason why this int is not `const` (or for that matter, why the int above (in `setFirstColumnHandlePos) _is_?)
> </details>
> FYI: Both methods are dead code. Removing them. Otherwise there is no reason to pass a simple type "const" or by reference, unless the method should explicitly update a value. In this case, I've seen often a pointer to be used instead. I think, just because it is more clear.

---

_[src/persistence/settings.cpp, line 1453 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvWgiBH2lGCov0Mfk:-KOPmW2pA-8HqlkIuuXX:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1453)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QString& format`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1470 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvamiTimSZRIsrjxk:-KOPmwZvztzJZBlt4Ag1:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1470)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QString& format`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1504 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOveJ9Xgm3Xbx7ktb2:-KOPn2kkI7I7Eo45nvyM:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1504)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1521 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvk6DaiYb3zE7GeTx:-KOPnECtbsrUuttjO4Qt:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1521)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1572 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvr4m7JG67mfey2Lt:-KOPnFfaNwYgsOKly_uW:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1572)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1589 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvz8CvehU_KjONbj6:-KOPnxIjqa1Bn2P2JrJV:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1589)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1606 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOw1Poky_EwcgwA_wf:-KOPnzAXr6nt9_mWx9Q6:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1606)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1623 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOw2k1bUDnF9v79VGV:-KOPo0FqUTXwdAOAo3ZM:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1623)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QByteArray& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KOPo4M2pl_0dLlccXZv:b-514wa0) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Any reason why this is not just `dB != audioInGainDecibel`?
> </details>
> FYI: Yeah because llvm warns about floating point checks on `==` and `!=`.

---

_[src/persistence/settings.cpp, line 1793 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwdcasBKCMJMOnmV6:-KOPpQeGTqBCabQdGPur:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1793)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QRect& value`.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1855 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOx-574DlTIdKkU6rD:-KOPpj3V8BZs4IanaKlG:bxb8foa) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1855)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Formatting: `const QString& publicKey`.
> </details>
> Done.

Also fixed typo in "Adress" -> "Address"

---

_[src/persistence/settings.h, line 106 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOzUF3W9pPmHQlF911:-KOPpywxBWDz88rhOgoz:b-csauup) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.h#L106)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Is this deliberately commented out? Doesn't look like it's done...
> </details>
> Done.

It was, because the enums were not known to Qt's meta type system. Now they are and I extended this. Btw. the properties are not urgently required, but make life easier, if we decide one day to use settings from QML or some other script language. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
@initramfs I decided to reconnect the settings to settings instance (see 1d8b246).

initramfs:
Reviewed 7 of 11 files at r2, 2 of 2 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

_[src/core/corestructs.cpp, line 11 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_Cf6gtXKEGkwYc5m2:-KO_Cf6hVVwbdIC8zfp0:b-ood8as) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/core/corestructs.cpp#L11)):_

> ``` cpp
> /**
> @file corestructs.h
> @brief Some headers use Core structs but don't need to include all of core.h
> ```

As per #3559, the new comment style now includes leading stars. since you've changed this comment, might want to add them here. The rest should be picked up by Diadlo's PR #3577.

---

_[src/core/corestructs.cpp, line 16 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_DTqTrz74b8F9eDje:-KO_DTqU9fuoU6naYlej:b43txyn) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/core/corestructs.cpp#L16)):_

> ``` cpp
> 
> 
> @var uint8_t ToxFile::fileKind
> ```

Remove the variable type from the comment, so just: `@var ToxFile::fileKind`. This is prone to error (getting outdated with type changes) and doxygen works without variable types.

---

_[src/persistence/settings.cpp, line 1265 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOtNyWsp8xHDOd9xZX:-KO_JcITfQm7h7gj2u2c:btg6ma8) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1265)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Done.
> 
> (see previous comment about "enum class")
> </details>
> Unless I'm missing something you're saying, this isn't an enum (or enum class) though. Meaning it's still possible to give out-of-bounds value to cause an incorrect `autoAwayTimeChanged()` signal from being emitted.

You solved the toxme string function by discarding the invalid value and and printing a warning, couldn't you do the same here?

---

_[src/persistence/settings.cpp, line 1436 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOvIUasOqhnuObq2hu:-KO_LBZJL8ExeXzPCun6:b-zrpjt) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1436)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> FYI: Both methods are dead code. Removing them. Otherwise there is no reason to pass a simple type "const" or by reference, unless the method should explicitly update a value. In this case, I've seen often a pointer to be used instead. I think, just because it is more clear.
> </details>
> Ok, I see.

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KO_Nkdfu_MSEreh2JUb:b-swjzey) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> FYI: Yeah because llvm warns about floating point checks on `==` and `!=`.
> </details>
> Strange I don't get any warnings (using `clang++` on `-Wall` and `-Wpedantic`). Nor does the static analyzer pick this up. Try surrounding it in another set of braces: `if((dB != audioInGainDecibel))`, if that doesn't work just leave it.

---

_[src/persistence/settings.cpp, line 1855 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOx-574DlTIdKkU6rD:-KO_N3lDG788PKGB9xnZ:bch7hhi) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1855)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Done.
> 
> Also fixed typo in "Adress" -> "Address"
> </details>
> OK. And I completely didn't notice that when reading that line... <facepalm>

---

_[src/persistence/settings.cpp, line 1103 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_IcXYy5f98rGecMjh:-KO_IcXYy5f98rGecMji:b-y5xnvw) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/persistence/settings.cpp#L1103)):_

> ``` cpp
>     {
>         toxmePass = pass;
>         // DO NOT EXPOSE THE CHANGED PASSWORD!
> ```

Not really an issue but perhaps change the comment to like:

``` c++
// Password information not exposed for security
```

instead of the one now, sounds less like you're shouting to people. :)

---

_[src/persistence/settings.cpp, line 1266 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_KW3I2Wd9Qf8IDmZ2:-KO_KW3JABQdBNsjRv9f:bfz0eo7) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/persistence/settings.cpp#L1266)):_

> ``` cpp
> }
> 
> void Settings::setAutoAwayTime(int newValue)
> ```

Whilst, ideally all settings should be documented. This one in particular probably needs something to indicate that it takes minutes. If documentation is not an option, perhaps change the parameter name to something like `newValueMinutes` or `minutes` or something of that liking for now.

---

_[src/persistence/settings.h, line 132 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_ODL6bzkWOZCwdZuN:-KO_ODL6bzkWOZCwdZuO:b74q3p7) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/persistence/settings.h#L132)):_

> ``` objc
> 
> public:
>     enum class ProxyType {ptNone = 0, ptSOCKS5, ptHTTP};
> ```

I noted for the initialization of global settings, the code uses integer values to represent defaults, e.g:

``` c++
proxyType = static_cast<ProxyType>(s.value("proxyType", 0).toInt());
```

Though this is trivial, for better clarity, I'd suggest manually numbering each enum so that defaults can quickly be matched. Also, additions of new enums won't screw up the defaults. So, something like:

``` c++
enum class ProxyType {ptNone = 0, ptSOCKS5 = 1, ptHTTP = 2};
```

---

_[src/persistence/settings.h, line 133 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_OptohWBrjUMg4ibO:-KO_OptpAIzmSeE4gcGg:bul9gw2) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/persistence/settings.h#L133)):_

> ``` objc
> public:
>     enum class ProxyType {ptNone = 0, ptSOCKS5, ptHTTP};
>     enum class StyleType {NONE = 0, WITH_CHARS, WITHOUT_CHARS};
> ```

Same as above here.

---

_[src/widget/form/settings/generalform.cpp, line 402 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_Qcxjo4NFXOsP3xBK:-KO_Qcxjo4NFXOsP3xBL:b-byany8) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/widget/form/settings/generalform.cpp#L402)):_

> ``` cpp
> void GeneralForm::onStyleUpdated()
> {
>     Settings::StyleType styleType =
> ```

You can probably just use `auto` here to shorten the declaration (and put the initialization on the same line).

---

_[src/widget/form/settings/generalform.cpp, line 436 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_QoMR5kYyD3hl-8p1:-KO_QoMR5kYyD3hl-8p2:b-doc2ac) ([raw file](https://github.com/qtox/qtox/blob/1d8b2463d1da48f645132bce48b33264b460830d/src/widget/form/settings/generalform.cpp#L436)):_

> ``` cpp
> void GeneralForm::onUseProxyUpdated()
> {
>     Settings::ProxyType proxytype =
> ```

Same as above, use `auto`.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

initramfs:
Oh and, might want to tweak the name of the last commit a bit:

```
refactor(settings): fix connected settings signals
```

Is a bit confusing to whether it's an refactor operation or a fix. Seems more like a refactor to me.

---

Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590#-:-KO_SFCgRF0Vs7Ez8MBy:b1un9rd)_

antis81:
Review status: all files reviewed at latest revision, 11 unresolved discussions, some commit checks failed.

---

_[src/persistence/settings.cpp, line 1265 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOtNyWsp8xHDOd9xZX:-KOdYhbsFjUDPua041WN:b-liyyul) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1265)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Unless I'm missing something you're saying, this isn't an enum (or enum class) though. Meaning it's still possible to give out-of-bounds value to cause an incorrect `autoAwayTimeChanged()` signal from being emitted.
> 
> You solved the toxme string function by discarding the invalid value and and printing a warning, couldn't you do the same here?
> </details>
> Huh? Somehow that change got reverted (see in commits for #3590)… I'll check that.

About the enum (class) and boundary checks:
It is "insane" to do! Here's why:
In both cases an enum (class) is a primitive`int` type and gets compiled exactly the same way. As you probably know, an enum can be used in two ways:
1. Represents a value range (…,-1,0,1,…) without combinations.
2. Represents a flag combination (1 & 2 & 4 & 8…)

Both representations may contain "gaps", which is used widely to represent flag combination presets and keep version compatibility. A safe boundary check would require to check against each and every value instead of a range (e.g. by switch statement). In addition, this includes a pretty high chance of invalidating the check by adding another value - which might be even more "dangerous", than no check at all.

The single major advantage of the enum class over the enum is, that it solves the "name ambiguity" problem and adds namespace-like usage, which is regularly used e.g. for an "Enum::None" value, but also by other values of course. No longer wonder, why the compiler moans about ambiguity. :) 

Now to secure the value boundaries, it would have to be done rather in settings (de-)serialization - but also in any place, where `static_cast<int/EnumType>` occurs.

Seems there's a current draft in c++ standard, which integrates that check into the language (found good explanation [on StackOverflow](http://stackoverflow.com/questions/24184262/c-enum-class-initialization-from-underlying-type)).

So the best thing we can do is to "enforce" using the enum class from within our program and only cast to/from int, when absolutely required (which is also in a QComboBox using that as an "int" value).

Btw. in recent Qt versions this is (almost) solved by "enum from/to string" (de-)serialization (see Q_ENUM macro), making boundary checks obsolete unless in high security scenarios mybe. However this was introduced in Qt 5.5 and so our beloved building machine with Qt 5.3 on it cannot handle it - yet.

Enough confusion?… :)

---

_[src/persistence/settings.cpp, line 1103 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_IcXYy5f98rGecMjh:-KOdXM5BCS8_0DvN6Ojv:bfdb985) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.cpp#L1103)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Not really an issue but perhaps change the comment to like:
> 
> ``` c++
> // Password information not exposed for security
> ```
> 
> instead of the one now, sounds less like you're shouting to people. :)
> </details>
> Done.

Ok… just thought it is an important message. I'll try to be kind… :)

---

_[src/persistence/settings.cpp, line 1266 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_KW3I2Wd9Qf8IDmZ2:-KOdYD2H3JYGSAe-WqUV:bhd9khm) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.cpp#L1266)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Whilst, ideally all settings should be documented. This one in particular probably needs something to indicate that it takes minutes. If documentation is not an option, perhaps change the parameter name to something like `newValueMinutes` or `minutes` or something of that liking for now.
> </details>
> Documentation will be fine.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
Review status: 12 of 19 files reviewed at latest revision, 10 unresolved discussions.

---

_[src/persistence/settings.cpp, line 1265 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOtNyWsp8xHDOd9xZX:-KOdpji-eNj50IVHvrPO:b-9utpb7) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1265)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Huh? Somehow that change got reverted (see in commits for #3590)… I'll check that.
> 
> About the enum (class) and boundary checks:
> It is "insane" to do! Here's why:
> In both cases an enum (class) is a primitive`int` type and gets compiled exactly the same way. As you probably know, an enum can be used in two ways:
> 1. Represents a value range (…,-1,0,1,…) without combinations.
> 2. Represents a flag combination (1 & 2 & 4 & 8…)
> 
> Both representations may contain "gaps", which is used widely to represent flag combination presets and keep version compatibility. A safe boundary check would require to check against each and every value instead of a range (e.g. by switch statement). In addition, this includes a pretty high chance of invalidating the check by adding another value - which might be even more "dangerous", than no check at all.
> 
> The single major advantage of the enum class over the enum is, that it solves the "name ambiguity" problem and adds namespace-like usage, which is regularly used e.g. for an "Enum::None" value, but also by other values of course. No longer wonder, why the compiler moans about ambiguity. :) 
> 
> Now to secure the value boundaries, it would have to be done rather in settings (de-)serialization - but also in any place, where `static_cast<int/EnumType>` occurs.
> 
> Seems there's a current draft in c++ standard, which integrates that check into the language (found good explanation [on StackOverflow](http://stackoverflow.com/questions/24184262/c-enum-class-initialization-from-underlying-type)).
> 
> So the best thing we can do is to "enforce" using the enum class from within our program and only cast to/from int, when absolutely required (which is also in a QComboBox using that as an "int" value).
> 
> Btw. in recent Qt versions this is (almost) solved by "enum from/to string" (de-)serialization (see Q_ENUM macro), making boundary checks obsolete unless in high security scenarios mybe. However this was introduced in Qt 5.5 and so our beloved building machine with Qt 5.3 on it cannot handle it - yet.
> 
> Enough confusion?… :)
> </details>
> Done.

(Sorry, I think I misunderstood the "enum" thing. there's no boundary check required other than negative value, which defaults to 10 minutes here. Anyway the information is good to know about the places, where that boundary check is now removed.)

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KOdqQCJo2nputxydAIM:bgfgxvb) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Strange I don't get any warnings (using `clang++` on `-Wall` and `-Wpedantic`). Nor does the static analyzer pick this up. Try surrounding it in another set of braces: `if((dB != audioInGainDecibel))`, if that doesn't work just leave it.
> </details>
> Try llvm's static analyzer. Compiler doesn't warn… :)

---

_[src/persistence/settings.cpp, line 1266 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_KW3I2Wd9Qf8IDmZ2:-KOdxKSXa1MJbDnD2uqm:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.cpp#L1266)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Documentation will be fine.
> </details>
> Done.

---

_[src/persistence/settings.h, line 132 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_ODL6bzkWOZCwdZuN:-KOdrOuSb6lSWic6-LlW:bp1v6rh) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.h#L132)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> I noted for the initialization of global settings, the code uses integer values to represent defaults, e.g:
> 
> ``` c++
> proxyType = static_cast<ProxyType>(s.value("proxyType", 0).toInt());
> ```
> 
> Though this is trivial, for better clarity, I'd suggest manually numbering each enum so that defaults can quickly be matched. Also, additions of new enums won't screw up the defaults. So, something like:
> 
> ``` c++
> enum class ProxyType {ptNone = 0, ptSOCKS5 = 1, ptHTTP = 2};
> ```
> 
> </details>
> See my comment above about enum boundary checks jada blabla… :)

From Qt 5.5 enums can be represented as string ("ProxyType::ptNone"), which is used e.g. in QJSValue's conversion from/to QVariant. I'll have a look, if the property can be used to store/read the value instead. However this would change user's settings file from int to string and just make it incompatible. However if it works, I wouldn't have a problem with that.

Otherwise :+1: for the fixed numbering.

---

_[src/persistence/settings.h, line 133 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_OptohWBrjUMg4ibO:-KOdukSPlXA3SheYk-Jb:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.h#L133)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same as above here.
> </details>
> Done.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
Review status: 12 of 19 files reviewed at latest revision, 10 unresolved discussions.

---

_[src/widget/form/settings/generalform.cpp, line 402 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_Qcxjo4NFXOsP3xBK:-KOdxeFMS-RxrmgVMYYY:b-78wglb) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/widget/form/settings/generalform.cpp#L402)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> You can probably just use `auto` here to shorten the declaration (and put the initialization on the same line).
> </details>
> Actually I do like to see the type and avoid using the "auto" keyword. It is not part of this PR anyway, thus leaving it.

---

_[src/widget/form/settings/generalform.cpp, line 436 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_QoMR5kYyD3hl-8p1:-KOdyR1NEZEunYeWpBv6:b-lfhur7) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/widget/form/settings/generalform.cpp#L436)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Same as above, use `auto`.
> </details>
> Same as above… :)

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

initramfs:
Reviewed 4 of 7 files at r5.
Review status: 16 of 19 files reviewed at latest revision, 5 unresolved discussions.

---

_[src/persistence/settings.h, line 132 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_ODL6bzkWOZCwdZuN:-KP2-08CcCQFw-RIXGGE:be9lkzv) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/persistence/settings.h#L132)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> See my comment above about enum boundary checks jada blabla… :)
> 
> From Qt 5.5 enums can be represented as string ("ProxyType::ptNone"), which is used e.g. in QJSValue's conversion from/to QVariant. I'll have a look, if the property can be used to store/read the value instead. However this would change user's settings file from int to string and just make it incompatible. However if it works, I wouldn't have a problem with that.
> 
> Otherwise :+1: for the fixed numbering.
> </details>
> Ok. Numbering looks good.

---

_[src/widget/form/settings/generalform.cpp, line 402 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_Qcxjo4NFXOsP3xBK:-KOeQYWtTO4Lv6_WA6N4:b-6d0mph) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/widget/form/settings/generalform.cpp#L402)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Actually I do like to see the type and avoid using the "auto" keyword. It is not part of this PR anyway, thus leaving it.
> </details>
> Ok. Though I'd argue the type is trivially deducable in this scenario given the next call to `setStylePreference()` but I can definitely see your point, leave it.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

initramfs:
Reviewed 3 of 7 files at r5.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[src/core/corestructs.cpp, line 11 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_Cf6gtXKEGkwYc5m2:-KP70lodk1GgeOB8XOeQ:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/core/corestructs.cpp#L11)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> As per #3559, the new comment style now includes leading stars. since you've changed this comment, might want to add them here. The rest should be picked up by Diadlo's PR #3577.
> </details>
> Done.

---

_[src/core/corestructs.cpp, line 16 [r4]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KO_DTqTrz74b8F9eDje:-KP70rx-bFjuKNIoWJD3:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/06ecbc12f97b659acb47d147eb814e0994e72336/src/core/corestructs.cpp#L16)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Remove the variable type from the comment, so just: `@var ToxFile::fileKind`. This is prone to error (getting outdated with type changes) and doxygen works without variable types.
> </details>
> Done.

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KP70xb7Maakwuqdv__1:b2nhei9) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Try llvm's static analyzer. Compiler doesn't warn… :)
> </details>
> FYI

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KP71a3CpzTSsFSVj5Eb:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> FYI
> </details>
> Done.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
Reviewed 4 of 9 files at r1, 6 of 11 files at r2, 2 of 4 files at r4, 7 of 7 files at r5.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

initramfs:

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KP72Mc9CuJSUIYI7glx:b-6mheec) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> Done.
> </details>
> Ok. Can't reproduce (scan-build right? Or is there some other static analyzer?) but I believe you.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:

---

_[src/persistence/settings.cpp, line 1712 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3590#-KOOwQP6AWqhv3-3WzWz:-KP71QqZdW9LQkhl8ICp:bv69gm7) ([raw file](https://github.com/qtox/qtox/blob/a28568454ed47c6aa8da0c72f79631199c28462e/src/persistence/settings.cpp#L1712)):_
<details><summary>

<i>Previously, initramfs wrote…</i></summary>



> Ok. Can't reproduce (scan-build right? Or is there some other static analyzer?) but I believe you.
> </details>
> Hm… it might be due to different llvm versions in use. Looks like the warning changed. Have a look [here](https://llvm.org/bugs/show_bug.cgi?id=13330) or [here](https://llvm.org/bugs/show_bug.cgi?id=23717) and definitely [here](http://stackoverflow.com/questions/11421756/weverything-yielding-comparing-floating-point-with-or-is-unsafe). I don't know. I'd say it is safe now and it doesn't harm and that's fine for me.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590)_

antis81:
@initramfs Are we done here then? I'd like to merge.

initramfs:
:lgtm:

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3590#-:-KP7Jt6OFJSfZ9l921bA:bnfp4nl)_

PR #3623, opened 2016-08-12
fix: use `qAbs()` instead of `abs()` for better platform compatibility

zetok:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3623)

PR #3624, opened 2016-08-12
Prefer std::abs and correctly include it

tux3:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3624)

### DIFF
--- a/src/video/videomode.cpp
+++ b/src/video/videomode.cpp
@@ -18,6 +18,7 @@
 */
 
 #include "videomode.h"
+#include <cstdlib>
 
 /**
 @struct VideoMode
--- a/src/widget/tool/screengrabberchooserrectitem.cpp
+++ b/src/widget/tool/screengrabberchooserrectitem.cpp
@@ -167,7 +167,7 @@ void ScreenGrabberChooserRectItem::mouseRelease(QGraphicsSceneMouseEvent* event)
         setCursor(QCursor(Qt::OpenHandCursor));
 
         QPointF delta = (event->scenePos() - startPos);
-        if (qAbs(delta.x()) < MinRectSize || qAbs(delta.y()) < MinRectSize)
+        if (std::abs(delta.x()) < MinRectSize || std::abs(delta.y()) < MinRectSize)
         {
             rectWidth = rectHeight = 0;
             mainRect->setRect(QRect());
--- a/src/widget/tool/screenshotgrabber.cpp
+++ b/src/widget/tool/screenshotgrabber.cpp
@@ -29,6 +29,7 @@
 #include <QMouseEvent>
 #include <QScreen>
 #include <QTimer>
+#include <cstdlib>
 
 #include "screengrabberchooserrectitem.h"
 #include "screengrabberoverlayitem.h"


zetok:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624#-:-KP-hFIvEhs-7JzXaS9F:bnfp4nl)_

initramfs:
If you want to standardize the use of `std::abs()` instead of `qAbs()`, you've missed one instance of `qAbs()` in [src/widget/form/settings/avform.cpp#L78](https://github.com/qTox/qTox/blob/master/src/widget/form/settings/avform.cpp#L78).

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624#-:-KP1aoA9gkhmyT7LcJMa:bi45chs)_

initramfs:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_[src/widget/tool/screengrabberchooserrectitem.cpp, line 170 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3624#-KP2Kh5w5epT-b5NSqhT:-KP2Kh5xuGcW3pBpKzlI:b7pmseb) ([raw file](https://github.com/qtox/qtox/blob/f76998d37c2ed7b4f9c7f8a9d4c319cf8416d522/src/widget/tool/screengrabberchooserrectitem.cpp#L170)):_

> ``` cpp
> 
>         QPointF delta = (event->scenePos() - startPos);
>         if (std::abs(delta.x()) < MinRectSize || std::abs(delta.y()) < MinRectSize)
> ```

You are calling `std::abs()` with a floating point argument, this will not work if you don't include `<cmath>` as well. (`#include <cstdlib>` includes the integral abs, `#include <cmath>` includes the floating point types). If you want to be explicit, you could also change this call to `std::fabs()`.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624)_

iphydf:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_[src/video/videomode.cpp, line 67 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3624#-KP7HxoWqNNSymjctFN2:-KP7HxoXoC2W-1KxiZ-Y:b5zmh5b) ([raw file](https://github.com/qtox/qtox/blob/f76998d37c2ed7b4f9c7f8a9d4c319cf8416d522/src/video/videomode.cpp#L67)):_

> ``` cpp
> uint32_t VideoMode::norm(const VideoMode &other) const
> {
>     return abs(this->width-other.width) + abs(this->height-other.height);
> ```

Someday you're going to come back to this to add the `std::` because C++ libraries have started being clean enough to do it.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624)_

initramfs:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_[src/video/videomode.cpp, line 67 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3624#-KP7HxoWqNNSymjctFN2:-KP7KEhD1PWhlxpQ665w:b8zasaa) ([raw file](https://github.com/qtox/qtox/blob/f76998d37c2ed7b4f9c7f8a9d4c319cf8416d522/src/video/videomode.cpp#L67)):_
<details><summary>

<i>Previously, iphydf wrote…</i></summary>



> Someday you're going to come back to this to add the `std::` because C++ libraries have started being clean enough to do it.
> </details>
> Actually, whilst you're at that, mind putting spaces between each subtraction operation? E.g:

``` c++
std::abs(this->width - other.width)
```

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624)_

Diadlo:
Why we should use std::abs instead of qAbs?

**EDIT:** I think, we should write in `CONTRIBUTING.md` that project prefer std structures and functions instead of Qt (or conversely). Now we use many Qt analogs (QList, QVector, etc)

initramfs:
@Diadlo From what I saw on IRC, the decisions wasn't something of a hard-and-fast rule.

From tux3 on IRC regarding the issue (before the PR was made):

> For basic things like these or standard integer types, I'd rather we use the standard headers than Qt shims
> But that's more of a preference than an objective rule :)

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3624#-:-KP7wzxCFy9NirOHeT2s:b5j5jzy)_

zetok:
@initramfs there are IRC logs: https://github.com/qTox/qtox-irc-logs/blob/master/2016/08/%23qtox_20160812.log.txt#L507,L518 :)

Diadlo:
I think, we should merge this PR or #3623 as hotfix
And open issue-discussion

antis81:
> I think, we should write in CONTRIBUTING.md that project prefer std structures and functions instead of Qt (or conversely). Now we use many Qt analogs (QList, QVector, etc)

No we shouldn't! Coders actually have to learn and decide what fits best - which can be either of them in each particular case. As a rule of thumb: Write "with" the API and not "against" it. That means: The less you have to convert, the better it is!

So, use Qt and STL where appropriate without preferring one over the other without good reason. Nothing more, nothing less.

tux3:
Superseded by #3623, closing

PR #3631, opened 2016-08-16
[Obsolete] ui redesign

antis81:
# Obsolete! Please see upstream branch `ui/redesign` (PR #3639)

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3631)

### DIFF
--- a/qtox.pro
+++ b/qtox.pro
@@ -303,6 +303,7 @@ HEADERS  += \
     src/video/videomode.h \
     src/video/genericnetcamview.h \
     src/video/groupnetcamview.h \
+    src/widget/contentwidget.h \
     src/widget/emoticonswidget.h \
     src/widget/style.h \
     src/widget/tool/croppinglabel.h \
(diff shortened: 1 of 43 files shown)


Diadlo:
> Separate "FriendWidget" from Friend data class (just like with ChatForm).

Now I work with it

antis81:
@Diadlo 
If I understand you right, you want to work on "separating the FriendWidget from Friend", right? While this is likely not possible without the work done in this PR, I can setup a new upstream PR (on the main repo), which we both merge/rebase on. This PR is not going to be rebased unless we both agree, this makes sense. I'll try to split it into smaller chunks, though I doubt this is possible in some places, because some things are completely rewritten. Let's see what I can do here…

Diadlo:
@antis81 Yep. 
And may be add to todo:
- [ ] Move `GroupList` and `FriendList` from global scope to `Core`
- [ ] Create superclass to `Friend` and `Group` (we have some code duplication for work with both)
- [ ] Separate `FriendWidget` and `ChatForm` from Group

**EDIT:** I never try teamwork with upstream. We should just push new commits in the branch without any discussion?

Issue #3621, opened 2016-08-12
Incorrect group name in the invitation


PKEv:
##### Brief Description

OS: Windows 
qTox version: v1.5.0-2-gf05b7e0.
Commit hash:  f05b7e040a073c7afadaf763dda5beaf130c6dbd
Qt:  5,5,1

Reproducible: Always 
##### Steps to reproduce
1. invite in group 1 and invite in group 2 (Groupbot and Kalina)
2. go to group's page with invites
3. Both invitations will be from the same group (same name)
##### Additional Info

![default](https://cloud.githubusercontent.com/assets/6053197/17621992/3108ef66-60a0-11e6-873d-6f1ac5f10d5c.PNG)

ProMcTagonist:
I've noticed this too.

PR #3639, opened 2016-08-20
[WIP] ui redesign

antis81:
**Attention:** Something **huge** is about to make it's way into qTox! :smile:

While dealing with current settings(-ui), audio, video… issues, I thought about how much qTox would improve by simplifying the (UI) code and requesting (memory and other) resources when needed and freeing them, when no longer required. My first thought was to just implement this for the settings UI. While this works fine, it highly suffers from the "spaghetti crap code" in our beloved "Widget" class, which actually is a really destructive thing, which is hard - very hard! - to control. It is so much spaghetti, that refactoring is not possible and feature implementations… well, forget about that. So I decided, it deserves a "redesign" and gave that a try silently in this branch during the last weeks.

While there's quite an amount of work to be done here and it is unfinished, most of the stuff actually works now and I'd like to share, so you can see for yourself. You may ask, why I'm going to do such an "insane" thing. So here's some answers to that:
- In current state, it shrinks down qTox's memory footprint by >50% (60M) - and it will even shrink more! If you wonder, why qTox easily eats up ~120M and more RAM (even without any active chat) - this is why!
- As a neat side effect, we get a visible performance improvement, when resizing the qTox window(s).
- Code-Flow gets **a lot** easier to understand and to follow for contributors. Lots of "redundant" code gets smashed and completely removed, making qTox a more lightweight experience.
- It opens future opportunities, which currently seem impossible.

So, what features do we get?
- Fluid, responsive user interface.
- Clean resource management: The "resource subscription" problem is solved, which is mentioned in so many issues.

Visible features:
- As said above: Overall more "fluid" user experience.
- Finally! -> Reliable representation of a friend's actual "online status".
- More accurate positioning of attached and detached widgets on screen!
- No more suddenly "disappearing" widgets. 
- A neat side effect of the now centralized "(sub)widget display management" now allows for an (optional) "auto-flip" feature, depending on the main window's x-position on screen. This works in attached and detached (!) mode - I'd say it's ~95% done, but has few minor glitches.
- Not checked, but offline message delivery may work more accurately now. At least it works minimum as before.

Features for coders:
- No more "show/hide/activateWindow/raise/…" workarounds! It is cleanly handled by `Widget::showContentWidget()`.
- Widgets are "free'd" from the "ContentLayout / ContentDialog" burden and do what they are meant to do again. However most widgets now inherit "ContentWidget", which provides a unified look.
- Create widgets as you need them and free them, when you're done. When using the `ContentWidget` base class, the widget destroys automatically after closing it. So, if the close is rejected (e.g. by the infamous "Save changes before close?" dialog), the widget stays visible and won't get destroyed!
# TODO (known issues):

There's quite some work to finish, until this can be considered stable.
## Style
- [ ] For some reason, the widget style seems to randomly fallback to system's theming during `ContentWidget` creation. This might be due to mixed use of QPalette and CSS style-sheets. Probably we should create a style-sheet for this purpose.
- [x] Due to the fact, that we now use a lot less layouts, the bottom and right margins need to be adjusted. Anyway we now make good use of available space.
## FriendListWidget
- [ ] Friend/Group doesn't "un-highlight", when switching between chats. An active chat should be marked with "white background", and inactive with the current style's default background. Note, that in "multiple windows" mode, multiple chats can be visible (selected -> white BG).
## Chat (single & group):
- [x] Chat appears empty, after switching to another chat (-> load history).
- [x] Call/Video-Call buttons don't reflect the call's status:
  - Buttons stay "green" or "grey", depending on the friends status after switching to another chat). We need to request the ToxCall's state on widget initialization.
- [x] Typing status falsely reports "[user] is typing…" instead of user name.
- [x] The button icons for "mute microphone" and "mute output volume" unfortunately contain the status colors (green, red, grey, …) - contrary to the "AV call" buttons, which use a background color. Someone has the time to draw two lousy SVG's?
- [x] The video widget is not re-initialized and disappears after switching to another chat.
- [x] The video call button disappears as well when switching to another chat.
- [x] Separate "FriendWidget" from Friend data class (just like with ChatForm).
- [ ] Make sure everything works properly with group chats as well.
## ProfileForm:
- [x] No profile Information is displayed at all.
## SettingsWidget & it's sub-widgets:

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3639)


antis81:
@Diadlo Here it is - a brandnew qTox feature branch! About your question…

> I never try teamwork with upstream. We should just push new commits in the branch without any discussion?

Please setup a PR **in your personal repository** like usual. The only difference is, that you select `ui/redesign` as the **base** branch on Github.

src/widget/widget.cpp
```diff
@@ -1079,20 +1080,27 @@ void Widget::onFriendUsernameChanged(int friendId, const QString& username)
...
+    FilterCriteria filter = getFilterCriteria();
     switch (s)
     {
-        case Status::Offline:
-            friendWidget->searchName(searchContactText->text(), filterOffline(filter));
-        default:
-            friendWidget->searchName(searchContactText->text(), filterOnline(filter));
+    case Status::Offline:
+        friendWidget->searchName(searchContactText->text(), filterOffline(filter));
+                                 filterOffline(filter));
```
Diadlo, line 1019: Can't be compiled with it `expected ';' before ')' token`

src/widget/contentlayout.cpp
```diff
@@ -71,15 +71,11 @@ void ContentLayout::clear()
     QLayoutItem* item;
     while ((item = mainHead->layout()->takeAt(0)) != 0)
     {
-        item->widget()->hide();
-        item->widget()->setParent(nullptr);
```
Diadlo, line 75: Did you try build qtox with this commit?
This lines are needed, otherwise it break UI of ChatForm.

antis81, line 75: Actually I don't care much, because ContentLayout will vanish anyway after this is done. Anyway, it works here (Qt 5.7) - and if it doesn't, something else is wrong, because:

The `delete item` (the contained widget actually) removes the widget from it's parent's children list and an additional hide is not required. This example is actually given in Qt's docs.

Diadlo, line 75: Do you agree, that every commit should be a correct version? If not -- no question anymore

spectacle m10381

antis81, line 75: Guess you're talking about that "overlayed" layout. Yes, that's fine. The commit doesn't fix everything, but  represents an interims state (which obviously is even compilable, yay :)). It doesn't make sense to fix a bug, which actually does not exist. Effort would simply be huge to compile and check every time…

src/widget/widget.cpp
antis81, line 1019: And here we have an example of what can happen, if conflicts are not resolved correctly. Damn I was sure I had fixed that during the nightly rebase session… :)

I'll rebase it again and remove the duplicate `filterOffline(filter));`, but you'll have to rebase your branch as well - which is why I like to avoid that. Anyway, it's fine in this case and makes every commit compilable at least.

initramfs:
A few bugs I found so far, apologies if I report something already known:

---

Upon loading, I immediately get presented with a RTL window instead of the expected LTR:

rtl layout

When choosing languages that are traditionally RTL (e.g. Arabic), it swtiches back to LTR (suggesting a reversed preference).

Upon creating a new profile, this setting is retained (i.e. it's still RTL)

---

Clicking on settings dramatically resizes the friend's list to be much wider (takes up 50% of the screen width):

wide-friendlist

---

The fusion style is also lost, defaulting to Qt's default style.

---

When attempting to create a brand new profile with no existing tox directory (e.g. No directory under `~/.config/tox/`), qTox crashes with `SIGABRT`. Signal is raised within `qlist.h` used from `widget.cpp:355` (`Widget::init()`). Backtrace is attached:

[new-profile-backtrace.txt](https://github.com/qTox/qTox/files/428782/new-profile-backtrace.txt)

initramfs:
Review status: 0 of 43 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_[src/widget/widget.cpp, line 353 [r5]](https://reviewable.io:443/reviews/qtox/qtox/3639#-KPg_rWtyihhpU8O-VGi:-KPg_rWu2yBbCWqL2X6m:b-nwfspi) ([raw file](https://github.com/qtox/qtox/blob/6e15d98861050404f71dc585e34420150ee33d78/src/widget/widget.cpp#L353)):_

> ``` cpp
>         constexpr int spWidthPc = 33;
>         mainSplitter->resize(size());
>         QList<int> sizes = mainSplitter->sizes();
> ```

In my previous comment I mentioned a crash with `SIGABRT`. This turns out to be due to the fact that the `sizes` list here somehow only contains 1 element after this PR. As such indexing `sizes[1]` causes `SIGABRT` to be raised by Qt.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3639)_

antis81:
@initramfs Thanks a lot for your feedback! Don't worry about reporting "known (TODO) bugs". I definitely have an eye on making this feature complete up to the current state!

Btw. anyone having OSX available? I cannot test this and build fails -> will add to TODO list as well.

Let me try to answer your findings/questions:

> Upon loading, I immediately get presented with a RTL window instead of the expected LTR

Please don't mix that up with LTR/RTL - which also was my first approach to achieve this. It won't work! The "layout direction" (Qt::LayoutDirection) is clearly part of localization and the flip algorithm is completely independent of that! Instead, it actually flips (rearranges) the widget x-positions on screen or within the splitter. It checks vs. "x-center" and "right border" as reference points. Thus, the answer to your finding is: "It is a (known) TODO, which is actually easy to fix by not flipping the widget arrangement in maximized or full-screen mode.". In addition, the check should be done vs. "current screen geometry" instead of "virtual desktop's geometry", which spans multiple screens and thus performs an unwanted flip in multi-screen setups. Would you like improve (finish) that in open another PR on `ui/redesign`? I won't touch this before the more important stuff is ready and there's a good chance, this won't generate any conflicts.

> The fusion style is also lost, defaulting to Qt's default style.

Have a look at the TODO list in description -> [style topic](#Style). Btw. "Fusion" is Qt's default style, but obviosly it falls back to system's style. I'll try and write a style sheet instead of that current QPalette based stuff.

> When attempting to create a brand new profile with no existing tox directory (e.g. No directory under ~/.config/tox/), qTox crashes with SIGABRT.

Thanks! I'll track down that bugger. :smile: Btw. I just "adjusted" the ProfileForm, which now works even better than before. So, another puzzle piece, that can be considered done.

To not having to rebase this PR, I put the upcoming changes in another PR. I'll link that in the description (about to be created).

antis81:
@iphydf Hey, do you have a mac available? If, yout can probably help making this compile on mac again. I'd have to do changes blind and it currently fails…

zetok:
@antis81 it doesn't even seem to compile on Linux?

antis81:
@zetok i know - at least on travis. as you can see from the screenshots, it compiles on anybody's local computer - strange eh?

zetok:
@antis81 Not really. AFAIK no one who has been building it did it on sufficiently old distro, i.e. ubuntu 14.04.

iphydf:
I'm going to try to build it here. Also, wow you have many git tags..

iphydf:
[20:54:00.459 UTC] :0 : Warning: QLayout: Attempting to add QLayout "" to GenericChatForm "", which already has a layout

iphydf:
Fix:

```
diff --git a/src/widget/contentdialog.h b/src/widget/contentdialog.h
index e43a0ca..abf0ab0 100644
--- a/src/widget/contentdialog.h
+++ b/src/widget/contentdialog.h
@@ -75,6 +75,7 @@ public:
 #ifdef Q_OS_MAC
 signals:
     void activated();
+    void windowStateChanged(Qt::WindowStates state);
 #endif

 public slots:
```

iphydf:
A (perhaps known) bug: selection in the friend list is never deselected. After selecting a friend, it stays selected forever.

antis81:
> Also, wow you have many git tags.

You mean commits? To the nature of redesign, this is not possible by "simple" changes. It is actually more like a rewrite of stuff.

> [20:54:00.459 UTC] :0 : Warning: QLayout: Attempting to add QLayout "" to GenericChatForm "", which already has a layout

Yeah, I can read the log. :) Did not have the time to look at it. Somewhere, the layout is set twice in GenericChatForm c'tor context. I'll probably redo layout related things here (did it right for `ProfileForm`). But all in all I'd like to focus on more important stuff first, before fixing this up. E.g. ToxCall status is not correctly tracked (in old and new code) and I'm currently working on that.

iphydf:
Some more bugs (I gave up, it's scary):
- Sometimes when a friend is typing, it says `"[user] is typing"` instead of the name.
- When someone uses /me for an action, and I switch to another friend and back, it will say the name twice in the action, e.g. `"* Zetok Zetok does something"`.

antis81:
> Some more bugs (I gave up, it's scary):

Thanks! Don't worry. I actually know what I'm doing here. Everything will be fine! :)

antis81:
@zetok Branch `ui/redesign` is not available to travis. Could you have a look, please? OSX build should now work again.

initramfs:
@antis81 OSX builds seem to pass on travis, but Ubuntu (linux) builds still don't

toxlover:
OMG is it going to look like this?

https://github.com/qTox/qTox/pull/3639#discussion_r75578305

That is so much nicer. That old header is so useless.

antis81:
> OMG is it going to look like this?

@toxlover Thanks for positive feedback! And yes, it already does look like this - and cleaned up. Mind to compile or even help improve?

antis81:
@zetok @initramfs @sudden6 @toxlover @iphydf @Diadlo @agilob 
Here's a little progress update to this PR: Overall situation with (single) calls now is further improved (less "partner cannot hear me" moans ^^) and the buttons reflect the actual status of an active call instead of just assuming it. Users would maybe also notice, that if a friend's status changes, the buttons are now updated without any delay. Progress, yay!

However, there's still some todo's and I probably could use some support on those:
1. The button icons for "mute microphone" and "mute output volume" unfortunately contain the status colors (green, red, grey, …) - contrary to the "AV call" buttons, which use a background color. Someone has the time to draw two lousy SVG's? I'll add a TODO to the description above.
2. The video widget is not re-initialized and disappears after switching to another chat.
3. Somewhat surprisingly, the video call button disappears as well together with point 2.

@initramfs Probably you could help with 2 and 3??

Diadlo:
To add to TODO:
- [ ] Create GroupInviteWidget (to fix #3621)

And another bug: http://diadlo.dyndns.org/owncloud/s/DWVCLJIFnn2uXmt
It's appear between b76c958a5a86c2e81a9ef3432555fa7d6b3ed068 and b454edb1ab2e6d467a8d9ea55bc2594796a654ba

initramfs:
Just a FYI, after merging in the 8 commits from @Diadlo's PR, I'm getting a segmentation violation from uninitialized variable access in function `Group::isAvGroupchat()`, caller: `GroupChatForm` constructor, variable is a bool `Group::avGroupchat`. Happen upon attempting to select a contact after the main interface loads.

---

Review status: 0 of 87 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3639#-:-KSa8Kb1R068L9EfTRft:b-619gbg)_

antis81:
@initramfs Thank you. Yeah I noticed that immediately of course and fixed it locally. Just didn't push that yet. I'll try to push it and cherry-pick from my work-branch.

antis81:
@diadlo I had to rebase this branch onto `master` to make it compatible with the new programming guidelines (namely ~~dynamic_cast~~ -> `static_cast/qobject_cast`).

**EDIT:** Due to the nature of this PR this will raise a few hundred conflicts. Could you check if #3690 was already picked and if not, pick just your commits and add them on top? I'm here to help, if needed.

Diadlo:
To avoid a lot of conflicts, this PR will be broken on smaller. I can't (and don't want) rebase it on current master (too many conflict) => I will add prefix `[merged]` to commit message, which was cherry-picked to master. 
I will do it to make navigation here easier

antis81:
Due to the fact, that we don't have an official milestone branch available yet, I am closing this one and treat development, as if it was a usual feature PR with a single commit. It is not possible to perform this rather important step without "breaking" master! @zetok We should really think about stabilizing the master branch and work with a development branch, where all the work for the next milestone goes into.

PR #3645, opened 2016-08-23
fix(video): Fallback if mode settings are invalid

ghost:
This fixes a problem where the video stream is not displayed and debug indicates that the pixel format is invalid. If opening the camera device with mode settings fails then try a fallback using only the device name. In tests on different systems, this always succeeds.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3645)

### DIFF
--- a/src/video/camerasource.cpp
+++ b/src/video/camerasource.cpp
@@ -280,8 +280,13 @@ bool CameraSource::openDevice()
 
     if (!device)
     {
-        qWarning() << "Failed to open device!";
-        return false;
+        // fallback if opening with mode settings fails
+        device = CameraDevice::open(deviceName);
+        if (!device)
+        {
+            qWarning() << "Failed to open device!";
+            return false;
+        }
     }
 
     // We need to open the device as many time as we already have subscribers,
--- a/src/video/camerasource.cpp
+++ b/src/video/camerasource.cpp
@@ -287,6 +287,7 @@ bool CameraSource::openDevice()
             qWarning() << "Failed to open device!";
             return false;
         }
+        qWarning() << "Selected video mode could not be opened!";
     }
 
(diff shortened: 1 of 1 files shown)


initramfs:
Reviewed 1 of 1 files at r1.

---

_[src/video/camerasource.cpp, line 284 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3645#-KPtIFVkcayLu6dAht1V:-KPtIFVlZ5zH4Kjro-11:b-j8n2hq) ([raw file](https://github.com/qtox/qtox/blob/c7f25601d7887514f91721010600353908984294/src/video/camerasource.cpp#L284)):_

> ``` cpp
>     {
>         // fallback if opening with mode settings fails
>         device = CameraDevice::open(deviceName);
> ```

This however doesn't set/indicate the new open mode anywhere, effectively causing a default mode to be opened anytime a given mode doesn't work. This is usually fine for differing stream types or pixel formats, but what about differing resolutions?

Imagine this:
1. User has a webcam that goes up to 1080p.
2. For some reason, the 1080p mode can no longer be opened (say it's an issue with ffmpeg).
3. User selects 1080p on their video selection list, instead the webcam defaults to some jarringly low default resolution like 240p with no indication from qTox that anything was wrong.

Whilst this PR does fix a very important issue, I don't think "silently ignoring the set video mode" should be the way to solve the issue at hand. At the very least, change the video mode combobox to indicate the new mode being used, or blatantly refusing to load the video (with an error message of some sort). Also, logging with `qWarning()` that the selected mode could not be opened should also be done.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3645)_

zetok:

PR #3657, opened 2016-08-26
Fix problem with audio call on settings tab

### LINKED ISSUE #3538: Partner don't hear me while I'm not in Audio/Video tab
##### Brief Description

OS: Windows 10
qTox version: v1.4.1-1-g06d78dc
Commit hash:  06d78dcdee7bfe9a320fe91662f34227d74c3916
toxcore:   0.0.0
Qt:  5.5.1
Hardware:  Realtek High Definition Audio

Reproducible:  Almost Always 
##### Steps to reproduce
1.  Log in
2.  Audio call
3.  Go to Audio/Video tab
4.  Capture Device is always None(This is issue also) so I need to choose my microphone
5.  When microphone is chosen and i go to other tab my partner don't hear me anymore. So i need to stay at audio/video tab. 

Sorry but i didn't find log files, i will be grateful if you'll help me.

Diadlo:
fix(audio): partial revert of 81df534c9a1b392cab136c6b84a58d2e5028a425

Fix #3538.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3657)

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -352,10 +352,9 @@ bool Audio::initInput(const QString& deviceName)
     const uint32_t chnls = AUDIO_CHANNELS;
     const ALCsizei bufSize = (frameDuration * sampleRate * 4) / 1000 * chnls;
 
-    const ALchar* tmpDevName = deviceName.isEmpty()
-                               ? nullptr
-                               : deviceName.toUtf8().constData();
-    alInDev = alcCaptureOpenDevice(tmpDevName, sampleRate, stereoFlag, bufSize);
+    if (!deviceName.isEmpty())
+        alInDev = alcCaptureOpenDevice(deviceName.toUtf8().constData(),
+                                       sampleRate, stereoFlag, bufSize);
 
     // Restart the capture if necessary
     if (!alInDev)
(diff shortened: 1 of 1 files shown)


antis81:
NIce try, but problem is actually located in `ToxCall` copy/move c'tors. Otherwise it is unrelated. :)

antis81:
Btw. if you're at it, remove those qDebug's. We get the warnings anyway, if something goes wrong.

antis81:
@Diadlo I don't know how to say it other than what it actually is: Every single line in this PR is simply wrong. Please read OpenAL docs to understand how stuff works. The "nullptr" actually opens the default device (in qTox context, this equals `QString()` - OpenAL uses `nullptr`instead of empty strings; otherwise it would think it is actually a device name). Further, the Audio code has nothing to do with settings. The fact, that a call works while AV settings is visible is purely coincidental! The problem is located in `ToxCall` default/copy/move c'tors and the d'tor. Some combination leads to "unsubscribe" and this is not possible to fix unless you know the exact code flow, which again is conincidental! That means, if someone adds a "fix", things will break again.

Closing as invalid. Please open another PR, fixing stuff in ToxCore.

Issue #2407, opened 2015-10-16
qTox fails to compile with Qt 5.2


zetok:
reported by @abbat,

https://github.com/tux3/qTox/commit/073d4f0ad329511dd0cf3211e18b60dd3d6f2d32#diff-6891a8673df6107b3f58a72f73bd7006R114 ← this line breaks it, since it's available only in Qt >=5.4
@antis81 

So, the question is, how long qTox is supposed to support Qt <5.4 ?

AFAIK there are reasons to bump minimal Qt version to 5.4, e.g. [accessibility support](https://github.com/tux3/qTox/milestones/Accessibility%20support), bugfixes, etc.

Now, the only reason against it, AFAICS is to support "easy" way of compiling qTox on some older distros, e.g. debian 8.

abbat:
Pay attention what Qt < 5.4 is at least in these active (not EOL) distros:
- Debian 7 (Qt 5.3.2)
- Debian 8 (Qt 5.3.2)
- Ubuntu 14.04 LTS (Qt 5.2.1)
- Fedora 21 (Qt 5.3.2)

PR #3661, opened 2016-08-27
docs(INSTALL.md): Fix minimal Qt version

abbat:
Fix minimal Qt version in dependencies to 5.4.x, see #2407 for details.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3661)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -40,7 +40,7 @@
 
 | Name          | Version     | Modules                                           |
 |---------------|-------------|-------------------------------------------------- |
-| Qt            | >= 5.3.0    | core, gui, network, opengl, sql, svg, widget, xml |
+| Qt            | >= 5.4.0    | core, gui, network, opengl, sql, svg, widget, xml |
 | GCC/MinGW     | >= 4.8      | C++11 enabled                                     |
 | toxcore       | most recent | core, av                                          |
 | FFmpeg        | >= 2.6.0    | avformat, avdevice, avcodec, avutil, swscale      |


zetok:
Is there a need to bump it if things seem to work with 5.3?

initramfs:
Debian stable is still at Qt 5.3.2.

antis81:
@zetok
As long as we're not using QML, there's few changes to the widget engine between Qt versions and we'd barely notice them. However, there's some nice stuff, that got introduced around 5.5.1. I actually don't understand why Debian won't upgrade their stable branch to the 5.6.1-1 LTS version. Maybe they'll do in the future.

As much as I hate that, but I also suggest staying with Qt 5.3.2 provides the most compatible basis to build on.

abbat:
So, we have to fix code to compile qTox with 5.3.2, now it not compiled.

initramfs:
@abbat Perhaps I've set up my enviornment wrong, but it does compile for me? I'll test it on a Debian VM later on. Compiling with gcc goes fine without issue, compiling with clang yields 11,500 issues (from QtCreator) but the application still launches are performs normally.

antis81:
I know there's some (partially painful) shortcomings with Qt <5.4. If the official Debian backports repository supports that, it might be a valid alternative to make that a requirement on the target platform.

Side note:
Btw. I cross-compiled Qt 5.5.1 with DRI/EGL support for a commercial project to get rid of those shortcomings on armhf (Raspberry Pi) platform with Debian 7/8. Result allowed for seamless (Qt-)quick deployment. It's a pity this project had to die. However, compiling the Qt runtime library for every desktop (and probably mobile) target is a tough task to accomplish. It might be an option though, if we found enough people supporting that idea. Probably Docker would also be an option?

Talkless:
I would suggest same as @antis81  - fix any incompatibilities with 5.3. At least until Debian 8 goes into LTS (~May 2018, around year after 9 is released), unless someone creates Qt 5.6 LTS available in backports.

By the way I've compiled on Debian 8 today successfully, so it's fixed?

zetok:
Closing, since there seems to be a consensus that Qt 5.3 works and should still be supported at least until Debian 9 is stabilized or if Qt 5.6 gets backported to Debian 8.

@antis81 

> I actually don't understand why Debian won't upgrade their stable branch to the 5.6.1-1 LTS version. Maybe they'll do in the future.

Well, Debian aims for something labeled as "stability", which in Debian land means basically sticking to old versions – their API/ABI, bugs, etc. – as if your life depended on it. Which, considering how widely Debian & stuff based on is spread, is most likely the case for some people..

Not saying that it's reasonable for every single package, but there are reasons.

PR #3667, opened 2016-08-29
[ui/redesign] Change buttons

Diadlo:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3667)


antis81:
Reviewed 13 of 14 files at r1.
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

_[src/widget/form/chatform.cpp, line 875 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3667#-KQLn6kox4gRyZuXlAMq:-KQLn6kpInxEHaeAHu6z:b-qn0py7) ([raw file](https://github.com/qtox/qtox/blob/3fed9fd4da236c1f7614782dba4a93d8cf8520f2/src/widget/form/chatform.cpp#L875)):_

> ``` cpp
>     volButton->setEnabled(av->isCallActive(f));
> 
>     if (volButton->isEnabled())
> ```

Oops, nice typo, thanks! :+1:

---

_[src/widget/form/chatform.cpp, line 888 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3667#-KQLnQB1Tv4YzVD_Xpz_:-KQLnQB2ypbgS8WIws0k:bqe7z0j) ([raw file](https://github.com/qtox/qtox/blob/3fed9fd4da236c1f7614782dba4a93d8cf8520f2/src/widget/form/chatform.cpp#L888)):_

> ``` cpp
>         }
>     }
>     else
> ```

By fixing the "enabled" state, this should not be required any longer. The ":disabled" pseudo-state should overrule the other stuff (exactly like in callButton/videoButton). Could you check that (also for micButton)? Otherwise code looks good.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3667)_

Diadlo:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

_[src/widget/form/chatform.cpp, line 888 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3667#-KQLnQB1Tv4YzVD_Xpz_:-KQN5YZXi14HGb2iqLOd:bidz3xb) ([raw file](https://github.com/qtox/qtox/blob/3280e4be1cdf970bfe51d335bf5214aff9cda538/src/widget/form/chatform.cpp#L888)):_
<details><summary>

<i>Previously, antis81 (Nils Fenner) wrote…</i></summary>



> By fixing the "enabled" state, this should not be required any longer. The ":disabled" pseudo-state should overrule the other stuff (exactly like in callButton/videoButton). Could you check that (also for micButton)? Otherwise code looks good.
> </details>
> Yes, I know, but I could not make it work without name clearing. Have no idea where the problem (maybe upstream problem?)

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3667)_

Diadlo:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

_[src/widget/form/chatform.cpp, line 888 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3667#-KQLnQB1Tv4YzVD_Xpz_:-KQN8pj8ROEBdarXm5Dc:bnkdky0) ([raw file](https://github.com/qtox/qtox/blob/3280e4be1cdf970bfe51d335bf5214aff9cda538/src/widget/form/chatform.cpp#L888)):_
<details><summary>

<i>Previously, Diadlo (Diadlo) wrote…</i></summary>



> Yes, I know, but I could not make it work without name clearing. Have no idea where the problem (maybe upstream problem?)
> </details>
> `:disabled` don't work to callButton/videoButton too. It brokes after button take objectName.
> Step:
> 0. Alise is offline (buttons is gray, no object name)
> 1. Alise comes online (buttons is green, object name is "green")
> 2. Alise goes offline (buttons is green, object name is "green")

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3667)_

Diadlo:
Review status: 13 of 14 files reviewed at latest revision, 2 unresolved discussions.

---

_[src/widget/form/chatform.cpp, line 888 [r1]](https://reviewable.io:443/reviews/qtox/qtox/3667#-KQLnQB1Tv4YzVD_Xpz_:-KQNp_h9uOq_d0vbf9KP:b-f7cqhc) ([raw file](https://github.com/qtox/qtox/blob/3280e4be1cdf970bfe51d335bf5214aff9cda538/src/widget/form/chatform.cpp#L888)):_
<details><summary>

<i>Previously, Diadlo (Diadlo) wrote…</i></summary>



> `:disabled` don't work to callButton/videoButton too. It brokes after button take objectName.
> Step:
> 0. Alise is offline (buttons is gray, no object name)
> 1. Alise comes online (buttons is green, object name is "green")
> 2. Alise goes offline (buttons is green, object name is "green")
> </details>
> I found reason of this but. `element#id` css-rule has higher priority, than `element:pseudo-class`
> In web-css we should use `!important`, but Qt currently doesn't implement this.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3667)_

antis81:
> I found reason of this but. element#id css-rule has higher priority, than element:pseudo-class
> In web-css we should use !important, but Qt currently doesn't implement this.

Thanks for the info! :+1: If you look into the concepts of QtQuick, it is unlike this will ever be implemented, because we have better solutions available and to my knowledge, QSS didn't evolve much over the past ~10 years. In the past I used to like CSS, but it highly lacks some important features: It is static by design and cannot reflect any dynamic logic (like "if status == call.active then background = red, else background = green"). And I know there's some parsers outside, that try to solve it (I wrote one myself, which is able to handle nested expressions etc.). But… it's the wrong approach actually. ^^

So, what is the right approach then? For Qt/C++, this is for sure the QStyle/QProxyStyle inheritance, which allows us to get a flexible control over colors, icons, borders, fonts, etc. and is more efficient. I started an implementation, but of course this cannot be done in a single day and requires to completely ditch CSS (maybe with one or two exceptions).

And just to "wet some appetite", in QML the follwing is valid:

``` qml
// this is only simple example of a button, you can enhance it with real states (like "active", "pending", "inactive", …)
Rectangle {
    id: root // this helps accessing properties

    property bool active: false

    signal clicked()

    // this is a property binding, it can hold a simple variable or even a complete method. Powerful stuff!
    color: active ? "red" : "green"

    MouseArea {
        // this is actually a slot, redirecting the signal
        onClicked: root.clicked()
    }
}
```

antis81:
Btw. QML example is working code. You can really use it in a surrounding context.

antis81:
Merged.

PR #3674, opened 2016-08-30
refactor: Disable RecursiveSignalBlocker if not available.

iphydf:
This is the second step to ensuring Qt 5.2 compatibility.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3674)

### DIFF
--- a/src/core/recursivesignalblocker.cpp
+++ b/src/core/recursivesignalblocker.cpp
@@ -20,7 +20,9 @@
 #include "recursivesignalblocker.h"
 
 #include <QObject>
+#if HAVE_QSIGNAL_BLOCKER
 #include <QSignalBlocker>
+#endif
 
 /**
 @class  RecursiveSignalBlocker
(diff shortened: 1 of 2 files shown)


Diadlo:
cc @antis81

antis81:
According to Qt documentation`QSignalBlocker` was introduced in Qt 5.3, which is also our minimum requirement to be able to build qTox. Further, simply removing RecursiveSignalBlocker functionality messes up settings. What's the point of this change?

iphydf:
@antis81 according to the PR that added it, this "may not be necessary anymore", so if that is not correct, can you tell me how this messes up settings?

The point is to make 5.2 the minimum requirement. Until the signal blocker was added, there was no reason to make 5.3 the minimum required version, and I think dropping support for Ubuntu 14.04 (which ships with 5.2) is a decision that should not be taken lightly.

initramfs:
@iphydf I think with Ubuntu 14.04 not shipping ffmpeg, compliation on such an OS isn't really possible anyway, especially with all the newer FFmpeg function calls being used in the video code. I haven't tried though nor am I considering manually compiling ffmpeg/libav.

iphydf:
ppa:mc3man/trusty-media has ffmpeg. I use it on the TokTok/qTox build.

antis81:
Rather than opening several PR's to support backwards compatibility, consider taking part in the discussion in #3661, which was just opened a few days ago. AFAIK qTox builds and installs on 14.04 (I actually used both setups until few moths ago upgraded to 16.04).

iphydf:
Having read the discussion on #3661, I see what kind of sentiment I'm dealing with here, so I understand that this PR and any discussion around it is useless.

PR #3685, opened 2016-09-04
chore: Statically link the windows platform plugin.

iphydf:
When building statically linked versions of qTox, we need to link in all
plugins it uses. Linking the windows platform plugin will make linking
succeed, and the program will start. Some additional plugins may be
needed for e.g. image formats. These require some testing on windows and
are out of scope for this PR.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3685)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -45,6 +45,11 @@
 #include "platform/install_osx.h"
 #endif
 
+#if defined(Q_OS_WIN32) && defined(QT_STATIC)
+#include <QtPlugin>
+Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin);
+#endif
+
 #ifdef LOG_TO_FILE
 static QAtomicPointer<FILE> logFileFile = nullptr;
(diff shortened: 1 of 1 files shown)


zetok:
@towlie mind testing?

towlie:
I don't think I can test this, I only have shared qtox dependencies. I once tried static but the qtox app crashed when I did this.

iphydf:
It crashed because it couldn't find the platform plugin. That's why this links the platform plugin statically.

towlie:
Oh this is interesting. Well, then I should try it again. But this needs some time.

towlie:
How does your change work? When shall one define QT_STATIC? Looks like I have to modify something more than you changed.

towlie:
I am unable to test this. Because qtox compiliation already fails without these PR changes when I try to make a static build. (I get some weird ld error at the last step)

iphydf:
No static builds then.

PR #3714, opened 2016-09-13
fix(settings): correct empty listbox audio device

### LINKED ISSUE #3709: Empty fields
##### Brief Description

OS: Windows 
qTox version: 1.5.1-97
Reproducible: Always 
##### Observed Behavior

In the absence of playback devices fields are empty
##### Additional Info

default

PKEv:
fix incorrect checking index
fix #3709

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3714)

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -455,10 +455,14 @@ void AVForm::getAudioInDevices()
     inDevCombobox->addItems(deviceNames);
     inDevCombobox->blockSignals(false);
 
-    int idx = Settings::getInstance().getAudioInDevEnabled()
-              ? deviceNames.indexOf(Settings::getInstance().getInDev())
-              : 0;
-    inDevCombobox->setCurrentIndex(idx < 0 ? 1 : idx);
+    int idx = 0;
+    if (Settings::getInstance().getAudioInDevEnabled() && deviceNames.size() > 1)
+    {
+        idx = deviceNames.indexOf(Settings::getInstance().getInDev()) + 1;
+        if (idx <= 0)
+            idx = 1;
+    }
+    inDevCombobox->setCurrentIndex(idx);
 }
 
 void AVForm::getAudioOutDevices()
@@ -471,10 +475,14 @@ void AVForm::getAudioOutDevices()
     outDevCombobox->addItems(deviceNames);
     outDevCombobox->blockSignals(false);
 
-    int idx = Settings::getInstance().getAudioOutDevEnabled()
-              ? deviceNames.indexOf(Settings::getInstance().getOutDev())
-              : 0;
-    outDevCombobox->setCurrentIndex(idx < 0 ? 1 : idx);
+    int idx = 0;
+    if (Settings::getInstance().getAudioOutDevEnabled() && deviceNames.size() > 1)
+    {
+        idx = deviceNames.indexOf(Settings::getInstance().getOutDev()) + 1;
+        if (idx <= 0)
+            idx = 1;
+    }
+    outDevCombobox->setCurrentIndex(idx);
 }
 
 void AVForm::on_inDevCombobox_currentIndexChanged(int deviceIndex)


Diadlo:
Reviewed 1 of 1 files at r1.

---

_[src/widget/form/settings/avform.cpp, line 461 at r1](https://reviewable.io:443/reviews/qtox/qtox/3714#-KRY_tnrAXZ0vsvpLoA1:-KRY_tnrAXZ0vsvpLoA2:b-7oiagu) ([raw file](https://github.com/qtox/qtox/blob/3c1ea1b8a705882b084c3a5ddf20d5ff2af02afe/src/widget/form/settings/avform.cpp#L461)):_

> ``` cpp
>               ? deviceNames.indexOf(Settings::getInstance().getInDev())
>               : 0;
>     inDevCombobox->setCurrentIndex(idx <= 0 ? 1 : idx);
> ```

0 is valid index. I think there should be:
`idx < 0 ? 0 : idx`

And maybe add check on max value?

---

_[src/widget/form/settings/avform.cpp, line 477 at r1](https://reviewable.io:443/reviews/qtox/qtox/3714#-KRY_zTBn3CNNloGDR2a:-KRY_zTBn3CNNloGDR2b:b-ya81ng) ([raw file](https://github.com/qtox/qtox/blob/3c1ea1b8a705882b084c3a5ddf20d5ff2af02afe/src/widget/form/settings/avform.cpp#L477)):_

> ``` cpp
>               ? deviceNames.indexOf(Settings::getInstance().getOutDev())
>               : 0;
>     outDevCombobox->setCurrentIndex(idx <= 0 ? 1 : idx);
> ```

Same as above

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714)_

PKEv:
correct

Diadlo:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714#-:-KRYfWOJmzjuulcSE1fU:bnfp4nl)_

antis81:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

_[src/widget/form/settings/avform.cpp, line 461 at r1](https://reviewable.io:443/reviews/qtox/qtox/3714#-KRY_tnrAXZ0vsvpLoA1:-KRZ8KiUbBNd9dXSoe2C:b-2vd12y) ([raw file](https://github.com/qtox/qtox/blob/3c1ea1b8a705882b084c3a5ddf20d5ff2af02afe/src/widget/form/settings/avform.cpp#L461)):_
<details><summary>

<i>Previously, Diadlo (Diadlo) wrote…</i></summary>



> 0 is valid index. I think there should be:
> `idx < 0 ? 0 : idx`
> 
> And maybe add check on max value?
> </details>
> First, there's an extraneous space after the first brace. Please remove. Then there's something you probably overlooked: idx == 1 means, that the default audio device will be selected, if the string from settings is not found in the list of available devices. The audio device should only be disabled (idx == 0), if there really are no audio devices available. As is, this PR solves a problem by introducing another one -> be prepared for users to moan about regression. ^^

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714)_

antis81:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

_[src/widget/form/settings/avform.cpp, line 477 at r1](https://reviewable.io:443/reviews/qtox/qtox/3714#-KRY_zTBn3CNNloGDR2a:-KRZAYJQoDz8ckoJ0BKn:b-80n7p6) ([raw file](https://github.com/qtox/qtox/blob/3c1ea1b8a705882b084c3a5ddf20d5ff2af02afe/src/widget/form/settings/avform.cpp#L477)):_
<details><summary>

<i>Previously, Diadlo (Diadlo) wrote…</i></summary>



> Same as above
> </details>
> So, if the device is disabled, you want to select the default device?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714)_

antis81:
@Diadlo Please do not merge this PR! It **will** introduce regression and that is not what we want to achiev, is it?

@PKEv I really like to merge fixes, but if you really want to fix a problem, you should understand it completely and not just part of it. We are actually dealing with two settings values here:
1. the audio enabled flag (independent of the selected device!)
2. the name of the selected audio device

The algo fails, if the list of available audio devices is empty (we don't have a default device then), which should never be the case -> check your log file for "warning" entries.

PKEv:
@antis81 
I don't understand the problem, I foresee the following scenario:
1. audio device missing - is what we fixed
2. audio device derived from the stored parameters none - this turns off the audio output. Use the default device I believe is not correct because user did not choose it

antis81:
@PKEv Ok, initialize your settings to a state, as if you just downloaded and installed qTox for the first time. If you don't have a VM at hand, you can do the following: Open your your local `~/.config/tox/qtox.ini` file and simply remove the `[Audio]` block. Mine looks as follows:

``` ini
[Audio]
audioInDevEnabled=true
audioOutDevEnabled=true
inDev=ALSA Default
inGain=11.9
outDev="HDA Intel, STAC9221 A1 Analog (CARD=Intel,DEV=0)"
outVolume=97
```

Now start qTox and see what happens -> my guess is, that the audio devices will probably be "disabled" by default instead of trying to select the default device, if available.

Let me explain in code (probably becomes the complete solution :smile:):

``` c++
method() {
    int idx = 0; // init with "device disabled" status (default is "enabled -> select default device")
    if (Settings::getInstance().getAudioInDevEnabled() && deviceNames.size() > 1) {
        idx = deviceNames.indexOf(Settings::getInstance().getInDev()) + 1; // <-- the bug was here, as the index was "off-by-one"!!
        if (idx <= 0)
            idx = 1; // select the available default device, which is the first entry in the name list
    }

    inDevCombobox->setCurrentIndex(idx);
```

Please try that code and see for yourself. If you go through any possible path you probably will notice there's quite a number of combinations to check… :)

Also note, you need to apply that code twice (audio input & output device).

Hope you understand better now.

antis81:
@PKEv Uhm… btw. have you noticed, that the "ALSA Default" string is not quoted as string? This might have an impact on the behaviour as well.

PKEv:
@antis81 
now I can't compile, but I think more correct is:

```
if (Settings::getInstance().getAudioInDevEnabled() && deviceNames.size() > 1) {
```

antis81:
@PKEv Yeah, sorry. Guess I wrote a little too quick. Positive condition is correct of course.

**EDIT:** … and `if (idx <= 0)` instead `idx == 0` is better. Corrected both in the example, so confusion should be complete. :P

PKEv:
correct

antis81:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714#-:-KRm1bAjTSCfh97d88CS:bnfp4nl)_

antis81:
@Diadlo Now the fix is complete. Mind to merge?

Diadlo:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3714#-:-KRtaBwAnHQYlNr7I4Hh:bnfp4nl)_

PR #3734, opened 2016-09-22
fix: send offline message history once

### LINKED ISSUE #2726: Messages are sent to me twice on some occasions

antis81:
~~closes 2726~~

@GrayHatter Would you review this? Moving the "deliverOfflineMsgs" call to the "non-ui classes" - where it belongs - is not possible in master yet. Unfortunately UI and data are highly bound and by far not separated, which is WIP in the "ui/redesign" branch). IOW: This cannot be completely solved/fixed yet.

Have a look at current implementation in `widget.cpp`:

``` c++
// This slot is called from "Core" when a friend's connection status changed (same trigger as for "corefile.cpp#L472").
void Widget::onFriendStatusChanged(/*…*/) {
  //… some code above

  // f = instance of Friend
  // i removed that distracting bullshit code around, checking if a status really changed, firing a singleshot timer etc.
  f->getChatForm()->getOfflineMsgEngine()->deliverOfflineMsgs();
}
```

In #3639 this completely vanished and is triggered only by the status change (from "Core"). So both changes are pointing in the same direction.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3734)

### DIFF
--- a/src/persistence/offlinemsgengine.cpp
+++ b/src/persistence/offlinemsgengine.cpp
@@ -18,106 +18,115 @@
 */
 
 #include "offlinemsgengine.h"
-#include "src/friend.h"
-#include "src/persistence/settings.h"
+
 #include "src/core/core.h"
+#include "src/friend.h"
 #include "src/nexus.h"
 #include "src/persistence/profile.h"
+#include "src/persistence/settings.h"
+
 #include <QMutexLocker>
-#include <QTimer>
 
-/**
- * @var static const int OfflineMsgEngine::offlineTimeout
- * @brief timeout after which faux offline messages get to be re-sent.
- * Originally was 2s, but since that was causing lots of duplicated
- * messages on receiving end, make qTox be more lazy about re-sending
- * should be 20s.
- */
-
-
-const int OfflineMsgEngine::offlineTimeout = 20000;
 QMutex OfflineMsgEngine::globalMutex;
 
-OfflineMsgEngine::OfflineMsgEngine(Friend *frnd) :
-    mutex(QMutex::Recursive),
-    f(frnd)
+class OfflineMsgEngine::Private
+{
+public:
+    Private(Friend* frnd)
+        : f(frnd)
+    {
+    }
+
+    void registerReceipt(int receipt, int64_t messageID, ChatMessage::Ptr msg,
+                         const QDateTime& ts = QDateTime::currentDateTime())
+    {
+        receipts[receipt] = messageID;
+        undeliveredMsgs[messageID] = {msg, ts, receipt};
+    }
+
+public:
+    Friend* f;
+    QHash<int, int64_t> receipts;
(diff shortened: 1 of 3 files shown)


zetok:
on 57d4411c0a0ccd3c2e09c20d51c76292d7dd6ced:

Not sure if I'm testing it right, but..

I've tried to send 11 messages, receiving end got those 11×3 messages.

Also:
- [ ] spectacle ch7295
  ↑ those messages were delivered only twice.
  
  Steps to reproduce:
  1. Quickly send messages, a few(?) per second. It seems that it can be triggered with >=3 messages per second.

# 

`sending | receiving`
spectacle lg7295
↑ Messages `1, 2, 3, 4` were delivered at the first try, but sending side marked only `1` as delivered, which resulted in sending again `2, 3, 4`.

---

@antis81 um, `offline messages` is too long for commit message scope

Btw, there's also https://github.com/qTox/qTox/blob/077a84446129076537b34d9f3b5d7143d7019b0f/src/widget/widget.cpp#L124, with a proper fix it could be decreased to 15s.

antis81:
> um, offline messages is too long for commit message scope

Changed to "chat".

antis81:
@zetok Thanks for putting this to a test. However by taking a closer look into the DB code, the PR cannot fix anything on this side. Looks like whoever wrote that stuff didn't actually know what he/she wanted to write… Why? A chat message has a "sent" status, which is not cleanly set. And this is where problems just begin… In addition, instead of adding a "send_status" column in the chat history, a `ROWID` is added into another table, which only has a single column -> resembling a generated (wtf?) message id. And now the "OfflineMsgEngine" comes into play, which is in turn absolutely not sync'ed to that not-sync'ed table. And worst: If the history is disabled, another code path is taken. Why not do the same thing, just one time with history enabled and the other time without the history. Who likes to reimplement this? :smile:

GrayHatter:
lol

GrayHatter:
Rip the whole thing out and start over.

Or just use uTox

GrayHatter:
I've been thinking about this. I'm assuming that qTox sends the message off to a different thread, so it doesn't block the UI while waiting to "post" the message to toxcore?

If that's right, I'm also going to guess that something in the process is clobbering qTox's method of tracking which messages were sent?

Otherwise I'm going to need more information if you want me to be more helpful.

zetok:

Diadlo:
This PR didn't fix the problem and inactive over 2 month. Thus closing. Open new one, if you want to continue

PR #3791, opened 2016-10-06
fix(avform): Fixed empty value in autio and video lists

Diadlo:
Removed extra `+1` for index and small refactoring
@antis81 [Here](https://github.com/qTox/qTox/pull/3714#issuecomment-246987436) you said, that `idx = deviceNames.indexOf(Settings::getInstance().getInDev()) + 1; // <-- the bug was here, as the index was "off-by-one"!!` Why?

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3791)

### DIFF
--- a/src/widget/form/settings/avform.cpp
+++ b/src/widget/form/settings/avform.cpp
@@ -456,11 +456,11 @@ void AVForm::getAudioInDevices()
     inDevCombobox->blockSignals(false);
 
     int idx = 0;
-    if (Settings::getInstance().getAudioInDevEnabled() && deviceNames.size() > 1)
+    bool enabled = Settings::getInstance().getAudioInDevEnabled();
+    if (enabled && deviceNames.size() > 1)
     {
-        idx = deviceNames.indexOf(Settings::getInstance().getInDev()) + 1;
-        if (idx <= 0)
-            idx = 1;
+        QString dev = Settings::getInstance().getInDev();
+        idx = qMax(deviceNames.indexOf(dev), 1);
     }
     inDevCombobox->setCurrentIndex(idx);
 }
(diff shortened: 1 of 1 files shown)


sudden6:
typo in commit title `audio`

sudden6:
Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3791)_

Diadlo:
@sudden6 commit message fixed

sudden6:
:lgtm:

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3791#-:-KTQpEtiraS-glNEZ3qU:bnfp4nl)_

antis81:
> Here you said, that idx = deviceNames.indexOf(Settings::getInstance().getInDev()) + 1; // <-- the bug was here, as the index was "off-by-one"!! Why?

Because ~~`combobox.count() != deviceNames.count()`~~ <\- this is bullshit:

| combobox index | AL name index | device name (avform) | settings "in/out enabled" | settings "device" |
| --- | --- | --- | --- | --- |
| 0 | n.a. => `QList::indexOf("\<device name\>") == -1` | "Disabled" | false | <unchanged> => initialized with default (e.g. "ALSA default") |
| >=1 | >=0 | <Device Name> | true|false | <Device Name> |

Now, when trying to open an audio device (e.g. by starting a phone call => AVForm is invisible), audio API checks the `enabled` flag, first and even if a device name was set, doesn't open the audio device.

This is horribly realized in the UI part -> by using a separate checkbox, the behaviour becomes transparent and clear - and actually useful. You wouldn't have had to ask at all. ^^

antis81:
**Follow-up:**
Actually I was sleeping :innocent: => `deviceNames.count() == Audio::outDevNames().count() + 1`.

The combobox is filled by the `deviceNames` list, which contains this nasty first dummy item. This item only controls the "enabled" flag and does not carry a device name. Before introducing this flag, the code was actually messing with an ambiguous device name (`QObject::tr("None")`). Thus, device lookup was unreliable. Now the device name always points to a physical device. The enabled state is a logical value, allowing to disable audio in/output without touching the device name at all. Handling via combobox actually has several disadvantages:
1. Selecting an audio device implicitly set it "enabled", which can be unwanted behaviour.
2. Implied by 1: The enabled flag cannot have full effect, when disabling the device and user has to reselect a device to enable regardless of the fact there already was one selected in settings.
3. Algorithm is harder to understand. With a simple additional checkbox, it becomes easy.

Guess it's all clear. If you like, feel free to add the checkbox. Otherwise , removing the `+1` is correct and the only logical change, this PR contains.

sudden6:
@antis81 Is you assigning this issue to me equal to you acknowledging the changes? I'm a little confused by this^^

Diadlo:
IMO: checkbox will make UI more complicated

PR #3818, opened 2016-10-18
chore(osx): link to ffmpeg and sqlcipher statically

zetok:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3818)


sudden6:
The macOS linker doesn't seem to be as flexible as the GNU linker when mixing static and dynamic linking, see https://stackoverflow.com/questions/4576235/mixed-static-and-dynamic-link-on-mac-os

PR #3820, opened 2016-10-18
style: remove unnecessary curly braces in `Widget::updateIcons()`

zetok:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3820)


Diadlo:
In real, I prefer user bracket everywhere. It make code style the same and help avoid mistakes, when you add new operator in `if` statement, but forget about brackets

zetok:
Ok, so we need to change that in style guidelines and code.

zetok:
Deprecated by #3826

Issue #3795, opened 2016-10-07
Update noptification icon properly


ghost:
Stupid Github doesn't let me upload an image file.
http://expirebox.com/download/277c38163de6bc3e92ed601611df48f6.html

zetok:
image from link:
![1922d177de47318d55087e8fc853affa](https://cloud.githubusercontent.com/assets/3148759/19211458/b4c74fee-8d34-11e6-85ba-cf80ae406989.jpg)

AliceGrey:
@zetok @GitHubHatesTor I am unable to reproduce this issue. I told qtox to connect to a proxy that was down and this is what rendered.
Edit: The icon still displays properly with ipv6 enabled. I just usually leave it off and double checked with it on.
![notification bar](http://i.imgur.com/MRZGeVJ.png)

zetok:
@AliceGrey Right.

Steps:
1. Connect to working proxy
2. Take proxy down
3. ![spectacle ea7954](https://cloud.githubusercontent.com/assets/3148759/19229254/18da7e26-8ec4-11e6-86ac-14a56fae4389.png)

AliceGrey:
@zetok Should this be closed since it's not reproducible and the creator hasn't responded?

zetok:
@AliceGrey err? It's reproducible, I've posted the steps.

AliceGrey:
@zetok You said > Right. I thought you meant you weren't able to reproduce it.

zetok:
Sorry, I should have been more verbose. What I meant was "Oh, so it's not obvious what the steps are. Right. Here are the steps"

AliceGrey:
@zetok No worries, I'll see what I can do to get this one fixed.

AliceGrey:
@zetok I've done some poking around in my free time and it looks like after being disconnected from the proxy and DHT `updateIcons()` never gets called. It does get called when it connects though.

AliceGrey:
@zetok From what I've read in the code, `updateIcons()` gets called by `onStatusSet()` which is connected to `Core::statusSet`. 

Where I get lost is when DHT gets disconnected when you take down the proxy, `emit disconnected();` gets called which then calls `onDisconnected` which  does `emit statusSet(Status::Offline);` which should trigger `onStatusSet()` and call `updateIcons()` but for some reason it never gets to `updateIcons()`

AliceGrey:
@zetok After A LOT of digging and debugging I found the culprate. I'll submit a PR :+1:

[...]

PR #3829, opened 2016-10-25
fix(settings): Properly update taskbar on disconnect

AliceGrey:
Fixes: #3795

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3829)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -626,7 +626,7 @@ void Widget::onConnected()
 void Widget::onDisconnected()
 {
     ui->statusButton->setEnabled(false);
-    emit statusSet(Status::Offline);
+    emit Core::getInstance()->statusSet(Status::Offline);
 }
 
 void Widget::onFailedToStartCore()


Diadlo:
@AliceGrey LGTM, but take a look on [my comment](https://github.com/qTox/qTox/issues/3795#issuecomment-255946921) and maybe better change `Core::setStatus`?

---

Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3829#-:-KUuT_IAdOtpYt5JdtsS:b-l6byol)_

Diadlo:
:lgtm:

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3829#-:-KUuVo7nfV2f0REC-8r1:bnfp4nl)_

AliceGrey:
@Diadlo I'll try it with your version and see if I can get it to work. It would probably be best to have core accept it as an option than hacking around it.

AliceGrey:
@Diadlo So it looks like in core there is no TOX_USER_STATUS_OFFLINE so we'd have to change the case statement a bit to get it to work without being called uninitalized. https://github.com/irungentoo/toxcore/blob/master/toxcore/tox.h#L289

AliceGrey:
@Diadlo Ok so it fixes the issue if you add `case Status::Offline: userstatus = TOX_USER_STATUS_NONE;break;` but that's technically incorrect because TOX_USER_STATUS_NONE is supoosed to signify "User is online and available" but it works. It looks like it might be an issue for the new toktok migration.

Edit: As Diadlo mentioned adding `emit statusSet(status);` to the `default`case would also work without having to incorrectly use toxcore.

Diadlo:
I'm not sure, what will be better. Next two points leads to the problem.
1) toxcore doesn't consider `offline` as status. It's just disconnected user.
2) qTox -- consider `offline` as status.

If you whant to change the core, you can add `emit statusSet(status)` in `default` case
@sudden6 @initramfs What you think?

initramfs:
Doesn't toxcore expose a `tox_friend_get_connection_status()` function? This returns an enum:

``` c
/**
 * Protocols that can be used to connect to the network or friends.
 */
typedef enum TOX_CONNECTION {

    /**
     * There is no connection. This instance, or the friend the state change is
     * about, is now offline.
     */
    TOX_CONNECTION_NONE,

    /**
     * A TCP connection has been established. For the own instance, this means it
     * is connected through a TCP relay, only. For a friend, this means that the
     * connection to that particular friend goes through a TCP relay.
     */
    TOX_CONNECTION_TCP,

    /**
     * A UDP connection has been established. For the own instance, this means it
     * is able to send UDP packets to DHT nodes, but may still be connected to
     * a TCP relay. For a friend, this means that the connection to that
     * particular friend was built using direct UDP packets.
     */
    TOX_CONNECTION_UDP,

} TOX_CONNECTION;
```

Can't friends appear to be offline if there is no connection? I feel like the purpose of `tox_friend_get_status()` is to get the state of the friend when there is an active connection, the results being invalid otherwise (since how would we know if a friend is away/busy/available if we have no connection to them.)

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3829#-:-KV-Re-Zx-GF0mW3JIHE:b-ekc0hr)_

Diadlo:
@initramfs So, Are you agree, that the current implementation is right?

initramfs:
@Diadlo Sorry I'm not sure what you mean. Do you the implementation provided by this PR or the existing implementation?

I just had a look at the whole status stuff in qTox and it's an absolute mess. I can't seem to keep track of what signals are connected to which slots and when, especially when you have the same named signals in different classes.

AFAIK, to toxcore "offline" is not a status, it is a state, completely different things. Try to merge the two into "statuses" in general will make it very hard to have it do the right thing without going through hoops of stuff. In this case the PR fixes the behaviour to be correct, but at the cost of emitting a signal from `Core` when it actually came from `Widget`. This solution is also brittle that if this `Widget` function is removed or refactored, you now have broken behaviour from `Core` again. It would also be hard to track down the fact that `Widget` was actually emitting signals `Core` should be.

Whilst this PR can serve as a hotfix, I'd prefer to have separate constructs for a user/friend's status and their connection type (None, TCP, UDP). It seems that `Core::onConnectionStatusChanged()` exists and essentially shoehorns connection type into status (done incorrectly too, forced a friend's status to be either `Online` or `Offline` rather than the spectrum of statuses). In short, any code that checks for the user's status or friend's status must consult both the connection type and the currently set status under the following pseudocode:

```
if(connection is TCP || UDP){
    // Update status as set by user status
}else{
    // Always show offline regardless of user status
}
```

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3829#-:-KV94QgRKtf6GcnNSjTT:bx8z9xj)_

Diadlo:
Ok. It will merge it as hotfix today

PR #3870, opened 2016-11-11
fix(chatlog): silence warning about unused parameters

sudden6:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3870)

### DIFF
--- a/src/chatlog/content/text.cpp
+++ b/src/chatlog/content/text.cpp
@@ -150,6 +150,8 @@ QRectF Text::boundingRect() const
 
 void Text::paint(QPainter* painter, const QStyleOptionGraphicsItem* option, QWidget* widget)
 {
+    Q_UNUSED(option);
+    Q_UNUSED(widget);
 
     if (!doc)
         return;


Diadlo:
:lgtm:
But what you think about using [`__attribute__((__unused__))`](http://stackoverflow.com/a/12891181/6273701) instead of `Q_UNUSED`?

---

Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3870#-:-KWJ1N5lnxs6JU3rY53S:b-ik3s1f)_

sudden6:
Not much, since we're using Qt anyway we might as well use `Q_UNUSED`. It's also used in the rest of the codebase.

Diadlo:
We can use variable after `Q_UNUSED`, so it can say wrong information to the reader (i.e. variable marked as unused, but in real, is used)

---

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/3870#-:-KWJi3xxwesIoxZXsiOb:b-vq07ci)_

sudden6:
Ok, didn't know that. But the attribute looks like something GCC specific, what about Clang?

Diadlo:
Seems, clang [support it](http://clang.llvm.org/docs/AttributeReference.html#maybe-unused-unused-gnu-unused)

sudden6:
ok, maybe we should provide our own `UNUSED` macro then in the future. Should I still merge this as an intermediate solution?

Diadlo:
Of course. That a reason, why I mark it as lgtm :)

initramfs:
@Diadlo @sudden6 Please don't use compiler specifics that aren't part of the language definition, just because it's supported by some compiler X doesn't mean it's safe for general use. For example, what if qTox was ported to MSVC, can you guarantee support? It would be tragic if qTox fails to compile on certain systems for a "feature" that doesn't even impact code generation.

As much as I am for silencing warnings, introducing additional complexity (especially with macros) simply causes more problems than it solves (imagine how much fun it would be to debug a bug with the `UNUSED` macro which somehow fails only under certain conditions). As far as I know, omitting the variable name in most cases will shut the compiler up and prevent accidental use (since how can you use the variable if it's not named).

If you really want an official way to silence the compiler warning whilst keeping the variable name, you're stuck with `Q_UNUSED` from Qt, the `[[maybe_unused]]` attribute (but that's C++17) or assigning things to `std::ignore` (which is probably the closest you can get without writing ugly hacks within C++ itself).

sudden6:
@initramfs wouldn't then be a macro that is if defined to Q_UNUSED or the attribute a good idea? But I'm also ok with only Q_UNUSED atm.

initramfs:
@sudden6 Again, what's wrong with simply removing the variable name if it's truly unused?

This PR could have avoided the whole discussion if it had simply changed the function definition signature to:

``` c++
void Text::paint(QPainter* painter, const QStyleOptionGraphicsItem*, QWidget*)
```

But anyway, what is the objective of such a macro? How are you going to conditionally check for the existence attribute specifier and have it be applied correctly? Are you sure you can handle all expressions correctly? What if I tried to mark a function as unused, would `Q_UNUSED` work there? If so, does it have the same meaning as the GCC attribute specifier for unused variables?

I still see this as additional complexity that neither improves development nor helps the build system and I can definitely see it biting people in the back in the future. What is so inherently evil about the compiler warning you that variables are unused that causes a whole PR with a comment discussion to be created just to discuss the issue?

PR #3902, opened 2016-11-21
refactor: major ui redesign

antis81:
Mileston PR for UI redesign development.

While dealing with current settings(-ui), audio, video… issues, I thought about how much qTox would improve by simplifying the (UI) code and requesting (memory and other) resources when needed and freeing them, when no longer required. My first thought was to just implement this for the settings UI. While this works fine, it highly suffers from the "spaghetti crap code" in our beloved "Widget" class, which actually is a really destructive thing, which is hard - very hard! - to control. It is so much spaghetti, that refactoring is not possible and feature implementations… well, forget about that. So I decided, it deserves a "redesign" and gave that a try silently in this branch during the last weeks.

While there's quite an amount of work to be done here and it is unfinished, most of the stuff actually works now and I'd like to share, so you can see for yourself. You may ask, why I'm going to do such an "insane" thing. So here's some answers to that:
- In current state, it shrinks down qTox's memory footprint by >50% (60M) - and it will even shrink more! If you wonder, why qTox easily eats up ~120M and more RAM (even without any active chat) - this is why!
- As a neat side effect, we get a visible performance improvement, when resizing the qTox window(s).
- Code-Flow gets **a lot** easier to understand and to follow for contributors. Lots of "redundant" code gets smashed and completely removed, making qTox a more lightweight experience.
- It opens future opportunities, which currently seem impossible.

So, what features do we get?
- Fluid, responsive user interface.
- Clean resource management: The "resource subscription" problem is solved, which is mentioned in so many issues.

Visible features:
- As said above: Overall more "fluid" user experience.
- Finally! -> Reliable representation of a friend's actual "online status".
- More accurate positioning of attached and detached widgets on screen!
- No more suddenly "disappearing" widgets. 
- A neat side effect of the now centralized "(sub)widget display management" now allows for an (optional) "auto-flip" feature, depending on the main window's x-position on screen. This works in attached and detached (!) mode - I'd say it's ~95% done, but has few minor glitches.
- Not checked, but offline message delivery may work more accurately now. At least it works minimum as before.

Features for coders:
- No more "show/hide/activateWindow/raise/…" workarounds! It is cleanly handled by `Widget::showContentWidget()`.
- Widgets are "free'd" from the "ContentLayout / ContentDialog" burden and do what they are meant to do again. However most widgets now inherit "ContentWidget", which provides a unified look.
- Create widgets as you need them and free them, when you're done. When using the `ContentWidget` base class, the widget destroys automatically after closing it. So, if the close is rejected (e.g. by the infamous "Save changes before close?" dialog), the widget stays visible and won't get destroyed!
# TODO (known issues):

There's quite some work to finish, until this can be considered stable.
## Style
- [ ] For some reason, the widget style seems to randomly fallback to system's theming during `ContentWidget` creation. This might be due to mixed use of QPalette and CSS style-sheets. Probably we should create a style-sheet for this purpose.
- [x] Due to the fact, that we now use a lot less layouts, the bottom and right margins need to be adjusted. Anyway we now make good use of available space.
## FriendListWidget
- [ ] Friend/Group doesn't "un-highlight", when switching between chats. An active chat should be marked with "white background", and inactive with the current style's default background. Note, that in "multiple windows" mode, multiple chats can be visible (selected -> white BG).
## Chat (single & group):
- [x] Chat appears empty, after switching to another chat (-> load history).
- [x] Call/Video-Call buttons don't reflect the call's status:
  - Buttons stay "green" or "grey", depending on the friends status after switching to another chat). We need to request the ToxCall's state on widget initialization.
- [x] Typing status falsely reports "[user] is typing…" instead of user name.
- [x] The button icons for "mute microphone" and "mute output volume" unfortunately contain the status colors (green, red, grey, …) - contrary to the "AV call" buttons, which use a background color. Someone has the time to draw two lousy SVG's?
- [x] The video widget is not re-initialized and disappears after switching to another chat.
- [x] The video call button disappears as well when switching to another chat.
- [x] Separate "FriendWidget" from Friend data class (just like with ChatForm).
- [ ] Make sure everything works properly with group chats as well.
## ProfileForm:
- [x] No profile Information is displayed at all.
## SettingsWidget & it's sub-widgets:

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3639)


zetok:

Diadlo:

PR #3937, opened 2016-12-05
fix(UI): prevent status message from stretching profile corner

zetok:
Profile corner could stretch vertically if status message contained
newlines.

Nothing else beside setting a maximum vertical size worked.

currently:
spectacle pe2998

with PR:
spectacle rw2355

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3937)

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -880,6 +880,12 @@ QSplitter:handle{
                    <verstretch>0</verstretch>
                   </sizepolicy>
                  </property>
+                 <property name="maximumSize">
+                  <size>
+                   <width>16777215</width>
+                   <height>15</height>
+                  </size>
+                 </property>
                  <property name="palette">
                   <palette>
                    <active>


zetok:
This probably can introduce a regression with high DPI settings, so someone who can should test.

sudden6:
Is this rectangle always displayed over the other contacts? Because if it is, it looks even worse IMO.

antis81:
@sudden6 It's a tooltip… ^^

sudden6:
oh, sorry then, it wasn't that clear from the picture :)

Diadlo:
Now not quite clear, that status message partially hidden. I think we should add `...` in the end of message if it's multiline

zetok:
>Now not quite clear, that status message partially hidden. I think we should add ... in the end of message if it's multiline

Right. Dunno how to do that.

Diadlo:
I will help you

Diadlo:
@zetok Sorry, but seems it's impossible with current implementation

PR #4001, opened 2016-12-22
refactor: core & nexus

antis81:
A long overdue cleanup of the core code flow. As a side effect, this greatly helps to better understand qTox code (especially for newcomers), improves stability etc.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4001)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -53,29 +53,39 @@
 
 const QString Core::CONFIG_FILE_NAME = "data";
 const QString Core::TOX_EXT = ".tox";
-QThread* Core::coreThread{nullptr};
 
 #define MAX_GROUP_MESSAGE_LEN 1024
 
-Core::Core(QThread *CoreThread, Profile& profile) :
-    tox(nullptr), av(nullptr), profile(profile), ready{false}
+Core::Core(Profile& profile)
+    : tox(nullptr)
+    , av(nullptr)
+    , toxTimer(new QTimer)
+    , profile(profile)
+    , ready{false}
+    , coreThread(new QThread)
 {
-    coreThread = CoreThread;
+    coreThread->setObjectName("qTox Core");
+    connect(coreThread, &QThread::started,
+            this, &Core::onStarted);
+    connect(coreThread, &QThread::finished,
+            coreThread, &QThread::deleteLater);
 
-    toxTimer = new QTimer(this);
     toxTimer->setSingleShot(true);
-    connect(toxTimer, &QTimer::timeout, this, &Core::process);
-    connect(&Settings::getInstance(), &Settings::dhtServerListChanged, this, &Core::process);
+    connect(toxTimer, &QTimer::timeout,
+            this, &Core::process);
 
+    connect(&Settings::getInstance(), &Settings::dhtServerListChanged,
+            this, &Core::process);
+
+    moveToThread(coreThread);
+    toxTimer->moveToThread(coreThread);
 }
 
 void Core::deadifyTox()
 {
-    if (av)
-    {
-        delete av;
-        av = nullptr;
-    }
+    delete av;
+    av = nullptr;
+
     if (tox)
     {
         tox_kill(tox);
@@ -85,22 +95,21 @@ void Core::deadifyTox()
 
 Core::~Core()
 {
-    if (coreThread->isRunning())
-    {
-        if (QThread::currentThread() == coreThread)
-            killTimers(false);
-        else
-            QMetaObject::invokeMethod(this, "killTimers", Qt::BlockingQueuedConnection,
-                                      Q_ARG(bool, false));
-    }
-    coreThread->exit(0);
-    while (coreThread->isRunning())
+    coreThread->exit();
+    while (!coreThread->wait(500))
     {
         qApp->processEvents();
-        coreThread->wait(500);
     }
 
     deadifyTox();
+}
+
+/**
+ * @brief Starts the core thread.
+ */
+void Core::start()
+{
+    coreThread->start();
 }
 
 /**
@@ -238,7 +247,7 @@ void Core::makeTox(QByteArray savedata)
 /**
  * @brief Initializes the core, must be called before anything else
  */
-void Core::start()
+void Core::onStarted()
 {
     bool isNewProfile = profile.isNewProfile();
     if (isNewProfile)
@@ -374,8 +383,9 @@ void Core::process()
         tolerance = 3*CORE_DISCONNECT_TOLERANCE;
     }
 
-    unsigned sleeptime = qMin(tox_iteration_interval(tox), CoreFile::corefileIterationInterval());
-    toxTimer->start(sleeptime);
+    unsigned sleeptime = qMin(tox_iteration_interval(tox),
+                              CoreFile::corefileIterationInterval());
+    toxTimer->start(static_cast<int>(sleeptime));
 }
 
 bool Core::checkConnection()
@@ -1496,19 +1506,11 @@ void Core::setNospam(uint32_t nospam)
     emit idSet(getSelfId().toString());
 }
 
-/**
- * @brief Returns the unencrypted tox save data
- */
-void Core::killTimers(bool onlyStop)
+void Core::killTimers()
 {
-    assert(QThread::currentThread() == coreThread);
+    assert(QThread::currentThread() == thread());
     av->stop();
     toxTimer->stop();
-    if (!onlyStop)
-    {
-        delete toxTimer;
-        toxTimer = nullptr;
-    }
 }
 
 /**
@@ -1517,14 +1519,14 @@ void Core::killTimers(bool onlyStop)
  */
 void Core::reset()
 {
-    assert(QThread::currentThread() == coreThread);
+    assert(QThread::currentThread() == thread());
 
     ready = false;
-    killTimers(true);
+    killTimers();
     deadifyTox();
 
     emit selfAvatarChanged(QPixmap(":/img/contact_dark.svg"));
     GUI::clearContacts();
 
-    start();
+    onStarted();
 }
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -38,18 +38,18 @@
 #include <limits>
 #include <functional>
 
+#include <QBuffer>
+#include <QCoreApplication>
 #include <QDebug>
 #include <QDir>
 #include <QFile>
+#include <QDateTime>
+#include <QList>
+#include <QMutexLocker>
 #include <QSaveFile>
 #include <QStandardPaths>
 #include <QThread>
 #include <QTimer>
-#include <QCoreApplication>
-#include <QDateTime>
-#include <QList>
-#include <QBuffer>
-#include <QMutexLocker>
 
 const QString Core::CONFIG_FILE_NAME = "data";
 const QString Core::TOX_EXT = ".tox";
@@ -104,19 +104,21 @@ Core::~Core()
 }
 
 /**
- * @brief Returns the global widget's Core instance
+ * @brief Returns the profile's Core instance.
+ * @return The profile's Core instance or nullptr, if profile not initialized.
  */
 Core* Core::getInstance()
 {
-    return Nexus::getCore();
+    Profile* profile = Nexus::getInstance().getProfile();
+    return profile ? profile->getCore() : nullptr;
 }
 
-const CoreAV *Core::getAv() const
+const CoreAV* Core::getAv() const
 {
     return av;
 }
 
-CoreAV *Core::getAv()
+CoreAV* Core::getAv()
 {
     return av;
 }
@@ -282,7 +284,7 @@ void Core::start()
 
     // TODO: This is a backwards compatibility check,
     // once most people have been upgraded away from the old HistoryKeeper, remove this
-    if (Nexus::getProfile()->isEncrypted())
+    if (Nexus::getInstance().getProfile()->isEncrypted())
         checkEncryptedHistory();
 
     loadFriends();
@@ -613,7 +615,7 @@ void Core::requestFriendship(const QString& friendAddress, const QString& messag
             if (message.length())
                 inviteStr = tr("/me offers friendship, \"%1\"").arg(message);
 
-            Profile* profile = Nexus::getProfile();
+            Profile* profile = Nexus::getInstance().getProfile();
             if (profile->isHistoryEnabled())
                 profile->getHistory()->addNewMessage(userId, inviteStr, getSelfId().getPublicKeyString(), QDateTime::currentDateTime(), true, QString());
             emit friendAdded(friendId, userId);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -305,12 +305,15 @@ int main(int argc, char *argv[])
     else if (eventType == "save")
         handleToxSave(firstParam.toUtf8());
 
-    // Run
     int errorcode = a.exec();
 
     Nexus::destroyInstance();
     CameraSource::destroyInstance();
-    qDebug() << "Clean exit with status" << errorcode;
+
+    if (errorcode)
+    {
+        qWarning() << "qTox exited with error code" << errorcode;
+    }
 
 #ifdef LOG_TO_FILE
     logFileFile.store(nullptr);   // atomically disable logging to file
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -310,7 +310,6 @@ int main(int argc, char *argv[])
 
     Nexus::destroyInstance();
     CameraSource::destroyInstance();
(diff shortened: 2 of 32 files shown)


src/widget/form/chatform.cpp
```diff
@@ -854,7 +856,8 @@ void ChatForm::onScreenshotTaken(const QPixmap &pixmap) {
 
 void ChatForm::onLoadHistory()
 {
-    if (!Nexus::getProfile()->isHistoryEnabled())
+    Profile* profile = Nexus::getInstance().getProfile();
+    if (!profile->isHistoryEnabled())
```
Diadlo: Add brackets

src/persistence/profile.h
```diff
@@ -26,8 +26,12 @@
 #include <QByteArray>
 #include <QPixmap>
 #include <QObject>
-#include <tox/toxencryptsave.h>
+
 #include <memory>
+
+#include <tox/toxencryptsave.h>
+
+#include <core/core.h>
```
Diadlo: Use `"` for local include

src/persistence/historykeeper.cpp
```diff
@@ -205,7 +207,8 @@ QString HistoryKeeper::getHistoryPath(QString currentProfile, int encrypted)
     if (currentProfile.isEmpty())
         currentProfile = Settings::getInstance().getCurrentProfile();
 
-    if (encrypted == 1 || (encrypted == -1 && Nexus::getProfile()->isEncrypted()))
+    const Profile* profile = Nexus::getInstance().getProfile();
+    if (encrypted == 1 || (encrypted == -1 && profile->isEncrypted()))
```
Diadlo, line 213: Same as above

src/persistence/historykeeper.cpp
```diff
@@ -91,8 +91,10 @@ bool HistoryKeeper::checkPassword(std::shared_ptr<Tox_Pass_Key> passkey, int enc
     if (!Settings::getInstance().getEnableLogging() && (encrypted == -1))
         return true;
 
+    const Profile* profile = Nexus::getInstance().getProfile();
     if ((encrypted == 1) || (encrypted == -1))
-        return EncryptedDb::check(passkey, getHistoryPath(Nexus::getProfile()->getName(), encrypted));
+        return EncryptedDb::check(passkey, getHistoryPath(profile->getName(),
```
Diadlo, line 97: Same as above

src/nexus.cpp
```diff
@@ -263,18 +239,11 @@ Profile* Nexus::getProfile()
...
 
-/**
- * @brief Get desktop GUI widget.
- * @return nullptr if not started, desktop widget otherwise.
- */
-Widget* Nexus::getDesktopGUI()
-{
-    return getInstance().widget;
+    if (mProfile)
+        Settings::getInstance().loadPersonal(mProfile);
```
Diadlo, line 248: Add braces

src/nexus.cpp
```diff
@@ -150,40 +152,44 @@ void Nexus::start()
  */
 void Nexus::showLogin()
 {
+    loginScreen = new LoginScreen();
+
+#ifdef Q_OS_MAC
+    connect(loginScreen, &LoginScreen::windowStateChanged, this,
```
Diadlo: Can move `this` on the next line?

src/nexus.cpp
```diff
@@ -49,26 +49,31 @@
...
     widget{nullptr},
     loginScreen{nullptr}
 {
 }
 
 Nexus::~Nexus()
 {
-    delete widget;
-    delete loginScreen;
-    delete profile;
```
Diadlo, line 70: Why removed?

Diadlo (commented):
Made a brief overview. Also, can you add docs to methods which you modified?

antis81:
@Diadlo Style corrections done! Whatsoever… :smile:

antis81:
Review status: 0 of 32 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/nexus.cpp, line 158 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fO-r1-158:-KZgRJZHrDoxcpmDL9Xi:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/nexus.cpp#L158)):*

Can move `this` on the next line?
</blockquote></details>

Done.

---

*[src/nexus.cpp, line 246 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fO-r1-246:-KZgRRnMz_DpUa19FBQt:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/nexus.cpp#L246)):*

Add braces
</blockquote></details>

Done.

---

*[src/persistence/historykeeper.cpp, line 96 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fQ-r1-96:-KZgRSt84WdToQt8Q3Jh:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/persistence/historykeeper.cpp#L96)):*

Same as above
</blockquote></details>

Done.

---

*[src/persistence/historykeeper.cpp, line 211 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fQ-r1-211:-KZgRUqHT7Y4vCIpjLHj:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/persistence/historykeeper.cpp#L211)):*

Same as above
</blockquote></details>

Done.

---

*[src/persistence/profile.h, line 34 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fT-r1-34:-KZgRVmHPTH8LWrJMvfN:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/persistence/profile.h#L34)):*

Use `"` for local include
</blockquote></details>

Done.

---

*[src/widget/form/chatform.cpp, line 860 at r1](https://reviewable.io:443/reviews/qtox/qtox/4001#-KZaX54GcV9Qw9RmQ0fb-r1-860:-KZgRWai28CoEdA_CuNj:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/55a095e48914741391c886cc59b7053ff9270a75/src/widget/form/chatform.cpp#L860)):*

Add brackets
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4001)*

src/nexus.cpp
antis81: 1. Have a look [here](https://github.com/qTox/qTox/pull/4001/files#diff-39d71983c1fc16a6b3191ee87174baaeR96) and [here](https://github.com/qTox/qTox/pull/4001/files#diff-1e0d837757bc75752ac84d02d7e8ed44R40) -> double free -> crash
2. Widget and LoginScreen instances are not "owned" by the Nexus instance, but rather (mis-)used for page switching. So one of both members is always a `nullptr`. This is a classic use case for `QMainWindow::setCentralWidget`. Probably too much for the moment… :)

EDIT: One more thing!
The Nexus instance is destroyed after (!) the main event queue is left (-> QApplication::exec()). IOW: The widget (whichever of them was active) would be in an undefined state, which is very bad.

zetok:

antis81:
@initramfs Reassigned to @Diadlo for review. If you find time, feel free to review…

zetok:

src/main.cpp
```diff
@@ -307,6 +307,8 @@ int main(int argc, char *argv[])
 
     int errorcode = a.exec();
 
+    Settings::getInstance().saveGlobal();
```
Diadlo, line 310: I'm not against it, but could you tell why moved?
Also add explantaion in commit text

src/nexus.h
```diff
@@ -41,18 +41,27 @@ class Nexus : public QObject
...
+    void start();
+    void showMainGUI();
+
+public:
+    inline Widget* getDesktopGUI() const
+    {
+        return widget;
+    }
+
+    inline Profile* getProfile() const
```
Diadlo, line 57: Why made `inline` and moved in to the `.h`?

src/nexus.h
```diff
@@ -89,7 +98,7 @@ public slots:
     ~Nexus();
 
 private:
-    Profile* profile;
+    Profile* mProfile;
```
Diadlo, line 100: Why it was renamed?

src/nexus.cpp
```diff
@@ -160,31 +152,35 @@ void Nexus::start()
...
     widget = nullptr;
 
     delete mProfile;
     mProfile = nullptr;
 
     loginScreen->reset();
-    loginScreen->move(QApplication::desktop()->screen()->rect().center() - loginScreen->rect().center());
     loginScreen->show();
-    ((QApplication*)qApp)->setQuitOnLastWindowClosed(true);
+    loginScreen->move(QApplication::desktop()->screen()->rect().center() -
```
Diadlo, line 171: I think, it should be moved before show? It will prevent showing in wrong place on few second if system will produse some freeze

src/core/core.cpp
```diff
@@ -53,29 +53,39 @@
...
 const QString Core::CONFIG_FILE_NAME = "data";
 const QString Core::TOX_EXT = ".tox";
-QThread* Core::coreThread{nullptr};
 
 #define MAX_GROUP_MESSAGE_LEN 1024
 
-Core::Core(QThread *CoreThread, Profile& profile) :
-    tox(nullptr), av(nullptr), profile(profile), ready{false}
+Core::Core(Profile& profile)
+    : tox(nullptr)
```
Diadlo, line 60: What about using `{}`?

Diadlo:
Last commit should be renamed to `style: cleanup style` ;)

src/persistence/profile.h
```diff
@@ -45,7 +46,11 @@ class Profile : public QObject
     static Profile* createProfile(QString name, QString password);
     ~Profile();
 
-    Core* getCore();
+    inline Core* getCore()
```
Diadlo, line 49: Why inlined?

src/persistence/historykeeper.cpp
```diff
@@ -209,9 +211,13 @@ QString HistoryKeeper::getHistoryPath(QString currentProfile, int encrypted)
 
     const Profile* profile = Nexus::getInstance().getProfile();
     if (encrypted == 1 || (encrypted == -1 && profile->isEncrypted()))
+    {
         return baseDir.filePath(currentProfile + ".qtox_history.encrypted");
+    }
     else
```
Diadlo, line 217: What the reason of `else` if we use `return` in both cases?
Also, it can be simplified:
```
QString path = currentProfile + QStringLiteral(".qtox_history")
if (encrypted == 1 || (encrypted == -1 && profile->isEncrypted())) 
{
    path += QStringLiteral(".encrypted");
}

return baseDir.filePath( + ext);
```

src/nexus.h
sudden6: `inline` functions need to be in the *.h file IIRC

src/nexus.cpp
sudden6: I don't think a not-shown window can be moved

src/nexus.h
Diadlo: Yes. More correct question is: why made `inline`?

src/nexus.cpp
Diadlo: @sudden6 But it was (see old code)

antis81: Moving/resizing a window will fail before the **first** show(), which also initializes the widget position and size. This is a logical consequence of the fact, that move/resize events are **deferred** as long as a widget is invisible. Especially on X11 based systems, there are further factors playing a role. See Qt docs for [QWidget::setGeometry](http://doc.qt.io/qt-5/qwidget.html#geometry-prop) and [Window Geometry](http://doc.qt.io/qt-5/application-windows.html#x11-peculiarities) for further information.

src/nexus.h
antis81: @Diadlo The correct question is: Why don't we inline getters in general? :smile:

Basically there's two approaches:
1. No PIMPL => getters **should be** inlined in general
2. PIMPL => getters cannot be inline'd (due to the nature of a private class residing in implementation)

Inline means nothing but compiling the function body into the code, which is (tada!!) a pointer or reference. However, if the compile algorithm finds it is not possible, it won't do it anyway and add use a function pointer instead. Btw. you won't find much inline'd getters in the Qt internal framework, as the concept is highly pimp(l)'ed. :smile:

src/nexus.h
antis81: Well, how would you read the following (valid) code?

```c++
Game::Game(Ball* ball)
{
    ball = ball; // "noop" (even more: this->ball is not initialized!)
}
```

…and so does the compiler! :smile:

To differenciate, the name must be different or scoped. Scoping however includes a high risk of failure. Thus it is done by prefixing a member, because it is a) clear, b) less error prone and c) can be used in c'tor initializer lists:

```c++
// works
mProfile = profile;

// also works
this->profile = profile;

// doesn't work
SomeClass(Ptr* profile) 
: this->profile(profile)
{
}
```

The question is: "Why qTox decided against a prefix in class members?" (This is a recommended practice in OOP.)

sudden6: Maybe add that to the Coding Style PR?

Diadlo: Agree with @sudden6
IMO, this practice should be everywhere or nowhere

src/nexus.h
Diadlo: I just don't like methods in `.h` file. If you think, that it's make code better -- ok.

src/nexus.cpp
sudden6: @antis81 After reading the linked documentation, I think the correct approach would be to call `show()` after `move()` because it will apparently not fail as I thought. I think @Diadlo is right that it'll provide a more glitch free experience.

sudden6:
Reviewed 15 of 26 files at r6, 1 of 3 files at r7, 4 of 6 files at r10, 3 of 7 files at r11.
Review status: 23 of 32 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4001)*

Diadlo:
Reviewed 3 of 32 files at r1, 2 of 5 files at r2, 1 of 4 files at r3, 1 of 9 files at r4, 6 of 32 files at r5, 13 of 26 files at r6, 2 of 3 files at r7, 3 of 6 files at r10, 5 of 7 files at r11, 2 of 4 files at r12.
Review status: 29 of 32 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/nexus.cpp, line 59 at r5](https://reviewable.io:443/reviews/qtox/qtox/4001#-K__MOhlz8eEE9DMeqPH:-K__MOhlz8eEE9DMeqPI:bctv03p) ([raw file](https://github.com/qtox/qtox/blob/2c1b47b263789f7872dd72c355a449245ae9b471/src/nexus.cpp#L59)):*
> ```cpp
> 
> Nexus::Nexus() :
>     QObject(),
> ```

Seems, now `QObject` redundant here

---

*[src/persistence/profile.h, line 35 at r12](https://reviewable.io:443/reviews/qtox/qtox/4001#-K__NWavI_h8E5wB3M9x:-K__NWavI_h8E5wB3M9y:bjviy2e) ([raw file](https://github.com/qtox/qtox/blob/3f87cd6d17dd7863020f1b3feb68a21a9963c97b/src/persistence/profile.h#L35)):*
> ```objc
> 
> #include "core/core.h"
> #include "persistence/history.h"
> ```

[Use `src/`](https://github.com/qTox/qTox/pull/3986)

---

*[src/persistence/toxsave.cpp, line 20 at r12](https://reviewable.io:443/reviews/qtox/qtox/4001#-K__OMV7hdgxRdjuR86B:-K__OMV8CRkxrqYjw5i0:b-p6o91p) ([raw file](https://github.com/qtox/qtox/blob/3f87cd6d17dd7863020f1b3feb68a21a9963c97b/src/persistence/toxsave.cpp#L20)):*
> ```cpp
> */
> 
> #include "toxsave.h"
> ```

Use `src/`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4001)*

Diadlo:

Diadlo:

PR #4107, opened 2017-01-29
chore(macdeploy): make qtox build with cmake and mac os

sudden6:
Patch provided by vit9696 from #qtox@freenode

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4107)

### DIFF
--- a/cmake/Installation.cmake
+++ b/cmake/Installation.cmake
@@ -22,6 +22,12 @@ if(APPLE)
   execute_process(COMMAND ${MACDEPLOYQT_PATH}/macdeployqt ${BUNDLE_PATH} -no-strip)
   message(STATUS \"Updating library paths\")
   execute_process(COMMAND ${CMAKE_SOURCE_DIR}/osx/macfixrpath ${BUNDLE_PATH})
+  " COMPONENT Runtime
+  )
+  
+  install(FILES "img/icons/*.icns" DESTINATION "${BUNDLE_PATH}/Contents/Resources/")
+  
+  install(CODE "
   message(STATUS \"Creating dmg image\")
   execute_process(COMMAND hdiutil create -volname ${PROJECT_NAME} -srcfolder ${BUNDLE_PATH} -ov -format UDZO ${PROJECT_NAME}.dmg)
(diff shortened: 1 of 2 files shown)


Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---

*[osx/qTox-Mac-Deployer-ULTIMATE.sh, line 253 at r1](https://reviewable.io:443/reviews/qtox/qtox/4107#-Kbf66mvqvmk1lfYaV0N:-Kbf66mvqvmk1lfYaV0O:bf264uu) ([raw file](https://github.com/qtox/qtox/blob/bca00a4834e97842f8ef3fd84ced13e424f8f840/osx/qTox-Mac-Deployer-ULTIMATE.sh#L253)):*
> ```Shell
>     $MACDEPLOYQT qTox.app
>     $MACFIXRPATH qTox.app
>     cp $MACICONS qTox.app/Contents/Resources/
> ```

I'm not sure, that it's right way to do it.
Is it possible to do it by `install`? Like [here](https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake#L42)?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4107#-:-Kbf6N1uUnB27gsaLGNV:b-jbsf8z)*

sudden6:
updated, but to be honest I have no idea how mac-deployer is supposed to work. It seems it does some things that also cmake does, like the three lines you mentioned.

vit9696:
@Diadlo this is nonsense. cmake install is not used by your scripts, all the actions are performed separately. Therefore you have just suggested to break macOS support once more.

If you look carefully at [Installation.cmake](https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake) you will see that it also calls macfixrpath, so it is a good reason to add icon copying there as well. However, it definitely should be in two places.

sudden6:
I think the script should make use of the stuff cmake already provides. @vit9696 can we ping you when we are done trying to fix this?

sudden6:
Closing because #4108 does it the right way

PR #4108, opened 2017-01-29
Deployment fixes for macOS

vit9696:
Hopefully brings all the necessary fixes (see #4107):
* Fixed version setting (enabled plist autoformatting so the diff is a bit large)
* Changed the deployer script to rely on cmake code
* Fixed cmake installer to generate a proper bundle (with icons and valid paths)
* Fixed cmake code to call a separate script to build a correct dmg (also corrected its layout)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4108)


zetok:
Reviewed 1 of 6 files at r1.
Review status: 1 of 6 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[osx/createdmg, line 2 at r1](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kbg4zaXupIRJuixN4_G:-Kbg4zaXupIRJuixN4_H:b53d2th) ([raw file](https://github.com/qtox/qtox/blob/bd09ce4b5f77cd4049532f82d9bf8931b719ecad/osx/createdmg#L2)):*
> ```
> #!/bin/bash
> ```

missing copyright header

---

*[osx/createdmg, line 2 at r1](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kbg52-4QFZ6lylbbiLr:-Kbg52-4QFZ6lylbbiLs:bwk4u3i) ([raw file](https://github.com/qtox/qtox/blob/bd09ce4b5f77cd4049532f82d9bf8931b719ecad/osx/createdmg#L2)):*
> ```
> #!/bin/bash
> ```

add `set -eu -o pipefail` and make the code work with it

---

*[osx/DS_Store-DMG, line 0 at r1](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kbg58scutjt0-4qUzbD:-Kbg58sdPlzONdM4Y0wT:b-qprq75) ([raw file](https://github.com/qtox/qtox/blob/bd09ce4b5f77cd4049532f82d9bf8931b719ecad/osx/DS_Store-DMG#L0)):*
Why is a binary blob needed?
Can it be automatically generated instead?

---

*[osx/update-plist-version.sh, line 40 at r1](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kbg5fgR7E5hQULnItST:-Kbg5fgSqkyRQ3h1yBPo:b-ai8omn) ([raw file](https://github.com/qtox/qtox/blob/bd09ce4b5f77cd4049532f82d9bf8931b719ecad/osx/update-plist-version.sh#L40)):*
> ```Shell
> # append version to .plist file(s) after the right line
> update_version() {
>     ver=$(get_version)
> ```

`s/ver/local version/`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4108)*

vit9696:
>Why is a binary blob needed?
>Can it be automatically generated instead?

https://en.wikipedia.org/wiki/.DS_Store
TL;DR You cannot really generate it automatically without going with non-standard proprietary tools.
It is needed to set the visual look of the dmg file contents when it opens (background, icon size and positioning).

vit9696:
Fixed. The issue in travis does not seem to be caused by my changes, `git describe` fails from `$(get_version)`, and this is certainly not my code. Perhaps fork specifics matters.

zetok:
>https://en.wikipedia.org/wiki/.DS_Store
>TL;DR You cannot really generate it automatically without going with non-standard proprietary tools.
It is needed to set the visual look of the dmg file contents when it opens (background, icon size and positioning).

According to that wikipedia entry:

>Its internal structure is proprietary.

I'm not an expert, but doesn't that mean that it can't be included in the project? Doesn't GPL disallow bundling that?

---

Review status: 1 of 6 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[osx/qTox-Mac-Deployer-ULTIMATE.sh, line 229 at r2](https://reviewable.io:443/reviews/qtox/qtox/4108#-KbgrLSoJpRmMj8A6oEc:-KbgrLSoJpRmMj8A6oEd:b-neahkt) ([raw file](https://github.com/qtox/qtox/blob/0a53331b855dffb05a34ad43139ed7b1b546803c/osx/qTox-Mac-Deployer-ULTIMATE.sh#L229)):*
>Fixed. The issue in travis does not seem to be caused by my changes, git describe fails from $(get_version), and this is certainly not my code. Perhaps fork specifics matters.

When `cmake` runs the script, current working directory is not qTox git repo, which causes `git describe` to fail. Code above in `build()` would need to be modified to fix that. Possibly other code also needs to be adjusted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4108#-:-Kbgst6g9UonRmM6uLHo:b61edya)*

vit9696:
> When cmake runs the script, current working directory is not qTox git repo, which causes git describe to fail. Code above in build() would need to be modified to fix that. Possibly other code also needs to be adjusted.

If you look carefully you will see that no changes related to the paths were made. I have just added the code to ensure we are in qTox dir, and it still failed. The issue is pretty certainly in the fact that travis builds from the fork in my opinion.

> I'm not an expert, but doesn't that mean that it can't be included in the project? Doesn't GPL disallow bundling that?

Me neither, although I suspect you could have misunderstood GPL. Regardless this file already exists in your repo, it was added by whoever but not me, so please do not make it my problem to deal with it. If qTox has issues with DS_Store presence, the right way is to create an issue and wait for someone to contribute a fix.

sudden6:
@vit9696 I don't quite understand what  the purpose of the `.DS_Store` file is, why does qTox need it? Isn't the icon of the application determined by the `.icns` file?

vit9696:
@sudden6 that would be hard for a non-macOS user. In brief, it is a standard practice on how dmgs are packed and how they should look.

I have to repeat that I have no interest in discussing .DS_Store stuff. If you would like to have any discussion on this topic, direct it to the person, who added this file to the codebase. I did not change anything design-wise in the original .DS_Store, just fixed the background size and altered the position offsets.

sudden6:
After some reading I've come to the conclusion that there's no way to create this file without using a mac and we're best of to just include this file into the dmg.

For the rest of the PR, @zetok can you review the scripts? I'm not that familiar with bash.

zetok:
Reviewed 1 of 6 files at r1, 2 of 2 files at r4.
Review status: 3 of 6 files reviewed at latest revision, 3 unresolved discussions.

---

*[osx/createdmg, line 24 at r4](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kc7ZSuEOF2ALzxPZdMg:-Kc7ZSuEOF2ALzxPZdMh:bfrsddt) ([raw file](https://github.com/qtox/qtox/blob/07639011daa5ff84b61e76b0ad0dc8e659c5a68b/osx/createdmg#L24)):*
> ```
> pushd "$2" &> /dev/null
> BUNDLE_PATH="$(pwd)"
> popd &>/dev/null
> ```

Why use `pushd/popd` and send it's output to `/dev/null` ?

Same effect can be achieved by using just `cd`, in a lot more clear way, i.e.:

```
cd "$2"
BUNDLE_PATH=$(pwd)
cd -
```

Same goes for other occurrences.

---

*[osx/qTox-Mac-Deployer-ULTIMATE.sh, line 224 at r4](https://reviewable.io:443/reviews/qtox/qtox/4108#-Kc7YP5BZekEL-mn87OI:-Kc7YP5BZekEL-mn87OJ:b-f1axa5) ([raw file](https://github.com/qtox/qtox/blob/07639011daa5ff84b61e76b0ad0dc8e659c5a68b/osx/qTox-Mac-Deployer-ULTIMATE.sh#L224)):*
> ```Shell
>         fcho "Now working in ${PWD}"
>         fcho "Setting the version ..."
>         ./tools/update-versions.sh
> ```

This can fail regardless of whether it runs on travis or not – if someone clones the repo with `--depth=1` or downloads a tarball, it would fail, so it's better to just remove that part from the script.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4108)*

vit9696:
I have an opposing view on cd usage in scripts, in my opinion it should never be used at all. As for why pushd, it is much safer in case of further changes and the need to change directory more than once. Your project your rules though.

As for version updating, just again it was there, and it is not my addition. Removed it if you think it is better.

zetok:
Reviewed 1 of 2 files at r2, 2 of 2 files at r5.
Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[cmake/Installation.cmake, line 9 at r5](https://reviewable.io:443/reviews/qtox/qtox/4108#-KcCjXJPuMsmROs8So7J:-KcCjXJPuMsmROs8So7K:bx2gisp) ([raw file](https://github.com/qtox/qtox/blob/bf8b23e373275f0d34d99ff79d0637f2fb558d18/cmake/Installation.cmake#L9)):*
> ```CMake
> if(APPLE)
>   set_target_properties(${PROJECT_NAME} PROPERTIES
>     MACOSX_BUNDLE_INFO_PLIST "${CMAKE_SOURCE_DIR}/osx/info.plist")
> ```

Is it still needed, given that lines above were removed?

---

*[osx/update-plist-version.sh, line 45 at r5](https://reviewable.io:443/reviews/qtox/qtox/4108#-KcCj00dBofiXwCvTRin:-KcCj00dBofiXwCvTRio:bt7gd7m) ([raw file](https://github.com/qtox/qtox/blob/bf8b23e373275f0d34d99ff79d0637f2fb558d18/osx/update-plist-version.sh#L45)):*
> ```Shell
>     plutil -convert xml1 info.plist
> }
> update_version
> ```

I think that this script can be removed, and part of cmake that updated the version should be restored. Or was what cmake was doing not working?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4108)*

vit9696:
> Is it still needed, given that lines above were removed?

if I remember correctly, it copies the plist to the bundle so is important.

> I think that this script can be removed, and part of cmake that updated the version should be restored. Or was what cmake was doing not working?

the last sentence is correct

zetok:
:lgtm_strong: 

Could you use reviewable next time?

---

Reviewed 1 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4108#-:-KcClZ3iPj3oaBEp-WLm:blajauf)*

vit9696:
> Could you use reviewable next time?

Of course, just this was quicker~

PR #4122, opened 2017-02-02
feat(UI): Move filetransferwidget buttons side-by-side.

### LINKED ISSUE #2597: filetransferwidget: Move buttons side-by-side.
Here's something I made for my builds:

qtox_filetransferwidget_buttons_changed_01

The background to this is that, due to the widget moving upwards after me sending more messages after a file, a friend often clicked 'cancel' when they wanted to 'accept'.

As it is, it might break the layout a bit, looking a bit odd, and it definitely doesn't need to be like in the picture above.  The point is to have the buttons side-by-side.  If nothing else, this does give a bit more room for the file-name and friends above.  ^^

That said, I was going to make the top-left corner of the left button rounded (pause in the image), so it would sort of play along with the 'Send a screenshot/Send file(s)' combo, but didn't get to it yet!

Thoughts?

Chiitoo:
This aims to prevent accidental clicks on the incorrect button when accepting a file transfer, and the chatlog moves due to new messages being received at the time of the click.

Closes #2597.

Took only 14 months, but I finally got to it!  (Where does the time go...)

Originally I had only figured out how to move the buttons around, but I wanted to also make the left button rounded from the upper left corner, similar to what the 'Send file(s)' and 'Send a screenshot' buttons have going.  The way I managed to get it done feels a bit odd, as if drawing three different rectangles, but not quite?  Looking around it seems it's not that unusual.

The right button still uses the old code mostly, since it gets the lower right corner rounded for free due to the background.

qtox_filetransferwidget_buttons_changed_02

(I wonder if this could be done by using css stylesheets instead, but I imagine there is a reason for it not having been done like so in the first place.)

I'm thinking with the extra space on top of the buttons, the file name and/or other information could be extended there, or something completely different, but that's for another pull request.

I will not be surprised at all if the code looks absolutely bonkers to someone who knows more about Qt and stuff than I do.  At least I finally got the idea out there, which was my main goal.  :]

Tested on Gentoo Linux and Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4122)

### DIFF
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -40,6 +40,11 @@
 
 #include <math.h>
 
+// The leftButton is used to accept, pause, or resume a file transfer, as well as to open a
+// received file.
+// The rightButton is used to cancel a file transfer, or to open the directory a file was
+// downloaded to.
+
 FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
     : QWidget(parent)
     , ui(new Ui::FileTransferWidget)
@@ -47,6 +52,7 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
     , lastTick(QTime::currentTime())
     , backgroundColor(Style::getColor(Style::LightGrey))
     , buttonColor(Style::getColor(Style::Yellow))
+    , buttonBackgroundColor(Style::getColor(Style::White))
     , active(true)
 {
     ui->setupUi(this);
@@ -93,8 +99,10 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
             &FileTransferWidget::fileTransferRemotePausedUnpaused);
     connect(Core::getInstance(), &Core::fileTransferBrokenUnbroken, this,
             &FileTransferWidget::fileTransferBrokenUnbroken);
-    connect(ui->topButton, &QPushButton::clicked, this, &FileTransferWidget::onTopButtonClicked);
-    connect(ui->bottomButton, &QPushButton::clicked, this, &FileTransferWidget::onBottomButtonClicked);
+    connect(ui->leftButton, &QPushButton::clicked, this,
+            &FileTransferWidget::onLeftButtonClicked);
+    connect(ui->rightButton, &QPushButton::clicked, this,
+            &FileTransferWidget::onRightButtonClicked);
     connect(ui->previewButton, &QPushButton::clicked, this,
             &FileTransferWidget::onPreviewButtonClicked);
 
@@ -190,8 +198,8 @@ void FileTransferWidget::setButtonColor(const QColor& c)
 
 bool FileTransferWidget::drawButtonAreaNeeded() const
 {
-    return (ui->bottomButton->isVisible() || ui->topButton->isVisible())
-           && !(ui->topButton->isVisible() && ui->topButton->objectName() == "ok");
+    return (ui->rightButton->isVisible() || ui->leftButton->isVisible())
+           && !(ui->leftButton->isVisible() && ui->leftButton->objectName() == "ok");
 }
 
 void FileTransferWidget::paintEvent(QPaintEvent*)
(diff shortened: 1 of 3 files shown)


sudden6:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/chatlog/content/filetransferwidget.h, line 72 at r1](https://reviewable.io:443/reviews/qtox/qtox/4122#-Kc0YwfFCBjx034BENDN:-Kc0YwfGq7bYiid8-0dp:b-r8y4wm) ([raw file](https://github.com/qtox/qtox/blob/7d578f24667b22661886eefedd38f7b1ad10eb96/src/chatlog/content/filetransferwidget.h#L72)):*
> ```objc
> 
> private slots:
>     void onLeftButtonClicked();
> ```

Can you maybe rename them according to their function and not according to their position?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4122)*

Chiitoo:
I was thinking about that, too, but couldn't make up my mind and left them closer to what they were.

Something like 'acceptButton' and 'cancelButton'?  Or longer, 'acceptTransferButton' and 'cancelTransferButton'?

With less buttons, 'acceptTransfer' and 'cancelTransfer'?

Something completely different?

I might like keeping the 'button' word there in some form, but that's me.

sudden6:
`acceptButton` and `cancelButton` is fine for me

---

Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4122#-:-Kc3io0a_c24SO5Q42Xt:bxt3iy5)*

zetok:
on 7d578f2

buttons seem to be too small: 
spectacle fn1992

Chiitoo:
You would prefer the buttons to reach the top from the bottom?  Or perhaps be the original size, but either centered, or on the bottom or on top?

It looked a bit... weird I think when I first tried them reaching the top from the bottom, and there is the idea of opening up some space for other things when they're smaller.

That said, I probably don't mind which way they will go.

zetok:
> Or perhaps be the original size, but either centered, or on the bottom or on top?

Original size sounds good, perhaps located on top?

ProMcTagonist:
Like this?

new ft buttons mockup

zetok:
@ProMcTagonist looks good.

Chiitoo:
That might be exactly the way I have it done already, but I want to look into some other changes while I'm at it (apologies for the delay, again).  One of them is letting the transfer progress go under the buttons, which might seem like a simple thing to do, but isn't quite so (or I'm not seeing the simple way yet, heh).

sudden6:
@Chiitoo what do you think about merging this now to fix the urgent issue and doing the style improvements in another PR?

Chiitoo:
Yeah, I think it's about ready for now.

I put some time into figuring out ways to make the white lines around the buttons happen, as is in the original style (previously I just made the background go under the buttons).

The only thing missing visually, is making the progress bar go below the buttons, which seems to require a bit of re-making of the whole layout internally.

That can wait, but there is still also the naming-of-things thing I've been meaning to ask about.  I haven't been able to come up with a phrase that would catch all the functions the buttons have.  The only thing really in my mind, is leave them be 'leftButton' and 'rightButton' etc., and add a comment somewhere that makes it clear what they do...

Any ideas on that?

sudden6:
If there's really no way we can name them according to their function, then name them 'leftButton' and 'rightButton'.

Chiitoo:
Had a bit of merge conflict-fun with the recent changes, so it took even longer to get this ready.  Apologies for the delay yet again!.

The comment about the buttons and their functions is now after the includes.  Should it be above even those, or is this good?

sudden6:
@Diadlo checks seem to fail on Travis for OSX with a mysterious error message, any idea?

Diadlo:
@sudden6 https://github.com/libcheck/check/issues/110 ?
Seems osx has old version of libcheck and [use clang](https://travis-ci.org/qTox/qTox/jobs/206224276#L65)
May be issue somehow related with adding clang-format?

sudden6:
Maybe `clang-format` reformatted in a strange way? Or there have been some changes in Travis or the tools we use, because it worked a few days ago.

zetok:
@Chiitoo workaround has been merged, rebase the PR to make it pass travis.

sudden6:
Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4122)*

sudden6:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4122#-:-KeCqq7ctWybcISQt1y_:b-jbsf8z)*

PR #4126, opened 2017-02-03
Add link to Chocolatey package

pascalberger:
Add link to Chocolatey package as Windows installation option.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4126)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ Windows | Linux | OS X | FreeBSD
 **[64 bit installer]**, [signature][sig-64] | **[CentOS, Debian, Fedora, openSUSE, Ubuntu]** | **[Building instructions]** | **[Port]**
 [32 bit installer], [signature][sig-32] | **[Arch]**, **[Gentoo]** | |
 [64 bit][64portable], [32 bit][32portable] portable | [Other] | |
+[Chocolatey Package][chocolateypackage] | | |
 
 _**Bold** options are recommended._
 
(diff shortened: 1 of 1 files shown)


sudden6:
What is chocolatey and why should it get added to the official build sources?
How can we ensure nobody adds malware to the package while building it?

pascalberger:
Chocolatey is a package manager for Windows, like apt-get for Windows. See https://chocolatey.org.

[Here](https://github.com/chocolatey/chocolatey-coreteampackages/blob/master/automatic/qtox/update.ps1) is the script which builds the package. It actually scrapes your site, looks for a new version and packs it into the Chocolatey package (which actually is not that easy since I couldn't find a single page where we could determine the version and download the file). So actually you don't have any guarantee that there is no malware in the package, except trusting this process. If you want you can build the package in your build pipe though and publish it to the Chocolatey community repository. If you're interested you can take the existing metadata file and I can support in integrating it in your pipeline.

sudden6:
Ok, that sounds like something that should be added to the community builds section in INSTALL.md

Additionally it would be nice if you could add a link to our website `https://qtox.github.io/` to your package description.

Edit: Also please adjust the commit message to https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#commit-message-format since it's used to generate the changelog.

yurivict:
Does chocolatey sign package downloads? Because if it doesn't, attackers can substitute the file in the MITM attack.

pascalberger:
@yurivict Chocolatey currently uses checksums. Package signing is on the [roadmap](https://chocolatey.org/docs/road-map).

sudden6:
@pascalberger I think without package signing we can't really put it into the main README, but feel free to add it to the community builds in INSTALL.md

Diadlo:

PR #4166, opened 2017-02-11
chore(build): adjust version updating script to not rely on git

zetok:
Relying on git doesn't work in awful lot of cases, e.g. when there's a
shallow `git clone`.

Instead have "hardcoded" version in files, and update it before release.

Closes #4165.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4166)

PR #4167, opened 2017-02-11
chore(build): make cmake update version

zetok:
Depends on #4166 to work.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4167)

### DIFF
--- a/cmake/Installation.cmake
+++ b/cmake/Installation.cmake
@@ -4,6 +4,11 @@
 #
 ################################################################################
 
+# set correct version to update files
+set(QTOX_VERSION "1.8.1")
+execute_process(COMMAND ${CMAKE_SOURCE_DIR}/tools/update-versions.sh
+    ${QTOX_VERSION})
+
 if(APPLE)
   set_target_properties(${PROJECT_NAME} PROPERTIES
     MACOSX_BUNDLE_INFO_PLIST "${CMAKE_SOURCE_DIR}/osx/info.plist")


sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4167#-:-Kci9NPNdi6YbkppPIJe:b-jbsf8z)*

yurivict:
tools/update-versions.sh appears to be for only OSX and Windows. Why do you call it unconditionally in cmake then?

zetok:
>tools/update-versions.sh appears to be for only OSX and Windows.

It can be used on any system that has GNU sed.

>Why do you call it unconditionally in cmake then?

To update the version.

yurivict:
Could you add cmake variable like NO_VERSION_UPDATE? Because on some systems (BSDs) network connections aren't allowed during build.

zetok:
>Could you add cmake variable like NO_VERSION_UPDATE? Because on some systems (BSDs) network connections aren't allowed during build.

Why would network connection be needed in the first place?

yurivict:
Ok, maybe network connection isn't needed, but git shouldn't be required during build. On FreeBSD for example it isn't installed from the git repository.

```
# tools/update-versions.sh
fatal: Not a git repository (or any parent up to mount point /usr/ports)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
./update-plist-version.sh: line 41: defaults: command not found
```

You shouldn't assume there is a git repository there.

yurivict:
So the problems are 1. this patch adds the unconditional dependency on git executable 2. it assumes the build is done in a git repository directory 3. it functionally duplicates the already existing GIT_DESCRIBE variable that can currently be supplied by the caller.

zetok:
>You shouldn't assume there is a git repository there.

I don't. You're running the old script from master before #4166 was merged. This PR literally depends on #4166 to work.

yurivict:
Now it just fails:
```
# ./tools/update-versions.sh
Not a version: 
```

zetok:
Yep.

PR #4180, opened 2017-02-16
Undeprecate video

tux3:
This needs testing!
Works On My Machine, but video can be fickle.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4180)


zetok:
Since travis build fails, probably minimal ffmpeg version needs to be bumped (both in docs and build scripts).

tux3:
It seems the new API was introduced in FFMpeg 3.1, in March 2016.
So it might be a bit early to switch to it, but the code is there.

sudden6:
Can we make this a build time decision and mark the old code as deprecated?

tux3:
We could check LIBAVCODEC_VERSION_INT >= 3747941 at build time, I suppose

Diadlo:
Thanks for PR. Close since #4266 was merged

PR #3839, opened 2016-10-28
Implement clang-format and updated coding style document

initramfs:
This PR introduces a basic clang-format configuration along with a updated coding guidelines document.

The document loosely mirrors the structure of the clang-format config such that it's easy to see the corresponding rules in clang-format. The document also replaces the coding styles section found in `CONTRIBUTING.md` (instead, a link to the new document can be found in contributing right now).

The coding style itself is an extension of the existing style in qTox, with somewhat stricter requirements for conformity. I've attempted to duplicate as much of the formatting rules toktok/c-toxcore uses except in the cases where it seems nonsensical, unnecessarily or contradictory to the established coding convention by qTox.

This however by no means implies that this is the final scheme. Discussion regarding certain stylistic features should occur in the comment thread of this PR such that all interested parties come to some sort of consensus before this gets merged. **After** this gets merged, I'll work on slowly adopting the existing codebase to the new scheme, such that by the time I'm done, the qTox codebase will be consistent.

Closes #3601 when merged.

---

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/3839)

PR #4183, opened 2017-02-17
Fixes a variety of Tox URI & IPC handling bugs, deadlocks & infinite loops.

### LINKED ISSUE #1925: qTox doesn't support `tox:ID` URI
Clicking on an example ↓

```
<a href="tox:581A49A38D3EEA6DC4A3725C3DF9C93692B7B54A3644404132712CFE6A16CF5703FE7DAEA0BC">CLICKBAIT</a>
```

should result in qTox client being opened in FR "window".

https://github.com/Tox/Tox-STS/blob/master/STS.md#tox-uri-scheme

tux3:
Fixes #1925 in the first commit.

The second commit fixes some more bugs, there was a baroque situation with processEvents() loops running before the main event loop, causing an infinite loop due to QApplication::quit() being 1) ignored before the first call to exec() and 2) just plain useless in a processEvents() loop (this is the #1926 100% CPU bug), this was often triggered by IPC event senders timing out because the handler was blocking, just pure Fun™ all around!
I also fixed a potential deadlock between the Core and Main thread in the case of an early exit, that I hit while fixing the other problems.

@zetok Can you confirm that this fixes it for you?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4183)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -295,7 +295,8 @@ int main(int argc, char *argv[])
         }
     }
 
-    Nexus::getInstance().start();
+    Nexus& nexus = Nexus::getInstance();
+    nexus.start();
 
     // Event was not handled by already running instance therefore we handle it ourselves
     if (eventType == "uri")
@@ -303,8 +304,10 @@ int main(int argc, char *argv[])
     else if (eventType == "save")
         handleToxSave(firstParam.toUtf8());
 
-    // Run
-    int errorcode = a.exec();
+    // Run (unless we already quit before starting!)
+    int errorcode = 0;
+    if (nexus.isRunning())
+        errorcode = a.exec();
 
     Nexus::destroyInstance();
     CameraSource::destroyInstance();
--- a/src/net/toxuri.cpp
+++ b/src/net/toxuri.cpp
@@ -51,16 +51,25 @@ bool toxURIEventHandler(const QByteArray& eventData)
  */
 bool handleToxURI(const QString &toxURI)
 {
-    Core* core = Core::getInstance();
+    Nexus& nexus = Nexus::getInstance();
+    Core* core = nexus.getCore();
 
     while (!core)
     {
-        core = Core::getInstance();
+        if (!nexus.isRunning())
+            return false;
+
+        core = nexus.getCore();
         qApp->processEvents();
     }
 
     while (!core->isReady())
+    {
+        if (!nexus.isRunning())
+            return false;
+
         qApp->processEvents();
+    }
 
     QString toxaddr = toxURI.mid(4);
 
@@ -70,18 +79,26 @@ bool handleToxURI(const QString &toxURI)
         toxId = Toxme::lookup(toxaddr);
         if (!toxId.isValid())
         {
-            QMessageBox::warning(0, "qTox",
-                                         ToxURIDialog::tr("%1 is not a valid Toxme address.")
-                                         .arg(toxaddr));
+            QMessageBox *messageBox = new QMessageBox(QMessageBox::Warning, "qTox",
+                                   QMessageBox::tr("%1 is not a valid Toxme address.")
+                                   .arg(toxaddr), QMessageBox::Ok, nullptr);
+            messageBox->setButtonText(QMessageBox::Ok, QMessageBox::tr("Ok"));
+            QObject::connect(messageBox, &QMessageBox::finished, messageBox, &QMessageBox::deleteLater);
+            messageBox->show();
             return false;
         }
     }
 
-    ToxURIDialog dialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
+    ToxURIDialog *dialog = new ToxURIDialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
                                                 "Default message in Tox URI friend requests. Write something appropriate!")
                         .arg(Nexus::getCore()->getUsername()));
-    if (dialog.exec() == QDialog::Accepted)
-        Core::getInstance()->requestFriendship(toxId, dialog.getRequestMessage());
+    QObject::connect(dialog, &ToxURIDialog::finished, [=](int result) {
+        if (result == QDialog::Accepted)
+                Core::getInstance()->requestFriendship(toxId, dialog->getRequestMessage());
+
+        dialog->deleteLater();
+    });
+    dialog->open();
 
     return true;
 }
--- a/src/net/toxuri.cpp
+++ b/src/net/toxuri.cpp
@@ -33,6 +33,7 @@
 #include <QPlainTextEdit>
 #include <QPushButton>
 #include <QCoreApplication>
+#include <QThread>
 
 bool toxURIEventHandler(const QByteArray& eventData)
 {
@@ -61,6 +62,7 @@ bool handleToxURI(const QString &toxURI)
 
         core = nexus.getCore();
         qApp->processEvents();
+        QThread::msleep(10);
     }
 
     while (!core->isReady())
@@ -69,6 +71,7 @@ bool handleToxURI(const QString &toxURI)
             return false;
 
         qApp->processEvents();
+        QThread::msleep(10);
     }
 
     QString toxaddr = toxURI.mid(4);
--- a/src/net/toxuri.cpp
+++ b/src/net/toxuri.cpp
@@ -64,13 +64,17 @@ bool handleToxURI(const QString &toxURI)
 
     QString toxaddr = toxURI.mid(4);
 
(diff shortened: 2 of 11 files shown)


zetok:
on ce70b6815878f7064c01a74b553f6214904bf67c #1925 seems to be fixed, I've tested only with an ID.

#1926 seems to be still there, but at least the deadlock is fixed.

I've started patched qTox instance beside an older, not patched instance, dunno if that affects #1926.

It's interesting though that qTox would try to add my own ID: 
spectacle rx1098
which results in:
spectacle ed1098

zetok:
NB: 100% CPU usage seems to happen only when there is no patched qTox instance running prior to clicking on a `tox:` URI.

sudden6:
Reviewed 4 of 11 files at r1.
Review status: 4 of 11 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4183)*

tux3:
@zetok 
>It's interesting though that qTox would try to add my own ID:

Fixed.

>#1926 seems to be still there, but at least the deadlock is fixed.

Can you describe exactly what happens and how to reproduce?

zetok:
on a06f084a7424c4403533044e5b2b3f7fae13906f

#1926 – 100% CPU usage.

Steps to reproduce:
1. Make sure that there are no qTox instances running
2. Click on the `tox:` URI in the browser, e.g. from #1925; qTox profile selector is started

Before profile is chosen, qTox profile selector uses 100% CPU.

tux3:
Oh, right! I'll just throw a msleep in there real quick. Thanks.

zetok:
Review status: 4 of 11 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/net/toxuri.cpp, line 61 at r3](https://reviewable.io:443/reviews/qtox/qtox/4183#-KdBlpNjxUygRPjH78Ee:-KdBlpNjxUygRPjH78Ef:bo2n7p1) ([raw file](https://github.com/qtox/qtox/blob/4611e287bec684324864863750e1da1f90cb762c/src/net/toxuri.cpp#L61)):*
> ```cpp
>     {
>         if (!nexus.isRunning())
>             return false;
> ```

When there is code after the `if` statement that doesn't have curly braces, there should be added a newline before the code outside of `if` statement, since without it code can be confusing. Same goes for other occurrences.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4183)*

tux3:
>When there is code after the if statement that doesn't have curly braces, there should be added a newline before the code outside of if statement, since without it code can be confusing.

I suppose. Done.

sudden6:
Reviewed 5 of 11 files at r1, 2 of 2 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4183)*

sudden6:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4183#-:-KdINnts5xi34Qaq_aAx:b-jbsf8z)*

src/widget/gui.cpp
```diff
@@ -322,25 +322,34 @@ QString GUI::passwordDialog(const QString& cancel, const QString& body)
 
 void GUI::_clearContacts()
 {
-    Nexus::getDesktopGUI()->clearContactsList();
+    Widget* w = Nexus::getDesktopGUI();
+    if (w)
+        w->clearContactsList();
```
Diadlo, line 327: Brackets. And other places

src/persistence/profile.cpp
```diff
@@ -211,7 +211,10 @@ Profile::~Profile()
         saveToxSave();
     }
 
-    delete core;
+    core->deleteLater();
+    while (coreThread->isRunning())
+        qApp->processEvents();
```
Diadlo, line 216: Brackets

src/nexus.cpp
```diff
@@ -310,6 +343,12 @@ void Nexus::showLoginLater()
     QMetaObject::invokeMethod(&getInstance(), "showLogin", Qt::QueuedConnection);
 }
 
+void Nexus::onLastWindowClosed()
+{
+    if (quitOnLastWindowClosed)
+        quit();
```
Diadlo, line 349: Brackets

src/net/toxuri.cpp
```diff
@@ -51,33 +52,70 @@ bool toxURIEventHandler(const QByteArray& eventData)
...
 
-    ToxURIDialog dialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
+    ToxURIDialog *dialog = new ToxURIDialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
                                                 "Default message in Tox URI friend requests. Write something appropriate!")
                         .arg(Nexus::getCore()->getUsername()));
-    if (dialog.exec() == QDialog::Accepted)
-        Core::getInstance()->requestFriendship(toxId, dialog.getRequestMessage());
+    QObject::connect(dialog, &ToxURIDialog::finished, [=](int result) {
+        if (result == QDialog::Accepted)
+                Core::getInstance()->requestFriendship(toxId, dialog->getRequestMessage());
```
Diadlo, line 114: Brackets

src/net/toxuri.cpp
```diff
@@ -51,33 +52,70 @@ bool toxURIEventHandler(const QByteArray& eventData)
...
+                                                     "When trying to add your own Tox ID as friend"),
+                                                  QMessageBox::Ok, nullptr);
+        messageBox->setButtonText(QMessageBox::Ok, QMessageBox::tr("Ok"));
+        QObject::connect(messageBox, &QMessageBox::finished, messageBox, &QMessageBox::deleteLater);
+        messageBox->show();
         return false;
     }
 
-    ToxURIDialog dialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
+    ToxURIDialog *dialog = new ToxURIDialog(0, toxaddr, QObject::tr("%1 here! Tox me maybe?",
```
Diadlo, line 109: * should follow the type

src/net/toxuri.cpp
```diff
@@ -51,33 +52,70 @@ bool toxURIEventHandler(const QByteArray& eventData)
...
+            return false;
+
+        core = nexus.getCore();
         qApp->processEvents();
+        QThread::msleep(10);
     }
 
     while (!core->isReady())
+    {
+        if (!nexus.isRunning())
```
Diadlo, line 70: Same as above

src/net/toxuri.cpp
```diff
@@ -51,33 +52,70 @@ bool toxURIEventHandler(const QByteArray& eventData)
...
 bool handleToxURI(const QString &toxURI)
 {
-    Core* core = Core::getInstance();
+    Nexus& nexus = Nexus::getInstance();
+    Core* core = nexus.getCore();
 
     while (!core)
     {
-        core = Core::getInstance();
+        if (!nexus.isRunning())
```
Diadlo, line 60: Same as above

src/main.cpp
```diff
@@ -295,16 +295,19 @@ int main(int argc, char *argv[])
...
     if (eventType == "uri")
         handleToxURI(firstParam.toUtf8());
     else if (eventType == "save")
         handleToxSave(firstParam.toUtf8());
 
-    // Run
-    int errorcode = a.exec();
+    // Run (unless we already quit before starting!)
+    int errorcode = 0;
+    if (nexus.isRunning())
```
Diadlo, line 309: You forgot about the brackets

src/widget/gui.cpp
tux3: Is that really a requirement these days? I took a quick look at the code and the extra braces seems to be used very inconsistently.
Methinks it would be more acceptable if we were using K&R braces, but with Allman it seems a bit excessive in vertical space used.

Diadlo: 1) It's coding style. What the reason to have coding style if we will not stick it?
2) I'm absolutely [agree](https://github.com/qTox/qTox/pull/3839#issuecomment-277912589) with you about K&R braces. Can you vote on #3839 ?

tux3: 1. Absolutely, and that works both ways :) 
We clearly don't stick to this one coding style rule (by looking at the code), and it doesn't seem very useful. 
Why have it?
2. I'm neutral on that issue in general

Diadlo: We don't update code just to update style. We write new code with new code style. And when we change some code, we update code style too.
> Why have it?

Because code written in one style easy to read. And sometime can help to avoid mistakes (like rule: brackets everywhere)

PR #4202, opened 2017-02-24
chore(cmake): Add windows cmake toolchain

Diadlo:
@nurupo Can you test it?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4202)


nurupo:
Thanks but no. I have never built qTox and I don't want to. Just getting the dependencies for cross-compilation is painful enough.

PR #4238, opened 2017-03-06
Wrapped Tox API in macros.

yurivict:
This is in sync with https://github.com/TokTok/c-toxcore/issues/503, though it doesn't depend in it.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4238)


noavarice:
What is the point of this PR?

yurivict:
The point is to allow to use overloaded Tox APIs. I am going to create the mid-level library removing the message length limit.

Currently code can only reside in the base Tox library, or in the client. This approach allows mid-level libraries to overload Tox API.

sudden6:
That is an interesting approach to handle an API, can you explain this in a little bit more detail? i.e. idea behind it, improvements, drawbacks?

yurivict:
@sudden6 
The idea is to allow variants of the API provided by separate libraries, so that they can potentially be stacked on top of each other.

I saw this approach somewhere, but don't remember where.

Benefits: flexibility to add layers, ease of creation of such layers, minimal code in them. ex. no need to have stubs of every function.
Drawbacks: lost CPU cycles due to the dereferencing on each call. But this is a minor drawback.

Such library can be initialized this way:
```
ToxcoreApi mylib_initialize(const ToxcoreApi *api) {
    ToxcoreApi x_toxcore_api;
    // save the base API
    base_toxcore_api = *api;
    // overload API
    x_toxcore_api = *api;
    x_toxcore_api.tox_friend_send_message = &mylib_friend_send_message;
    x_toxcore_api.tox_callback_friend_read_receipt = &mylib_callback_friend_read_receipt;
    return x_toxcore_api;
}
```

Diadlo:
1) Move new `*.h` files under `src/core/`
2) I'm still not unserstand, how this macros help to avoid creating stubs of every function?
3) I don't think, that changing core library is so oftern action that we need macros to fast library change
4) Also, you can use something like this: `#define TOX(name, args…) tox_##name(args)` to avoid `TOX(name)(params)`

yurivict:
> Move new *.h files under src/core/

Moved

> I'm still not unserstand, how this macros help to avoid creating stubs of every function?

Because the underlying functions are silently reused. Only the ones that need to be replaced are replaced.

> I don't think, that changing core library is so oftern action that we need macros to fast library change

It's not often, it's rather the optional replacement depending on the cmake variable.

> Also, you can use something like this: #define TOX(name, args…) tox_##name(args) to avoid TOX(name)(params)

```TOX(name)(params)``` shows that ```name``` isn't just an argument.

yurivict:
Please merge this for the tox defragmenter: https://github.com/TokTok/c-toxcore/issues/492#issuecomment-285888816

sudden6:
Closing since this never really took off.

Issue #4214, opened 2017-02-27
Multi-line messages aren't properly copied into the clipboard: U+2028 LINE SEPARATOR (e2 80 a8) is inserted


yurivict:
It inserts this character in place of the newline. Newlines are commonly encoded as 0x0a byte.

1.8.1 (rev.60e5375) on FreeBSD
Qt version: 5.6.2

zetok:
For some reason I couldn't even copy full multiline messages, only first selected line was copied.

Note: when copying one should not select whole message (nick, timestamp), but only the text.

When copying whole message (nick, text, timestamp), newlines are copied correctly.

yurivict:
You should paste into the editor, save it as a file, and then look at its hexdump. The character U+2028 makes editors only show the first line, while the whole text is actually there.

zetok:
I did `hexdump -C <<EOF`, pasted, pressed enter, `EOF`, enter, and hexdump showed only the first line. I guess I should get a better terminal?

yurivict:
I am pasting into vim, saving it, then running hexdump on the saved file.

ezavod:
Since qTox version 1.9.0 (Win10 64bit; toxcore: 0.1.7; Qt: 5.6.1) this is fixed for me.
@yurivict can you confirm as well, because I do not really know what has fixed that.

yurivict:
Yes, 1.9.0 seems to have this fixed.

PR #4239, opened 2017-03-07
fix: Copy multiple lines to clipboard

ezavod:
This fix is related to issue #4214 (see  @zetok comment).
The PR fixes the issue that in Linux only the first line of a multiple line message can be copied to clipboard.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4239)

### DIFF
--- a/src/chatlog/content/text.cpp
+++ b/src/chatlog/content/text.cpp
@@ -337,31 +337,38 @@ QString Text::extractSanitizedText(int from, int to) const
         return "";
 
     QString txt;
-    QTextBlock block = doc->firstBlock();
 
-    for (QTextBlock::Iterator itr = block.begin(); itr != block.end(); ++itr) {
-        int pos =
-            itr.fragment()
-                .position(); // fragment position -> position of the first character in the fragment
+    QTextBlock begin = doc->findBlock(from);
+    QTextBlock end = doc->findBlock(to);
+    for (QTextBlock block = begin; block != end.next() && block.isValid(); block = block.next()) {
+        for (QTextBlock::Iterator itr = block.begin(); itr != block.end(); ++itr) {
+            int pos =
+                itr.fragment()
+                    .position(); // fragment position -> position of the first character in the fragment
 
-        if (itr.fragment().charFormat().isImageFormat()) {
-            QTextImageFormat imgFmt = itr.fragment().charFormat().toImageFormat();
-            QString key = imgFmt.name(); // img key (eg. key::D for :D)
-            QString rune = key.mid(4);
+            if (itr.fragment().charFormat().isImageFormat()) {
+                QTextImageFormat imgFmt = itr.fragment().charFormat().toImageFormat();
+                QString key = imgFmt.name(); // img key (eg. key::D for :D)
+                QString rune = key.mid(4);
 
-            if (pos >= from && pos < to) {
-                txt += rune;
-                ++pos;
-            }
-        } else {
-            for (QChar c : itr.fragment().text()) {
-                if (pos >= from && pos < to)
-                    txt += c;
+                if (pos >= from && pos < to) {
+                    txt += rune;
+                    ++pos;
+                }
+            } else {
+                for (QChar c : itr.fragment().text()) {
+                    if (pos >= from && pos < to)
+                        txt += c;
 
-                ++pos;
+                    ++pos;
+                }
             }
         }
+
+        txt += '\n';
     }
+
+    txt.chop(1);
 
     return txt;
 }


zetok:
Fixes the bug for me.

src/chatlog/content/text.cpp
```diff
@@ -337,30 +337,36 @@ QString Text::extractSanitizedText(int from, int to) const
...
+                for (QChar c : itr.fragment().text()) {
+                    if (pos >= from && pos < to)
+                        txt += c;
+
+                    ++pos;
+                }
             }
         }
+
+        if (block == doc->findBlock(to))
```
Diadlo: Seems, it should be moved in to the second `for` part with `block.isValid()`

Diadlo (dismissed):

src/chatlog/content/text.cpp
ezavod: Could you give me some more details please, because I do not fully understand.

First I would not move `block.isValid()` because it is the needed exit condition for the outer `for`. The last block has an invalid `.next()` member. The `block.isValid()` statement could be replaced by `block != doc->findBlock(to).next()` but this is a very ugly solution (although we get rid of the breaking  `if`) because we end up with an additional linebreak at the and an might run into problems if something goes wrong.
Second the breaking `if`has to be after the inner `for` because the it should parse the last needed block (which is `doc->findBlock(to)`) first and then exit the outer `for`. If I move the breaking `if`inside the inner `for`this could generate too many linebreaks at the end.

Diadlo:
Reviewed 1 of 1 files at r1.

---

*[src/chatlog/content/text.cpp, line 342 at r1](https://reviewable.io:443/reviews/qtox/qtox/4239#-KeeiUavio1emCBcA-be:-KeeiUavio1emCBcA-bf:bap6rsj) ([raw file](https://github.com/qtox/qtox/blob/fbf3b83598c94433d9f446ca865292b5aefba4c6/src/chatlog/content/text.cpp#L342)):*
> ```cpp
> 
>     for (QTextBlock block = doc->findBlock(from); block.isValid(); block = block.next()) {
>         for (QTextBlock::Iterator itr = block.begin(); itr != block.end(); ++itr) {
> ```

What about `for (const QTextFragment& f : block)` ?

---

*[src/chatlog/content/text.cpp, line 366 at r1](https://reviewable.io:443/reviews/qtox/qtox/4239#-Ked9T2Iliw31lLI9jqE-r1-366:-KeegpGiKV8GzkZg2h3L:b-miivok) ([raw file](https://github.com/qtox/qtox/blob/fbf3b83598c94433d9f446ca865292b5aefba4c6/src/chatlog/content/text.cpp#L366)):*
<details><summary><i>Previously, ezavod wrote…</i></summary><blockquote>

Could you give me some more details please, because I do not fully understand.

First I would not move `block.isValid()` because it is the needed exit condition for the outer `for`. The last block has an invalid `.next()` member. The `block.isValid()` statement could be replaced by `block != doc->findBlock(to).next()` but this is a very ugly solution (although we get rid of the breaking  `if`) because we end up with an additional linebreak at the and an might run into problems if something goes wrong.
Second the breaking `if`has to be after the inner `for` because the it should parse the last needed block (which is `doc->findBlock(to)`) first and then exit the outer `for`. If I move the breaking `if`inside the inner `for`this could generate too many linebreaks at the end.
</blockquote></details>

You can use iterator syntax. Also, not sure, that `.isValid()` is required
```
QTextBlock begin = doc->findBlock(from);
QTextBlock end = doc->findBlock(from);
for (QTextBlock block = start; block != end; block++) 
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4239)*

Diadlo:

---

*[src/chatlog/content/text.cpp, line 366 at r1](https://reviewable.io:443/reviews/qtox/qtox/4239#-Ked9T2Iliw31lLI9jqE-r1-366:-Keejl8LM_0gXLUMJiA7:bgw4pb5) ([raw file](https://github.com/qtox/qtox/blob/fbf3b83598c94433d9f446ca865292b5aefba4c6/src/chatlog/content/text.cpp#L366)):*

You can use iterator syntax. Also, not sure, that `.isValid()` is required
```
QTextBlock begin = doc->findBlock(from);
QTextBlock end = doc->findBlock(from);
for (QTextBlock block = start; block != end; block++) 
```
</blockquote></details>

Edit:
```
QTextBlock end = doc->findBlock(from).next();
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4239)*

ezavod:
> You can use iterator syntax.

This does not work for me, because `block++` causes errors. I think it is not possible to increment a `QTextBlock`, it shoult be an iterator of some sort. Maybe it is good enought to use
`for (QTextBlock block = begin; block != end; block = block.next())`?

> What about `for (const QTextFragment& f : block)`?

First this requires `block` to be `QTextBlock::iterator` which leads us to the outer `for` again.
Second this is original code which I have not changed. So to preserve code consistency it might be better to refactor this in a seperate PR because this same code might exist elsewhere (eg. in following function `Text::extractImgTooltip`).

Diadlo:
Sorry, misread Qt docs. `block = block.next()` is fine. What about `isValid()`? With new condition `&&` will seens not so ugly.
About second `for`: block shouldn't be `QTextBlock::iterator` it should provide `begin()` and `end()`. But feel free to leave it

ezavod:
I added `isValid()` condition, which should not be necessary but prevents problems if something breaks (eg. a block is unexpected invalid).

> About second `for`: block shouldn't be `QTextBlock::iterator` it should provide `begin()` and `end()`.

I will leave this for now. Maybe open an separate issue so that this could be refactored on all occurrences.

Diadlo:
Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/chatlog/content/text.cpp, line 368 at r2](https://reviewable.io:443/reviews/qtox/qtox/4239#-KehVhSGDUoD_gcyHxRZ:-KehVhSGDUoD_gcyHxR_:b-wjhp8p) ([raw file](https://github.com/qtox/qtox/blob/313458fa42cd107362666f034e1bd0fd7ea3ac42/src/chatlog/content/text.cpp#L368)):*
> ```cpp
>         }
> 
>         if(block != end)
> ```

IMO too many useless checks. What do you think about remove this check and add [`remove`](http://doc.qt.io/qt-5/qstring.html#remove) after loop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4239)*

ezavod:
> What do you think about remove this check and add `remove` after loop?

It is a fair point that the `if` adds many useless checks but I suspect that `remove` may cause some strange behavior on some platforms: Some [(people say)](http://stackoverflow.com/questions/27620038/qstring-new-line/27620074#27620074) that `QString` automatically translates `\n` to whatever newline character(s) your system uses. In Windows this would mean `\n` -> `\r\n`. So by just removing the last char the string would end with an extra `\r` which was not there in the original message. But I do not know this for sure, since I have no Windows machine for build at the moment.
(Note: Instead of `txt.remove(txt.size() - 1, 1);` one should use `txt.chop(1);` but has same problems).

Maybe it is a good compromise to use `txt = txt.trimmed();`. This would get rid of all whitespaces (`\r` and `\n`) at the end, but this would also remove all other whitespaces at the beginning and the end. This might be a problem if you want to copy a message which has some newlines (or other whitespaces) at the beginning or at the end (but I do not see any realistic scenario where this might be the case for someone).

So I would suggest to use `trimmed`.
(Using `chop` would be better if some could test this on Windows).

Diadlo:
@ezavod Thanks to notify me about `chop` I have tried to find something like this, but can't
`trimmed` looks good to me. I think it's ok to trim extra newlines in the end. @zetok ?

zetok:
@ezavod 
>(…) (but I do not see any realistic scenario where this might be the case for someone).

Use case: using qTox to send/receive code written in [whitespace](https://en.wikipedia.org/wiki/Whitespace_%28programming_language%29) in messages.

@Diadlo 
>I think it's ok to trim extra newlines in the end. @zetok ?

I wouldn't be so sure about that. E.g. the `Hello world` example from wikipedia page linked above seems to rely on newline chars at the end to make things work.

ezavod:
@zetok Since you are referring to code one can assume that the sender would use code message styling by surrounding the code in single backticks. `trimmed` just removes whitespaces at the beginning/end so the actual code would not be touched.

Diadlo:
@ezavod From your [link](http://stackoverflow.com/questions/27620038/qstring-new-line/27620074#27620074)
> GUI controls and the like if needed

So if I understand it correctly QString will store only `\n`. It will be very strange if `.append('\n')` will add 2 bytes. Also [the proof](https://github.com/radekp/qt/blob/master/src/corelib/tools/qstring.cpp#L1517)

ezavod:
> QString will store only \n.

You are right. So we are good to go with `chop(1)` that removes the last `\n`.

Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4239#-:-KemofBcbqE3iGn3E1A9:b-jbsf8z)*

Diadlo:
Can you squash commit?
I plan to merge it

ezavod:
Done. Squashed commits into one.

PR #4287, opened 2017-03-25
refactor(profile): reduce code duplication

sudden6:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4287)

PR #4289, opened 2017-03-25
Remove plaintext passwords

sudden6:
This PR removes all instances where a passwords were stored as `QString` and replaces them with `ToxEncrypt` passkeys. This improves performance (key derivation is only run once) as well as security (can't leak plaintext passwords).

@Diadlo @noavarice I'm not entirely sure what I do with `const ToxEncrypt* Profile::getPasskey()` is the correct way to make the `ToxEncrypt` object usable for encryption and decryption. Is there maybe a better way not involving raw pointers?

Depends on #4287

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4289)

### DIFF
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -51,7 +51,6 @@ QVector<QString> Profile::profiles;
 
 Profile::Profile(QString name, const QString& password, bool isNewProfile, const QByteArray& toxsave)
     : name{name}
-    , password{password}
     , newProfile{isNewProfile}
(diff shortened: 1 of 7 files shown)


Diadlo:
:lgtm_strong:

---

Reviewed 9 of 9 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4289#-:-KhDkP3LdmQ4oqLM-kPR:b-jbsf8z)*

Issue #4295, opened 2017-03-27
Qtox includes quote marks in URL highlight


ghost:
Expected behavior: exclude double quote marks (") from the blue URL highlight.
But they are consistently included. It's a minor problem, but I assume it's also an easy fix, with a slight tweaking to the regex.

Commit hash: f839abf33fb666f586fef3bda06efd49165c0b2d
qTox version v1.9.0-18-gf839abf

zetok:
I've got the regexp in #4297, but that addresses the problem only partially, i.e. string that is being matched against regexp has escaped HTML, so while regexp doesn't match against `"`, it does match against `&quot;` … No idea how to fix that.

When not matching against letters:
![spectacle aq1995](https://cloud.githubusercontent.com/assets/3148759/24360105/4a22d302-12fe-11e7-80d5-60122f781caf.png)

Same text, but matched against letters:
![spectacle jg9088](https://cloud.githubusercontent.com/assets/3148759/24360479/517b2c16-12ff-11e7-8844-a84989f7c6fb.png)

PR #4297, opened 2017-03-27
fix: use proper regexp for detecting URIs according to RFC 3986

zetok:
Partially addresses #4295, but doesn't fix it, since regexp is still
being matched against an escaped HTML.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4297)

### DIFF
--- a/src/chatlog/textformatter.cpp
+++ b/src/chatlog/textformatter.cpp
@@ -88,10 +88,15 @@ static const QVector<QPair<QRegularExpression, QString>> textPatternStyle{
     REGEX_MARKDOWN_PAIR(STRIKE, 2),
     {QRegularExpression(MULTILINE_CODE), htmlPatterns[CODE]}};
 
+// RFC 3986: https://tools.ietf.org/html/rfc3986#section-2
+static const QString uriPattern = "[\\w:/?#[\\]@!$&'\\(\\)*+,;=\\-._~]+";
+
 static const QVector<QRegularExpression> urlPatterns {
-    QRegularExpression("((\\bhttp[s]?://(www\\.)?)|(\\bwww\\.))"
-                       "[^. \\n]+\\.[^ \\n]+"),
-    QRegularExpression("\\b(ftp|smb)://[^ \\n]+"),
+    QRegularExpression(                   "("
+        "((\\bhttp[s]?://)|(\\bwww\\.))"  "|"
+        "(\\b(ftp|smb)://)"               ")"
+        + uriPattern,
+        QRegularExpression::UseUnicodePropertiesOption),
     QRegularExpression("\\bfile://(localhost)?/[^ \\n]+"),
     QRegularExpression("\\btox:[a-zA-Z\\d]{76}"),
     QRegularExpression("\\b(mailto|tox):[^ \\n]+@[^ \\n]+")
@@ -152,6 +157,8 @@ static void processUrl(QString& str, std::function<QString(QString&)> func)
 {
     int startLength = str.length();
     int offset = 0;
+    // FIXME: regex is matched against escaped HTML, it should be matched
+    // against raw text instead
     for (QRegularExpression exp : urlPatterns) {
         QRegularExpressionMatchIterator iter = exp.globalMatch(str);
         while (iter.hasNext()) {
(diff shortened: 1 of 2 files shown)


yurivict:
I have a question about ```(\\bhttp[s]?://(www\\.)?)```:
If the text begins with ```http(s)://```, why do you need to match www after it? URL hostnames don't have to begin with www.

Also, why do you only match http[s], ftp, smb? There are many other schemes. Shouldn't it be ```[a-z][a-z0-9-]*[a-z0-9]```? https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

zetok:
>I have a question about ```(\\bhttp[s]?://(www\\.)?)```:
>If the text begins with ```http(s)://```, why do you need to match www after it? URL hostnames don't have to begin with www.

Changed.

>Also, why do you only match http[s], ftp, smb? There are many other schemes. Shouldn't it be ```[a-z][a-z0-9-]*[a-z0-9]```? https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Those have been requested, so they're supported. So far there have been no requests to support other kinds of URLs, so..

sudden6:
This seems to break the tests, maybe the tests need updating?

zetok:
Yeah.

sudden6:
@noavarice can you review the changes in the tests?

noavarice:
You can just remove this test case completely, it's unnecessary since backtick is not allowed in URL. Actually, I have started my own refactoring after this PR, including new regexp for `http` and `ftp` schemes and test changing, but now I don't sure - should I make it further or not

zetok:
I think that it would be better to retain test case with backticks to ensure that they're not parsed as a part of the URL.

yurivict:
This might be beyond the scope of this PR, but I think the regular expression should be more sophisticated.

For example, trailing punctuation should be excluded. For example, in this text URLs are selected wrong: see here http://www.xxx.com, and here http://www.xxx.com.

Another problem is brackets and braces. Practically speaking, unmatched opening and closing  brackets and braces should be excluded from URLs too, for example: qTox (https://github.com/qTox) is cool.

yurivict:
This person, John Gruber, researched the problem, and came up with the solution that is free of the above 2 problems (I verified this).

http://daringfireball.net/2010/07/improved_regex_for_matching_urls

Maybe we should use his regexp instead?

zetok:
@yurivict I briefly looked at the regex, and it would seem that (at least without modification) it's limited to currently existing TLDs – it wouldn't support domains from e.g. https://www.opennicproject.org/ , not to mention all kinds of stuff on intranet.

yurivict:
@zetok  Could you provide an example of failing URL?
I tried http://xxx.游戏/xxx  (with the new Mandarin TLD) and http://сайт.сайт/xxx (with the new Cyrillic TLD) and they work ok.
I think, as long as alpha-numeric ranges are not used, regular expressions should be UTF8-safe.

ezavod:
@zetok 
What do you think about handling `file://` links the same like `smb://` and `ftp://`?
Using `(ftp|smb|file)` one could fix #4445.

zetok:
@ezavod sounds good to me.

Talkless:
Reviewed 2 of 2 files at r1.

---

*[src/chatlog/textformatter.cpp, line 92 at r1](https://reviewable.io:443/reviews/qtox/qtox/4297#-KrunvsLlVuSU0Ov0d6z:-KrunvsMwVlXWnJvLodD:be09w1) ([raw file](https://github.com/qtox/qtox/blob/dde9148235576ee1f0dba5aab7a1335665aaf27c/src/chatlog/textformatter.cpp#L92)):*
> ```cpp
> 
> // RFC 3986: https://tools.ietf.org/html/rfc3986#section-2
> static const QString uriPattern = "[\\w:/?#[\\]@!$&'\\(\\)*+,;=\\-._~]+";
> ```

QStringLiteral("[....
Also, I would suggest to use C++11 raw string literals, to avoid double-backslashes. In that thay regexp will be copy-pastable to debugging tools such as RegexCoatch or Kiki:
QStringLiteral(R"([\w:/?#[\]@!$&'\(\)*+,;=\-._~]+)")

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4297)*

anthonybilinski:
>trailing punctuation should be excluded

>unmatched opening and closing brackets and braces should be excluded from URLs

Both comments are addressed by merged https://github.com/qTox/qTox/pull/4951 (mixed regex and iterative solution). https://github.com/qTox/qTox/pull/4951 more aggressively matches than RFC 3986. Are there remaining changes in this PR, or is it safe to close?

yurivict:
Just close it, imo. If additional changes are needed, they could go into a separate PR.

Issue #4303, opened 2017-03-28
Avira hit while running bootstrap.bat


nweyand:
Avira detects something called „TR/Agent.eweqd“ (which has helpfully no description on the Avira website) while the bootstrap script works with
`qTox\libs\openal-soft-1.16.0\build\CMakeFiles\CMakeTmp\<random_filename>.exe`

You might want to add a warning about that to the website.

zetok:
Would you mind adding it to [wiki](https://github.com/qTox/qTox/wiki/Problematic-antiviruses)?

nweyand:
done!

nweyand:
The pull request linked to above adds a small warning about there possibly being false positives on some anti virus software and refers to the wiki page zetok referred me to.

sudden6:
Seems closed.

PR #4291, opened 2017-03-26
[WIP] Replace Windows build guide with fresh one

DX37:
Relates to #3372 and partially fixes it (still need to add static build manual).

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4291)

PR #4304, opened 2017-03-28
docs(INSTALL): updated Windows build instructions

nweyand:
There was some folder mixup, probably due to some parts of the instruction being written out of a 32 bit Windows perspective (where “C:\Program Files” is the default location for installed programs) and other parts written out of 64 bit perspective (where there are two distinct folders for 32 and 64 bit content).

I’ve made sure all paths are compliant with the x64 version now (I’ve left out x86 instructions as I would have no way to test them, x86 is slowly but surely dying out AND people compiling on older systems should be aware anyway that instructions referring to install paths including “C:\Program Files (x86)\” will need some adjustment on their systems).

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4304)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -778,14 +778,14 @@ version of MinGW, corresponds to the version of the QT component!
 
 Download the Wget installer for Windows from
 http://gnuwin32.sourceforge.net/packages/wget.htm. Install them. The following
-steps assume that Wget is installed at `C:\Program Files\GnuWin32\`. If you
+steps assume that Wget is installed at `C:\Program Files (x86)\GnuWin32\`. If you
 decided to choose another location, replace corresponding parts.
 
 ### UnZip
 
 Download the UnZip installer for Windows from
 http://gnuwin32.sourceforge.net/packages/unzip.htm. Install it. The following
-steps assume that UnZip is installed at `C:\Program Files\GnuWin32\`. If you
+steps assume that UnZip is installed at `C:\Program Files (x86)\GnuWin32\`. If you
 decided to choose another location, replace corresponding parts.
 
 ### Setting up Path
@@ -797,7 +797,7 @@ select tab `Advanced system settings` -> button `Environment Variables`). In the
 second box search for the `PATH` variable and press `Edit...`. The input box
 `Variable value:` should already contain some directories. Each directory is
 separated with a semicolon. Extend the input box by adding
-`;C:\MinGW\bin;C:\MinGW\msys\1.0\bin;C:\Program Files (x86)\CMake 2.8\bin;C:\Program Files\GnuWin32\bin;C:\Program Files (x86)\GnuWin32\bin`.
+`;C:\MinGW\bin;C:\MinGW\msys\1.0\bin;C:\Program Files (x86)\CMake 2.8\bin;C:\Program Files (x86)\GnuWin32\bin`.
 The very first semicolon must only be added if it is missing. CMake may be added
 by installer automatically. Make sure that paths containing alternative `sh`, 
 `bash` implementations such as `C:\Program Files\OpenSSH\bin` are at the end of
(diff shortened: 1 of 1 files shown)


zetok:
Are windows instructions with fixed paths sufficient to compile qTox?

According to @DX37 current instructions don't work, and so #4291 replaces with working instructions that use MSYS2.

If instructions with fixed path do the job, it would be good to have both ways of compiling listed in INSTALL.md.

DX37:
@zetok As I said in IRC, the current instruction is not that it does not work - it uses very old libraries. And in general, the usual MinGW is pretty old, while the optimized and improved MinGW-w64 is still alive.

nweyand:
> Are windows instructions with fixed paths sufficient to compile qTox?

I am not sure. I have not proceeded so far yet: Due to a problem while downloading something within bootstrap.bat (or rather bootstrap.sh) I haven’t been able to finish compilation. But this fixes everything at least up to the point that one can run the bootstrap shell script without having unused directory entries in the path.

EDIT: As for mingw... Why not switch to clang? It produces better error messages 90% of the time and also generates faster executables. (The only possible drawback is that you might have to compile Qt yourself ATM – but that is supported out of the box.)

nweyand:
This addresses #4303 and adds a small warning about there possibly being false positives on some anti virus software and refers to the appropriate [wiki page](https://github.com/qTox/qTox/wiki/Problematic-antiviruses).

DX37:
@nweyand Instead of using separated programs, with PATH editing, some system changes and etc, can we just use MSYS2, which just installs in one folder, use separated shell that doesn't affect system's PATH and can provide a simple way to install dependencies?

DX37:
Just one new working instruction instead of patching an old and ancient one.

nweyand:
Before reading your comments I wasn’t aware that this was being discussed. Is there already a pull request for your solution?

In any way my suggestion would be to merge my changes as a quick-fix and then start discussing and testing your changes and later use them to overwrite (part of) what I’ve suggested here. The latest addition I made (about possible false positives on some AV SW) stays valid though, even if the process of getting the dependencies is fixed.

DX37:
@nweyand It's #4291.

nweyand:
Ah, your suggestion really looks promising.

I fully support your version over the current instructions provided there is a way to statically build qTox (i.e. take the .exe (and necessary .dll’s), put them into a .7z file and pass them to a friend and be assured that it will work on his PC afterward).
Also if you would like to, I could suggest some grammar and spelling improvements?

DX37:
> Also if you would like to, I could suggest some grammar and spelling improvements?

Sure. I know English, but not very well :smile:

nweyand:
Cool. Would you like me to make a lot of comments in your pull request or may I get access to your repo so that I can modify it directly? Or would you prefer for me to add a pull request to your repo?

EDIT: Maybe we should migrate this discussion to some place that is more appropriate – it has strayed far from the original topic. :)

DX37:
@zetok I think you can close this PR 🤔

nweyand:
> @zetok I think you can close this PR :thinking:

no.

> In any way my suggestion would be to merge my changes as a quick-fix and then start discussing and testing your changes and later use them to overwrite (part of) what I’ve suggested here. The latest addition I made (about possible false positives on some AV SW) stays valid though, even if the process of getting the dependencies is fixed.

As I said: Use this as a quick-fix ;)

nweyand:
Any chance this could be merged? #4304 will take a while still before it is ready for merging and in the meantime there is no working instruction on how to get qTox to compile properly.

zetok:
:lgtm_strong:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4304#-:-KgSGZeiRfI_cKcovLsd:b-jbsf8z)*

Issue #4003, opened 2016-12-23
Mark message with triple click


NicoHood:
OS: ArchLinux
qTox version: 1.6

If you triple click a message I expect the GUI to mark the whole line/message. Test it on this github message with your browser. But qtox only knows the double click which marks a single word.

PR #4305, opened 2017-03-28
feat(chatform): mark message with triple click

ezavod:
This is an implementation of proposal #4003. It allows the user to select a line of a chat by triple click.
Note: This will select received lines not displayed lines. So if you receive a long line, which is displayed as multiple lines you still select the long line which visually spans several lines on your display.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4305)

### DIFF
--- a/src/chatlog/chatlinecontent.cpp
+++ b/src/chatlog/chatlinecontent.cpp
@@ -56,6 +56,10 @@ void ChatLineContent::selectionDoubleClick(QPointF)
 {
 }
 
+void ChatLineContent::selectionTripleClick(QPointF)
+{
+}
+
 void ChatLineContent::selectionFocusChanged(bool)
 {
 }
--- a/src/chatlog/chatlinecontent.h
+++ b/src/chatlog/chatlinecontent.h
@@ -45,6 +45,7 @@ public:
     virtual void selectionStarted(QPointF scenePos);
     virtual void selectionCleared();
     virtual void selectionDoubleClick(QPointF scenePos);
+    virtual void selectionTripleClick(QPointF scenePos);
     virtual void selectionFocusChanged(bool focusIn);
     virtual bool isOverSelection(QPointF scenePos) const;
     virtual QString getSelectedText() const;
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -105,6 +105,12 @@ ChatLog::ChatLog(QWidget* parent)
     workerTimer->setInterval(5);
     connect(workerTimer, &QTimer::timeout, this, &ChatLog::onWorkerTimeout);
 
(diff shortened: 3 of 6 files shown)


sudden6:
IMO it would be better to only select a line like it is displayed, I think it will confuse users if lines are selected with boundaries they can't see.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4305#-:-KgVvSMeU4O6jyc8HDwS:b2qtgpg)*

ezavod:
> I think it will confuse users if lines are selected with boundaries they can't see

Valid point but I still think the current implementation is better because of several reasons:

- Triple clicking is a kind of advanced user interaction, so people who are using this are probably aware of the difference between an automatic line break for display purpose and a line break intended by a sender. They would not expect to just select the displayed line.
- Selecting just the line, which is displayed would cause major confusion eg. if I send some code to a friend and he wants to select a certain line of code by triple clicking it would be wired if he selects just a random chunk depending on his window size and not the original line of code.
- It would also lead to inconsistency across programs because many programs (all I have tested so far) use triple click to select the "next logical unit" above double clicking a word (which would be a line/sentence/paragraph).
You could test this yourselves by triple click something in your browser. When I triple click the `Note:` paragraph in my first comment Firefox selects all two lines.
- **Most importantly** is that Qt implements this already in the same way (tested on 5.7.1). You could use `Add Friends -> Message` to type a text consisting of multiple long lines or just copy this comment. Triple click on such a long line that spans visually over multiple rows selects all displayed lines (Note: `SHIFT + ENTER` does not count as a new line).
- Many design guidelines for different OS/platforms also suggest the triple click to select a whole line/paragraph. (You could have a look at [this](http://stackoverflow.com/questions/2575758/conventions-for-the-behavior-of-double-or-triple-click-to-select-text-features/2584774#2584774) fantastic comment about this topic but sadly it is quite old so some links are broken).

But of course we do not have to stick with old traditions if there are some arguments that overweight the current implementation.

sudden6:
Ok, the arguments seem reasonable. I'll need some time to review the PR.

sudden6:
Reviewed 5 of 6 files at r1.
Review status: 5 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4305)*

sudden6:
:lgtm_strong:

---

Reviewed 1 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4305#-:-Kh14iEhXXw1IwkZ0V_3:b-jbsf8z)*

PR #4328, opened 2017-04-10
fix(includes): Added missing #include <ctime>

yurivict:
time(3) requires this include. Found when qTox failed to build on the FreeBSD due to the missing include.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4328)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -32,6 +32,7 @@
 #include <QTimer>
 
 #include <cassert>
+#include <ctime>
 
 const QString Core::CONFIG_FILE_NAME = "data";
 const QString Core::TOX_EXT = ".tox";
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -24,6 +24,7 @@
 #include <QThread>
 #include <random>
 #include <unistd.h>
+#include <ctime>
 
 /**
  * @var time_t IPC::lastEvent
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -40,6 +40,7 @@
 
 #include <sodium.h>
 #include <stdio.h>
+#include <ctime>
 
 #if defined(Q_OS_OSX)
 #include "platform/install_osx.h"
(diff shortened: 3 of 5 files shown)


ezavod:
@yurivict what do you think about using [`#include <ctime>`](http://www.cplusplus.com/reference/ctime/) instead?
IMO it is better to use `ctime` because the use of `time.h` seems to be deprecated:

> For compatibility with the Standard C library, the C++ Standard library provides the 18 C headers [...], but their use is deprecated in C++.

(see Working Draft, Standard for Programming Language C++, [section C.2.1](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1905.pdf#section.C.2) and for a complete list [section D.5](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1905.pdf#section.D.5) and [section 17.4.1.2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2005/n1905.pdf#section.17.4)).

Also `ctime` is [already used in qTox](https://github.com/qTox/qTox/search?l=C%2B%2B&q=ctime&type=&utf8=%E2%9C%93) (eg. in _src/ipc.h_ and _src/net/toxme.cpp_).

yurivict:
@ezavod  Agreed. I updated the patch.

sudden6:
:lgtm_strong:

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4328#-:-KhR4kUOAEFr0e8tqp1E:b-jbsf8z)*

noavarice:
If this header is missed only on FreeBSD build, would it be better to include `<ctime>` using `#define`?

yurivict:
@noavarice  time.h or ctime inclusion is required on all OSes. On linux headers are different and  probably have time.h included from some other header.

PR #4333, opened 2017-04-15
refactor(widget): wrap content layout in a "weak" pointer

Diadlo:
Like every layout, the content layout is parented and owned by a widget.
This owner is responsible for layout destruction.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4333)


Diadlo:
If you think, that this change is useless -- write it

Issue #3100, opened 2016-04-04
On initial startup "qTox - Settings" is shown when "qTox - Add friend" should be shown


ghost:
I have recently found that when first starting up qTox it incorrectly labels the section at the top as `qTox - Settings` when it should be labelling it as `qTox - Add friend`:

![qtox-settingsincorrectlyshowing](https://cloud.githubusercontent.com/assets/11575271/14250961/9d37d50a-fa78-11e5-911b-839c309e4c99.png)

Though when going to another section and then coming back to that section it is labelled correctly:

![qtox-addfriendcorrectlyshowing](https://cloud.githubusercontent.com/assets/11575271/14250968/a346c06e-fa78-11e5-8086-8c995422db3b.png)

I am running Ubuntu GNOME 15.10 with GNOME 3.18 and qTox version `1.3.0-138-g32ed2ac-1`.

Issue #2758, opened 2015-12-31
qTox should use the system's language by default


ovalseven8:
Hello,

currently, the default language is English.
In my opinion, qTox should check the system's language at the very first start and using this.
Of course, English should be the fallback if there's no translation yet.

At the moment, the user has to create a profile first (with English login window) and load the profile to be able to change the settings. But what if the user doesn't know English at all? He probably have to look for a few minutes to find the settings (if the profile creation was successful).

Nurmagoz:
but that would produce many problems if qtox going to use this feature by default dont u think ? 

for example what about if qtox didnt detect the system language ? like for e.g if we put qtox inside a virtualmachine? (not real system)

also if X user is using qtox as a portable app and he open it inside cabinetwork or different laptops and let us assume that one laptop has default language Spanish and other one is English , should qtox react to each pc with its language ? 

in theory , this is very possible. but i think on reality this will cause issues regarding if qtox read correctly the system language or not.

and imagine the reaction of someone opened qtox and qtox mistakenly switched the language to China and the user from America = how happy that user will be? (i think even Google-translatore wont help him)

zetok:
> (…) qTox should check the system's language at the very first start and using this.

Weird, I thought that qTox was already doing that.

# 

@TNTBOMBOM 

> for example what about if qtox didnt detect the system language ?

Always can fallback to English.

> like for e.g if we put qtox inside a virtualmachine? (not real system)

?
VM still has language set. If not, see above.

> also if X user is using qtox as a portable app

That's not first start, and qTox should detect current settings.

> and imagine the reaction of someone opened qtox and qtox mistakenly switched the language to China and the user from America = how happy that user will be? (i think even Google-translatore wont help him)

Eh.

tux3:
By default we already try to use the system language, and fallback on English if it fails.
If this doesn't work, it's a bug.

The only part where we don't do it yet is the login window, so this what needs to be added for this issue.

Nurmagoz:
> Always can fallback to English. @zetok
> fallback on English if it fails. @tux3

these lines solved my issuing statements above = good & thnx.

ovalseven8:
@tux3: Just downloaded the static binary from tox.chat and it doesn't use the system's language by default?

tux3:
@ovalseven8 Did you make sure to erase the settings file? Otherwise qTox will use whatever is set in the settings.
If you did and qTox fell back to english instead of using the system language, that sounds like a bug.

[...]

PR #4340, opened 2017-04-19
fix(settings): changing language sets title to "Add Friend"

### LINKED ISSUE #3708: Changing language set title to "Add Friend"
##### Brief Description

OS: Arch Linux
qTox version: v1.4.0-2-g24a7f4d.
Commit hash: 24a7f4d51960a73cb7234cec9a4a2811e9175c6c
toxcore: 0.0.0
Qt: 5.6.0

Reproducible: Always
##### Steps to reproduce
1. Setup one window mode
2. Change langugage
##### Observed Behavior

Window title changed on `qTox - Add friend`
##### Expected Behavior

Title should be `qTox - Settings`

ezavod:
This fixes #3708.
Changing language sets title of settings widget to "Settings" instead
of wrong "Add Friend".

Note that there is an other user who committed similar changes to his fork, but he has not been active for over half a year now, so I think its OK to get this minor bug fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4340)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -2318,8 +2318,9 @@ void Widget::retranslateUi()
     actionQuit->setText(tr("Exit", "Tray action menu to exit tox"));
     actionShow->setText(tr("Show", "Tray action menu to show qTox window"));
 
-    if (!Settings::getInstance().getSeparateWindow())
-        setWindowTitle(fromDialogType(DialogType::AddDialog));
+    if (!Settings::getInstance().getSeparateWindow() && (settingsWidget && settingsWidget->isShown())) {
+        setWindowTitle(fromDialogType(DialogType::SettingDialog));
+    }
 
     friendRequestsUpdate();
     groupInvitesUpdate();


Diadlo:
Sorry, but this solution will produce #3100
So it's not fix the problem

ezavod:
@Diadlo Sorry, did not noticed that.

Issue #3100 should not happen anymore.
So on startup title is translated as "Add Friend". Only after settings are visited at least once, title is translated as "Settings".

Diadlo:
I don't think, that it's right solution. Can you replace it on current widget detection + set title in accordance with it?

ezavod:
Did some code cleanup and added explicit check if settings are currently shown.
I don not think that it is necessary to check for other possible widgets, because title translation is handled correctly under all other circumstances except one changes the language from within a widget. That is only possible from settings.

Diadlo:
@ezavod What if we will add new place to change language? For example it will change according to system language change? Sould we add more if's?

ezavod:
@Diadlo 

> What if we will add new place to change language?

The way how it is currently handled by qTox felt a bit weird to me, but I have no idea how this could be improved. This is probably better for someone, who has a deeper understanding of how the qTox code works as a whole, than I have at the moment.

> change according to system language change?

I think this would also be a big task to do, because there are some problems.
For example: The user has the option to change the language in settings. So if a user sets the language to `A` in settings but the system-language is set to `B` should one use the user setting from qTox or the system setting?
A solution could be to add a new option to the language menu that is something like `Use system language` (and maybe make it default on fresh installs [see #2758]).

> Sould we add more if's?

No, this is not required at the moment. All titles are translated properly if the currently visible form/widget is changed.
This breaks if the language is changed from within a form/widget, because the title was set when the form/widget was set as active, but since this point in time the language was changed, but the title was not.
The only way this could happen is if the user changes the language from settings. Therefor it is sufficient to check if settings are visible to filter the language set event emitted at startup from user changes.
There is no option to change the language in one of the other possible forms/widgets (`AddFriendForm`, `GroupInviteForm`, `ProfileForm`, `FilesForm`,  `GenericChatroomWidget`) so there is no need to test for them.

**In short:** I think one should use this as a fix for #3708 (of course only if this works for everyone).
Refactoring the whole widget translation or detecting system-language is out of the scope of this PR, I think.

sudden6:
@Diadlo @ezavod Should this be merged for 1.10?

Diadlo:
I don't think, that it's the right way. But can be merged as hotfix

sudden6:
ok, I'll merge it then.

sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4340#-:-KjSNJ0jATFWXGwbjtIR:b-jbsf8z)*

ezavod:
> I don't think, that it's the right way. But can be merged as hotfix

This was meant as hotfix, so that the uses is not bothered with strange title changes anymore. Sorry that I can not do more, due to a lag of time and qTox code knowledge.

Issue #300, opened 2014-09-19
Accessibility support


tyrylu:
Hello,
it would really help if qtox would support screen reader accessibility. QT accessibility can be done and for testing you can use NVDA on Windows and Orca under Linux.
The current situation is as follows:
- [x] Button labels are read correctly
- [x] The presence of some edit box is announced
- [x] Labels aren't read (see #2101)
- [x] Focus behavior is sometimes quite strange (see #2100)

zetok:
from #1215:

> The QTox Windows Client is not accessible for Users with Assistive Technologies.
> See: http://qt-project.org/doc/qt-4.8/accessible.html

.

> Or consider upgrading to the latest QT5 version, there were some substantial accessibility improvements and the entire qtaccessibblewidgets stuff was simplified.

@tyrylu which Qt version has those improvements?

tyrylu:
Now, when the builds are using QT5, it's better to create smaller issues.

suhr:
Wait. Let this issue be a roadmap of these smaller issues (consider editing issue message).

tyrylu:
Alright, i'll update the issue descripttion with my current findings.

ezavod:
@tyrylu Is there anything left to do in this issue, since #2100 has been closed?
I would say this one could be closed.

PR #4341, opened 2017-04-19
fix(chatform): disable Tab in add friend message text  area

### LINKED ISSUE #2100: Focus management during initial setup
When an user creates a new account, the settings window is likely opened (inferred from the title), but the keyboard focus is on the main window elements. After tabbing a while, the user reaches some edit control which can't be left using the keyboard. I have no idea if the NVDA screen reader messes that somehow. Tried on the latest windows build from the tox wiki, 64-bit version.

ezavod:
This is related to #2100 that is part of #300.
Tab is no longer accepted as input in add friend message text area,
instead it changes the focus.

I do not know if this fixes #2100 completely, but it solves the problem that the user gets stuck in the text area when using Tab to change focus.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4341)

### DIFF
--- a/src/widget/form/addfriendform.cpp
+++ b/src/widget/form/addfriendform.cpp
@@ -85,6 +85,7 @@ AddFriendForm::AddFriendForm()
     messageLabel.setAccessibleDescription(tr("Friend request message"));
     message.setAccessibleDescription(tr(
         "Type message to send with the friend request or leave empty to send a default message"));
+    message.setTabChangesFocus(true);
 
     retranslateUi();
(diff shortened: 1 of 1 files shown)


sudden6:
:lgtm_strong: 

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4341#-:-KiKEp--5c4neNlUzgzd:b-uwkowd)*

PR #4438, opened 2017-06-02
test: Replace check on Qt test

Diadlo:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4438)


Diadlo:
Qt Test is better, because if tested class use `QMetaObject` and some other Qt stuff then test required started `QApplication`.

sudden6:
Do tests work on Windows? I think we should not run them by default there.

Diadlo:
1) It should work
2) They doesn't runned by default (same as before)

sudden6:
:lgtm_strong: then

---

Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4438#-:-Kljhzvcx8zJUD1aGWEj:b1hmrgw)*

PR #4441, opened 2017-06-03
chore(build): don't update version numbers via cmake

### LINKED ISSUE #4439: cmake and FreeBSD
This sed call in osx/update-plist-version.sh breaks on the FreeBSD:
```
        sed -i -r "\\R$v\$R,+1 s,(<string>)[0-9\\.]+(</string>)$,\\1$@\\2," \
            "./info.plist"
 ```
This is an invalid syntax for the FreeBSD sed.

It appears to be wrong that tools/update-versions.sh calls update_osx and update_windows on all systems. osx/info.plist contains Apple DTD for some reason, yet it is used on linux and FreeBSD, this isn't right.

Replacing sed -> gsed causes cmake build to succeed but this isn't a proper workaround.

osx/update-plist-version.sh should only be called on OSX, same with Windows script.

I don't understand what it is doing, but IMO the way how it is called on all OSes is clearly wrong.

zetok:
Scripts updating versions in source depend on GNU sed which is sometimes
not set as the default on some OSes.

With the change, one will need to manually call the update script before
releases.

Fixes #4439.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4441)

### DIFF
--- a/cmake/Installation.cmake
+++ b/cmake/Installation.cmake
@@ -4,11 +4,6 @@
 #
 ################################################################################
 
-# set correct version to update files
-set(QTOX_VERSION "1.10.1")
-execute_process(COMMAND ${CMAKE_SOURCE_DIR}/tools/update-versions.sh
-    ${QTOX_VERSION})
-
 if(APPLE)
   set_target_properties(${PROJECT_NAME} PROPERTIES
     MACOSX_BUNDLE_INFO_PLIST "${CMAKE_SOURCE_DIR}/osx/info.plist")


sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4441#-:-KljdZqTmISSgFGRJ157:b-jbsf8z)*

nurupo:
Pretty sure you could replace the GNU sed with a CMake script. CMake can read/write to file, supports regex, etc., so all you'd need is just to have CMake on the target system.

>You can read a file using FILE (READ) and then replace stuff using 
>STRING(REPLACE) and STRING(REGEX REPLACE) and write it using FILE(WRITE).
> [...]
>I would suggest using plain configure_file().

https://cmake.org/pipermail/cmake/2007-August/015936.html

(CMake can also do things like downloading a file given URL, calculating hashes of files, etc.)

Although more portable, it would be harder to write and maintain than a bash script, so chose your poison.

Issue #4406, opened 2017-05-14
[Regression] Warning: QObject::connect: invalid null parameter


ezavod:
I am getting an error: `:0 : Warning: QObject::connect: invalid null parameter` on startup.
This is consistently reproducible since 7e3ad7ba873615f4b7a65071d678ac936f7504e7.
As far as I can tell the problem seems to be that `av` is `null` in lines 237-238 in above mentioned commit:
```cpp
    CoreAV* av = core->getAv();
    connect(av, &CoreAV::avEnd, this, &Widget::onCallEnd);
```
I think this is because `Core::getAv` is called **before** `Core::start` initializes the returned `av`.

Full log: [log.txt](https://github.com/qTox/qTox/files/999624/log.txt)

anthonybilinski:
I couldn't reproduce this on tip at the location mentioned, but https://github.com/qTox/qTox/blob/c49959711e6ed7dc6c0c244c34d323ab30fd0170/src/widget/widget.cpp#L240 is definitely null at connect time and throwing the same error.

PR #4472, opened 2017-06-17
 fix(core, global): add mutex for guard Core::start 

ghost:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4472)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -99,11 +99,13 @@ Core* Core::getInstance()
 
 const CoreAV* Core::getAv() const
 {
+    initializeMutex.tryLock(1000);
     return av;
 }
 
 CoreAV* Core::getAv()
 {
+    initializeMutex.tryLock(1000);
     return av;
 }
 
@@ -253,6 +255,8 @@ void Core::makeAv()
  */
 void Core::start(const QByteArray& savedata)
 {
+    initializeMutex.lock();
+
     bool isNewProfile = profile.isNewProfile();
     if (isNewProfile) {
         qDebug() << "Creating a new profile";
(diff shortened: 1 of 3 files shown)


ghost:
This fixed #4406 .

ezavod:
@QweEwq2016 
Sorry, but this does not seem to work right.
Startup of qTox is lot longer than normal (~3sec before, ~15sec with your changes).

As far as I can tell this is because in `Core::start()` you lock `initializeMutex` and then call `Core::getAv()` (on my system this happens 19 times, at least one of them is triggered via `Core::makeAv()`).
Obviously your lock is set at this point and one waits for timeout in `Core::getAv()`.
Within your lock you call `GUI::setEnabled()` -> `QMetaObject::invokeMethod()` which blocks until all tries to lock in `Core::getAv()` resume.

After a brief glance at the code, the lock does not seem to have a use anyway (please correct me, if i miss something). Issue #4406 is fixed by your reordering in nexus.

ghost:
@ezavod 
Hello.
"Issue #4406 is fixed by your reordering in nexus." - maybe, not only reordering? startCore() call in other thread, and getAv() in Widget::init() maybe return nullptr, because startCore() maybe not finished. I do not quite understand how to solve this problem without a mutex. Any ideas?

ezavod:
@QweEwq2016 
Sorry for my delayed answer.
I do not know any good solution either.

Maybe one could test if the reordering fixes #4406 reliably.
It seems so to me, but maybe someone else should do proper testing.
This would not prevent other possible nullpointers, but would at least get rid of this occurring always at startup.

Diadlo:
@zetok As I said, I prefer github PR body autofill. Most of the PR sends by regular contributors who know the rules. (Frequency PR of regular contributers) * (time to set "X" in checkbox) >> (Frequency PR of new contributers) * (time to fix commit message)
I think this PR should be closed

anthonybilinski:
The referenced issue https://github.com/qTox/qTox/issues/4406 was fixed by https://github.com/qTox/qTox/pull/4651 which has been merged.

PR #4477, opened 2017-06-18
fix: Add unix signals handler

### LINKED ISSUE #4470: qTox does not handle common POSIX signals and quits abruptly
##### Brief Description

OS: Linux (and probably macOS?)
qTox version: current HEAD
Qt:  5.3.2, 5.9.0

Reproducible: Always

##### Steps to reproduce

**Case #1:** 
1. Launch qTox in terminal.
2. Close terminal.
3. qTox quits with `SIGHUP` signal, without `Debug: Clean exit with status 0` to log - abruptly.

**Case #2:**
1. Launch qTox in terminal.
2. Press `Ctrl + C`.
3. qTox quits with `SIGINT` signal, again abruptly.

**Case #3:**
1. Launch qTox in any way
2. Terminate it using your favourite GUI task manager or with `killall`
3. qTox quits with `SIGTERM` signal, again abruptly.

##### Observed Behavior

Terminating qTox with POSIX signal does not write last debug message, this is evidence that `main`() did not finish, so Qt event loop did not quit, destructors not called, etc, etc. This might corrupt user data.

##### Expected Behavior

qTox should quit cleanly when terminated with common POSIX signals such as `SIGINT`, `SIGHUP` and `SIGTERM`.

Diadlo:
Fix #4470
Thanks @azadkuh for the article
(https://gist.github.com/azadkuh/a2ac6869661ebd3f8588)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4477)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -42,6 +42,11 @@
 #include <sodium.h>
 #include <stdio.h>
 
+#if defined(Q_OS_UNIX)
+# include <signal.h>
+# include <unistd.h>
+#endif
+
 #if defined(Q_OS_OSX)
 #include "platform/install_osx.h"
 #endif
@@ -120,6 +125,17 @@ void logMessageHandler(QtMsgType type, const QMessageLogContext& ctxt, const QSt
 #endif
 }
 
+void catchUnixSignals(const std::vector<int>& quitSignals) {
+
+    auto handler = [](int sig) {
+        qDebug() << "Signal" << sig << "received. Quit";
+        QApplication::quit();
+    };
+
+    for (int sig : quitSignals)
+        signal(sig, handler);
+}
+
 int main(int argc, char* argv[])
 {
 
@@ -134,6 +150,10 @@ int main(int argc, char* argv[])
     a.setApplicationName("qTox");
     a.setOrganizationName("Tox");
     a.setApplicationVersion("\nGit commit: " + QString(GIT_VERSION));
+
+#if defined(Q_OS_UNIX)
+    catchUnixSignals({SIGQUIT, SIGINT, SIGTERM, SIGHUP});
+#endif
 
     // Install Unicode 6.1 supporting font
     // Keep this as close to the beginning of `main()` as possible, otherwise


Talkless:

---

*[src/main.cpp, line 132 at r1](https://reviewable.io:443/reviews/qtox/qtox/4477#-KmvNLaj0rSIOD22vWEx:-KmvNLaj0rSIOD22vWEy:bepxf6p) ([raw file](https://github.com/qtox/qtox/blob/fcad16571b8bc90c15fb6bb312953e14afc415db/src/main.cpp#L132)):*
> ```cpp
>     auto handler = [](int sig) {
>         qDebug() << "Signal" << sig << "received. Quit";
>         QApplication::quit();
> ```

NO. Only signal-safe functions can be called within signal handler. See https://doc.qt.io/qt-5/unix-signals.html .

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4477)*

Talkless:
I would suggest to create helper class to handle Unix, and Windows native "quting". 

On Windows, it should use [QCoreApplication::installNativeEventFilter](https://doc.qt.io/qt-5/qcoreapplication.html#installNativeEventFilter) to catch some desktop notification (I do not recall message name at the moment) when user logs out. 

Meanwhile, on *nix, it should start QSignalNotifier to get event from signal handling function through dedicated anonymous socket / pipe, as in Qt example: https://doc.qt.io/qt-5/unix-signals.html

Diadlo:
@Talkless Thanks you!

azadkuh:
@Talkless it could be safe to call `QCoreApplication::quit()` in signal handler by some minor modification. i've updated the gist and the reference.

although it could lead to UB for other purposes.

Diadlo:
@azadkuh Can you proof, that `QCoreApplication::quit()` call is safe?

azadkuh:
dear @diadlo:
i can't proof! but it is generally safe to call a blocking function by properly setting the mask (blocking mask) of `sigaction`

please have a look at GNU c lib doc (find the link at the end of the gist)

Talkless:
@azadkuh Blocking is nice, but other problem is that we do not know _when_ will signal be triggered.

What if it triggers when low-level event dispatcher implementation (and Qt has number of implementations for various systems) is holding some mutex for example (because Qt has thread-safe signal/slot mechanism across threads boundries), and then within your signal handler a `QCoreApplication::quit()` is called, then `QCoreApplication::exit()`, and then `QEventLoop::exit()` and now we are back into same event dispatcher implementation and then what, a deadlock?

Even `std::atomic<>'s` may not be used in signal handlers, if implementation is not actually lock-free (probably not the case ever for achitectures qTox targets, but still).

Who-knows-what jungle begins, what race conditions to expect, etc, etc...

azadkuh:
@talkless
thank you for your notes. 

Qt is supposed to release (free, unlock) all low level file descriptors, or handles, terminating threads, ... by destructors (due to RAII design), and those `dtors` are being called before the `QEventLoop::exit()` returns to the caller.

AFAIK there is not a strong guarantee, but i'm wondering if there is reproducible case to break this assumption.

Talkless:
> AFAIK there is not a strong guarantee

So is there any reason to risk and venture into that uncertainty? What to we loose by doing it by-the-book?

> but i'm wondering if there is reproducible case to break this assumption.

Yes that would be interesting experiment. Or we could ask for experience in Qt developer/user mailing lists, but I believe it's just waste of time...

I would suggest to implement in the way that is _correct by design_, rather than choosing implementation that is _maybe_ correct.. _by chance_... until Qt implementation changes or something :-)

PR #4493, opened 2017-06-26
Update packages versions

Diadlo:
Since new Debian stable was released, we can drop support of old packages

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4493)

PR #4494, opened 2017-06-26
fix(ui): wrong size of svg images on buttons

ezavod:
Replaced size in percentage with size in pixels in call and emote buttons.
Added margin to mic and volume buttons.

After updating to Qt 5.9.0 some buttons were broken (I am not sure which Qt version was the first to break, but maybe merge this before #4493 to avoid this issue).
Seems to be because setting size in percent is no longer working.
Current state and fixed:
buttons_top   buttons_top_fixed
button_bottom   buttons_bottom_fixed

~~Did also some cleanup of SVG code,~~ but this should be better done systematically with all images (I am very busy at the moment, but maybe I could have a look at this in future).

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4494)

### DIFF
--- a/ui/callButton/callButton.svg
+++ b/ui/callButton/callButton.svg
@@ -8,8 +8,8 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    version="1.1"
-   width="100%"
-   height="100%"
+   width="25"
+   height="23"
    viewBox="0 0 25 23"
    id="Layer_1"
    xml:space="preserve"><metadata
--- a/ui/emoteButton/emoteButton.svg
+++ b/ui/emoteButton/emoteButton.svg
@@ -8,8 +8,8 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    version="1.1"
-   width="100%"
-   height="100%"
+   width="17"
+   height="17"
    viewBox="0 0 17 17"
    id="Layer_1"
    xml:space="preserve"><metadata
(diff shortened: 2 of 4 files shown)


Diadlo:
@ezavod Can you split svg cleanup and icons fix in 2 commits?
Also, I don't think that there is really need to cleanup SVG, but feel free to continue

ezavod:
Removed cleanup for now, only necessary changes for replacing percentage values and adding margins.

> I  don't think that there is really need to cleanup SVG

Sure, it is not really an issue at all, but it might be still good to do so:
First it is possible to shrink the file size of the images by stripping useless [metadata](https://github.com/qTox/qTox/blame/2aed859bdb6d5204678a29f472aa24f430536db0/ui/volButton/volButton.svg#L33) (this would not do much, but would not harm either).
Second it seems that Inkscape has changed its default DPI value from 90 to 96 and maybe some other subtle changes in Inkscape and Qt (like the one in this PR).
I have a working script, that does this cleanup automated, but I do not have the time to check if those changes would not break something, so I leave that untouched for now.

Diadlo:
@ezavod What is DPI in vector image?)

ezavod:
Sorry, my comment was misleading, it seems that inkscape saves your preferred DPI value for export in the file as metadata. This just causes inkscape to display a warning, that you are using an old version of save format. It asks you if you want to convert to the new version. As far as I can tell this has no effect on the image itself, but may mess with viewbox attribute (at least this is what the warning says). I am not an graphics artist so I do not know, if this is such a big issue (and if this is even needed for Qt).

But this shows that there is a bulk of metadata, which is not needed.

Diadlo:
:lgtm_strong:

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4494#-:-Kn_eW7JxJBYS9er3hXu:b-jbsf8z)*

PR #4504, opened 2017-07-02
CMake Windows builds on Jenkins

sudden6:
Builds can be tested from [here](https://build.tox.chat/view/qtox/job/qTox-cmake_build_windows_x86-64_release/)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4504)


nurupo:
Your toolchain file is only for `x86_64-w64-mingw32`, you might want to add another one for `i686-w64-mingw32`. That said, many cmake-based projects don't provide toolchain files and expect whoever cross-compiles to write their own, since, well, there are so many possible cross-compilations (many architectures of android, musl, may arms, x32, ppc, sparc, etc.) that it doesn't make sense to provide toolchain files for all of them, not to mention that in one distro a 32-bit mingw cross-compiler might be named `i686-w64-mingw32`, while in another `i486-w64-mingw32`, meaning that your toolchain files might not work for everyone. I'm not saying that it's bad to include them into the repository or anything, it's for you to decide if it makes sense to include them. Maybe you want to explicitly say that you support Windows cross-compilation by providing those, or/and maybe you will write a cross-compilation documentation that uses those, or maybe you will make Travis cross-compile qTox -- those are good reasons to include them in the repo.

Another thing I want to note is that in the Jenkins build script your cmake command is

```sh
CMAKE_OPTIONS="-DCMAKE_PREFIX_PATH=$WORKSPACE/libs \
			   -DENABLE_STATUSNOTIFIER=False \
               -DENABLE_GTK_SYSTRAY=False \
               -DCMAKE_FIND_ROOT_PATH=$WORKSPACE/libs \
               -DCMAKE_BUILD_TYPE=Release \
               -DCMAKE_TOOLCHAIN_FILE=./windows/toolchain-x86_64-w64-mingw32.cmake"

cmake ${CMAKE_OPTIONS} ./
```

Ignoring the fact that you are not doing out-of-tree build, the issue here are the ` -DENABLE_STATUSNOTIFIER=False` and `-DENABLE_GTK_SYSTRAY=False` switches. I think that they should be automatically disabled for Windows, so that you don't have to explicitly turn them off.

Also, having `CMAKE_BUILD_TYPE` set to `Release` by default would be nice, though I know many cmake projects that default to `Debug`. I think cmake itself defaults to debug build, doesn't it? Anyway, it's my personal preference, I tend to overwrite the default to `Release` in my own cmake projects. You might or might not want to do that.

nurupo:
A bit offtopic, but [here is a debug version of the Windows build](https://build.tox.chat/job/qTox-cmake_build_windows_x86-64_debug/) that could use some testing too. The difference from the release one is that it is built with `-DCMAKE_BUILD_TYPE=Debug`, `qtox.exe` is not stripped, `gdb.exe` is included, as well as `debug-automatic.bat` and `debug-manual.bat` scripts and the qtox's `src` directory for gdb to pick up source files from. The user is supposed to run qtox using the `debug-automatic.bat` and `debug-manual.bat` scripts, instead of running `qtox.exe` directly. These scripts run `qtox.exe` in gdb, logging the output. You can read more on these scripts in [mingw-w64-debug-scripts repo](https://github.com/nurupo/mingw-w64-debug-scripts).

towlie:
Fixes https://github.com/qTox/qTox/issues/4150
But only when you compiled c-toxcore with cmake. The autotools variant seems to be deprecated.

sudden6:
@nurupo I think we're going to ship all the stuff that's needed for the jenkins builds in the repo, this will also include the toolchain files. IMO they can serve as good examples for others.

I fixed the switches to only be enabled on linux + freebsd.

I prefer to have the default build type as `Debug` because then it's easier to debug.

Thank you for the debug builds :)

---

Review status: 0 of 3 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4504#-:-KocJS--_2Rc99lbwuI5:b-4z1n33)*

zetok:
:lgtm_strong:

---

Reviewed 1 of 3 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4504#-:-KpA-BQh1YVjVwNEWdWP:b-jbsf8z)*

PR #4532, opened 2017-07-23
feat(audio): Added the configuration variable to enable/disable audio

yurivict:
The new configuration variable ENABLE_AUDIO with the default ON allows to disable audio.

The reasons:
1. [security] People who want to only use text and are afraid that their sound will accidentally get transmitted should be able to disable linking with audio library.
2. [debugging] It is better to have a way to disable various libraries to be able to isolate potential problems.
3. [flexibility] Users have another option to choose what they want.

Caveat:
* It looks like toxcore still doesn't support capability flags, so callers can always initiate audio calls, regardless of the peer ability/willingness to accept them. Currently choosing audio devices "None" also doesn't prevent callers from initiating audio calls.

Future:
* Same should be done with video, there should be ENABLE_VIDEO_OUT that will disable camera.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4532)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ cmake_minimum_required(VERSION 2.8.11)
 set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
 option(PLATFORM_EXTENSIONS "Enable platform specific extensions, requires extra dependencies" ON)
+option(ENABLE_AUDIO "Enable audio conversations" ON)
 
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Debug)
@@ -181,10 +182,6 @@ execute_process(COMMAND ${CMAKE_COMMAND} -E copy_if_different
   ${CMAKE_CURRENT_BINARY_DIR}/translations.qrc)
 
 set(${PROJECT_NAME}_SOURCES
-  src/audio/audio.cpp
-  src/audio/audio.h
-  src/audio/backend/openal.cpp
-  src/audio/backend/openal.h
   src/chatlog/chatlinecontent.cpp
   src/chatlog/chatlinecontent.h
   src/chatlog/chatlinecontentproxy.cpp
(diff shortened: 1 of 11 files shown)


sudden6:
1. I don't think this much better than selecting `None` as Input/Output device. IMO it would be the same to disable audio calls when Input/Output device are `None`

2. A much cleaner way to implement your solution would be to implement a dummy audio backend that just doesn't do anything and doesn't include any audio library, see https://github.com/qTox/qTox/pull/4367/commits/455c644e9ff3f63867e2dd0a2b7a346305120f85 for an example how this could be done.

yurivict:
Having a package option is more reliable. Run-time settings selection isn't as reliable. Application can crash and lose settings while to change the package option you need to recompile.

This is just anther, more reliable, way to disable audio. Plus it has other benefits.

qTox chose OpenAL as a sound selection/multiplexor mechanism. OpenAL has Null backend that does the same. It isn't logical to have another place to select audio backend.

Diadlo:
@yurivict, I think @sudden6 talk, that instead of encircling code with a lot of `ifdef` better use it once to select instance to create

yurivict:
How can I disable audio in https://github.com/qTox/qTox/commit/455c644e9ff3f63867e2dd0a2b7a346305120f85 ? It has to be a compile option, not settings. Settings sometimes change randomly.

Diadlo:
@yurivict You can change the `Audio::getInstance` code in compile time

Diadlo:
@yurivict comment from 455c644e9ff3f63867e2dd0a2b7a346305120f85
> This should be a compile time option. Why enable both old and new at the same time?
> NullBackend should be one of the choices.
> When NullBackend is selected at compile time, Audio Settings and Audio Call button should be grayed out.

Diadlo:
1) You remove all not `None` backends from `CMakeLists.txt`
2) You change `Audio::getInstance` on compile time on code, which will create dummy backend
2.1) `None` backend will return empty device list => in settings will be only one possible device `Disabled`
3) Make buttons gray if `Disabled` audio selected

PR #4533, opened 2017-07-24
Fix sqlcipher header inclusion

nurupo:
SQLCipher sets include path to `-I${includedir}/sqlcipher` https://github.com/sqlcipher/sqlcipher/blob/master/sqlcipher.pc.in#L13 , which you can see in cmake

```sh
-- Checking for one of the modules 'sqlcipher'
-- SQLCIPHER LIBRARY_DIRS: /workdir/x86_64/build/../deps/libsqlcipher/lib
-- SQLCIPHER INCLUDE_DIRS: /workdir/x86_64/build/../deps/libsqlcipher/include/sqlcipher
```

while the header inclusion in `rawdatabase.cpp` assumes that it's set to `-I${includedir}`. This might accidentally work as long as you have libsqlcipher-dev installed system-wide, as gcc will look for header `sqlcipher/sqlite3.h` in default system include directory and will accidentally find it there, but if you don't have it installed system-wide, which is pretty common when cross-compiling, it will look into `sqlcipher-prefix/include/sqlcipher/sqlcipher/sqlite3.h` and not find it, as it's actually one level below, at `sqlcipher-prefix/include/sqlcipher/sqlite3.h`.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4533)

### DIFF
--- a/src/persistence/db/rawdatabase.cpp
+++ b/src/persistence/db/rawdatabase.cpp
@@ -34,7 +34,7 @@
 #define SQLITE_HAS_CODEC
 #define SQLITE_TEMP_STORE 2
 
-#include <sqlcipher/sqlite3.h>
+#include <sqlite3.h>
 
 
 /**


Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4533#-:-Kpod2UfiUw9fOYD4ugS:b-jbsf8z)*

PR #4553, opened 2017-07-30
fix(build): remove newline from timestamp

sudden6:
If this newline makes it into the build flags, following flags are ignored

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4553)

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -179,6 +179,7 @@ if (NOT TIMESTAMP)
   execute_process(
     COMMAND date +%s
     OUTPUT_VARIABLE TIMESTAMP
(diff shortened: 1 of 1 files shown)


Diadlo:
Unfortunately, I don't understand the reason and cast of this change. But seems, nothing broken. :lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4553#-:-KqWQ-EOQdfiRh6UF4q5:bnldsgq)*

sudden6:
The reason is that when a compiler flag with the `add_definition()` cmake function is added after the timestamp, the following flag is ignored, because makefile commands end at newlines.

PR #4570, opened 2017-08-11
fix(dependency): remove unused qt sql dependency

### LINKED ISSUE #4567: qTox doesn't seem to use libQt5Sql, but installation instructions, cmake, etc. say that it's required
Was confused by this https://github.com/voidlinux/void-packages/commit/03bbc343309ad5e687e6fed7973123229c82fe36 commit as why they link against Qt SQL libraries, since, as far as I can see (`grep -i -R --colour 'sql' *`), the only SQL library that qTox uses is `libsqlcipher`, which is a fork of `libsqlite` with encryption added on top, so you don't even have to install a vanilla `libsqlite`, not to mention Qt5's SQL libraries. If this is correct, then the sqlite and Qt5 SQL library requirements must be removed from all instructions, build scripts and cmake files, as they are not being used by qTox.

sudden6:
fix #4567

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4570)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -40,18 +40,18 @@
 
 ## Dependencies
 
-| Name          | Version     | Modules                                           |
-|---------------|-------------|-------------------------------------------------- |
-| [Qt]          | >= 5.5.0    | core, gui, network, opengl, sql, svg, widget, xml |
-| [GCC]/[MinGW] | >= 4.8      | C++11 enabled                                     |
-| [toxcore]     | = 0.1.\*    | core, av                                          |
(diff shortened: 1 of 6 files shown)


nurupo:
Missed NSIS installer scripts `windows/qtox.nsi` and `windows/qtox64.nsi`, they do something with `sqldrivers` Qt plugin .dlls.

sudden6:
Should be fixed

nurupo:
:lgtm_strong:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4570#-:-KrIh_oLPGAi3p2BIVD9:b-jbsf8z)*

PR #4617, opened 2017-09-01
fix(toxme): Remove HTML tags from ID to un-break toxme integration

LittleVulpix:
Due to the colours, the current implementation sends a bunch of html tags along with the ID, which obviously isn't desirable.
To remove these tags, I am using the same implementation as in copyIdClicked .

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4617)

### DIFF
--- a/src/widget/form/profileform.cpp
+++ b/src/widget/form/profileform.cpp
@@ -512,6 +512,7 @@ void ProfileForm::onRegisterButtonClicked()
     bodyUI->toxmeUpdateButton->setText(tr("Update (processing)"));
 
     QString id = toxId->text();
+    id.remove(QRegularExpression("<[^>]*>"));
     QString bio = bodyUI->toxmeBio->text();
     QString server = bodyUI->toxmeServersList->currentText();
     bool privacy = bodyUI->toxmePrivacy->isChecked();


LittleVulpix:
@noavarice Changed commit message from https://github.com/qTox/qTox/pull/4616

sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4617#-:-Kt1YDAluKbea7d3KyW9:b-jbsf8z)*

Talkless:

---

*[src/widget/form/profileform.cpp, line 515 at r1](https://reviewable.io:443/reviews/qtox/qtox/4617#-Kt23SQVjNyuT9atuUnD:-Kt23SQW9JvaNGQoCf2l:bav0ahw) ([raw file](https://github.com/qtox/qtox/blob/5af1b4af73d05e8da4bbc174410f0854a810ecd6/src/widget/form/profileform.cpp#L515)):*
> ```cpp
> 
>     QString id = toxId->text();
>     id.remove(QRegularExpression("<[^>]*>"));
> ```

QRegularExpression(**QStringLiteral(**"<[^>]*>"**)**);
Please see https://github.com/qTox/qTox/blob/master/doc/coding_standards.md#strings

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4617)*

LittleVulpix:
@Talkless This is literally your (=qTox's) existing code that I copied.

https://github.com/qTox/qTox/blob/master/src/widget/form/profileform.cpp#L215

Talkless:
Someone one day might refactor old code for new coding guidelines, but *new* code (this is new added line) could be better by default. 

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4617#-:-Kt26f0TyV3YXK5mF_UJ:ba6gr2z)*

LittleVulpix:
@Talkless When someday, someone is refactoring the old code, they should get rid of the stupid way that the &id is a reference and thus it is being overwritten in https://github.com/qTox/qTox/blob/master/src/widget/form/profileform.cpp#L242 for the sole purpose of pretty colours, so that we don't have to parse the modified string in https://github.com/qTox/qTox/blob/master/src/widget/form/profileform.cpp#L212 (and also newly in https://github.com/qTox/qTox/blob/master/src/widget/form/profileform.cpp#L503 in my commit) and instead use a local variable to set it up. At that time, it will be easier if the code uses the same way of doing things cause then it'll be much simpler to replace it.

Until then, I stand by keeping my commit as-is.

PR #4621, opened 2017-09-03
fix(chatform): Update member info when group created

kitech:
this should be fix issue https://github.com/qTox/qTox/issues/4503

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4621)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1865,6 +1865,8 @@ void Widget::onEmptyGroupCreated(int groupId)
     if (Widget::getInstance()->groupsVisible()) {
         groupWidgets[groupId]->editName();
     }
+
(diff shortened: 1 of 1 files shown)


Diadlo:
@kitech Probably, this method should be called in Group constructor?

kitech:
There are some other calls from widget.cpp, so I put here.
But not very sure.

noavarice:
@Diadlo meant that, since you want to update member info at the moment
group created, it makes sense to do within group constructor

kitech:
if this method called in Group constructor, how about the invited group?

Diadlo:
Implemented in #4978

anthonybilinski:
Closing since fix is merged.

PR #4666, opened 2017-09-20
fix(build): Fixed header file of libfilteraudio not being found

### LINKED ISSUE #4587: cmake doesn't look for filteraudio headers
cmake can't find filteraudio headers when filteraudio is not system-wide installed, which is the case when, for example, cross-compiling qTox to Windows. You can reproduce my issue by cross-compiling qTox to Windows with https://github.com/qTox/qTox/pull/4588 changes.

filteraudio doesn't provide pkg-config files, so cmake [falls back to using `find_library`](https://github.com/qTox/qTox/blob/5d159e0203bdd0e72153cce9ca5c0fe8483ba0db/cmake/Dependencies.cmake#L62). It does find the library successfully, but the issue is that it only looks for the library, when it also should be looking for the headers in the addition. This is what I get

```
-- FILTERAUDIO LIBRARY_DIRS: 
-- FILTERAUDIO INCLUDE_DIRS: 
-- FILTERAUDIO CFLAGS_OTHER: 
-- FILTERAUDIO LIBRARIES:    /workspace/i686/dep-cache/libfilteraudio/lib/libfilteraudio.a
-- FILTERAUDIO found
-- using filteraudio
```

As you can see, `INCLUDE_DIRS` is empty. And then I get

```
In file included from /build/qtox/src/audio/audio.cpp:23:0:
/build/qtox/src/audio/backend/openal2.h:41:26: fatal error: filter_audio.h: No such file or directory
compilation terminated.
```

To fix this, [besides just doing `find_library()` call]((https://github.com/qTox/qTox/blob/5d159e0203bdd0e72153cce9ca5c0fe8483ba0db/cmake/Dependencies.cmake#L62)), it should be doing something like

```
find_path(${pkg}_INCLUDE_DIRS NAMES filter_audio.h)
```

You'd want to either pass `filter_audio.h` as an extra argument to `search_dependency()`, or write a [`FindFilteraudio.cmake` module](https://cmake.org/Wiki/CMake:How_To_Find_Libraries) and use `find_package(filteraudio)`.

nurupo:
Fixes #4587

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4666)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -538,7 +538,7 @@ if(${ENABLE_GTK_SYSTRAY})
 endif()
 
 if(${USE_FILTERAUDIO})
-    search_dependency(FILTERAUDIO LIBRARY filteraudio OPTIONAL)
+    search_dependency(FILTERAUDIO LIBRARY filteraudio HEADER filter_audio.h OPTIONAL)
     if(${FILTERAUDIO_FOUND})
         set(${PROJECT_NAME}_SOURCES ${${PROJECT_NAME}_SOURCES}
             src/audio/backend/openal2.cpp
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -35,7 +35,7 @@ include(Qt5CorePatches)
 
 function(search_dependency pkg)
   set(options OPTIONAL)
-  set(oneValueArgs PACKAGE LIBRARY FRAMEWORK)
+  set(oneValueArgs PACKAGE LIBRARY FRAMEWORK HEADER)
   set(multiValueArgs)
   cmake_parse_arguments(arg "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
@@ -60,7 +60,8 @@ function(search_dependency pkg)
   # Last, search for the library itself globally.
   if(NOT ${pkg}_FOUND AND arg_LIBRARY)
     find_library(${pkg}_LIBRARIES NAMES ${arg_LIBRARY})
-    if(${pkg}_LIBRARIES)
+    find_path(${pkg}_INCLUDE_DIRS NAMES ${arg_HEADER})
+    if(${pkg}_LIBRARIES AND ${pkg}_INCLUDE_DIRS)
       set(${pkg}_FOUND TRUE)
     endif()
   endif()


sudden6:
Is there a reason not to reuse the whole cmake find module?

nurupo:
Yes there is. The `search_dependency()` function aims to be an universal find module, it alredy does all the things a generic find module does -- looking up a library in pkg-config and then falling back to `find_library()` , except that whoever wrote it seem to have forgotten to also call `find_path()` to find the include directory, or they just assumed that not looking for a custom include directory was ok, i.e. they assumed that the headers are always system-wide installed, which turned out to not always being the case, thus this PR. Also, `search_dependency()` is used by qTox to find many different libraries already, so using one method of finding libraries (`search_dependency()`) in one cases and using another (`find_package()` with find module) in other cases doesn't look uniform, it looks more like two different people with different approaches wrote two parts of a software and glued their work together so that it would somehow work. Not to mention that `search_dependency()` already does most things a `find_package()` approach would do, so there will be unnecessary code duplication, and when adding `find_package()` you'd also need to replicate the printing of library information (libraries, cflags, include dirs, library dir) to stdout, so code duplication again.

I actually did reuse the whole cmake find module at first, but I found that to be a poor fix to the problem (due to the above).

nurupo:
Btw, I think I can remove the `set()` call, it's there purely because we find just one include dir, while the variable we need to set is named  `_INCLUDE_DIRS`, i.e. plural. Now that I think about it, there is probably no significance to it being plural, it must be just a naming convention, I can't think of cases when a library would actually have multiple include dirs, and if there is a significance to it, the code I have added for searching an include dir finds only one anyway, so might as well set `_INCLUDE_DIRS` directly, rather than doing `_INCLUDE_DIR` -> `_INCLUDE_DIRS` hop.

nurupo:
Should be good to merge.

sudden6:
Ok, the reasoning makes sense. :lgtm_strong: 

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4666#-:-KuY_90BqqWQ5SDUD-Xq:bkbo6r1)*

PR #4671, opened 2017-09-21
fix(settings): Delete dynamically allocated members on destruction

### LINKED ISSUE #4670: Error log on exit: "Problem creating accessible interface for txtChatFont"
Commit hash:  b7921ef2

Reproducible: Always

##### Steps to reproduce

1. Open qTox
2. Close qTox

##### Observed Behavior
```
[18:30:23.786 UTC] main.cpp:65 : Debug: Cleanup success
[18:30:23.786 UTC] main.cpp:341 : Debug: Exit with status 0
[18:30:23.805 UTC] :0 : Warning: Problem creating accessible interface for:  QFontComboBox(0xea534599d0, name = "txtChatFont") 
Make sure to deploy Qt with accessibility plugins.
```

##### Expected Behavior
No error.

anthonybilinski:
Fixes #4670

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4671)

### DIFF
--- a/src/widget/form/settingswidget.cpp
+++ b/src/widget/form/settingswidget.cpp
@@ -19,10 +19,6 @@
 
 #include "settingswidget.h"
 
-#include <QLabel>
-#include <QTabWidget>
-#include <QWindow>
-
 #include "src/video/camerasource.h"
 #include "src/widget/contentlayout.h"
 #include "src/widget/form/settings/aboutform.h"
@@ -34,6 +30,12 @@
 #include "src/widget/translator.h"
 #include "src/widget/widget.h"
 
+#include <QLabel>
+#include <QTabWidget>
+#include <QWindow>
+
+#include <memory>
+
 SettingsWidget::SettingsWidget(QWidget* parent)
     : QWidget(parent, Qt::Window)
 {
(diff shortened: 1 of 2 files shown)


Diadlo:
AFAIK we haven't strict rule about using Qt class instead of std one, but we prefer Qt one. What do you think about using `QPointer` instead of `std::unique_ptr`?

anthonybilinski:
In general, it seems like std is more widely tested, has more people familiar with it, more online resources etc. If the project prefers Qt over std where equivalent, I'll switch to that.

For unique_ptr in particular, I found this: https://stackoverflow.com/questions/25786778/c-qt-wheres-the-stdunique-ptr-qt-version.
Basically unique_ptr has move semantics, which QScopedPtr doesn't.

Diadlo:
:lgtm_strong:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4671#-:-KuaFNaWR3FmcxxuIpNA:b-jbsf8z)*

sudden6:
:lgtm_strong: maybe we should prefer `std::*_ptr` types over Qt ones in general? Would have to extend the coding style.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4671#-:-Kuae0iwUWZ8SIM8WvY7:b-7l099p)*

Talkless:
I wonder, why QObject parent-child ownership "does not work" in this case?

Difference is that these forms are now destroyed in `~SettingsWidget()`, not it `~QObject()`.

And most forms doesn't even have "this" as parent.

And that `bodyLayout`, how does it even work? It does not have a parent widget? Maybe that's the core issue?

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4671#-:-KueSYKznnsg7WJ0Qb0C:b-v1wrpa)*

anthonybilinski:
Those are some good questions. I'm a noob at Qt (only since starting on this project), so the whole parent/child ownership model is new to me. I (or someone more familiar with QObject's?) can look  into this and see if there's a more Qt way to destruct safely.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4671#-:-Kuek_VN-VuPajB-FzRZ:b-k2f6cu)*

sudden6:
After reading https://doc.qt.io/qt-5/qobject.html#QObject I wonder why `UserInterfaceForm` is not correctly destructed, even though it has `this` as parent object and should be destructed from `SettingsWidget` destructor.

Talkless:
Maybe it's parent is changed, for example it's that layout, which does not have a parent itself?

PR #4736, opened 2017-10-14
feat(travis): Windows cross-compilation

nurupo:
~~~This is a Work In Progress (WIP) PR.~~~ It adds Windows cross-compilation build(s?) to Travis, since Windows builds frequently get broken without anyone realizing it. ~~~I will be messing with Travis and modifying scripts until I get Travis stages and the cross-compilation to work, so no point in reviewing before I'm done.~~~

**EDIT:** the PR is ready.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/qtox/qtox/4736)
<!-- Reviewable:end -->

### DIFF
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -66,7 +66,7 @@ then
   exit 1
 fi
 
-if [ "$(id -u)" != "0" ]
+if [[ "$(id -u)" != "0" ]]
 then
   echo "Error: This script must be run as root."
   exit 1
@@ -140,6 +140,17 @@ then
                     g++-mingw-w64-x86-64 \
                     gcc-mingw-w64-x86-64
 fi
+
+
+# Install wine to run qTox tests in
+set +u
+if [[ "$TRAVIS_CI_STAGE" == "stage3" ]]
+then
+  dpkg --add-architecture i386
+  apt-get update
+  apt-get install -y wine wine32 wine64
+fi
+set -u
 
 
 # Create the expected directory structure
@@ -275,7 +286,7 @@ then
   # which happens when building Qt
   CONFIGURE_EXTRA=""
   set +u
-  if [ "$TRAVIS_CI_STAGE" == "stage1" ]
+  if [[ "$TRAVIS_CI_STAGE" == "stage1" ]]
   then
     CONFIGURE_EXTRA="-silent"
   fi
@@ -355,7 +366,7 @@ fi
 
 # Stop here if running the first stage on Travis CI
 set +u
-if [ "$TRAVIS_CI_STAGE" == "stage1" ]
+if [[ "$TRAVIS_CI_STAGE" == "stage1" ]]
 then
   # Strip to reduce cache size
   strip_all
@@ -957,7 +968,7 @@ fi
 
 # Stop here if running the second stage on Travis CI
 set +u
-if [ "$TRAVIS_CI_STAGE" == "stage2" ]
+if [[ "$TRAVIS_CI_STAGE" == "stage2" ]]
 then
   # Strip to reduce cache size
   strip_all
@@ -1008,6 +1019,13 @@ echo "
     SET(CMAKE_FIND_ROOT_PATH /usr/$ARCH-w64-mingw32 $CMAKE_FIND_ROOT_PATH)
 " > toolchain.cmake
 
+set +u
+if [[ "$TRAVIS_CI_STAGE" == "stage3" ]]
+then
+  echo "SET(TEST_CROSSCOMPILING_EMULATOR /usr/bin/wine)" >> toolchain.cmake
+fi
+set -u
+
 if [[ "$BUILD_TYPE" == "release" ]]
 then
   cmake -DCMAKE_TOOLCHAIN_FILE=./toolchain.cmake \
@@ -1041,6 +1059,25 @@ cp $OPENSSL_PREFIX_DIR/bin/ssleay32.dll \
 cp /usr/lib/gcc/$ARCH-w64-mingw32/*-posix/libgcc_s_*.dll $QTOX_PREFIX_DIR
 cp /usr/lib/gcc/$ARCH-w64-mingw32/*-posix/libstdc++-6.dll $QTOX_PREFIX_DIR
 cp /usr/$ARCH-w64-mingw32/lib/libwinpthread-1.dll $QTOX_PREFIX_DIR
+
+set +u
+if [[ "$TRAVIS_CI_STAGE" == "stage3" ]]
+then
+  # Setup wine
+  if [[ "$ARCH" == "i686" ]]
+  then
+    export WINEARCH=win32
+  elif [[ "$ARCH" == "x86_64" ]]
+  then
+    export WINEARCH=win64
+  fi
+  winecfg
+  # Add libgcc_s_*.dll, libwinpthread-1.dll, QtTest.dll, etc. into PATH env var of wine
+  export WINEPATH=`cd $QTOX_PREFIX_DIR ; winepath -w $(pwd)`\;`winepath -w $QT_PREFIX_DIR/bin/`
+  export CTEST_OUTPUT_ON_FAILURE=1
+  make test
+fi
+set -u
 
 cd ..
 
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -177,6 +177,20 @@ check_sha256()
   fi
 }
 
+# If it's not a tarball but a git repo, let's check a hash of a file containing hashes of all files
+check_sha256_git()
+{
+  # There shoudl be .git directory
+  if [ ! -d ".git" ]
+  then
+    echo "Error: this function should be called in the root of a git repository."
+    exit 1
+  fi
+  # Create a file listing hashes of all the files except .git/*
+  find . -type f | grep -v "^./.git" | LC_COLLATE=C sort --stable --ignore-case | xargs sha256sum > /tmp/hashes.sha
+  check_sha256 "$1" "/tmp/hashes.sha"
+}
+
 # Strip binaries to reduce file size, we don't need this information anyway
 strip_all()
 {
@@ -356,17 +370,17 @@ set -u
 
 SQLCIPHER_PREFIX_DIR="$DEP_DIR/libsqlcipher"
 SQLCIPHER_VERSION=v3.4.1
+SQLCIPHER_HASH="4172cc6e5a79d36e178d36bd5cc467a938e08368952659bcd95eccbaf0fa4ad4"
 if [ ! -f "$SQLCIPHER_PREFIX_DIR/done" ]
 then
   rm -rf "$SQLCIPHER_PREFIX_DIR"
   mkdir -p "$SQLCIPHER_PREFIX_DIR"
 
-  git clone \
-    --branch $SQLCIPHER_VERSION \
-    --depth 1 \
-    https://github.com/sqlcipher/sqlcipher \
-    sqlcipher
-  cd sqlcipher
+  wget https://github.com/sqlcipher/sqlcipher/archive/$SQLCIPHER_VERSION.tar.gz -O sqlcipher.tar.gz
+  check_sha256 "$SQLCIPHER_HASH" "sqlcipher.tar.gz"
+  bsdtar -xf sqlcipher.tar.gz
+  rm sqlcipher.tar.gz
+  cd sqlcipher*
 
   sed -i s/'LIBS="-lcrypto  $LIBS"'/'LIBS="-lcrypto -lgdi32  $LIBS"'/g configure
   sed -i s/'LIBS="-lcrypto $LIBS"'/'LIBS="-lcrypto -lgdi32  $LIBS"'/g configure
@@ -405,7 +419,7 @@ EOF
   echo -n $SQLCIPHER_VERSION > $SQLCIPHER_PREFIX_DIR/done
 
   cd ..
-  rm -rf ./sqlcipher
+  rm -rf ./sqlcipher*
 else
   echo "Using cached build of SQLCipher `cat $SQLCIPHER_PREFIX_DIR/done`"
 fi
@@ -499,6 +513,7 @@ fi
 
 OPENAL_PREFIX_DIR="$DEP_DIR/libopenal"
 OPENAL_VERSION=b80570bed017de60b67c6452264c634085c3b148
+OPENAL_HASH="734ef00895a9c1eb1505c11d638030b73593376df75da66ac5db6aa3e2f76807"
 if [ ! -f "$OPENAL_PREFIX_DIR/done" ]
 then
   rm -rf "$OPENAL_PREFIX_DIR"
@@ -507,6 +522,10 @@ then
   git clone https://github.com/irungentoo/openal-soft-tox openal-soft-tox
   cd openal*
   git checkout $OPENAL_VERSION
+  check_sha256_git "$OPENAL_HASH"
+
+  mkdir -p build
+  cd build
 
   echo "
       SET(CMAKE_SYSTEM_NAME Windows)
@@ -530,11 +549,14 @@ then
     -DALSOFT_EXAMPLES=OFF \
     -DCMAKE_TOOLCHAIN_FILE=toolchain.cmake \
     -DDSOUND_INCLUDE_DIR=/usr/$ARCH-w64-mingw32/include \
-    -DDSOUND_LIBRARY=/usr/$ARCH-w64-mingw32/lib/libdsound.a
+    -DDSOUND_LIBRARY=/usr/$ARCH-w64-mingw32/lib/libdsound.a \
+    ..
 
   make
   make install
   echo -n $OPENAL_VERSION > $OPENAL_PREFIX_DIR/done
+
+  cd ..
 
   cd ..
   rm -rf ./openal*
@@ -547,6 +569,7 @@ fi
 
 FILTERAUDIO_PREFIX_DIR="$DEP_DIR/libfilteraudio"
 FILTERAUDIO_VERSION=ada2f4fdc04940cdeee47caffe43add4fa017096
+FILTERAUDIO_HASH="cf481e87c860aaf28b50d125195d84556f36d0ebb529d7ebdac39f8cc497256a"
 if [ ! -f "$FILTERAUDIO_PREFIX_DIR/done" ]
 then
   rm -rf "$FILTERAUDIO_PREFIX_DIR"
@@ -555,6 +578,7 @@ then
   git clone https://github.com/irungentoo/filter_audio filter_audio
   cd filter*
   git checkout $FILTERAUDIO_VERSION
+  check_sha256_git "$FILTERAUDIO_HASH"
 
   $ARCH-w64-mingw32-gcc -O2 -g0 -c \
                aec/aec_core.c \
@@ -741,20 +765,19 @@ fi
 # Opus
 
 OPUS_PREFIX_DIR="$DEP_DIR/libopus"
-OPUS_VERSION=v1.2.1
+OPUS_VERSION=1.2.1
+OPUS_HASH="cfafd339ccd9c5ef8d6ab15d7e1a412c054bf4cb4ecbbbcc78c12ef2def70732"
 if [ ! -f "$OPUS_PREFIX_DIR/done" ]
 then
   rm -rf "$OPUS_PREFIX_DIR"
   mkdir -p "$OPUS_PREFIX_DIR"
 
-  git clone \
-    --branch $OPUS_VERSION \
-    --depth 1 \
-    git://git.opus-codec.org/opus.git \
-    opus
-  cd opus
+  wget https://archive.mozilla.org/pub/opus/opus-$OPUS_VERSION.tar.gz
+  check_sha256 "$OPUS_HASH" "opus-$OPUS_VERSION.tar.gz"
+  bsdtar -xf opus*.tar.gz
+  rm opus*.tar.gz
+  cd opus*
 
-  ./autogen.sh
   CFLAGS="-O2 -g0" ./configure --host="$ARCH-w64-mingw32" \
                                --prefix="$OPUS_PREFIX_DIR" \
                                --disable-shared \
@@ -766,7 +789,7 @@ then
   echo -n $OPUS_VERSION > $OPUS_PREFIX_DIR/done
 
   cd ..
-  rm -rf ./opus
+  rm -rf ./opus*
 else
   echo "Using cached build of Opus `cat $OPUS_PREFIX_DIR/done`"
 fi
@@ -776,19 +799,18 @@ fi
 
 SODIUM_PREFIX_DIR="$DEP_DIR/libsodium"
 SODIUM_VERSION=1.0.15
+SODIUM_HASH="fb6a9e879a2f674592e4328c5d9f79f082405ee4bb05cb6e679b90afe9e178f4"
 if [ ! -f "$SODIUM_PREFIX_DIR/done" ]
 then
   rm -rf "$SODIUM_PREFIX_DIR"
   mkdir -p "$SODIUM_PREFIX_DIR"
 
-  git clone \
-    --branch $SODIUM_VERSION \
-    --depth 1 \
-    https://github.com/jedisct1/libsodium \
-    libsodium
-  cd libsodium
+  wget https://download.libsodium.org/libsodium/releases/libsodium-$SODIUM_VERSION.tar.gz
+  check_sha256 "$SODIUM_HASH" "libsodium-$SODIUM_VERSION.tar.gz"
+  bsdtar -xf libsodium*.tar.gz
+  rm libsodium*.tar.gz
+  cd libsodium*
 
-  ./autogen.sh
   ./configure --host="$ARCH-w64-mingw32" \
               --prefix="$SODIUM_PREFIX_DIR" \
               --disable-shared \
@@ -799,7 +821,7 @@ then
   echo -n $SODIUM_VERSION > $SODIUM_PREFIX_DIR/done
 
   cd ..
-  rm -rf ./libsodium
+  rm -rf ./libsodium*
 else
   echo "Using cached build of Sodium `cat $SODIUM_PREFIX_DIR/done`"
 fi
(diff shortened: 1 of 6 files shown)


Talkless:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion.

---

*[.travis/build-windows.sh, line 32 at r1](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwQsOM3_mg3aIkwMCoX:-KwQsOM3_mg3aIkwMCoY:b55ko09) ([raw file](https://github.com/qtox/qtox/blob/944a71e4b16eb935595ba0d225046d28ed30cf5e/.travis/build-windows.sh#L32)):*
> ```Shell
> then
>   # Clear the cache, removing all the pre-built dependencies
>   rm -rf $DEP_CACHE/*
> ```

I'd suggest to quote every variable usage to avoid problems with possible spaces in paths.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

windows/cross-compile/build.sh
```diff
@@ -211,30 +223,31 @@ then
...
   mkdir -p "$QT_PREFIX_DIR"
 
   QT_MIRROR=http://qt.mirror.constant.com
 
-  QT_MAJOR=5
-  QT_MINOR=6
-  QT_PATCH=3
-
-  QT_VERSION=$QT_MAJOR.$QT_MINOR.$QT_PATCH
   wget $QT_MIRROR/official_releases/qt/$QT_MAJOR.$QT_MINOR/$QT_VERSION/single/qt-everywhere-opensource-src-$QT_VERSION.tar.xz
```
sudden6, line 272: maybe add `--no-verbose` to produce less output?

nurupo: Nah, it's fine as it is, it doesn't hurt anything, not to mention that it's even beneficial to have a verbose output -- pretty sure whoever corss-compiles qTox on a slow internet would want to know the progress.

sudden6: this still displays errors, but hides the progress, which isn't that usefull imo

sudden6: e.g. in this job https://travis-ci.org/qTox/qTox/jobs/288231919 the download creates ~3k lines only for the progress :/

nurupo: This script is not only for Travis, but also for users and it's useful for them to see the progress. 3k for wget lines is a lot, I'm not sure why it does that on Travis, it doesn't print that many lines locally, but removing those ~3k lines won't help with anything, the entire build log is like 20k lines anyway, it doesn't even fit on Travis page, you have to access the log file directly. What would you gain by removing those 3k lines from Travis log and making users unable to track download progress?

nurupo:
Weird, I can't reproduce any build failures I'm having on Jenkins. It builds just fine on Travis without https://github.com/qTox/qTox/commit/2f1bf107808953f5f080de3458f88abb33ad247a and https://github.com/qTox/qTox/commit/d69023c9d78714f5e730e98bb251c8d0f723688a commits.

I'm actually not sure about what is going on with `posixsignalnotifier` test at all. My PR branch doesn't have the https://github.com/qTox/qTox/commit/d69023c9d78714f5e730e98bb251c8d0f723688a commit yet, i.e. [there is no `if()` around `posixsignalnotifier` test in `cmake/Testing.cmake`](https://github.com/nurupo/qTox/blob/travis-windows/cmake/Testing.cmake), it should be build unconditionally with all other tests, yet `posixsignalnotifier` test is not even mentioned in build logs, the last test that is mentioned is `toxmedata`, I don't understand why it skips `posixsignalnotifier`.

nurupo:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[.travis/build-windows.sh, line 32 at r1](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwQsOM3_mg3aIkwMCoX:-Kwf_LX9e1X0Invv_HXh:b-c7n2mi) ([raw file](https://github.com/qtox/qtox/blob/944a71e4b16eb935595ba0d225046d28ed30cf5e/.travis/build-windows.sh#L32)):*

I'd suggest to quote every variable usage to avoid problems with possible spaces in paths.
</blockquote></details>

Nah, it's fine. `CACHE_DIR` is set to either `/opt/build-windows/i686` or `/opt/build-windows/x86_64`, there are no spaces in them. There might be spaces in `$PWD`. so I quoted it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

nurupo:
Should be good to go. A few things of note:

- Travis fails due to qTox build generating an warning and then treating it as an error.
- See https://github.com/qTox/qTox/pull/4736#issuecomment-336752910.
- You can notice that Windows jobs are inconsistent with all other jobs in `.travis.yml`. Windows jobs don't define `env: JOB=...` and they have their own `script:` and they pass arguments to that `script:` which otherwise be specified in `env:`. This inconsistency is required for Windows jobs to share the cache. [In Travis, jobs can share cache only if they use exactly the same `env:`, os, language, etc.](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices) I actually used this to force cache separation between `i686` and `x86_64` jobs by defining `env: CACHE_TRICK_VARIABLE` variae with value different for both; without this separation the caches of one arch were overwriting the cache of another arch.
- Cache is invalidated when `build.sh` is modified, which is the behavior you want. For example, if you open a PR that bumps versions in `build.sh`, you don't need to do any extra work to clear the cache, Travis will detect that `build.sh` was modified, wipe the cache and rebuild everything from scratch, caching the results.
- I have changed the distro from Ubuntu 16.04 to Debian Stretch because when I bumped Qt from 5.6.3 to 5.9.2 it was failing to build using Ubuntu 16.04's MinGW (5.3.1) but worked with Debian Stretch's MinGW (6.3.0).

Talkless:
Review status: 0 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[.travis/build-windows.sh, line 32 at r1](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwQsOM3_mg3aIkwMCoX:-Kwk5GwDaZIO8Nnxj5dD:b-f4k0zx) ([raw file](https://github.com/qtox/qtox/blob/944a71e4b16eb935595ba0d225046d28ed30cf5e/.travis/build-windows.sh#L32)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Nah, it's fine. `CACHE_DIR` is set to either `/opt/build-windows/i686` or `/opt/build-windows/x86_64`, there are no spaces in them. There might be spaces in `$PWD`. so I quoted it.
</blockquote></details>

I believe it would be _good practice_ to quote, well, everything. To get a "reflex" for any other future shell code. Deciding case-by-case basis might be too bug prone, and dangerous.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

sudden6:
@nurupo can you check if  #4744 fixes the build error?

nurupo:
@sudden6 sure, let me cherry-pick your commit into this PR and see what Travis says. Btw, you could as easily cherry-picked my commit into your PR and wouldn't have to wait on me to do so.

nurupo:
Is anyone looking into https://github.com/qTox/qTox/pull/4736#issuecomment-336752910 ? The `posixsignalnotifier` test is a mystery to me.

sudden6:
@nurupo sorry, didn't think to cherry-pick it into my PR...

About the posixsignalnotifier, I have no idea why it isn't compiled. On the other hand it doesn't really make sense to build any of the tests, since we can't run them anyway (without using things like wine).

sudden6:
Reviewed 2 of 5 files at r3.
Review status: 2 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[.travis/build-windows.sh, line 32 at r1](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwQsOM3_mg3aIkwMCoX:-KwoDfkrtKm4KmFdEME5:b5iruwu) ([raw file](https://github.com/qtox/qtox/blob/944a71e4b16eb935595ba0d225046d28ed30cf5e/.travis/build-windows.sh#L32)):*

I believe it would be _good practice_ to quote, well, everything. To get a "reflex" for any other future shell code. Deciding case-by-case basis might be too bug prone, and dangerous.
</blockquote></details>

I agree with @Talkless, we may as well do it right from the beginning

---

*[windows/cross-compile/build.sh, line 251 at r2](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwPlUhvmKBK5vSjU-Pb-r2-251:-KwoOcxfqYLfAsUSJV4J:b-viwoid) ([raw file](https://github.com/qtox/qtox/blob/bdd43cb8c5a9075340e0bf56ea20bfe45e25c75e/windows/cross-compile/build.sh#L251)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

This script is not only for Travis, but also for users and it's useful for them to see the progress. 3k for wget lines is a lot, I'm not sure why it does that on Travis, it doesn't print that many lines locally, but removing those ~3k lines won't help with anything, the entire build log is like 20k lines anyway, it doesn't even fit on Travis page, you have to access the log file directly. What would you gain by removing those 3k lines from Travis log and making users unable to track download progress?
</blockquote></details>

Maybe only disable it on Travis then? I'll leave that up to you to decide though.

---

*[windows/cross-compile/build.sh, line 203 at r3](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwoP26TsqU0e8IQ62Rg:-KwoP26TsqU0e8IQ62Rh:bq1gvxu) ([raw file](https://github.com/qtox/qtox/blob/5010c21823579de7aa44bcc77072d3ded0d089c5/windows/cross-compile/build.sh#L203)):*
> ```Shell
> 
>   wget https://www.openssl.org/source/openssl-$OPENSSL_VERSION.tar.gz
>   check_sha256 "ce07195b659e75f4e1db43552860070061f156a98bb37b672b101ba6e3ddf30c" "openssl-$OPENSSL_VERSION.tar.gz"
> ```

Maybe move the hash into a variable and put it to the others at the beginning? groups all variables needed for an update together.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

sudden6:
The fix was merged, this should be rebased.

nurupo:
Ok, I will rebase. I delayed the rebase in hope that someone would figure out why of all tests `posixsignalnotifier` test wasn't even being mentioned by cmake and make, since I didn't pull the commit that added the `if()` condition around it, i.e. all tests are unconditional

```CMake
auto_test(core toxpk)
auto_test(core toxid)
auto_test(chatlog textformatter)
auto_test(net toxmedata)
auto_test(platform posixsignalnotifier)
```

Oh well, I guess the mystery will go unsolved.

nurupo:
Huh, weird. When I run the build in `debian:stretch-slim` container locally, it fails with

```
[ 40%] Building CXX object CMakeFiles/qtox_static.dir/src/net/toxmedata.cpp.obj
/build/qtox/src/net/toxmedata.cpp: In member function 'QString ToxmeData::createAddressRequest(ToxId, const QString&, const QString&, bool) const':
/build/qtox/src/net/toxmedata.cpp:123:63: error: 'time' was not declared in this scope
            "\",\"timestamp\":" + QString().setNum(time(nullptr)) +
                                                               ^
/build/qtox/src/net/toxmedata.cpp: In member function 'QString ToxmeData::deleteAddressRequest(const ToxPk&)':
/build/qtox/src/net/toxmedata.cpp:161:63: error: 'time' was not declared in this scope
            "\",\"timestamp\":" + QString().setNum(time(nullptr)) +
                                                               ^
CMakeFiles/qtox_static.dir/build.make:1267: recipe for target 'CMakeFiles/qtox_static.dir/src/net/toxmedata.cpp.obj' failed
make[2]: *** [CMakeFiles/qtox_static.dir/src/net/toxmedata.cpp.obj] Error 1
make[2]: *** Waiting for unfinished jobs....
CMakeFiles/Makefile2:219: recipe for target 'CMakeFiles/qtox_static.dir/all' failed
make[1]: *** [CMakeFiles/qtox_static.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
```

yet it doesn't on Travis. I'm confused as why. We use the same container that runs the same `build.sh` script. The only differences between my and Travis setup might be the Docker and Linux kernel versions.

nurupo:
Haven't rebased yet, want to look a bit deeper into it. It's as if Travis runs this script on qTox master where all Windows errors are fixed instead of my `travis-windows` branch.

nurupo:
I see. Turns out Travis merges my PR branch into master and then testes that. I always thought Travis was testing only my PR branch alone. In my PR branch there are code issues which should have failed the compilation, but they were fixed in master, and I couldn't figure out why in the world Travis wasn't failing. It acted as if it had the fixes from the master branch merged into my branch. Turns out this was indeed the case. Not sure if this is default Travis behavior or something that qTox repo has configured.

nurupo:
Well, that solves the `posixsignalnotifier` test mystery. After the merge there was an `if(UNIX)` around it, so no wonder cmake and make never mentioned it.

nurupo:
Review status: 2 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[.travis/build-windows.sh, line 32 at r1](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwQsOM3_mg3aIkwMCoX:-Kwwq-XtIpGD3D-esVJO:b-11bl1i) ([raw file](https://github.com/qtox/qtox/blob/944a71e4b16eb935595ba0d225046d28ed30cf5e/.travis/build-windows.sh#L32)):*

I agree with @Talkless, we may as well do it right from the beginning
</blockquote></details>

Ok, quoted every single variable, path or not path.

---

*[windows/cross-compile/build.sh, line 251 at r2](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwPlUhvmKBK5vSjU-Pb-r2-251:-KwwqVv_7KrUT2Gjc4Rg:b1a18eu) ([raw file](https://github.com/qtox/qtox/blob/bdd43cb8c5a9075340e0bf56ea20bfe45e25c75e/windows/cross-compile/build.sh#L251)):*

Maybe only disable it on Travis then? I'll leave that up to you to decide though.
</blockquote></details>

I think it's fine as it is.

---

*[windows/cross-compile/build.sh, line 203 at r3](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwoP26TsqU0e8IQ62Rg:-Kwwq9HHfAymB663lmIy:bfwrr91) ([raw file](https://github.com/qtox/qtox/blob/5010c21823579de7aa44bcc77072d3ded0d089c5/windows/cross-compile/build.sh#L203)):*

Maybe move the hash into a variable and put it to the others at the beginning? groups all variables needed for an update together.
</blockquote></details>

Done. Note that not all deps have hash. We get some of the deps through git, so there is really nothing to act as hash, aside from commit sha1, but we also use tag names in a few places.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

sudden6:
:lgtm_strong: @Diadlo can you also take a look?

---

Reviewed 3 of 3 files at r4.

---

*[windows/cross-compile/build.sh, line 203 at r3](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwoP26TsqU0e8IQ62Rg:-Kwy53kNG2TgVSdNexXM:b-rbjuga) ([raw file](https://github.com/qtox/qtox/blob/5010c21823579de7aa44bcc77072d3ded0d089c5/windows/cross-compile/build.sh#L203)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Done. Note that not all deps have hash. We get some of the deps through git, so there is really nothing to act as hash, aside from commit sha1, but we also use tag names in a few places.
</blockquote></details>

Ok this sounds good

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736#-:-Kwy6fFA82b-ejOZFNAT:bbzeprc)*

nurupo:
Btw, if I remember right, only OpenSSL and Qt5 provide sha256 hashes, I had to download all other tarballs and calculate it locally.

Diadlo:
Reviewed 1 of 5 files at r3, 1 of 3 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[.travis.yml, line 32 at r4](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwyXFKC749-hB-v6rPo:-KwyXFKC749-hB-v6rPp:b675wis) ([raw file](https://github.com/qtox/qtox/blob/062039f9e20afdb1d438448e366b65cb3f6f0bd3/.travis.yml#L32)):*
> ```YAML
>       os: linux
>       env: JOB=build-ubuntu-14-04
>     - stage: "Windows Stage 1: Dependencies (OpenSSL, Qt)"
> ```

What about adding architecure in stage name?
This will allow, when opening travis immediately understand which architecture have an error

---

*[windows/cross-compile/build.sh, line 190 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1l14fgTbl7RT7kYpC:-Kx1l14fgTbl7RT7kYpD:b-yi3tif) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L190)):*
> ```Shell
>   fi
>   # Create a file listing hashes of all the files except .git/*
>   find . -type f | grep -v "^./.git" | LC_COLLATE=C sort --stable --ignore-case | xargs sha256sum > /tmp/hashes.sha
> ```

Why not check the commit hash?

---

*[windows/cross-compile/build.sh, line 218 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1lT2PINqdgc0VF2yk:-Kx1lT2PINqdgc0VF2yl:b-9mzoox) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L218)):*
> ```Shell
>   mkdir -p "$OPENSSL_PREFIX_DIR"
> 
>   wget https://www.openssl.org/source/openssl-$OPENSSL_VERSION.tar.gz
> ```

What about creating function?
```
url=$1
hash=$2
wget $file
filename=`basename $url`
check_sha256 "$hash" "filename"

bsdtar -xf $filename
rm $filename
```

Or even whole "if [ -f done ] .... else fi" extract to a function with callbacks to build (not sure about this)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

nurupo:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

*[.travis.yml, line 32 at r4](https://reviewable.io:443/reviews/qtox/qtox/4736#-KwyXFKC749-hB-v6rPo:-Kx1nSTleb4TaTalKP3U:bsnyj5z) ([raw file](https://github.com/qtox/qtox/blob/062039f9e20afdb1d438448e366b65cb3f6f0bd3/.travis.yml#L32)):*

What about adding architecure in stage name?
This will allow, when opening travis immediately understand which architecture have an error
</blockquote></details>

Stages build all architectures though. Stage 1 builds both i686 and x86_64, same for stage 2 and stage 3.

---

*[windows/cross-compile/build.sh, line 190 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1l14fgTbl7RT7kYpC:-Kx1nf7ZiDwIuD3qbSP_:b-v2h0e6) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L190)):*

Why not check the commit hash?
</blockquote></details>

Something something SHA1 vulnerable something.

---

*[windows/cross-compile/build.sh, line 218 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1lT2PINqdgc0VF2yk:-Kx1npiKdC-lua3uJBYR:bx38pef) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L218)):*

What about creating function?
```
url=$1
hash=$2
wget $file
filename=`basename $url`
check_sha256 "$hash" "filename"

bsdtar -xf $filename
rm $filename
```

Or even whole "if [ -f done ] .... else fi" extract to a function with callbacks to build (not sure about this)
</blockquote></details>

I was thinking about it and thought to leave it as it is, since some deps are wget'ed, some deps are git cloned and other were shallow git cloned on specific branch -- you'd have to write 3 separate functions for them.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

Diadlo:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[windows/cross-compile/build.sh, line 218 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1lT2PINqdgc0VF2yk:-Kx1p7drGf7qvcE5pkz7:b-jrm956) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L218)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

I was thinking about it and thought to leave it as it is, since some deps are wget'ed, some deps are git cloned and other were shallow git cloned on specific branch -- you'd have to write 3 separate functions for them.
</blockquote></details>

Here, at least 10 `wget` call :P
And just 3 git (which can be extracted in one function with not necessary branch parameter)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

nurupo:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[windows/cross-compile/build.sh, line 218 at r5](https://reviewable.io:443/reviews/qtox/qtox/4736#-Kx1lT2PINqdgc0VF2yk:-Kx1plSRf57iTMoHiLDj:bpnzbsk) ([raw file](https://github.com/qtox/qtox/blob/33b978fc31c784563187fb90853f6650ea268cb6/windows/cross-compile/build.sh#L218)):*

Here, at least 10 `wget` call :P
And just 3 git (which can be extracted in one function with not necessary branch parameter)
</blockquote></details>

Yeah, I replaced all shallow git clones with wgets with my last commit. Anyway, feel free to open a PR that extracts things in functions with callbacks, I'm not doing it in this PR though, I don't even know how.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736)*

nurupo:
Built all the 4 qTox builds and ran them on Windows, all seem to work fine, including gdb debugging.

nurupo:
Btw, you can tell cmake to use wine for running tests by setting [`CMAKE_CROSSCOMPILING_EMULATOR` variable](https://cmake.org/cmake/help/v3.9/prop_tgt/CROSSCOMPILING_EMULATOR.html#prop_tgt:CROSSCOMPILING_EMULATOR) in the toolchain file. Want me to try setting this up?

Diadlo:
:lgtm_strong:

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4736#-:-Kx20Fgv2hPl-Im5oyFF:b-jbsf8z)*

nurupo:
Oh, I forgot to remove some debug stuff, don't merge yet.

nurupo:
Removed the stuff I used for debugging while the build script was failing.

nurupo:
Also, I have tried running the tests in wine, and succeeded, but we can't do this the proper way by setting the [`CMAKE_CROSSCOMPILING_EMULATOR` variable](https://cmake.org/cmake/help/v3.9/prop_tgt/CROSSCOMPILING_EMULATOR.html#prop_tgt:CROSSCOMPILING_EMULATOR), because it breaks Qt5 cmake files as they use `add_custom_command()` internally and prefixing those with `wine` breaks them. So I have added just `CROSSCOMPILING_EMULATOR`, which is not some CMake's variable my mine (I could rename it if you don't want confusion) and prefixed the `add_test()`'s commands with it. Anyway, see the next commit I push.

nurupo:
Also, the `${CHECK_LIBRARIES}` in `cmake/Testing.cmake` is bothering me. Where exactly does it come from? I suspect that it might always be empty. I haven't tried printing out contents of it yet, but I don't see what could be setting it, as there is no dependency on libcheck anywhere, nor it's set manually in some Travis bash script, like i do with `CROSSCOMPILING_EMULATOR`.

Diadlo:
@nurupo Probably I forgot to remove `${CHECK_LIBRARIES}` when migrate to Qt tests. Feel free to remove

Diadlo:
`CROSSCOMPILING_EMULATOR`. Wine is not emulator :smirk:
IMO better use `CROSSCOMPILING_STARTER` or `TEST_STARTER`

nurupo:
>CROSSCOMPILING_EMULATOR. Wine is not emulator

Well, although Wine Is Not an Emulator, it's named after [`CMAKE_CROSSCOMPILING_EMULATOR`](https://cmake.org/cmake/help/v3.9/prop_tgt/CROSSCOMPILING_EMULATOR.html#prop_tgt:CROSSCOMPILING_EMULATOR). I guess I will rename it to `TEST_CROSSCOMPILING_EMULATOR` or something like this, since it's used only in `add_test()`.

nurupo:
If you are building for arm on amd64, you could set it to an appropriate `qemu` command so that it would run your arm tests in an arm qemu vm, for example. If you feel adventurous enough, you could set it to a bash script which `scp`s the test binary to a remote server (i.e. raspberry pi) and then runs it through `ssh`. So it can be set to other things than just `wine`.

nurupo:
Ok, changed `CROSSCOMPILING_EMULATOR` -> `TEST_CROSSCOMPILING_EMULATOR`.

Diadlo:
@nurupo Is it ready to merge now?

nurupo:
Yes, but let's first see if it builds successfully on Travis.

nurupo:
If you want, we could also add `-gdwarf-2` to `CMAKE_CXX_FLAGS` for  `DEBUG and MINGW` configuration. `dwarf-2` is a debugging data format. This will allow wine's debugger to print a readable stack trace and show the line that resulted in a crash when running a test. Without setting this it will print some gibberish stack-trace and probably won't point to the correct line.

nurupo:
Added `-gdwarf-2`. This results in the debug `qtox.exe` being 3mb larger (83mb vs 86mb), but should show more info if a test crashes in wine. This flag is needed primarily for wine, the gdb.exe that the debug qTox build includes can read the debug symbols without `-gdwarf-2` just fine.

nurupo:
The PR is ready to be merged btw.

Diadlo:
@sudden6 Still LGTM?

sudden6:
@Diadlo yes

PR #4738, opened 2017-10-14
Fix displaying own username on detached windows.

Chiitoo:
After the changes in commit 'https://github.com/qTox/qTox/commit/add8d51a2913fa6979674247696da9d397d252b5', the name of the "local" user is empty, and as such, it is not displayed in the window title any longer in multiple windows mode.  This commit adds a change that brings it back.

There might be a better way, or/and a better place for this fix, but this is what I came up with.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4738)

### DIFF
--- a/src/widget/contentdialog.cpp
+++ b/src/widget/contentdialog.cpp
@@ -485,6 +485,8 @@ void ContentDialog::updateTitleAndStatusIcon()
         return;
     }
 
+    username = Core::getInstance()->getUsername();
+
     setWindowTitle(activeChatroomWidget->getTitle() + QStringLiteral(" - ") + username);
 
     bool isGroupchat = activeChatroomWidget->getGroup() != nullptr;


noavarice:

---

*[src/widget/contentdialog.cpp, line 488 at r1](https://reviewable.io:443/reviews/qtox/qtox/4738#-KwVKYw8Rwr_Jw7SbLw4:-KwVKYw8Rwr_Jw7SbLw5:b-ejvlds) ([raw file](https://github.com/qtox/qtox/blob/9b3bb7be3dfe332920dbf57209c7c02824512e6e/src/widget/contentdialog.cpp#L488)):*
> ```cpp
>     }
> 
>     username = Core::getInstance()->getUsername();
> ```

This is not the good idea. `ContentDialog::setUsername(QString newName)`, as it follows from function name, sets user name to `newName`, and then calls `updateTitleAndStatusIcon`, where `username` will be set back to global username. This is either wrong or redundant. ~~I think, you can place this assignment in the global scope, that is, [here](https://github.com/qTox/qTox/blob/master/src/widget/contentdialog.cpp#L47)~~ Most likely in constructor?..

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4738)*

noavarice:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4738)*

Chiitoo:
Indeed, I was/am a bit confused about 'newName', as it looked like it should be doing just this, but seemingly doesn't.

I also played around with the 'QString ContentDialog::username = "";', but setting the name there, like I did in the working version, resulted into a segfault instead.  :]

PR #4783, opened 2017-11-01
windows build with circleCI

zoff99:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4783)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,36 @@
+machine:
+  services:
+    - docker
+  timezone:
+    Europe/Berlin
+  environment:
+    MAKEFLAGS: "j4"
+    BUILD__: "x86_64"
+    BTYPE__: "release"
+dependencies:
+  pre:
+    - sudo apt-get update
+    - sudo apt-get install zip
+    - docker info
+compile:
+  override:
+    - mkdir -p ~/workspace
+    - mkdir -p ~/script
+    - cp -av ./windows/cross-compile/build.sh ~/script/
+
+    - docker run --rm
+      -v /home/ubuntu/workspace:/workspace
+      -v /home/ubuntu/script:/script
+      -v /home/ubuntu/qTox:/qtox
+      debian:stretch-slim
+      /bin/bash /script/build.sh "$BUILD__" "$BTYPE__"
+
+    - ls -al /home/ubuntu/workspace/"$BUILD__"/qtox/"$BTYPE__"/
+    - ls -al /home/ubuntu/workspace/"$BUILD__"/qtox/"$BTYPE__"/qtox.exe
+    
+    - mkdir -p $CIRCLE_ARTIFACTS/"$BUILD__"/
+    - cp -av /home/ubuntu/workspace/"$BUILD__"/qtox/"$BTYPE__"/qtox.exe $CIRCLE_ARTIFACTS/"$BUILD__"/qtox_"$BTYPE__".exe
+    - cd /home/ubuntu/workspace/"$BUILD__"/qtox/"$BTYPE__"/ ; zip -r $CIRCLE_ARTIFACTS/"$BUILD__"/qtox_"$BUILD__"_"$BTYPE__".zip *
+test:
+  override:
+    - echo "no tests yet"
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -893,7 +893,9 @@ then
 
   wget https://github.com/TokTok/c-toxcore/releases/download/v$TOXCORE_VERSION/c-toxcore-$TOXCORE_VERSION.tar.gz
   check_sha256 "$TOXCORE_HASH" "c-toxcore-$TOXCORE_VERSION.tar.gz"
+  set +euo pipefail
   bsdtar -xf c-toxcore*.tar.gz
+  set -euo pipefail
   rm c-toxcore*.tar.gz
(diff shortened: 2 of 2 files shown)


sudden6:
please squash and rename the commits according to https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#commit-message-format

also @Diadlo what do you think about adding CircleCI?

zoff99:
there seems to be a problem with windows cross compile.
when i use this binary it will only send the first video frame, and nothing after that
(to be precise i mean sharing my desktop as video)

who developed this script? did anybody test the resulting binary?

@nurupo can you help please?

nurupo:
Sorry, can't test desktop sharing, maybe someone else can.

I doubt that the video issue is because of something in the build script. We have been using almost exactly this build script to produce qTox binaries on Jenkins for a couple years now, which are the qTox binaries that qTox users use. It's more likely to be an issue with qTox, toxcore or ffmpeg rather than the script. It could also be an issue with your computer setup, I remember you complaining about a similar issue where you were getting 1 fps with the tox cam thing. Anyway, we need to get someone to try to reproduce this issue.

zoff99:
it works with the official qtox binary, but not with the one this cross compile produces.
hope someone can recreate.

nurupo:
Try the nightly cmake binaries, not the release ones.

sudden6:
@zoff99 maybe try with #4796 merged, this could potentially cause some very strange and random errors.

zoff99:
@sudden6 @Diadlo can you merge this?
it does no harm, and i could see if video is fixed when this is used on master.

you would need to login to https://circleci.com/gh/qTox once and enable the build

zoff99:
@sudden6 "Allow edits from maintainers" is checked. can u squash please?

Diadlo:
@zoff99 Rebase. As you can see it's not so easy to rebase such commit history. Please, use rebase instead of merge in the next time.

**ADDED:** To easy squash you can use something like this:
```
suqash() {
  git diff master $1 > /tmp/$1
  git checkout $1
  git reset --hard master
  patch -p1 < /tmp/$1
}

suqash master3
```

nurupo:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[windows/cross-compile/build.sh, line 896 at r1](https://reviewable.io:443/reviews/qtox/qtox/4783#-Kyq_-Roh8phUsMpApT-:-Kyq_-Rp5h8FLpzsLwTW:b-yvyhj3) ([raw file](https://github.com/qtox/qtox/blob/83c9679d96540322e1a3b823632f39105c9063d6/windows/cross-compile/build.sh#L896)):*
> ```Shell
>   wget https://github.com/TokTok/c-toxcore/releases/download/v$TOXCORE_VERSION/c-toxcore-$TOXCORE_VERSION.tar.gz
>   check_sha256 "$TOXCORE_HASH" "c-toxcore-$TOXCORE_VERSION.tar.gz"
>   set +euo pipefail
> ```

Why was this done? This shouldn't be necessary. What was wrong with the tar command? 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4783)*

zoff99:
@nurupo the tar command gives and error because it tries to change uid and gid of the extracted files.
on the version of docker on circleCI (and the btrfs) this is not working.

so the bash options need to be disabled for this command. otherwise everything will stop right there

nurupo:
>@nurupo the tar command gives and error because it tries to change uid and gid of the extracted files.
on the version of docker on circleCI (and the btrfs) this is not working.

Interesting. My guess is that the files inside toxcore tarball are made by a user whose uid and gid are not present on the system, which makes it to fail. Could you paste the error message Circle-CI is outputting?

>so the bash options need to be disabled for this command. otherwise everything will stop right there

There is no need to disable and the re-enable all of the options though, just disabling and re-enabling `e` should be enough.

A better fix would be to add `--no-same-owner` to all `bsdtar` commands, this will make tar create files with current user's uid and gid, which will make sure that the build script doesn't fail if any library decides to change uid and gid in their tarballs.

zoff99:
@nurupo its truncated, but if you leave it running it the browser you see the message

https://circleci.com/gh/zoff99/qTox/31

btw. why are you using "bsdtar" and not normal "tar" command?

sudden6:
@zoff99 there is some obscure bug with `tar` and docker, `bsdtar` isn't affected. There is an issue in somewhere in our bugtracker about it.

nurupo:
>its truncated, but if you leave it running it the browser you see the message

Sorry, I can't find it. Can you just post it here?

zoff99:
guys the PR is complete and works. if you need any changes you can do them.
edits from maintainers is ON

nurupo:
It's your responsobility to edit the PR until it's accepted though. Also, I'm not a maintainer, so I can't edit it.

Anyway, like I said in https://github.com/qTox/qTox/pull/4783#issuecomment-344181550, adding `--no-same-owner` to all `bsdtar` commands is a more future-proof solution than disabling bash flags around the single `bsdtar` command. (Not to mention that disabling some of them doesn't even make sense, do you even know what those flags do?). I'm not sure if `--no-same-owner` would actually fix the error you are getting, since you won't tell what the error is, but based on what you said it sounds like it should fix it.

zoff99:
#4892

PR #4784, opened 2017-11-02
fix(call): fix bugs introduced from ToxCall refactor

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4784)

### DIFF
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -240,7 +240,7 @@ bool CoreAV::answerCall(uint32_t friendNum, bool video)
 
     const uint32_t videoBitrate = video ? VIDEO_DEFAULT_BITRATE : 0;
     if (toxav_answer(toxav, friendNum, Settings::getInstance().getAudioBitrate(), videoBitrate, &err)) {
-        calls[friendNum].setActive(false);
+        calls[friendNum].setActive(true);
         return true;
     } else {
         qWarning() << "Failed to answer call with error" << err;
@@ -344,7 +344,7 @@ bool CoreAV::sendCallAudio(uint32_t callId, const int16_t* pcm, size_t samples, 
 
     ToxFriendCall& call = calls[callId];
 
-    if (call.getMuteMic() || call.isActive()
+    if (call.getMuteMic() || !call.isActive()
         || !(call.getState() & TOXAV_FRIEND_CALL_STATE_ACCEPTING_A)) {
         return true;
     }
@@ -379,15 +379,15 @@ void CoreAV::sendCallVideo(uint32_t callId, std::shared_ptr<VideoFrame> vframe)
 
     ToxFriendCall& call = calls[callId];
 
-    if (!call.getVideoEnabled() || call.isActive()
+    if (!call.getVideoEnabled() || !call.isActive()
         || !(call.getState() & TOXAV_FRIEND_CALL_STATE_ACCEPTING_V)) {
         return;
     }
 
-    if (call.getVideoEnabled()) {
+    if (call.getNullVideoBitrate()) {
         qDebug() << "Restarting video stream to friend" << callId;
         toxav_bit_rate_set(toxav, callId, -1, VIDEO_DEFAULT_BITRATE, nullptr);
-        call.setVideoEnabled(false);
+        call.setNullVideoBitrate(false);
     }
 
     ToxYUVFrame frame = vframe->toToxYUVFrame();
@@ -436,7 +436,7 @@ void CoreAV::toggleMuteCallOutput(const Friend* f)
 {
     if (f && (calls.find(f->getId()) != calls.end())) {
         ToxCall& call = calls[f->getId()];
-        call.setMuteVol(call.getMuteVol());
+        call.setMuteVol(!call.getMuteVol());
     }
 }
 
@@ -473,8 +473,10 @@ void CoreAV::groupCallCallback(void* tox, int group, int peer, const int16_t* da
     }
 
     Audio& audio = Audio::getInstance();
-    if (!call.getPeers()[peer])
+    if (!call.getPeers()[peer]) {
+        // FIXME: 0 is a valid sourceId, we shouldn't necessarily re-subscribe just because we have 0
         audio.subscribeOutput(call.getPeers()[peer]);
+    }
 
     audio.playAudioBuffer(call.getPeers()[peer], data, samples, channels, sample_rate);
 }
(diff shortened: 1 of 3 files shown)


sudden6:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/core/toxcall.cpp, line 69 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvlq8sQ5ML_hyraeEX:-Kxvlq8sQ5ML_hyraeEY:b-dv01o9) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L69)):*
> ```cpp
>         // if we're already valid, we need to cleanup our resources since we're going to inherit the resources
>         // that are being moved in
>         QObject::disconnect(audioInConn);
> ```

This is wrong because a few lines down we move/copy the active connection from `other` to `this`.

---

*[src/core/toxcall.cpp, line 70 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvkl75MphheqY6kkYz:-Kxvkl76qYzWnCxOFkpx:b-smy4m5) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L70)):*
> ```cpp
>         // that are being moved in
>         QObject::disconnect(audioInConn);
>         Audio::getInstance().unsubscribeInput();
> ```

unsubscribing here is wrong, because the subscription is not bound to the object we just don't unsubscribe on deletion if the object was moved from.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784)*

anthonybilinski:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/core/toxcall.cpp, line 69 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvlq8sQ5ML_hyraeEX:-KxxivuuUgLlbnM1kCLo:bcshkr2) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L69)):*

This is wrong because a few lines down we move/copy the active connection from `other` to `this`.
</blockquote></details>

When we receive a call, we default construct a ToxFriendCall, which subscribes. Then when we answer, we construct a ToxFriendCall in coreav.cpp:720, which subscribes again. We then move-assign the r-value ToxFriendCall, moving its subscription and connection into the existing default constructed ToxFriendCall, and invalidating the r-value ToxFriendCall, which then destructs and does not unsubscribe. This means that the l-value ToxFriendCall is now responsible for two subscriptions, and will only unsubscribe once on destruction.

This might not be the best way to do it, perhaps just removing the "valid" check altogether and keeping destruction pattern from before would be better, but we need some way to avoid 2 subscription increments with only 1 decrement on destruction that is caused in this case.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784)*

sudden6:
:lgtm_strong:

---

Reviewed 3 of 3 files at r2.

---

*[src/core/toxcall.cpp, line 69 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvlq8sQ5ML_hyraeEX:-KxyyN5YqkYBePE0Hgsa:bmb0io6) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L69)):*

When we receive a call, we default construct a ToxFriendCall, which subscribes. Then when we answer, we construct a ToxFriendCall in coreav.cpp:720, which subscribes again. We then move-assign the r-value ToxFriendCall, moving its subscription and connection into the existing default constructed ToxFriendCall, and invalidating the r-value ToxFriendCall, which then destructs and does not unsubscribe. This means that the l-value ToxFriendCall is now responsible for two subscriptions, and will only unsubscribe once on destruction.

This might not be the best way to do it, perhaps just removing the "valid" check altogether and keeping destruction pattern from before would be better, but we need some way to avoid 2 subscription increments with only 1 decrement on destruction that is caused in this case.
</blockquote></details>

ahh sorry, I totaly read the code wrong :/ I thought the check was for `other.valid` instead of `this.valid` your fix is entirely correct.

I think we should either init the object with `valid = false` and skip the resource allocation, or should prevent usage of the default constructor entirely. What do you think?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784#-:-Kxz-0FzqGLjZb-TI98U:b-jbsf8z)*

anthonybilinski:

---

*[src/core/toxcall.cpp, line 69 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvlq8sQ5ML_hyraeEX:-Kxz2Dm8FYVycIJFPg2j:bk4b7sz) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L69)):*
> either init the object with valid = false and skip the resource allocation, or should prevent usage of the default constructor
I think just removing the default constructor would be cleaner, then any ToxFriendCall we have we know is valid, rather than having to constantly check a bool to see if it's real or not. This would unfortunately mean quite a bit of changes in coreav where we currently index calls like an array all over the place, each of which can cause a default construction.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784)*

anthonybilinski:

---

*[src/core/toxcall.cpp, line 69 at r1](https://reviewable.io:443/reviews/qtox/qtox/4784#-Kxvlq8sQ5ML_hyraeEX:-Kxz34ohN9IwyQEZd7vO:b-8dqv2j) ([raw file](https://github.com/qtox/qtox/blob/9590c3aaf5410fc55cf9f031048e43526bf43381/src/core/toxcall.cpp#L69)):*

> either init the object with valid = false and skip the resource allocation, or should prevent usage of the default constructor
I think just removing the default constructor would be cleaner, then any ToxFriendCall we have we know is valid, rather than having to constantly check a bool to see if it's real or not. This would unfortunately mean quite a bit of changes in coreav where we currently index calls like an array all over the place, each of which can cause a default construction.
</blockquote></details>

:S meant to only quote the first line...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784)*

anthonybilinski:
I have a pretty major ToxCall refactor in PR that I'm rebasing on top of this. I can create that change to remove ToxCall default ctor as part of that, and make the correspoding coeav changes. I think the current implementation from this PR is still safe, even though a bit messy.

sudden6:
ok, if you already have a PR I'll merge this one as is.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4784#-:-Ky0tfSSkVX--yApKauf:b-b09h6)*

PR #4805, opened 2017-11-14
Update INSTALL.md

torvn77:
The version of sqlcipher in Debian 9 is now 3.2.0-2 => 3.2.0
Added it to the list of installed packages.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4805)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -281,7 +281,8 @@ sudo apt-get install \
     qt5-default \
     qttools5-dev \
     qttools5-dev-tools \
-    yasm
+    yasm \
+    sqlcipher
 ```
 
 <a name="fedora-other-deps" />


Diadlo:
Your PR is failing because of the Commit Guidlines https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#-git-commit-guidelines

And are you sure, that it needed to install it manually? Probably `sqlcipher` is dependency of `libsqlcipher-dev`

torvn77:
I checked the dependencies, the sqlcipher package itself in dependencies is not.
In dependencies it is necessary only library libsqlcipher0.

P.S.А вообще, сам этот sqlcipher нужен или нужна только его библиотека, зачем его ставят из гит?
tr:P.S.A in general, this sqlcipher itself is needed or only its library is needed, why is it being put out of git?

Diadlo:
@torvn77 Not sure. AFAIK, sqlcipher itself is not required.

Diadlo:
@sudden6 @nurupo Is it?

sudden6:
I don't think `sqlcipher` is needed, I downloaded the package and it just contains an executable, but we need the sqlcipher library.

Diadlo:
Close?

sudden6:
@Diadlo I think so

torvn77:
And why then in the manual it is written as sqlcipher to put from git?
Maybe this part is already outdated and now only confuses users?

nurupo:
Are you running an old Fedora Linux? If not, why are you following instructions for old Fedora? It clearly says:

>If you are not using an old version of Fedora, skip this section

PR #4807, opened 2017-11-15
feat: Add spell checking

### LINKED ISSUE #1301: [feature] spell checking
It may be interesting to add spell checking on text input (with suggestions on right click).

Diadlo:
Fix #1301

Spell checking implemented by KF5Sonnet

Seems KF5Sonnet doesn't work with my window manager (i3). Can someone test it?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4807)

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -15,6 +15,7 @@ find_package(Qt5Svg           REQUIRED)
 find_package(Qt5Test          REQUIRED)
 find_package(Qt5Widgets       REQUIRED)
 find_package(Qt5Xml           REQUIRED)
+find_package(KF5Sonnet)
 
 function(add_dependency)
   set(ALL_LIBRARIES ${ALL_LIBRARIES} ${ARGN} PARENT_SCOPE)
@@ -29,6 +30,11 @@ add_dependency(
   Qt5::Svg
   Qt5::Widgets
   Qt5::Xml)
+
+if(KF5Sonnet_FOUND)
+  add_definitions(-DSPELL_CHECKING)
+  add_dependency(KF5::SonnetUi)
+endif()
 
 include(CMakeParseArguments)
 include(Qt5CorePatches)
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -45,6 +45,10 @@
 #include <QKeyEvent>
 #include <QShortcut>
 
+#ifdef SPELL_CHECKING
+#include <KF5/SonnetUi/sonnet/spellcheckdecorator.h>
+#endif
+
 /**
  * @class GenericChatForm
  * @brief Parent class for all chatforms. It's provide the minimum required UI
@@ -126,6 +130,9 @@ GenericChatForm::GenericChatForm(QWidget* parent)
     connect(&s, &Settings::chatMessageFontChanged, this, &GenericChatForm::onChatMessageFontChanged);
 
     msgEdit = new ChatTextEdit();
+#ifdef SPELL_CHECKING
+    decorator = new Sonnet::SpellCheckDecorator(msgEdit);
+#endif
 
     sendButton = new QPushButton();
     emoteButton = new QPushButton();
@@ -222,6 +229,9 @@ GenericChatForm::GenericChatForm(QWidget* parent)
 
 GenericChatForm::~GenericChatForm()
 {
+#ifdef SPELL_CHECKING
+    delete decorator;
+#endif
     Translator::unregister(this);
 }
 
--- a/src/widget/form/genericchatform.h
+++ b/src/widget/form/genericchatform.h
@@ -52,6 +52,10 @@ namespace Ui {
 class MainWindow;
 }
 
+namespace Sonnet {
+class SpellCheckDecorator;
+}
+
 class GenericChatForm : public QWidget
 {
     Q_OBJECT
@@ -150,6 +154,7 @@ protected:
 
     ChatLog* chatWidget;
     ChatTextEdit* msgEdit;
+    Sonnet::SpellCheckDecorator* decorator;
     FlyoutOverlayWidget* fileFlyout;
     GenericNetCamView* netcam;
     Widget* parent;


tox-user:
How exactly can I test this, how does this feature work? I have installed libkf5sonnet-dev, do I need anything else?

Diadlo:
@tox-user Use `./test-pr.sh 4807` and build qTox (clear cmake cache before).
Also, you probably need backend for sonnet and dictionary for it

sudden6:
1) It compiles and is loaded, but according to the log messages:
```
[15:13:43.142 UTC] :0 : Warning: No language dictionaries for the language: "de"
```
can't find the dictionaries. I have all the dictionaries installed I can find, but still doesn't work.

2) No message if it's built with spell checking support in cmake
3) Please add an option to select the language, maybe based on qTox language. Because it defaults to system language but I mostly write in English in qTox.

tox-user:
I have those packages installed:
libkf5sonnet-dev libkf5sonnet-dev-bin libkf5sonnet5-data libkf5sonnetcore5 libkf5sonnetui5 sonnet-plugins
but I don't know how to install dictionaries. I'm getting the same log message as sudden6. Cmake shows a warning when it can't find kf5sonnet, but it doesn't say anything when it's installed.

Diadlo:
@sudden6, 
Sonnet autodetect language.

> It also supports automated language detection, based on a combination of different algorithms.
Source: https://api.kde.org/frameworks/sonnet/html/index.html

About warning: I have the same behaviour, seems it's not only i3 problem (or it works just with KDE?)

sudden6:
Is there no possibility to specify the language or is it just optional?

I was testing on KDE, but I think KDE frameworks libraries are not tied to KDE and should work on other DEs.

Maybe it has to do with the name of the dictionaries? In my package manager they are called `de_AT` and `de_DE`, but in the error message it says just `de`.

Diadlo:
Ok. Now it works for me.
@sudden6 What dictionaries do you have?
@tox-user Try to install something like `aspell-LANG` Where LANG is your language

Diadlo:
Sonnet supports at least this backends: aspell, hspell, hunspell, voikko

Diadlo:
image

sudden6:
`aspell --help` gives me:

```
  de_AT
  de_AT
  de_CH
  de_CH
  de_DE
  de_DE
```

Is there a way to check which backend is used?

Diadlo:
@sudden6 sonnet use all backends simultaneously

Diadlo:
@sudden6 `aspell --help` gives me a large help output. Are you sure, that you have used `help` flag?
```

Aspell 0.60.6.1.  Copyright 2000-2011 by Kevin Atkinson.

Usage: aspell [options] <command>

<command> is one of:
  -?|usage         display a brief usage message
  help             display a detailed help message
  -c|check <file>  to check a file
  -a|pipe          "ispell -a" compatibility mode
.....
```

Diadlo:
> @sudden6 sonnet use all backends simultaneously

Check content of /usr/lib/qt/plugins/kf5/sonnet
Here should be shared libs for all supported backends

Diadlo:
@sudden6 Seems, problem in language suffix. IDK is there good way to fix it. Can you try other backends?

tox-user:
Turns out I have the required dictionary installed and I have aspell too. There also is this log message:
```
core/core.cpp:130 : Debug: Core starting with IPv6 enabled
Hspell: can't open /usr/share/hspell/hebrew.wgz.sizes.
:0 : Warning: HSpellDict::HSpellDict: Init failed
```
I either missed it earlier or it wasn't there. It goes away after I install hspell, but the one about missing dictionary doesn't.

Diadlo:
@sudden6 We can force the language. But probably it will break autodetection

Diadlo:
@tox-user Can you test if hspell works outside of qTox? If not, check if this file exists, if so, check which package contains this file and reinstall it

tox-user:
@Diadlo it does work and the file exists when hspell is installed. My point was that qTox complained about hspell missing when hspell wasn't installed, but I had aspell. So I don't understand why it needed hspell.

Diadlo:
@tox-user Sonnet tries all backends

Diadlo:
If you want to disable backend, just remove it from `/usr/lib/qt/plugins/kf5/sonnet`

Diadlo:
@tox-user Is [it](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=204588) your issue?

tox-user:
@Diadlo no, that doesn't happen for me. The message I mentioned disappeared after I installed hspell, so that error is gone. I was just surprised that it required me to have hspell, since I had another backend already installed. The problem with "missing" dictionaries still remains.

Diadlo:
@tox-user What language do you use?

tox-user:
@Diadlo qTox says it's en-GB that's missing, but aspell lists it as en_GB (with underscore instead of dash).

tox-user:
I found some links related to this problem. Maybe they will help.
https://bugs.kde.org/show_bug.cgi?id=256896
https://bugs.kde.org/show_bug.cgi?id=171432

Diadlo:
So, what next? Drop Sonnet? Add ability to force some language?

sudden6:
After playing around with KDE regional settings I got it to work too. Or maybe the logout/login cycle helped.

I guess if we ship this as is we'll get a ton of bug reports about the problems we also had. I think a good idea would be to provide a list of installed and working dictionaries, where one can set the ones to use. Is this possible with Sonnet?

Diadlo:
Upstream issue: https://bugs.kde.org/show_bug.cgi?id=359265

tox-user:
Making a symlink with correct name in hunspell directory as explained in https://bugs.kde.org/show_bug.cgi?id=359265 solved the problem for me. It worked even after I removed hunspell, but the directory remained. It always complains when hspell is missing, but it works without it  too.

sudden6:
@Diadlo do you have an account for bugs.kde.org? Else I can comment with mine that we also experience this issue.

Diadlo:
@sudden6 AFAIK I haven't. Also sonnet works for me with de_DE

sudden6:
hmm, maybe it's outdated on my system then, will try to run a newer version and report back.

Diadlo:
@sudden6 I have test on `master` build of sonnet (https://cgit.kde.org/sonnet.git/)

sudden6:
same problem with sonnet version 5.40, seems it's only fixed in master or it's a problem with the configuration in OpenSUSE :/

Diadlo:
@sudden6 Can you try master?

tox-user:
@Diadlo I just tested it with Sonnet 5.40 stable (newest stable release) and it didn't solve the problem.

tox-user:
@Diadlo I checked with master too to be sure. It's the same.

Talkless:
Review status: 0 of 3 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/widget/form/genericchatform.cpp, line 233 at r1](https://reviewable.io:443/reviews/qtox/qtox/4807#-L12GWHpar4qxVRlw63C:-L12GWHqqHeeckILlNoS:b-n0n4yy) ([raw file](https://github.com/qtox/qtox/blob/de7ed495521e5ce800a66e49da2c3bca5f947008/src/widget/form/genericchatform.cpp#L233)):*
Is this explicit delete needed? 

>  @param textEdit the QTextEdit in need of spell-checking. It also is used as the QObject parent for the decorator.

https://github.com/KDE/sonnet/blob/e8fa82febb19f748b6dc9d54441465d01426fef5/src/ui/spellcheckdecorator.h#L49

---

*[src/widget/form/genericchatform.h, line 157 at r1](https://reviewable.io:443/reviews/qtox/qtox/4807#-L12Fysuf84uJRhxbXM5:-L12Fysuf84uJRhxbXM6:b-cw7o35) ([raw file](https://github.com/qtox/qtox/blob/de7ed495521e5ce800a66e49da2c3bca5f947008/src/widget/form/genericchatform.h#L157)):*
> ```objc
>     ChatLog* chatWidget;
>     ChatTextEdit* msgEdit;
>     Sonnet::SpellCheckDecorator* decorator;
> ```

`#ifdef` too? Could produce warnings as being not initialized variable, and also uses space.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4807)*

Diadlo:
Because many do not work I see no reason to continue. Creates only more problems

Issue #4741, opened 2017-10-16
Do not force avatar (daily builds)


fcore117:
I just tested daily build and found out that avatar is forced, please do not force avatar when you have not chosen own avatar. This feedback message is meant for upcoming versions.

sudden6:
To confirm I understand you correctly: You want a setting that disables the Identicon which is now displayed instead of the default avatar?

fcore117:
yes just no icon like before and like other clients has when no avatar in use.

tox-user:
Does that mean not forcing a generated avatar for your own profile only or do you want all avatars to be like they used to? 

I understand why someone would want to use a blank avatar for their own profile instead of a generated icon but is there a reason to not show them in other people's profiles?

fcore117:
tox-user: not for myself and no icon propagation to others too, like it was before. I could see other icons but when i turn off my icon then it wont be sent to others over tox network. In short, like it was before.

tox-user:
Why do you not want to see other people's avatars? Shouldn't users only have control over their own avatar?

fcore117:
tox-user: i did not say i do not want to see other avatars, i just did not want myself send out any avatar if not set.

sudden6: if you see that this feature is must have for future then just close this issue and let it be.

nurupo:
My comment is probably a little offtopic, but there was an identicon discussion brought up in #utox and someone asked me to write my thoughts about it in this issue, so here it goes, hopefully someone finds it useful.

---

The thing about identicons is that users might rely on them to uniquely identify contacts, something like Tox Id fingerprint, since they \*are\* Tox Id fingerprints, as they are uniquely generated from Tox Ids or public keys. A client allowing displaying of both identicons and avatars at the same time might cause issues, as once you allow a custom image as avatar you can't rely on identicons anymore, a user might set their avatar to someone else's identicon in attempt to impersonate that someone else. For example, say a user has Alice and Bob in their contact list. Alice has red dot identicon. Bob has a cat picture as avatar. Bob can "impersonate" Alice by changing his name and status to that of Alice, as well as changing his avatar from cat to a red dot.

Of course, even before identicons were added to qTox, with just avatar pictures, Bob could copy Alice's avatar in attempt to impersonate her. But the thing about the identicons is that they have this uniqueness property of being a Tox ID fingerprint, so I claim that it's a different matter from avatar impersonation before qTox had support for identicons.

[...]

PR #4818, opened 2017-11-21
feat(settings): add an option to toggle identicons

tox-user:
Solves #4741

Adds a setting to User Interface tab that allows to toggle displaying
identicons instead of default avatar picture.

This toggles identicons for all of user's contacts, not just for the 
user's own profile.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4818)

### DIFF
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -403,7 +403,7 @@ QPixmap Profile::loadAvatar(const ToxPk& owner)
         } else {
             pic.loadFromData(avataData);
         }
-    
+
     } else {
         pic.loadFromData(loadAvatarData(owner));
     }
@@ -477,10 +477,7 @@ void Profile::setAvatar(QByteArray pic, const ToxPk& owner)
             // with IDENTICON_ROWS=5 this gives a 160x160 image file
             const QImage identicon = Identicon(owner.getKey()).toImage(32);
             pixmap = QPixmap::fromImage(identicon);
-            QBuffer buf(&avatarData);
-            buf.open(QIODevice::WriteOnly);
-            identicon.save(&buf, "png");
-            
+
         } else {
             pixmap.load(":/img/contact_dark.svg");
         }
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -395,11 +395,17 @@ QPixmap Profile::loadAvatar()
 QPixmap Profile::loadAvatar(const ToxPk& owner)
 {
     QPixmap pic;
-    const QByteArray avataData = loadAvatarData(owner);
-    if(avataData.isEmpty()) {
-        pic = QPixmap::fromImage(Identicon(owner.getKey()).toImage(16));
+    if (Settings::getInstance().getShowIdenticons()) {
+
+        const QByteArray avataData = loadAvatarData(owner);
+        if (avataData.isEmpty()) {
+            pic = QPixmap::fromImage(Identicon(owner.getKey()).toImage(16));
+        } else {
+            pic.loadFromData(avataData);
+        }
+    
     } else {
-        pic.loadFromData(avataData);
+        pic.loadFromData(loadAvatarData(owner));
     }
 
     return pic;
@@ -467,12 +473,18 @@ void Profile::setAvatar(QByteArray pic, const ToxPk& owner)
         pixmap.loadFromData(pic);
         avatarData = pic;
     } else {
-        // with IDENTICON_ROWS=5 this gives a 160x160 image file
-        const QImage identicon = Identicon(owner.getKey()).toImage(32);
-        pixmap = QPixmap::fromImage(identicon);
-        QBuffer buf(&avatarData);
-        buf.open(QIODevice::WriteOnly);
-        identicon.save(&buf, "png");
+        if (Settings::getInstance().getShowIdenticons()) {
+            // with IDENTICON_ROWS=5 this gives a 160x160 image file
+            const QImage identicon = Identicon(owner.getKey()).toImage(32);
+            pixmap = QPixmap::fromImage(identicon);
+            QBuffer buf(&avatarData);
+            buf.open(QIODevice::WriteOnly);
+            identicon.save(&buf, "png");
+            
+        } else {
+            pixmap.load(":/img/contact_dark.svg");
+        }
+
     }
 
     saveAvatar(avatarData, owner);
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -211,6 +211,7 @@ void Settings::loadGlobal()
         groupchatPosition = s.value("groupchatPosition", true).toBool();
         separateWindow = s.value("separateWindow", false).toBool();
         dontGroupWindows = s.value("dontGroupWindows", false).toBool();
+        showIdenticons = s.value("showIdenticons", true).toBool();
 
         const QString DEFAULT_SMILEYS = ":/smileys/emojione/emoticons.xml";
         smileyPack = s.value("smileyPack", DEFAULT_SMILEYS).toString();
@@ -526,6 +527,7 @@ void Settings::saveGlobal()
         s.setValue("separateWindow", separateWindow);
         s.setValue("dontGroupWindows", dontGroupWindows);
         s.setValue("groupchatPosition", groupchatPosition);
+        s.setValue("showIdenticons", showIdenticons);
 
         s.setValue("smileyPack", smileyPack);
         s.setValue("emojiFontPointSize", emojiFontPointSize);
@@ -2188,6 +2190,22 @@ void Settings::setGroupchatPosition(bool value)
     if (value != groupchatPosition) {
         groupchatPosition = value;
         emit groupchatPositionChanged(value);
+    }
+}
+
+bool Settings::getShowIdenticons() const
+{
+    const QMutexLocker locker{&bigLock};
+    return showIdenticons;
+}
+
+void Settings::setShowIdenticons(bool value)
+{
+    const QMutexLocker locker{&bigLock};
+
+    if (value != showIdenticons) {
+        showIdenticons = value;
+        emit showIdenticonsChanged(value);
     }
 }
 
--- a/src/persistence/settings.h
+++ b/src/persistence/settings.h
@@ -75,6 +75,8 @@ class Settings : public QObject, public ICoreSettings, public IAudioSettings, pu
     Q_PROPERTY(QString style READ getStyle WRITE setStyle NOTIFY styleChanged FINAL)
     Q_PROPERTY(bool showSystemTray READ getShowSystemTray WRITE setShowSystemTray NOTIFY
                    showSystemTrayChanged FINAL)
+    Q_PROPERTY(bool showIdenticons READ getShowIdenticons WRITE setShowIdenticons NOTIFY
+                   showIdenticonsChanged FINAL)
 
     // ChatView
     Q_PROPERTY(bool groupchatPosition READ getGroupchatPosition WRITE setGroupchatPosition NOTIFY
@@ -209,6 +211,7 @@ signals:
     void styleChanged(const QString& style);
     void themeColorChanged(int color);
     void compactLayoutChanged(bool enabled);
+    void showIdenticonsChanged(bool enabled);
 
     // ChatView
     void useEmoticonsChanged(bool enabled);
@@ -499,9 +502,12 @@ public:
 
     bool getDontGroupWindows() const;
     void setDontGroupWindows(bool value);
-
+    
     bool getGroupchatPosition() const;
     void setGroupchatPosition(bool value);
+    
+    bool getShowIdenticons() const;
+    void setShowIdenticons(bool value);
 
     bool getAutoLogin() const;
     void setAutoLogin(bool state);
@@ -566,6 +572,7 @@ private:
     bool groupchatPosition;
     bool separateWindow;
     bool dontGroupWindows;
+    bool showIdenticons;
     bool enableIPv6;
     QString translation;
     bool makeToxPortable;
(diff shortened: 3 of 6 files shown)


Diadlo:
1) Please, add information box or somewere information, that change will be accepted after restart

2) Issue:
2.1) enable identicon
2.2) restart qTox
2.3) disable identicon
2.4) restart qTox
2.5) Observed: you avatar not reseted

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/persistence/profile.cpp, line 398 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJYN-Y5dUIBEq41pj:-KzVJYN-Y5dUIBEq41pk:begm0k3) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L398)):*
> ```cpp
> {
>     QPixmap pic;
>     if (Settings::getInstance().getShowIdenticons() == true) {
> ```

You don't need to compare bool with `true`. Just use
`if (Settings::getInstance().getShowIdenticons()) {`

---

*[src/persistence/profile.cpp, line 401 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJkeln7D4-RpYw5cx:-KzVJkeln7D4-RpYw5cy:biyh71h) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L401)):*
> ```cpp
> 
>         const QByteArray avataData = loadAvatarData(owner);
>         if(avataData.isEmpty()) {
> ```

Add space after `if`

---

*[src/persistence/profile.cpp, line 406 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJxG_S9vd1mTSzNI6:-KzVJxG_S9vd1mTSzNI7:bvt1fsf) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L406)):*
> ```cpp
>             pic.loadFromData(avataData);
>         }
>     
> ```

Seems tab here (or reviewable strange symbols). If there is tab, please, remove it. If not -- ignore

---

*[src/persistence/profile.cpp, line 476 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJt2AKO6NoHnTNYGD:-KzVJt2AKO6NoHnTNYGE:b49haa5) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L476)):*
> ```cpp
>         avatarData = pic;
>     } else {
>         if (Settings::getInstance().getShowIdenticons() == true) {
> ```

Same with `== true`

---

*[src/persistence/profile.cpp, line 483 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJvpRMDqhU6lWUJxE:-KzVJvpRMDqhU6lWUJxF:b7x1i6y) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L483)):*
> ```cpp
>             buf.open(QIODevice::WriteOnly);
>             identicon.save(&buf, "png");
>             
> ```

Same with tab

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818#-:-KzVMyMvR5aRypXK3am5:b49y07b)*

tox-user:
1. That information is already in the tootip (https://github.com/qTox/qTox/pull/4818/commits/905ca7708662c6fe5520ad2707715075ad1aa41d#diff-09d87ecacfd6805618183e4648cec68dR263).
2. This is because avatars are saved to a file. How do I distinguish if a file is a user avatar or an identicon? I'm stuck.

---

Review status: 5 of 6 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/persistence/profile.cpp, line 398 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJYN-Y5dUIBEq41pj:-KzVaqtCrUfP4nR-m6HM:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L398)):*

You don't need to compare bool with `true`. Just use
`if (Settings::getInstance().getShowIdenticons()) {`
</blockquote></details>

Done.

---

*[src/persistence/profile.cpp, line 401 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJkeln7D4-RpYw5cx:-KzVarfC2hdWME0KEewb:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L401)):*

Add space after `if`
</blockquote></details>

Done.

---

*[src/persistence/profile.cpp, line 406 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJxG_S9vd1mTSzNI6:-KzVaZPO_Px650ns6rHm:b-1chgip) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L406)):*

Seems tab here (or reviewable strange symbols). If there is tab, please, remove it. If not -- ignore
</blockquote></details>

There is no tab there.

---

*[src/persistence/profile.cpp, line 476 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJt2AKO6NoHnTNYGD:-KzVauFp9fu2Y_Rk4J3m:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L476)):*

Same with `== true`
</blockquote></details>

Done.

---

*[src/persistence/profile.cpp, line 483 at r1](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzVJvpRMDqhU6lWUJxE:-KzVahyte89Q_3uuI6Su:b-ci1cfc) ([raw file](https://github.com/qtox/qtox/blob/9011a21c420ab4429ca603a92363fa6664177c98/src/persistence/profile.cpp#L483)):*

Same with tab
</blockquote></details>

There is no tab there either.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818#-:-KzVb-n_eUCK4KJ4lRe0:b-cw7ley)*

Diadlo:
@sudden6 Probably, qTox should show identicon instead of default avatar, but not set it?

Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818#-:-KzXQ1yHbB3Jr_o5_IgX:b-jbsf8z)*

tox-user:
@Diadlo perhaps there should be one more option for user's own avatar? It would allow to switch between identicon and default picture for user's profile.

So there would be two independent options:
▢ Use identicons for contacts (when their avatar is empty)
▢ Use identicon for my own profile (when my avatar is empty)

I still don't know how it would work though.

Diadlo:
@tox-user I don't think, that the second point make sense in practice. If someone want to see identicon in one place, he wants in another places too.
But qTox shouldn't set identicon as avatar, just show it.

Diadlo:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/persistence/profile.cpp, line 406 at r2](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzZMUTZowDe6Gom3U2m:-KzZMUT_ACv6jcesnT4P:b-nekgxc) ([raw file](https://github.com/qtox/qtox/blob/905ca7708662c6fe5520ad2707715075ad1aa41d/src/persistence/profile.cpp#L406)):*
> ```cpp
>             pic.loadFromData(avataData);
>         }
>     
> ```

Oh.. it's spaces. Please, remove empty line here or spaces in the begin of the line

---

*[src/persistence/profile.cpp, line 483 at r2](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzZM_YeBAH55Kt5bEY0:-KzZM_YeBAH55Kt5bEY1:b-ya81ng) ([raw file](https://github.com/qtox/qtox/blob/905ca7708662c6fe5520ad2707715075ad1aa41d/src/persistence/profile.cpp#L483)):*
> ```cpp
>             buf.open(QIODevice::WriteOnly);
>             identicon.save(&buf, "png");
>             
> ```

Same as above

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818)*

sudden6:
@Diadlo IMO if it's shown to the user it should also be sent to others as avatar, less confusion possibilities.

tox-user:
@sudden6 so according to what you say if I have no avatar set and I have identicons enabled, someone who has them disabled will still see my identicon, because it has been set as my avatar. The setting will not work properly then. It will fail to disable identicons.

Right now it can't distinguish an avatar from an identicon. When user creates a new profile with no avatar, qTox will create an indenticon for them and set it as their avatar. This is because identicons are enabled by default. When the same user later decides they don't want to see identicons anymore and they disable them, their profile will still have an identicon as avatar.

This is all caused by the fact that our own identicon is saved to a file like any other avatar, so we can never be sure if user's avatar has actually been set or not, because if it hasn't qTox will set one for them. That means this PR in its current state doesn't work and we need a new approach.

Diadlo:
@sudden6 When we show default avatar, but not set, it doesn't confuse people, isn't it?

tox-user:
@Diadlo for users with no avatar who are on older versions of qTox or who use other clients we see their identicons even though their clients don't have that feature. So we already see something that is not there.

Diadlo:
@tox-user I know. I talk about @sudden6 argument with confusing. We didn't set qTox default avatar image as avatar before, we just showed it

tox-user:
I removed the part that saved the identicon as avatar. It solved the problem. If we want to be able to toggle identicons I don't see any other way.

---

Review status: 5 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

*[src/persistence/profile.cpp, line 406 at r2](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzZMUTZowDe6Gom3U2m:-Kz_mUCPMj3d9z37I8DN:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/905ca7708662c6fe5520ad2707715075ad1aa41d/src/persistence/profile.cpp#L406)):*

Oh.. it's spaces. Please, remove empty line here or spaces in the begin of the line
</blockquote></details>

Done.

---

*[src/persistence/profile.cpp, line 483 at r2](https://reviewable.io:443/reviews/qtox/qtox/4818#-KzZM_YeBAH55Kt5bEY0:-Kz_mUmup420ytcINI3L:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/905ca7708662c6fe5520ad2707715075ad1aa41d/src/persistence/profile.cpp#L483)):*

Same as above
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818#-:-Kz_mZYaxwQEd4MfzhiG:b-pprg0w)*

Diadlo:
@sudden6 So what is the decision? Two options in settings (I'm against this way) or not set identicon as avatar? Or set, when identicon enabled and remove, when disabled?

anthonybilinski:
I think it makes sense to treat identicons like the default avatar picture, where it's displayed to the user but not sent to different contacts like a real avatar picture is. It's true it would cause a different image to display for a user and their friend, but if their friend's client implements the identicon spec, they will see the same identicon if their client supports it and they have it enabled.

Note that there are some backwards compatibility issues here. Even if we stop saving identicons as avatar pictures, for any user that has run qTox between when identicons were added until when this gets merged, they will still have those identicons saved locally and will still load them as avatars even if they disable identicons in settings. This would affect any user that has used 0.12.*.

I'm not sure if that's worth a work-around, that could be when the setting is disabled by a user, we check if they already have the identicon image saved locally and if so delete it?

Diadlo:
@anthonybilinski qTox with identicon implementation wasn't released yet. Masten is unstable and its use is carried out at your own risk. We are not required to comply with backward compatibility with the master.

So, I think we should just remove setting identicon before release and everything will be fine

anthonybilinski:
Good point! I thought it was added right before 0.12, but it was really added right after.

tox-user:
> It's true it would cause a different image to display for a user and their friend, but if their friend's client implements the identicon spec, they will see the same identicon if their client supports it and they have it enabled.

Identicons don't make sense, if all clients don't use the same spec anyway.

> I'm not sure if that's worth a work-around, that could be when the setting is disabled by a user, we  check if they already have the identicon image saved locally and if so delete it?

@anthonybilinski, @Diadlo is kind of right about compatibility, but I think we should try not to break it if possible. The problem is how do we know for sure, that the image we want to delete is an identicon and not user's avatar? We don't want to start deleting people's avatars and I don't see how we could distinguish them. If we could somehow do that then we wouldn't have the compatibility problem anyway, because we could just keep saving them as avatars and delete them whenever users disable them.

anthonybilinski:
@tox-user  We can generate the identicon and check if it's binary equivalent to the saved PNG, and if so assume it was an auto-generated avatar that the user didn't set themselves. The user could still theoretically manually set their avatar to the PNG that is generated by the identicon algo, but that's pretty crazy.

sudden6:
@Diadlo I understand now what you mean and agree with you and @tox-user that it's better not to save the Identicon as a file.

@anthonybilinski This could also be a good idea

sudden6:
:lgtm_strong:

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4818#-:-Kzk36alhhwhft55eUp-:b-jbsf8z)*

PR #4831, opened 2017-11-24
feat: Prefere new line as message break

### LINKED ISSUE #4113: Chose intelligent place where to break long messages
qTox should be able to choose more intelligent places where to beak long messages apart than it does now.

OS: any
qTox version: 1.7.0
toxcore:  0.1.5
Qt:  5.5.1

Reproducible: Always

##### Steps to reproduce

1. Go to your email client
2. Hit Reply on a very long email message
3. Copy the quoted email from the reply window
4. Use qTox to send it to somebody

##### Observed Behavior

The email is longer than the maximum size for a single message and broken apart each n letters while somewhat respecting words in the sense that it tries to break at a space instead of the middle of a word.

In case of emails, this means that it might look somewhat like this:

```
> First line
[... lots of more lines here]
> this line is where the maximum message size is reached, so it is broken after the next word, which
is why this line is not preceded with a > and thus shown in black instead of being formatted as a quote
> but this line is preceded with a > again so it shows up green again
> and this line too
```

And this means you have large blocks of light green text (as qTox interprets lines starting with > as quotes and gives those a lighter green color) broken up by a normal black colored lines where the large message has been broken apart, breaking both the overall view as well as the original formatting of the quoted message.

##### Expected Behavior

If a long message includes line breaks (no matter whether it is a quote or not – quotes only allow to illustrate the problem nicely because of how they are colored), break up the message at a line break instead of doing so after a random word in the middle of a sentence.

This is something that can easily be implemented on the sending client side without having to touch toxcore at all.

Diadlo:
Fix #4113

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4831)

### DIFF
--- a/src/audio/backend/openal.cpp
+++ b/src/audio/backend/openal.cpp
@@ -554,23 +554,17 @@ void OpenAL::stopActive()
 }
 
 /**
- * @brief Called on the captureTimer events to capture audio
+ * @brief handles recording of audio frames
  */
-void OpenAL::doAudio()
+void OpenAL::doInput()
 {
-    QMutexLocker lock(&audioLock);
-
-    if (!alInDev || !inSubscriptions) {
-        return;
-    }
-
     ALint curSamples = 0;
     alcGetIntegerv(alInDev, ALC_CAPTURE_SAMPLES, sizeof(curSamples), &curSamples);
-    if (static_cast<ALuint>(curSamples) < AUDIO_FRAME_SAMPLE_COUNT) {
+    if (curSamples < static_cast<ALint>(AUDIO_FRAME_SAMPLE_COUNT)) {
         return;
     }
 
-    alcCaptureSamples(alInDev, inputBuffer, AUDIO_FRAME_SAMPLE_COUNT);
+    captureSamples(alInDev, inputBuffer, AUDIO_FRAME_SAMPLE_COUNT);
 
     float volume = getVolume();
     if (volume >= inputThreshold) {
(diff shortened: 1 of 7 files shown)


noavarice:

---

*[src/core/core.cpp, line 1323 at r1](https://reviewable.io:443/reviews/qtox/qtox/4831#-KzhSSt85GU6butqpKCM:-KzhSSt85GU6butqpKCN:b-aywu7a) ([raw file](https://github.com/qtox/qtox/blob/57abc7630e3dc1a6402a77d90c5d559e7cbda406/src/core/core.cpp#L1323)):*
> ```cpp
> }
> 
> QStringList Core::splitMessage(const QString& srcMessage, int maxLen)
> ```

I think, you can move this function out of `Core` as well

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831)*

Diadlo:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/core/core.cpp, line 1323 at r1](https://reviewable.io:443/reviews/qtox/qtox/4831#-KzhSSt85GU6butqpKCM:-KzhVY-22pHKw-lerH7Z:b-nvrvxh) ([raw file](https://github.com/qtox/qtox/blob/57abc7630e3dc1a6402a77d90c5d559e7cbda406/src/core/core.cpp#L1323)):*
<details><summary><i>Previously, noavarice wrote…</i></summary><blockquote>

I think, you can move this function out of `Core` as well
</blockquote></details>

Not in this PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831)*

Diadlo:
It's just fix before new release. Don't want to change logic

tox-user:
@Diadlo will \n work as line ending on all operating systems? https://github.com/qTox/qTox/pull/4831/commits/57abc7630e3dc1a6402a77d90c5d559e7cbda406#diff-5db0ca19c12b31af1e97555eeadc9eeaR1329

Diadlo:
@tox-user Yes. All messages are platform independent

sudden6:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/core/core.cpp, line 1329 at r2](https://reviewable.io:443/reviews/qtox/qtox/4831#-Kzk21vONk3silUvxmEC:-Kzk21vPTLZlM1o7SaKS:b-gk7yp5) ([raw file](https://github.com/qtox/qtox/blob/4a1881e0df7662a008de0e381b97de16ab0b90bf/src/core/core.cpp#L1329)):*
> ```cpp
>     QString message{srcMessage};
> 
>     while (message.size() > maxLen) {
> ```

`QString.size()` returns the size in characters, not in bytes as wee need here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831)*

Talkless:
Review status: 0 of 5 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/core/core.cpp, line 675 at r2](https://reviewable.io:443/reviews/qtox/qtox/4831#-L11sTi_HdSPJOSlntRk:-L11sTi_HdSPJOSlntRl:bszcfh8) ([raw file](https://github.com/qtox/qtox/blob/4a1881e0df7662a008de0e381b97de16ab0b90bf/src/core/core.cpp#L675)):*
> ```cpp
>     QStringList cMessages = splitMessage(message, MAX_GROUP_MESSAGE_LEN);
> 
>     for (auto& part : cMessages) {
> ```

`const auto& part`, also almost all variables (except `error`) could be const I belive.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831)*

Diadlo:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/core/core.cpp, line 675 at r2](https://reviewable.io:443/reviews/qtox/qtox/4831#-L11sTi_HdSPJOSlntRk:-L2yhJugGiPmJotIzscE:b-z89qtw) ([raw file](https://github.com/qtox/qtox/blob/4a1881e0df7662a008de0e381b97de16ab0b90bf/src/core/core.cpp#L675)):*

`const auto& part`, also almost all variables (except `error`) could be const I belive.
</blockquote></details>

Not in PR anymore

---

*[src/core/core.cpp, line 1329 at r2](https://reviewable.io:443/reviews/qtox/qtox/4831#-Kzk21vONk3silUvxmEC:-L2yjMOdQ9N6gfzgVpCr:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/4a1881e0df7662a008de0e381b97de16ab0b90bf/src/core/core.cpp#L1329)):*

`QString.size()` returns the size in characters, not in bytes as wee need here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831)*

sudden6:
:lgtm_strong:

---

Reviewed 1 of 2 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4831#-:-L30-_JFKqbNCBGoy8Gc:b-jbsf8z)*

PR #4845, opened 2017-11-30
fix(cpu): Reduce CPU usage by avatar render

### LINKED ISSUE #4843: High CPU load since 73dcbd2e60e98d305f4d1d36c6d2a1b4bcd9de13
OS: Archlinux
Qt: 5.9.3
Hardware:  RX 480 with latest mesa git

Reproducible: Always

CPU load is high in both qTox and X since commit 73dcbd2e60e98d305f4d1d36c6d2a1b4bcd9de13 as long as the GUI is visible.

MaskablePixmapWidget::paintEvent() is called a lot and it looks like QLabel::setPixmap() is an expensive function to call https://github.com/qTox/qTox/commit/73dcbd2e60e98d305f4d1d36c6d2a1b4bcd9de13#diff-ac3ece10b10583d2d275f26a143e91a4R97. Commenting that one line out makes CPU load very low again, but of course some icons are missing.

Diadlo:
Fix #4843

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4845)

### DIFF
--- a/src/widget/maskablepixmapwidget.cpp
+++ b/src/widget/maskablepixmapwidget.cpp
@@ -90,14 +90,6 @@ void MaskablePixmapWidget::setSize(QSize size)
     }
 }
 
-
-
-void MaskablePixmapWidget::paintEvent(QPaintEvent* e)
-{
-    QLabel::setPixmap(*renderTarget);
-    QLabel::paintEvent(e);
-}
-
 void MaskablePixmapWidget::mousePressEvent(QMouseEvent*)
 {
     if (clickable) {
@@ -118,4 +110,5 @@ void MaskablePixmapWidget::updatePixmap()
     painter.setCompositionMode(QPainter::CompositionMode_DestinationIn);
     painter.drawPixmap(0, 0, mask);
     painter.end();
+    QLabel::setPixmap(*renderTarget);
 }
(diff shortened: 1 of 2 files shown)


nurupo:
You might want to restart the Windows jobs, apparently some package server replied with HTTP 503 error. I wonder if there is an option in `apt-get install` to make it retry several times.

Diadlo:
@nurupo Thanks. Seems there is only one way to retry - using while.

sudden6:

---

*[src/widget/maskablepixmapwidget.cpp, line 97 at r1](https://reviewable.io:443/reviews/qtox/qtox/4845#-L-BQTFT_C94UYIFt9up:-L-BQTFUzt3SZtg3RCi7:b-qf34pm) ([raw file](https://github.com/qtox/qtox/blob/52a30c65a29495292cb7fed7e0c75598733995d0/src/widget/maskablepixmapwidget.cpp#L97)):*
> ```cpp
> void MaskablePixmapWidget::paintEvent(QPaintEvent* e)
> {
>     QLabel::paintEvent(e);
> ```

Do we even need to implement this function if we're just using the one from the base class anyway?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4845)*

sudden6:
:lgtm_strong:

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4845#-:-L-OIU58Gkm3bWYbbu1x:b-jbsf8z)*

PR #4881, opened 2017-12-28
fix(translations): Correction of the translation into Russian

hkarel:
The original phrase "Enable experimental audio backend" is translated
with a distortion of the original meaning, which misleads the user.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4881)

### DIFF
--- a/translations/ru.ts
+++ b/translations/ru.ts
@@ -89,11 +89,11 @@ which may lead to problems with video calls.</source>
     </message>
     <message>
         <source>Enables the experimental audio backend with echo cancelling support, needs qTox restart to take effect.</source>
-        <translation>Включает экспериментальную поддержку аудиоданных с поддержкой эхоподавления, требует перезагрузки qTox.</translation>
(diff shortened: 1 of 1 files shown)


Diadlo:
In theory, experimental audio backend is not only for echo canceling, probably better will be "Эксперементальная звуковая система"

hkarel:
I agree, it's also a good variant, but then tooltip necessary to clarify it. Now I will correct.

Diadlo:
Translation :lgtm_strong:. 
Not sure, maybe better use commit message `fix(l10n): ...`?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4881#-:-L1aXvDCExdYHS_dc-Gc:bgy7p2r)*

hkarel:
So fix it on 'fix(l10n)' or not?

PR #4925, opened 2018-01-22
[WIP] Implement global PTT using UGlobalHotkey

### LINKED ISSUE #903: Configure Push to Talk in Settings Panel
I think the user should be able to switch between normal voice settings and push to talk within qTox itself. Along with the option to change the keys for push to talk activation.

anthonybilinski:
(will) Fix #903

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4925)

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -104,6 +104,7 @@ search_dependency(LIBQRENCODE         PACKAGE libqrencode)
 search_dependency(LIBSODIUM           PACKAGE libsodium)
 search_dependency(LIBSWSCALE          PACKAGE libswscale)
 search_dependency(SQLCIPHER           PACKAGE sqlcipher)
+search_dependency(UGLOBALHOTKEY       LIBRARY UGlobalHotkey    OPTIONAL)
 search_dependency(VPX                 PACKAGE vpx)
 
 # Try to find cmake toxcore libraries
@@ -205,6 +206,12 @@ if (PLATFORM_EXTENSIONS)
     message(WARNING "Not using platform extensions, dependencies not found")
     set(PLATFORM_EXTENSIONS OFF)
   endif()
+endif()
+
+if (UGLOBALHOTKEY_FOUND)
+  add_definitions(
+    -DUGLOBALHOTKEY
+  )
 endif()
 
 add_definitions(
--- a/src/widget/form/chatform.h
+++ b/src/widget/form/chatform.h
@@ -70,6 +70,7 @@ public slots:
     void onAvEnd(uint32_t friendId, bool error);
     void onAvatarChange(uint32_t friendId, const QPixmap& pic);
     void onAvatarRemoved(uint32_t friendId);
+    void onMicMuteToggle();
 
 private slots:
     void clearChatArea(bool notInForm) override final;
@@ -84,7 +85,6 @@ private slots:
     void onVideoCallTriggered();
     void onAnswerCallTriggered(bool video);
     void onRejectCallTriggered();
-    void onMicMuteToggle();
     void onVolMuteToggle();
 
     void onFileSendFailed(uint32_t friendId, const QString& fname);
--- a/src/widget/form/groupchatform.cpp
+++ b/src/widget/form/groupchatform.cpp
@@ -372,28 +372,6 @@ GenericNetCamView* GroupChatForm::createNetcam()
     return view;
 }
 
-void GroupChatForm::keyPressEvent(QKeyEvent* ev)
-{
-    // Push to talk (CTRL+P)
-    if (ev->key() == Qt::Key_P && (ev->modifiers() & Qt::ControlModifier) && inCall) {
-        onMicMuteToggle();
-    }
-
-    if (msgEdit->hasFocus())
-        return;
-}
-
-void GroupChatForm::keyReleaseEvent(QKeyEvent* ev)
-{
-    // Push to talk (CTRL+P)
-    if (ev->key() == Qt::Key_P && (ev->modifiers() & Qt::ControlModifier) && inCall) {
-        onMicMuteToggle();
-    }
-
-    if (msgEdit->hasFocus())
-        return;
-}
-
 /**
  * @brief Updates users' count label text
  */
--- a/src/widget/form/groupchatform.h
+++ b/src/widget/form/groupchatform.h
@@ -40,11 +40,13 @@ public:
 
     void peerAudioPlaying(int peer);
 
+public slots:
+    void onMicMuteToggle();
+
 private slots:
     void onSendTriggered() override;
     void onScreenshotClicked() override;
     void onAttachClicked() override;
-    void onMicMuteToggle();
     void onVolMuteToggle();
     void onCallClicked();
     void onUserListChanged();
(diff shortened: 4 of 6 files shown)


anthonybilinski:
A couple problems with this implementation:

1. There's no "key-up" notification provided by UGlobalHotkey, so qTox could only support tapping a key to toggle mute, not a "hold key to talk" functionality.
2. UGlobalHotkey seems to consume the key press, so if you have a different window selected and try to type using your configured key, it gets broken.

Both of these could be worked around by qTox just only offering toggle, and having a reasonably complex key-combo to activate so that it won't interfere with typing. Alternatively these _might_ be fixable upstream.

Regardless of these downsides, UGlobalHotkey still seems like our best option for global push-to-talk.

Diadlo:
What about using Qt [keyReleaseEvent](http://doc.qt.io/archives/qt-4.8/qwidget.html#keyReleaseEvent)?

anthonybilinski:
The problem with all Qt events like keyReleaseEvent is that "A widget must accept focus initially and have focus in order to receive a key release event.", which makes push to talk impossible if the application isn't selected. This seems really limiting to me for some use cases, like in any gaming scenario where push to talk is often used.

Diadlo:
@anthonybilinski If I understood your "2." right, it still an issue with `UGlobalHotkey`

anthonybilinski:
Yes, both 1. and 2. are problems with `UGlobalHotkey`.

tox-user:
To fully solve this issue we would then have to extend the functionality of UGlobalHotkey? Mumble is written in Qt and offers push to talk. Maybe we could see how they have done it https://github.com/mumble-voip/mumble?

anthonybilinski:
@tox-user good thinking, thanks for the suggestion :)
I took a look at mumble. It looks like they have an invisible widget over the entire window:
```
 <widget class="QWidget" name="qwPTTButtonWidget">
  <property name="geometry">
   <rect>
    <x>0</x>
    <y>0</y>
    <width>100</width>
    <height>100</height>
```
which then captures key down / key up events and mutes/unmutes that way.
```
void PTTButtonWidget::closeEvent(QCloseEvent *e) {
	g.s.qbaPTTButtonWindowGeometry = saveGeometry();
	QWidget::closeEvent(e);
}

void PTTButtonWidget::on_qpbPushToTalk_pressed() {
	emit triggered(true, QVariant());
}

void PTTButtonWidget::on_qpbPushToTalk_released() {
	emit triggered(false, QVariant());
}
```
It seems a little hacky compared to registering with the OS to receive the hotkeys, but it gets them both key-down and key-up on all platforms, without a lot of code or external libs. I'll give implementing this in qTox a try.

anthonybilinski:
Thanks to Mumble's 3-clause BSD license, which looks like it's GPL-compatible, I _think_ we can grab as much of their code/implementation/ideas as we want. Do we have any license-savy people who can confirm?

anthonybilinski:
Never mind, PTTButtonWidget is exactly what it says it is, a button widget that can optionally appear on screen, which can be clicked to emulate the bound push to talk.

Mumble's actual push to talk is more what I expected: An interface that has a win, x11, and macx implementation with OS-specific behaviour.
[GlobalShortcut](https://github.com/mumble-voip/mumble/blob/master/src/mumble/GlobalShortcut.cpp)
impl by:
[GlobalShortcut_unix](https://github.com/mumble-voip/mumble/blob/master/src/mumble/GlobalShortcut_unix.cpp)
[GlobalShortcut_win](https://github.com/mumble-voip/mumble/blob/master/src/mumble/GlobalShortcut_win.cpp)
[GlobalShortcut_macx](https://github.com/mumble-voip/mumble/blob/master/src/mumble/GlobalShortcut_macx.mm)

Merging their impl with some changes is probably still worthwhile over trying to roll our own or trying to patch up UGlobalHotkey, as long as license allows.

anthonybilinski:
Re: combining the licenses, after some more research. It looks like
1. If all copyright holders agree, the license of the work could be changed.
    1.1 This is roughly impossible on open source projects, since there are so many copyright holders.
2. Without any of the copyright holders' permission, you can relicense 3-clause BSD under GPLvX
    2.1 This sometimes pisses off some holders of the BSD licenses, we should check with the mumble guys to prevent any bad blood.
3. You cannot remove the BSD license. You simply add GPL to the files as well. The whole project/binary is entirely GPL, the files contain licenses disclaimers for both, and the files are effectively only GPL since GPL is the more restrictive of the two.
4. The project would need to contain the BSD license
5. Any changes we made before changing license would be BSD, any changes after would be GPL.
6. Anyone could fork the files from our repo from a point before when GPL was added. (Of course they could just fork from Mumble too, so this doesn't really matter)

@qTox do any of you have philosophical or practical issue with moving forward with this?

sudden6:
I don't really have any issues with that. We need to add some more licenses for other 3rd party libs anyway.

Diadlo:
I don't understand much license stuff. I'm not against. Please be sure that there is no problem with toxcore license

tox-user:
If I understand correctly:
We will ask Mumble if we can relicense their code to GPL3. If they agree qTox will still be under GPL3 and the new code from them will also be under GPL3. Any future changes to qTox will continue to be under GPL3. There will also be BSD license in some files in the project in addition to GPL3, but because GPL is more restrictive, GPL will be the only active license? 

If my understanding is correct, I'm fine with it too :).

anthonybilinski:
@tox-user I think to ask "mumble" if we can change their BSD to GPL on these files we have to ask and get permission from every committer in mumble's history, which isn't really feasible.

Instead, because BSD is effectively "Include the BSD license" and "do whatever you want", and GPL is "Include the GPL license, publish changes of distributed work, distributions created using GPL code must be themselves GPL, derived work must be GPL, ship license with distributions etc. etc. etc." the union of the two licenses is "Include BSD license, include GPL license, and follow GPL rules." This wouldn't require mumble's permission since we're not breaking their BSD license at all by adding GPL.

The result would be:
Mumble's files with our GPL disclaimer saying the entire file is GPL
Under our GPL disclaimer, their existing BSD disclaimer saying parts of the files are covered by BSD
Any newly committed changes made to the files under GPL would be GPL (i.e. from our project, if needed). Optionally people could make modifications under BSD.
Our project, distributions, binaries etc. are all entirely GPL, 0% BSD.
Without significant effort to isolate the files, they will eventually have parts BSD and parts GPL, and get mixed to a point where no one can really use the BSD sections without pulling in some GPL code and having to follow GPL.

Asking if mumble is ok with us doing this is more of a pleasantry since some BSD holders can take offence to work being derived from theirs and being made "less free" with more "GPL freedom restrictions", but their license doesn't prevent it any way.

What I'm describing is section 2.2 of https://www.softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html

@Diadlo linking against GPL makes your result GPL, which won't change with the import of this code. I don't think this causes any conflicts or issues with using toxcore.

I'll start the work with talking with mumble and importing the code :)

anthonybilinski:
This PR has diverged a lot from the original code and the implementation plan has changed. I'm going to close this and will create a new PR shortly based on my new global hotkey issue: https://github.com/qTox/qTox/issues/4929

PR #4936, opened 2018-01-28
WIP feat(ui): Add global hotkey support, thanks to Mumble

### LINKED ISSUE #4929: Feature request: support global hotkeys, including global push to talk
This issue is a continuation from my previous PR https://github.com/qTox/qTox/pull/4925 which got carried away, changed topics a couple times, and ended up not being related to the original issue of hotkey configuration.

The want for global hotkeys was also brought up in the comments of https://github.com/qTox/qTox/issues/903

Our current push to talk implementation is only supported if the qTox window has focus. This is not very useful for any gaming use, where push to talk is often used and the game will always have focus.

Qt doesn't have a built in global hotkey solution, since it only captures key presses when it has focus. I've explored pulling in external libraries to handle the cross platform hotkey handling for us, instead of rolling our own which would both be a lot of work and seems like we must be re-doing work that has been done elsewhere.
Some candidates I've looked at:
1. Qxt Library: Large extension library that is no longer maintained. Not appealing.
2. [UGlobalHotkey](https://github.com/falceeffect/UGlobalHotkey): Doesn't offer key-up events so can't be used for hold to talk, and also consumes key presses interfering with typing.
3. [Mumble's implementation](https://github.com/mumble-voip/mumble): 3-clause BSD which is compatible with GPL. Has OS-specific implementations for win, macx and x11, and works well enough for them, offering key down and key up. I've done research on the license compatibility, and it looks like we must keep their BSD license disclaimers, add their BSD license into our source and have it available on releases, can add our own GPL, make our changes under GPL, our produced binaries would all still be GPL. Mumble would not be able to upstream any changes we made under GPL. Citation for strategy I'm looking at is under 2.2 of https://www.softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html

I talked briefly to some Mumble community and devs on IRC to double check that they agree it's within their license and also that they're ok with the idea of it. After a few hours in chat, there were a few comments agreeing it would be legal, and no comments disagreeing or expressing disapproval of the idea. This is no way releases their copyright or has any real legal meaning. Attached is a brief IRC log: [mumble-irc-log.txt](https://github.com/qTox/qTox/files/1657846/mumble-irc-log.txt)

There was also no disagreement from qTox members in the linked PR when asked about introducing some 3-clause BSD code.

I'm currently working on integrating Mumble's GlobalHotkey code and will PR shortly.

anthonybilinski:
once complete, will fix #4929 
complete so far:
- imported from mumble, added licensing.
- removed the majority of GlobalHotkey leaving only the bare minimum to support the API of adding or removing a hotkey, and linking that hotkey's keydown and keyup events.
- change GlobalHotkeyEngine from a static instance to a normal class
- change architecture by removing remap() logic, adding keys to map on add() call, and other confusing architectural decisions
- *nix implementation seems to be working pretty well, currently " ctrl + ` " toggles mute on calls on both key down and key up. Works to either "push to talk" or "push to mute" which can be toggled by just clicking the mute button in call

TODO:
- code full of TODO's added by me for parts of code hacked out
- win and macx implementations not yet worked on
- no UI to set shortcuts (@tox-user mentioned he may want to help out with this). Mumble had an implementation for this that used to be part of GlobalShortcut that I removed. We may want to pull that back in and re-use part of it.
- I don't understand the mapping from read-back keycodes to actual keys, making it tricky to set shortcuts. i.e. keycode values read are not ascii or UTF8. Maybe these key codes with align with our UI input so this won't be an issue.

Looking for input:
- API of GlobalShortcutEngine and GlobalShortcut
- Architecture of the classes
- Handling the cross-platform compilation, i.e. having a static platformInit() defined in GlobalShortcut.h which is implemented in whichever GlobalShortcut_(platform).cpp is compiled
- Testing on *nix

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4936)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -494,6 +494,14 @@ if (PLATFORM_EXTENSIONS)
     src/platform/timer_win.cpp
     src/platform/timer_x11.cpp
     src/platform/x11_display.cpp
+    src/platform/GlobalShortcut/GlobalShortcut.cpp
+    src/platform/GlobalShortcut/GlobalShortcut.h
+    src/platform/GlobalShortcut/GlobalShortcut_unix.cpp
+    src/platform/GlobalShortcut/GlobalShortcut_unix.h
+    src/platform/GlobalShortcut/GlobalShortcut_win.cpp
+    src/platform/GlobalShortcut/GlobalShortcut_win.h
+    src/platform/GlobalShortcut/GlobalShortcut_macx.cpp
+    src/platform/GlobalShortcut/GlobalShortcut_macx.h
   )
 endif()
 
@@ -585,6 +593,10 @@ if (MINGW)
   endif()
 endif()
 
+#set(${PROJECT_NAME}_SOURCES ${${PROJECT_NAME}_SOURCES}
+    #src/GlobalShortcut/GlobalShortcut_unix.cpp
+    #src/GlobalShortcut/GlobalShortcut_unix.h
+#)
 # the compiler flags for compiling C sources
 MESSAGE( STATUS "CMAKE_C_FLAGS: " ${CMAKE_C_FLAGS} )
 
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -124,6 +124,7 @@ if (PLATFORM_EXTENSIONS AND UNIX AND NOT APPLE)
   # Automatic auto-away support. (X11 also using for capslock detection)
   search_dependency(X11               PACKAGE x11 OPTIONAL)
   search_dependency(XSS               PACKAGE xscrnsaver OPTIONAL)
+  search_dependency(XI                PACKAGE xi)
 endif()
 
 if(APPLE)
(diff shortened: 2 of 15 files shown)


sudden6:
Keycodes could be the same as `QKeyEvent::nativeScanCode()`

anthonybilinski:
Thanks sudden6, printing out QKeyEvent::nativeScanCode() on keys captured do indeed match the values that GlobalHotkeyEngine is seeing. I'm not sure that's a good thing since [On Mac OS/X, this function is not useful, because there is no way to get the scan code from Carbon or Cocoa. The function always returns 1 (or 0 in the case explained above).](https://doc.qt.io/qt-5/qkeyevent.html#nativeScanCode), plus saving users shortcuts in terms of native scan codes might also cause problems if they transport their profile across OSes, something many do due to the lack of multi-device. I'll look into accepting adding shortcuts in terms of characters instead of scan codes.

anthonybilinski:
Closing in favour of https://github.com/qTox/qTox/pull/6185, offloading all of the cross-platform low level work to libuiohook rather than implementing it ourselves.

PR #4940, opened 2018-02-01
fix(chatlog): parse multi-length emoji properly

anthonybilinski:
Before, when multi-length emoji were next to any other character, they would not be transformed into images. With this change, emoji are replaced with images no matter where they are in the string. Parsing is now done by checking to see if two-character blocks are valid as a single UTF-32 character, or if they are truly two distinct characters. Because we no longer need white space before emoji, I also removed our addition of spaces on either side of an emoji when a user sends them.

[Before](https://imgur.com/a/kUaoU):  :(
[After](https://imgur.com/a/A1BHO):  :)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4940)

### DIFF
--- a/src/persistence/smileypack.cpp
+++ b/src/persistence/smileypack.cpp
@@ -258,21 +258,25 @@ QString SmileyPack::smileyfied(const QString& msg)
 {
     QMutexLocker locker(&loadingMutex);
     QString result(msg);
-    QRegularExpression exp("\\S+");
-    QRegularExpressionMatchIterator iter = exp.globalMatch(result);
-    int replaceDiff = 0;
-    while (iter.hasNext()) {
-        QRegularExpressionMatch match = iter.next();
-        QString key = match.captured();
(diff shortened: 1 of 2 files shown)


sudden6:
:lgtm_strong:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4940#-:-L4a5SuY68uYpdDtHAZj:b-jbsf8z)*

anthonybilinski:
Just want to point out that this will break backwards compatibility on some emoji with qTox clients from before this patch, since I removed the adding of spaces on either side of inserting emoji. Without those spaces, older qTox may parse emoji incorrectly and show just the font character. This seems correct since we don't want to have two unneeded spaces inserted every time someone adds an emoji, but will temporarily cause some new problems for some users.

PR #4941, opened 2018-02-01
Allow characters next to formatting characters

anthonybilinski:
NOTE: this PR is on top of https://github.com/qTox/qTox/pull/4940.

After https://github.com/qTox/qTox/pull/4940, adding formatting characters next to emoji no longer affects parsing. Because there's no longer a technical restriction on markdown characters next to other characters, and the discussion from https://github.com/qTox/qTox/issues/3047 seemed to lean towards it not being an intentional UX choice, I've enabled markdown next to other characters.
Before :(
image
After :)
image

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4941)

### DIFF
--- a/src/chatlog/textformatter.cpp
+++ b/src/chatlog/textformatter.cpp
@@ -23,51 +23,43 @@
 
 // clang-format off
 
-static const QString SINGLE_SIGN_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
-                                                          "[%1]"
+static const QString SINGLE_SIGN_PATTERN = QStringLiteral("[%1]"
                                                           "(?!\\s)"
                                                           "([^%1\\n]+?)"
                                                           "(?<!\\s)"
-                                                          "[%1]"
-                                                          "(?=$|[\\s\\n])");
+                                                          "[%1]");
 
-static const QString SINGLE_SLASH_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
-                                                           "/"
+static const QString SINGLE_SLASH_PATTERN = QStringLiteral("/"
                                                            "(?!\\s)"
                                                            "([^/\\n]+?)"
                                                            "(?<!\\s)"
-                                                           "/"
-                                                           "(?=$|[\\s\\n])");
+                                                           "/");
 
-static const QString DOUBLE_SIGN_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
-                                                          "[%1]{2}"
+static const QString DOUBLE_SIGN_PATTERN = QStringLiteral("[%1]{2}"
                                                           "(?!\\s)"
                                                           "([^\\n]+?)"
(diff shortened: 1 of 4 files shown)


anthonybilinski:
Working on failing tests

Diadlo:
@noavarice spent a lot of time on tests. BUG SHALL NOT PASS!

anthonybilinski:
@sudden6 @Diadlo 
I think I need some confirmation on intended behaviour before I can go ahead with changing markdown and the test.
1. What's the intended difference between single character and multi-character? i.e. `_test_` vs `__test__`
1.1 There are currently some difference in when we match, but they don't look intentional. i.e. `__test__test__` underlines the whole thing, but in the regex it looks like we're intentionally trying to non-greedily match `"([^\\n]+?)"` and end up continuing just because the next character is not whitespace, so `__test__ test__` stops at the first double underline.
1.2 On the other hand, with single characters, i.e. `_test_test_` we have no underline at all.

A lot of our difference in behaviour from single to double symbols is due to our requirements of whitespace, which when I removed changes behaviour.

Should double and single be different? If so what should the difference be? Is one greedy and one non-greedy? Or should it just be that `__` allows a single `_` with the underlined text, while `_` will stop at the first seen `_`? I don't see any mentioned difference of double and single in the [user manual](https://github.com/qTox/qTox/blob/master/doc/user_manual_en.md#message-styling)

sudden6:
I think the intention was to treat `__` the same as `_`, ie the two character syntax treats two chars as one.

Diadlo:
Agree with @sudden6 but probably it may be very hard to correct handle this sh*t

anthonybilinski:
That's easier to do than preserving the current behaviour, I'll make it happen :)

noavarice:
The only reason why was decided to add enclosing whitespace chars is that it is difficult to distinguish links and // markdown. This type of markdown should be removed, this character is using everywhere.
If someone will occasionally find markdown implemented with / somewhere, let us know) (but smth tells me that no one consider that character for markdown as good idea)

anthonybilinski:
I take it back, I was being [naive](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags). Dealing with links and other html tags makes things extremely complicated.

I agree with novarice, and I think a reasonable way to go forward with this would to follow a subset of a more official [markdown spec](http://spec.commonmark.org/0.28/), with standard characters, and use a proven markdown parsing algorithm or existing implementation instead of trying to do this in regex, which could also give us more features like escaping markdown characters.

I'm happy to close this PR, I don't think this is the right approach, and if we want to keep this approach, this isn't a good implementation.

anthonybilinski:
Closing PR in favour of changing markdown formatting more heavily and correctly, proposal [here](https://github.com/qTox/qTox/issues/4946)

PR #4951, opened 2018-02-08
fix(chatform): Broaden URL matching to any string following scheme

### LINKED ISSUE #4853: URLs containing certain characters (e.g., Unicode) aren't recognized properly anymore
(Split from #4420; #4593 seems to be the culprit.)

OS: Windows 7 Ultimate (64-bit), Service Pack 1
qTox version: v1.13.0
Commit hash: 531defd0aa66af1b128f7293bc08718ce2cc064f
toxcore: v0.1.10
Qt: v5.9.3

When a URL contains certain characters (like parentheses or Unicode), qTox now recognizes it only up to the offending character.

Try these links, for example:
https://en.wikipedia.org/wiki/Seal_(East_Asia)
https://ja.wikipedia.org/wiki/印章

anthonybilinski:
Fix #4853

Instead of searching strictly valid URLs, allow anything characters following scheme. This allows for UTF-8 characters used in other lanagues, as well as parentheses and other ASCII encoding. This will over-match some invalid URLs.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4951)

### DIFF
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -56,7 +56,7 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString& sender, const QSt
 
     // quotes (green text)
     text = detectQuotes(text, type);
-    text = highlightURL(text);
+    text = highlightURI(text);
 
     // text styling
     Settings::StyleType styleType = Settings::getInstance().getStylePreference();
--- a/src/chatlog/textformatter.cpp
+++ b/src/chatlog/textformatter.cpp
@@ -20,40 +20,42 @@
 #include "textformatter.h"
 
 #include <QRegularExpression>
+#include <QVector>
 
 // clang-format off
 
-static const QString SINGLE_SIGN_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
+// Note: escaping of '\' is only needed because QStringLiteral is broken by linebreak
+static const QString SINGLE_SIGN_PATTERN = QStringLiteral("(?<=^|\\s)"
                                                           "[%1]"
                                                           "(?!\\s)"
                                                           "([^%1\\n]+?)"
                                                           "(?<!\\s)"
                                                           "[%1]"
-                                                          "(?=$|[\\s\\n])");
+                                                          "(?=$|\\s)");
 
-static const QString SINGLE_SLASH_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
+static const QString SINGLE_SLASH_PATTERN = QStringLiteral("(?<=^|\\s)"
                                                            "/"
                                                            "(?!\\s)"
                                                            "([^/\\n]+?)"
                                                            "(?<!\\s)"
                                                            "/"
-                                                           "(?=$|[\\s\\n])");
+                                                           "(?=$|\\s)");
 
-static const QString DOUBLE_SIGN_PATTERN = QStringLiteral("(?<=^|[\\s\\n])"
+static const QString DOUBLE_SIGN_PATTERN = QStringLiteral("(?<=^|\\s)"
                                                           "[%1]{2}"
                                                           "(?!\\s)"
                                                           "([^\\n]+?)"
                                                           "(?<!\\s)"
                                                           "[%1]{2}"
-                                                          "(?=$|[\\s\\n])");
+                                                          "(?=$|\\s)");
 
-static const QString MULTILINE_CODE = QStringLiteral("(?<=^|[\\s\\n])"
+static const QString MULTILINE_CODE = QStringLiteral("(?<=^|\\s)"
                                                      "```"
                                                      "(?!`)"
                                                      "((.|\\n)+?)"
                                                      "(?<!`)"
                                                      "```"
-                                                     "(?=$|[\\s\\n])");
+                                                     "(?=$|\\s)");
 
 #define REGEXP_WRAPPER_PAIR(pattern, wrapper)\
 {QRegularExpression(pattern,QRegularExpression::UseUnicodePropertiesOption),QStringLiteral(wrapper)}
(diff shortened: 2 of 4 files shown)


tox-user:
A few more types of characters should be excluded from the url. 
Trailing `?!:`
and when the link is inside of ( ) or [ ] or quotes those surrounding characters should not be matched either. So I guess that also means trailing `) ] ' "`

tox-user:
Examples:
`(https://en.wikipedia.org/wiki/Tox_(protocol))` - trailing ) will be matched
`list of voip programs from https://en.wikipedia.org/wiki/Comparison_of_VoIP_software:` - trailing : will be matched
`href="https://tox.chat"` - the last " will be matched

anthonybilinski:
Good comments, have addressed them in the new version. Parsing ignores matching pairs of brackets etc. as well as ending punctuation, without ignoring brackets that are actually part of the URL. This doesn't add punycode on hover, although I think that would be a good feature.

anthonybilinski:
Just realized there's some overlap in functionality with the existing PR https://github.com/qTox/qTox/pull/4297 which has been open for months, it sounds like this addressed most of the comments from that as well.

sudden6:
@tox-user could you add the examples you gave to the test suite? This way we can ensure it doesn't get broken on accident.

tox-user:
Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/chatlog/textformatter.cpp, line 81 at r1](https://reviewable.io:443/reviews/qtox/qtox/4951#-L5eZE2wMYTr0rCVRTvw:-L5eZE2wMYTr0rCVRTvx:bu8tdse) ([raw file](https://github.com/qtox/qtox/blob/b29377e8d103ad1dae3864e3f17a8123313eebf0/src/chatlog/textformatter.cpp#L81)):*
>  QRegularExpression(QStringLiteral(R"((?<=^|[\s\n])[\S]*((www)\S+))"))

What does this part of regex do? `(?<=^|[\s\n])[\S]*`
`\s` should match all whitespace which also includes `\n`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4951)*

Diadlo:
@tox-user Yes, it should, but AFAIK it doesn't

anthonybilinski:
Tested [\s\n] vs \s and they both act equivalently (as they should according to regex spec), so removed the \n and [], updated PR.
>What does this part of regex do? (?<=^|[\s\n])[\S]*

`?<=` is a "positive lookbehind" https://www.regular-expressions.info/lookaround.html
it requires that the preceding character in this case matches either `^` (start of line) or (now) `\s`, whitespace. Then `[\S]*` means 0 or more of any non-whitespace characters, used to match leading characters like `(["` before the start of the scheme of the URI.

anthonybilinski:
I added tests, rebased to master tip, and fixed a bug I found with the tests where ending punctuation wouldn't be stripped if there were no surrounding characters :)

Diadlo:
`tox:0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF0123456789ABCDEF01234567890A` doesn't higlighted for me. Do you have any ideas?

anthonybilinski:
It was because we had `\\d` inside of a QStringLiteral for the ToxId matching. QStringLiteral already escapes the `\` so having a double `\\` breaks it. Fixed and updated PR, ToxId's didn't seem to match before this change either, but do now.

Diadlo:
Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/chatlog/textformatter.cpp, line 88 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAuNZRXcm4CiSP8dq:-L6BAuNZRXcm4CiSP8dr:b9as5bp) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L88)):*
> ```cpp
>     // Note: This does not match only strictly valid URLs, but we broaden search to any string following scheme to
>     // allow UTF-8 "IRI"s instead of ASCII-only URLs
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*((((http[s]?)|ftp)://)\S+))")),
> ```

Why do you think about splitting http[s] and ftp on 2 regeps? Or merge with `file|smb`

---

*[src/chatlog/textformatter.cpp, line 89 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BACsDuaSowtjK8VJX:-L6BACsDuaSowtjK8VJY:b-dg27kw) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L89)):*
> ```cpp
>     // allow UTF-8 "IRI"s instead of ASCII-only URLs
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*((((http[s]?)|ftp)://)\S+))")),
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*((file|smb)://([\S| ]*)))")),
> ```

Seems `[\S| ]` is equivalent to `\S`. Isn't it?

---

*[src/chatlog/textformatter.cpp, line 132 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BA_4WLaySUHmidgvP:-L6BA_4WLaySUHmidgvQ:b-sk8x6q) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L132)):*
> ```cpp
>     do {
>         matchFound = false;
>         for (QPair<QString, QString> const& surroundChars : URI_WRAPPING_CHARS)
> ```

Probably `auto` will be enought

---

*[src/chatlog/textformatter.cpp, line 156 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAhF3oVqEDdN5MpzZ:-L6BAhF3oVqEDdN5Mpz_:bjdxo3y) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L156)):*
> ```cpp
>     if (startOfBareUri != curValidationStartPos) {
>         strippedMatch.valid = false;
>     }
> ```

`} else {`

---

*[src/chatlog/textformatter.cpp, line 185 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAjZNRVBNU2zGyg_E:-L6BAjZNRVBNU2zGyg_F:b-ukjbp7) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L185)):*
> ```cpp
>             MatchingUri matchUri = stripSurroundingChars(match.capturedRef(uriWithWrapMatch),
>                    match.capturedStart(uriWithoutWrapMatch) - match.capturedStart(uriWithWrapMatch));
>             if (matchUri.valid == false) {
> ```

`!matchUri.valid`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4951)*

anthonybilinski:
Review status: 3 of 4 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/chatlog/textformatter.cpp, line 81 at r1](https://reviewable.io:443/reviews/qtox/qtox/4951#-L5eZE2wMYTr0rCVRTvw:-L6BK8wWtEKI3_x1zflH:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b29377e8d103ad1dae3864e3f17a8123313eebf0/src/chatlog/textformatter.cpp#L81)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

>  QRegularExpression(QStringLiteral(R"((?<=^|[\s\n])[\S]*((www)\S+))"))
What does this part of regex do? `(?<=^|[\s\n])[\S]*`
`\s` matches all whitespace which also includes `\n`.
</blockquote></details>

Done.

---

*[src/chatlog/textformatter.cpp, line 88 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAuNZRXcm4CiSP8dq:-L6BKbABdLoGPaG989pI:bqwydxc) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L88)):*
> Or merge with file|smb
As posted below, http/https have stricter criteria (no whitespace) than file/smb below. We could split all 4 into their own rules, but to me it seems like more duplication. If the only difference between different rules are the name of the schemes, combining rules with identical formats seems nicer to me.

---

*[src/chatlog/textformatter.cpp, line 89 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BACsDuaSowtjK8VJX:-L6BKL2h4S7RJb7zF6zF:bth5p94) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L89)):*

Seems `[\S| ]` is equivalent to `\S`. Isn't it?
</blockquote></details>

this is \S (uppercase S) not \s, which is the opposite. [\S| ] is saying "any non-whitespace character or the space character" since apparently file and smb URIs can have spaces in them.

---

*[src/chatlog/textformatter.cpp, line 132 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BA_4WLaySUHmidgvP:-L6BKzgyENPr_WQXc6v7:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L132)):*

Probably `auto` will be enought
</blockquote></details>

Done.

---

*[src/chatlog/textformatter.cpp, line 156 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAhF3oVqEDdN5MpzZ:-L6BLEE9GwhJAgm29LIf:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L156)):*

`} else {`
</blockquote></details>

Done.

---

*[src/chatlog/textformatter.cpp, line 185 at r2](https://reviewable.io:443/reviews/qtox/qtox/4951#-L6BAjZNRVBNU2zGyg_E:-L6BLZaiH01ykeD_oqXa:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/86b83bf532db83455d0010843f07040c86d89036/src/chatlog/textformatter.cpp#L185)):*

`!matchUri.valid`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4951)*

Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4951#-:-L6DWXCeVFEUJH7qK29E:b-jbsf8z)*

PR #4953, opened 2018-02-09
fix(core): Adapt qtox to new conferences state change callback.

Diadlo:
Patch from TokTok/qTox#7
Thanks @iphydf

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4953)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -503,7 +503,7 @@ void Core::onConnectionStatusChanged(Tox*, uint32_t friendId, TOX_CONNECTION sta
     }
 }
 
-void Core::onGroupInvite(Tox*, uint32_t friendId, TOX_CONFERENCE_TYPE type, const uint8_t* cookie,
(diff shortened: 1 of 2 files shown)


tox-user:
Last time I tested it, it was broken. Has anything changed?

Diadlo:
@tox-user AFAIK you tested TokTok/qTox
Not sure, that this change itself broke something

Diadlo:
@tox-user AFAIK you tested TokTok/qTox
Not sure, that this change itself broke something

tox-user:
The problem was with this PR + persistent group chats. This PR itself without persistent conferences causes segmentation fault on exit when you are in group chat.

Steps to reproduce:
1. Join group chat.
2. Exit qTox.
3. Notice the segmentation fault in the log.

The problem is either caused by this PR or by toxcore 0.2.0.

Diadlo:
It's strange. I tried it with self created empty group and with group with another person, but haven't segfault (master toxcore was used)

PR #4982, opened 2018-03-03
feat: Add the cmake option USE_CCACHE

yurivict:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/4982)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,6 +11,7 @@ option(PLATFORM_EXTENSIONS "Enable platform specific extensions, requires extra 
 option(USE_FILTERAUDIO "Enable the echo canceling backend" ON)
 # AUTOUPDATE is currently broken and thus disabled
 option(AUTOUPDATE "Enable the auto updater" OFF)
+option(USE_CCACHE "Use ccache when available" ON)
 
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE Debug)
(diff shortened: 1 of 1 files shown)


Diadlo:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/4982#-:-L6eUvpNCCzL5nNmOpqV:b-jbsf8z)*

CMakeLists.txt
```diff
@@ -69,10 +70,16 @@ endif()
...
+if (USE_CCACHE)
+  find_program(CCACHE_FOUND ccache)
+  if(CCACHE_FOUND)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+    message(STATUS "using ccache")
+  else()
+    message(STATUS "ccache not found")
+  endif()
+else()
+  message(STATUS "ccache not enabled")
```
mschilli87: This last message has very little information:
Imagine you try to build and see it. Would you understand what it means?
I guess it's not too important as you actively need to change a default to get there, but still it might be worth improving.
How about "ccache disabled; set option USE_CCACHE to use ccache if available"?

Diadlo: I'm ok with new text, but it doesn't make sense, because `USE_CCACHE` is on by default => you will see this message only if explicitly specified `-D USE_CCACHE=OFF`

yurivict: The text has been clarified.

PR #5009, opened 2018-03-16
docs(maint): Change patch release commit strategy to reduce duplication

anthonybilinski:
Prefer committing patch commits on previous patch and merging up to master head, instead of previously committing on master and cherry picking required changes back to patch head, duplicating each commit in history.

before, with commits duplicated and cherry picked back from master to make a patch release:
before
after this change, with commits made on patch and merged up to master:
after

I'm not very proficient with git, so let me know if I'm using the wrong terminology, or if this is a bad idea in general.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5009)

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -199,10 +199,11 @@ Follow steps for adding translations from Weblate up to step 5. Next:
 - Generate changelog with `clog`.
   - In a `MAJOR`/`MINOR` release tag should include information that changelog
     is located in the `CHANGELOG.md` file, e.g. `For details see CHANGELOG.md`
-- To release a `PATCH` version after non-fix changes have landed on `master`
-  branch, checkout latest `MAJOR`/`MINOR` version and `git cherry-pick -x`
-  commits from `master` that you want `PATCH` release to include. Once
-  cherry-picking has been done, tag HEAD of the branch.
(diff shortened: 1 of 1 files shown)


sudden6:
I prefer the current strategy, but am not strongly opposed to this change.

@Diadlo @tox-user what do you think?

tox-user:
I don't know any more. So many things about how we do releases are wrong. Will this help to reduce the amount of bugs and broken releases?

sudden6:
@tox-user this will definitely not change the amount of bugs :DD

A perfect release is very unlikely^^

Diadlo:
I'm satisfied with both stateges

sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/5009#-:-L7ot5tQF4WYfyM9mEzi:b-jbsf8z)*

anthonybilinski:
I like this less and less the more I think about it :P

My proposed strategy works if for each commit we know if it should be applied to the patch branch or not, which also requires knowing that we will have a patch release.

In reality, we're pushing both features and fixes to master daily, and if something serious comes up, we decide to do a patch release. I think it's easier to decide at that time what bug fixes to include in the patch release and backport them. Going to close without merging.

Issue #5013, opened 2018-03-16
Usernames are broken


tox-user:
Aliases in qTox have been broken for a long time (maybe forever). They would always be set with every name change of a friend and they were set to their new username. The result was that in addition to updating friend's name, we would also set their alias to their new name. Such behavior was impossible to notice, because of how qTox displays aliases - they replace a friend's name. There was also a second issue. When a friend changed their name, we would only see the changed name on friends list, but not in group chat.

PR https://github.com/qTox/qTox/pull/4970 was supposed to fix both of those problems and it almost did. It couldn't, because the previous broken behavior of aliases caused everyone's friends to have an alias set and the PR made aliases not reset with name change anymore. That means everyone running qTox 1.14.0 can't see name changes of their friends. They are blocked from seeing them, because of aliases that were automatically set for everyone. Most people do not realize they were using aliases without knowing. Usernames are simply broken now.

This problem is not easy to fix, but I propose we bring back the previous behavior of aliases for 1.14.1 release to make names usable for now. Only this time we should make aliases just reset on name change and not change them to a new name of a user. We should also add a new variable "wasAliasSet" to each friend, which would let us determine if an alias was automatic (set by previous broken behavior) or set by the user. In result we could reset all old, automatically set aliases in a future release with a little less damage to users.

So the plan would be to:
1. Make aliases reset (set them to an empty string) on name change in 1.14.1
2. Reset all automatic aliases in 1.15 or later and make user aliases stick forever - like they do now in 1.14

anthonybilinski:
I agree the behaviour of users not seeing name changes in v1.14.0 is broken. The complete solution would probably be to add some config file bool of "was this alias manually set by user" for new versions, if that's not set to true reset their alias and show the name, when the user sets a manual alias then set it to true and save the alias separate from the friend name.

I'm ok with "Make aliases reset (set them to an empty string) on name change in 1.14.1" since that should just be a 1-ish line change and will keep behaviour consistent with previous releases, where a friend changing their name resets your "alias". I don't want to include more than that into 1.14.1 since I want to get it out ASAP due to multiple major crash fixes after 1.14.0, so want to keep testing/complexity of change minimal.

PR #5015, opened 2018-03-17
fix(name): Clear alias on name change so that name changes are visible

anthonybilinski:
Temporary work around for #5013

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5015)

### DIFF
--- a/src/model/friend.cpp
+++ b/src/model/friend.cpp
@@ -52,6 +52,11 @@ void Friend::setName(const QString& _name)
 
     // save old displayed name to be able to compare for changes
     const auto oldDisplayed = getDisplayedName();
+    if (userName == userAlias) {
+        userAlias.clear(); // Because userAlias was set on name change before (issue #5013)
+                           // we clear alias if equal to old name so that name change is visible.
+                           // TODO: We should not modify alias on setName.
+    }
     if (userName != name) {
         userName = name;
         emit nameChanged(friendId, name);


anthonybilinski:
@tox-user does this seem like enough to work-around #5013 for the 1.14.1 release, while a more complete solution remembering if the alias was set by the user or automatically, and possibly having a way to view both the actual username and the alias together for 1.15?

@sudden6 do you think this is reasonable and doesn't hurt your recent alias + name major change?

sudden6:
I'm still not really in favor of permanently deleting the aliases, but I don't have a better solution, so I can live with that. I don't think it will conflict with my other changes.

Diadlo:
Maybe remove alias if it's equal to current username?

anthonybilinski:
I think that's a good idea, updated behaviour. Now the alias will be cleared on name change only if old name is equal to alias. This will fix all the cases of auto-set alias to name changes, so that name changes will remain visible. If a manual alias is set (as long as it never conflicts with the name), it will remain set through name change.

sudden6:
else :lgtm_strong: 

---

Reviewed 1 of 1 files at r1.

---

*[src/model/friend.cpp, line 55 at r1](https://reviewable.io:443/reviews/qtox/qtox/5015#-L7vW4bpMKeAC8RMXnsy:-L7vW4bqody0LuV_MhSW:bczityj) ([raw file](https://github.com/qtox/qtox/blob/fa71e6057302bc354ec7adead5a6b5a62e025bb4/src/model/friend.cpp#L55)):*
> ```cpp
>     // save old displayed name to be able to compare for changes
>     const auto oldDisplayed = getDisplayedName();
>     if (userName == userAlias) {
> ```

maybe add a TODO that this is a workaround

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/5015#-:-L7vWAAQZgPfqvDZKdQN:ba3ovo0)*

tox-user:
@anthonybilinski
>This will fix all the cases of auto-set alias to name changes

Not for people using 1.14.0, whose friends changed their name after updating to 1.14.0. Those people will have a different alias set than their friend's name.

tox-user:
This solution is not enough. Most people don't update qTox with every release, so it will not fix the problem for many people, who stayed on 1.14.0 for some amount of time. We should try to fix the problem for everyone. The original solution was better.

With this change the alias will most likely break when:
1. A friend with no alias, with user name XYZ changes their name to XYY.
2. We notice it and set their alias (for the first time) to XYZ, which is their previous name.
3. After some time they change their name back to XYZ.
4. Their alias will reset. This is because their userName is now equal to userAlias.

anthonybilinski:
Toxed with tox-user, going to summarize our chat to explain decision to keep using the current implementation for 1.14.1:
>Not for people using 1.14.0, whose friends changed their name after updating to 1.14.0. Those people will have a different alias set than their friend's name.

Yes, true. These people will have alias set and not see name changes, like the behaviour on 1.14.0
>With this change the alias will most likely break when

Yes, true again. This solution is more of an alternative to resetting on every name change, but it still sometimes does reset when it doesn't need to, since it doesn't add any versioning to alias to know if it was set pre or post 1.14.0.

The way I view it is that here there are corner cases that break aliases for people who updated from 1.13.0 to 1.14.1 (their alias can be wiped by their friend changing name to match their alias) and there are corner cases that break names for people who updated from 1.14.0 to 1.14.1 (their alias could not match the name of their friend if their friend changed names while they were on 1.14.0, in which case their alias will not be cleared). To me, these two rare cases are less bad than the case of always resetting alias on every name change. In @tox-user's view keeping 1.13.0 behaviour of always resetting alias would be better.

In either case, both cases will fixed when we add proper persistent versioning of alias (i.e. was it set automatically or manually set) in 1.15.0, which should cover all cases better.

Issue #5020, opened 2018-03-19
HTML in sent messages gets rendered to screen when reloading previous message in chatbox


Toj:
##### Brief Description

After sending a message containing HTML code and pressing the up arrow on the keyboard (to load the previous sent message again into the chatbox), the HTML code is stripped out, but it's effects are rendered to the new message assuming the entire message isn't stripped out.

Eg, if you send someone 'TEXT' that is styled with HTML, and press the up arrow in the chatbox to try and send it again, it will show 'TEXT' styled with whatever HTML was stripped out.

# [See this video](https://i.imgur.com/DpkNQlO.mp4)

##### Versions

OS: Affects both Linux & Windows on at least the latest qTox release
qTox version: 1.13, latest

##### Reproducible: 
Always

##### Expected Behavior
For the HTML code to not be stripped out or applied to the message; for it to simply load the previous sent message into the chatbox

tox-user:
I can reproduce this on 1.14 as well.

anthonybilinski:
This was merged as ebdc675cbc2dd0054306d461c382568c6148bf60.

PR #5022, opened 2018-03-20
fix(chattextedit): ChatTextEdit bug fix for issue #5020

HoraceBianchon:
In this bugfix last message in chatbox shows correctly.
About application of HTML-tags to the message: We should not do this.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5022)


sudden6:
What do you exactly mean by `About application of HTML-tags to the message: We should not do this.`? should we not use HTML for message styling? why? what's the alternative?

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/widget/tool/chattextedit.cpp, line 28 at r1](https://reviewable.io:443/reviews/qtox/qtox/5022#-L83chEJTCuV5HQG8uY4:-L83chEKwiEGcw_t9aQP:blonx0s) ([raw file](https://github.com/qtox/qtox/blob/ab61bb1bf776e77c28a0504012f05f12fab3279d/src/widget/tool/chattextedit.cpp#L28)):*
> ```cpp
> #include <QKeyEvent>
> #include <QMimeData>
> #include <QDebug>
> ```

This include seems unnecessary, please remove

---

*[src/widget/tool/chattextedit.cpp, line 62 at r1](https://reviewable.io:443/reviews/qtox/qtox/5022#-L83cp33SgxKkZrjUjWa:-L83cp33SgxKkZrjUjWb:b7ryyzh) ([raw file](https://github.com/qtox/qtox/blob/ab61bb1bf776e77c28a0504012f05f12fab3279d/src/widget/tool/chattextedit.cpp#L62)):*
> ```cpp
>     }
>     if (key == Qt::Key_Up && this->toPlainText().isEmpty()) {
> 	    
> ```

extra new line, please remove

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/5022#-:-L83cu9XylmipR5o9IV_:bh4a5au)*

HoraceBianchon:
I'm sure You know that HTML tags family very big, and some of them are 
have special interpretation. And admitting them all will lead to many problems.
One solution is admitting only simple tags, e.g h1, h2,--, h5, span, p, u, and so on.
But what would be if user want send message with html tags, e.f he/she want send example of html code to her/his colegue, in that situation what he/she must to do. And many use cases can be if admitting some or all html tags. That  is a reason why  not to do it.
Thanks.

sudden6:
As far as I'm aware of, no HTML tags will be specially rendered in qTox, if that's the case somewhere else, it's probably a bug.

Can you also please squash your commit into one?

sudden6:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/qtox/qtox/5022)*

anthonybilinski:
@HoraceBianchon I think the original fix in the PR is good. You can reopen this PR and update it by force pushing new commits to your master branch on your qTox repo (including removing commits), and they'll automatically update here. In the future, it's best for you if you use a new branch for the PR so that updating your master doesn't update your PR. I encourage you to re-open this PR :)

PR #5068, opened 2018-04-10
fix(audio): apply gain to both audio channels

### LINKED ISSUE #5039: Voice sounds robotic and distorted with non-default gain
##### Brief Description

OS: Windows
qTox version: v1.14.1
toxcore:  v0.2.1
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Audio call, group or friend.

##### Observed Behavior
1. My voice sounds very robotic/distorted.

##### Expected Behavior
1. Not that.

##### Additional Info
Mic loopback on Windows sounded fine. Network was not saturated, logs were not full of bandwidth notifications from toxcore. Other VOIP worked fine. Distortion was present across multiple days. Friend also on Windows v1.14.1 experiences the same issue. I never saw this issue on v1.13.0.

anthonybilinski:
Fix #5039

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5068)

### DIFF
--- a/src/audio/audio.h
+++ b/src/audio/audio.h
@@ -62,6 +62,8 @@ public:
 
     virtual qreal outputVolume() const = 0;
     virtual void setOutputVolume(qreal volume) = 0;
+    virtual qreal maxOutputVolume() const = 0;
+    virtual qreal minOutputVolume() const = 0;
 
     virtual qreal minInputGain() const = 0;
     virtual void setMinInputGain(qreal dB) = 0;
@@ -73,10 +75,7 @@ public:
     virtual void setInputGain(qreal dB) = 0;
 
     virtual qreal minInputThreshold() const = 0;
-    virtual void setMinInputThreshold(qreal dB) = 0;
-
     virtual qreal maxInputThreshold() const = 0;
-    virtual void setMaxInputThreshold(qreal dB) = 0;
 
     virtual qreal getInputThreshold() const = 0;
     virtual void setInputThreshold(qreal percent) = 0;
@@ -109,8 +108,9 @@ protected:
     // Public default audio settings
     static constexpr uint32_t AUDIO_SAMPLE_RATE = 48000;
     static constexpr uint32_t AUDIO_FRAME_DURATION = 20;
-    static constexpr uint32_t AUDIO_FRAME_SAMPLE_COUNT =
+    static constexpr uint32_t AUDIO_FRAME_SAMPLE_COUNT_PER_CHANNEL =
         AUDIO_FRAME_DURATION * AUDIO_SAMPLE_RATE / 1000;
+    uint32_t AUDIO_FRAME_SAMPLE_COUNT_TOTAL = 0;
 
 signals:
     void frameAvailable(const int16_t* pcm, size_t sample_count, uint8_t channels,
--- a/src/audio/backend/openal.cpp
+++ b/src/audio/backend/openal.cpp
@@ -32,6 +32,18 @@
 
 #include <cassert>
 
+namespace {
+    void applyGain(int16_t* buffer, uint32_t bufferSize, qreal gainFactor)
+    {
+        for (quint32 i = 0; i < bufferSize; ++i) {
+            // gain amplification with clipping to 16-bit boundaries
+            buffer[i] = qBound<int16_t>(std::numeric_limits<int16_t>::min(),
+                                   qRound(buffer[i] * gainFactor),
+                                   std::numeric_limits<int16_t>::max());
+        }
+    }
+}
+
 /**
  * @class OpenAL
  * @brief Provides the OpenAL audio backend
@@ -181,7 +193,7 @@ void OpenAL::setMaxInputGain(qreal dB)
 /**
  * @brief The minimum threshold value for an input device.
  *
- * @return minimum threshold percentage
+ * @return minimum normalized threshold
  */
 qreal OpenAL::minInputThreshold() const
 {
@@ -190,36 +202,14 @@ qreal OpenAL::minInputThreshold() const
 }
 
 /**
- * @brief Set the minimum allowed threshold percentage
+ * @brief The maximum normalized threshold value for an input device.
  *
- * @note Default is 0%; usually you don't need to alter this value;
- */
-void OpenAL::setMinInputThreshold(qreal percent)
-{
-    QMutexLocker locker(&audioLock);
-    minInThreshold = percent;
-}
-
-/**
- * @brief The maximum threshold value for an input device.
- *
- * @return maximum threshold percentage
+ * @return maximum normalized threshold
  */
 qreal OpenAL::maxInputThreshold() const
 {
     QMutexLocker locker(&audioLock);
     return maxInThreshold;
-}
-
-/**
- * @brief Set the maximum allowed threshold percentage
- *
- * @note Default is 40%; usually you don't need to alter this value.
- */
-void OpenAL::setMaxInputThreshold(qreal percent)
-{
-    QMutexLocker locker(&audioLock);
-    maxInThreshold = percent;
 }
 
 void OpenAL::reinitInput(const QString& inDevDesc)
@@ -311,15 +301,15 @@ bool OpenAL::initInput(const QString& deviceName, uint32_t channels)
     // TODO: Try to actually detect if our audio source is stereo
     this->channels = channels;
     int stereoFlag = channels == 1 ? AL_FORMAT_MONO16 : AL_FORMAT_STEREO16;
-    const uint32_t sampleRate = AUDIO_SAMPLE_RATE;
-    const uint16_t frameDuration = AUDIO_FRAME_DURATION;
     const int bytesPerSample = 2;
-    const int safetyFactor = 2;
-    const ALCsizei bufSize = AUDIO_FRAME_SAMPLE_COUNT * channels *  bytesPerSample * safetyFactor;
+    const int safetyFactor = 2; // internal OpenAL ring buffer. must be larger than our inputBuffer to avoid the ring
+                                // from overwriting itself between captures.
+    AUDIO_FRAME_SAMPLE_COUNT_TOTAL = AUDIO_FRAME_SAMPLE_COUNT_PER_CHANNEL * channels;
+    const ALCsizei ringBufSize = AUDIO_FRAME_SAMPLE_COUNT_TOTAL * bytesPerSample * safetyFactor;
 
     const QByteArray qDevName = deviceName.toUtf8();
     const ALchar* tmpDevName = qDevName.isEmpty() ? nullptr : qDevName.constData();
-    alInDev = alcCaptureOpenDevice(tmpDevName, sampleRate, stereoFlag, bufSize);
+    alInDev = alcCaptureOpenDevice(tmpDevName, AUDIO_SAMPLE_RATE, stereoFlag, ringBufSize);
 
     // Restart the capture if necessary
     if (!alInDev) {
@@ -327,7 +317,7 @@ bool OpenAL::initInput(const QString& deviceName, uint32_t channels)
         return false;
     }
 
-    inputBuffer = new int16_t[AUDIO_FRAME_SAMPLE_COUNT * channels];
+    inputBuffer = new int16_t[AUDIO_FRAME_SAMPLE_COUNT_TOTAL];
     setInputGain(Settings::getInstance().getAudioInGainDecibel());
     setInputThreshold(Settings::getInstance().getAudioThreshold());
 
@@ -542,17 +532,22 @@ void OpenAL::playMono16SoundCleanup()
  *
  * @param[in] buf   the current audio buffer
  *
- * @return volume in percent of max volume
+ * @return normalized volume between 0-1
  */
 float OpenAL::getVolume()
 {
-    quint32 samples = AUDIO_FRAME_SAMPLE_COUNT * channels;
-    float sum = 0.0;
+    const quint32 samples = AUDIO_FRAME_SAMPLE_COUNT_TOTAL;
+    const float rootTwo = 1.414213562; // sqrt(2), but sqrt is not constexpr
+    // calculate volume as the root mean squared of amplitudes in the sample
+    float sumOfSquares = 0;
     for (quint32 i = 0; i < samples; i++) {
         float sample = static_cast<float>(inputBuffer[i]) / std::numeric_limits<int16_t>::max();
-        sum += qAbs(sample);
+        sumOfSquares += std::pow(sample , 2);
     }
-    return sum / samples;
+    const float rms = std::sqrt(sumOfSquares/samples);
+    // our calculated normalized volume could possibly be above 1 because our RMS assumes a sinusoidal wave
+    const float normalizedVolume = std::min(rms * rootTwo, 1.0f);
+    return normalizedVolume;
 }
 
 /**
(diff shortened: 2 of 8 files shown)


sudden6:
Reviewed 2 of 8 files at r1.
Review status: 2 of 8 files reviewed at latest revision, all discussions resolved.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9pntR99Ay9B6xQAqnc:b-d4aro6) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*
> ```cpp
>     const uint16_t frameDuration = AUDIO_FRAME_DURATION;
>     const int bytesPerSample = 2;
>     const int safetyFactor = 2;
> ```

I see no reason why we need this safety factor, `inputBuffer` is only accessed with `AUDIO_FRAME_SAMPLE_COUNT_PER_CHANNEL*channel` elements. As long as `channel` is always correct this factor should not be needed.

---

*[src/audio/backend/openal.cpp, line 532 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pttMy-HIHlqrOxn1k:-L9pttMy-HIHlqrOxn1l:bajj8g4) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L532)):*
> ```cpp
> float OpenAL::getVolume()
> {
>     static const quint32 samples = AUDIO_FRAME_SAMPLE_COUNT_TOTAL;
> ```

I don't think static is correct here

---

*[src/audio/backend/openal.cpp, line 533 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pqJfJAXW8uLifz-gx:-L9pqJfJAXW8uLifz-gy:b497tyx) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L533)):*
> ```cpp
> {
>     static const quint32 samples = AUDIO_FRAME_SAMPLE_COUNT_TOTAL;
>     static const float twoOverRootTwo = 2/std::sqrt(2);
> ```

I think you can make this a `constexpr` to make the compiler do more optimization here, also the formula is equivalent to `std::sqrt(2)`. Where do you have this formula from? Especially the factor `2/sqrt(2)`?

---

*[src/audio/backend/openal.cpp, line 572 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pq3p-CklKVjXl1NMj:-L9pq3p-CklKVjXl1NMk:b-832x9d) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L572)):*
> ```cpp
>         isActive = true;
>         emit startActive(voiceHold);
>     } else if (!isActive) {
> ```

This change doesn't really make sense to me, can you explain why this is an `else if`?

---

*[src/widget/form/settings/avform.cpp, line 529 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppEYoB3INY4P1RV-a:-L9ppEYoB3INY4P1RV-b:b-skfgnv) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L529)):*
> ```cpp
> 
>     QString deviceName;
>     if (inputEnabled)
> ```

braces

---

*[src/widget/form/settings/avform.cpp, line 548 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppJUc4vpjgfGM7XRA:-L9ppJUc4vpjgfGM7XRB:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L548)):*
> ```cpp
> 
>     QString deviceName;
>     if (outputEnabled)
> ```

same

---

*[src/widget/form/settings/avform.cpp, line 561 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppSKNEhf9uM4b0V3N:-L9ppSKNEhf9uM4b0V3O:b-7s4sqq) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L561)):*
> ```cpp
> void AVForm::on_playbackSlider_valueChanged(int sliderSteps)
> {
>     int settingsVolume = getValueFromSteps(sliderSteps, audioSettings->getOutVolumeMin(), audioSettings->getOutVolumeMax());
> ```

const

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

anthonybilinski:
Review status: 2 of 8 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9px5fl8IQPfAf8x9v4:b-ti842x) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

I see no reason why we need this safety factor, `inputBuffer` is only accessed with `AUDIO_FRAME_SAMPLE_COUNT_PER_CHANNEL*channel` elements. As long as `channel` is always correct this factor should not be needed.
</blockquote></details>

This wasn't added by me, just named, but I think that we retrieve 20ms of data at a time, but if the OpenAL recording buffer was only 20ms, we'd have some lost audio from the time 20ms was available until the time when we retrieved it. By having this safety factor and having the OpenAL buffer larger, it gives us time to fetch the 20ms of data before the OpenAL buffer is full.

---

*[src/audio/backend/openal.cpp, line 532 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pttMy-HIHlqrOxn1k:-L9pxdgf5NjzzgbE8udg:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L532)):*

I don't think static is correct here
</blockquote></details>

Done.

---

*[src/audio/backend/openal.cpp, line 533 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pqJfJAXW8uLifz-gx:-L9py7E0-nvsN-9sjfTj:b-tl1k21) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L533)):*
> constexpr
good idea. I had it as static to only computer on first call then save the value, but constexpr is even better.
>equivalent to std::sqrt(2)
whoops. fixed.
>Where do you have this formula from
https://en.wikipedia.org/wiki/Root_mean_square
It represents the power of the sound, which is roughly equivalent to volume.
>Especially the factor 2/sqrt(2)
"Peak values can be calculated from RMS values from the above formula, which implies VP = VRMS × √2, assuming the source is a pure sine wave. "
Since our sound is roughly sinusoidal, RMS value of a peaking sine wave will still only be 0.7. Showing equivalent peak sound instead will better fill our entire volume display bar.

---

*[src/audio/backend/openal.cpp, line 572 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pq3p-CklKVjXl1NMj:-L9pz93KB79SeJR4furb:bbb6kst) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L572)):*
> can you explain why this is an else if
isActive may be true or false even if the first block is missed because it stays true for a period after being set.. A strict else would not take that time period of staying activated after volume threshold dips into account. The general reason for setting volume = 0 in this case is so that the volume display bar shows 0 when their voice is being clipped by activation, so a user can better see how they will sound to peers.

---

*[src/widget/form/settings/avform.cpp, line 529 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppEYoB3INY4P1RV-a:-L9pznF_AeEBko5THkSc:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L529)):*

braces
</blockquote></details>

Done.

---

*[src/widget/form/settings/avform.cpp, line 548 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppJUc4vpjgfGM7XRA:-L9pzrfbBbMtluIBjX3Y:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L548)):*

same
</blockquote></details>

Done.

---

*[src/widget/form/settings/avform.cpp, line 561 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9ppSKNEhf9uM4b0V3N:-L9q-2Nt7qlu71AZdvCT:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/widget/form/settings/avform.cpp#L561)):*

const
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

sudden6:
Reviewed 3 of 8 files at r1.
Review status: 5 of 8 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9q0qCaD9d_F1wZo59N:b-8ni6ss) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

This wasn't added by me, just named, but I think that we retrieve 20ms of data at a time, but if the OpenAL recording buffer was only 20ms, we'd have some lost audio from the time 20ms was available until the time when we retrieved it. By having this safety factor and having the OpenAL buffer larger, it gives us time to fetch the 20ms of data before the OpenAL buffer is full.
</blockquote></details>

If I read the OpenAL docs correctly the data is internally buffered and only read out in chunks with `captureSamples` in `doInput`. If you find time you can test it, else add a todo comment to recheck this.

---

*[src/audio/backend/openal.cpp, line 533 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pqJfJAXW8uLifz-gx:-L9q01Bc5uHizZnXVnpi:b82accd) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L533)):*

> constexpr
good idea. I had it as static to only computer on first call then save the value, but constexpr is even better.
>equivalent to std::sqrt(2)
whoops. fixed.
>Where do you have this formula from
https://en.wikipedia.org/wiki/Root_mean_square
It represents the power of the sound, which is roughly equivalent to volume.
>Especially the factor 2/sqrt(2)
"Peak values can be calculated from RMS values from the above formula, which implies VP = VRMS × √2, assuming the source is a pure sine wave. "
Since our sound is roughly sinusoidal, RMS value of a peaking sine wave will still only be 0.7. Showing equivalent peak sound instead will better fill our entire volume display bar.
</blockquote></details>

ok, this seems to match the german wikipedia so it must be right^^

---

*[src/audio/backend/openal.cpp, line 572 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pq3p-CklKVjXl1NMj:-L9q2Eqi9h3R42-W5qdN:b-o1agc7) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L572)):*

> can you explain why this is an else if
isActive may be true or false even if the first block is missed because it stays true for a period after being set.. A strict else would not take that time period of staying activated after volume threshold dips into account. The general reason for setting volume = 0 in this case is so that the volume display bar shows 0 when their voice is being clipped by activation, so a user can better see how they will sound to peers.
</blockquote></details>

ok, understand now

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

anthonybilinski:
Review status: 5 of 8 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9r0J1i1N4IuRJ3Ftue:b-8rkhbn) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

If I read the OpenAL docs correctly the data is internally buffered and only read out in chunks with `captureSamples` in `doInput`. If you find time you can test it, else add a todo comment to recheck this.
</blockquote></details>

From http://jerli.info/docu/Master_1/Son/OpenAL_Programmers_Guide.pdf on "alcCaptureStart" page, "alcCaptureStart will begin recording to an internal ring buffer of the size specified when opening the capture device.". So I think yes it is buffered internally, but the amount it buffers internally is what we're configuring with bufSize on alcCaptureOpenDevice.

---

*[src/audio/backend/openal.cpp, line 533 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pqJfJAXW8uLifz-gx:-L9r1azHD8pyoLF-6ydq:bchgeii) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L533)):*

ok, this seems to match the german wikipedia so it must be right^^
</blockquote></details>

undid constexpr because it fails to compile with travis for mac osx. Just changed it to a const "magic" number of 1.414213562 with a comment.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

sudden6:
Reviewed 1 of 8 files at r1, 1 of 2 files at r2.
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9sc2C1CNAxJr4ofVRm:b-7vlozt) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

From http://jerli.info/docu/Master_1/Son/OpenAL_Programmers_Guide.pdf on "alcCaptureStart" page, "alcCaptureStart will begin recording to an internal ring buffer of the size specified when opening the capture device.". So I think yes it is buffered internally, but the amount it buffers internally is what we're configuring with bufSize on alcCaptureOpenDevice.
</blockquote></details>

ah, I see now, could you add a comment to explain `safetyFactor`?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

anthonybilinski:
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9uvM7s9MU9aQeUF_N8:b-5pujfd) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

ah, I see now, could you add a comment to explain `safetyFactor`?
</blockquote></details>

Done. added a comment and renamed "bufSize" to "ringBufSize".

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068)*

sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r3.

---

*[src/audio/backend/openal.cpp, line 299 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5068#-L9pntR99Ay9B6xQAqnb:-L9v-Pmj7yweRK3An1rQ:b-ljprvs) ([raw file](https://github.com/qtox/qtox/blob/d47e9321f900d907440eea2df13e6a2781a0b8ec/src/audio/backend/openal.cpp#L299)):*

Done. added a comment and renamed "bufSize" to "ringBufSize".
</blockquote></details>

ok, perfect

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5068#-:-L9v06V_7Bsk2Nz8bKiU:b-jbsf8z)*

Issue #5078, opened 2018-04-13
Reduce message history memory usage


anthonybilinski:
##### Steps to reproduce
1. Load a week of history coming out of a db that's <1MB on disk.

##### Observed Behavior
1. 400MB+ of resident memory in qTox. Sometimes over 1GB resident with a couple months at once. Common use case when searching back in history.
2. Memory usage does not go down after clearing message history, forced to restart client to reclaim memory.

##### Expected Behavior
1. Use a sane amount of memory.
2. Release the memory when messages are deleted.
3. (Bonus) Not have to display massive amounts of history in text to see old history. Some kind of incremental loading and freeing as you scroll would be good.

##### Additional Info
Heap profile: [massif.qTox.log](https://github.com/qTox/qTox/files/1906088/massif.qTox.log). View with `massif-visualizer massif.qTox.log`
The good stuff is at
n = 68, time(i) = 724,197,684,190

#### Current suspects
1. We load a stylesheet off disk for every text line:
`, defStyleSheet(Style::getStylesheet(QStringLiteral(":/ui/chatArea/innerStyle.css"), font))`
```
| | |     ->22.26% (71,975,904B) 0x397919: Text::Text(QString const&, QFont const&, bool, QString const&, QColor) (text.cpp:42)
| | |     | ->07.42% (24,002,944B) 0x4B839D: Timestamp::Timestamp(QDateTime const&, QString const&, QFont const&) (timestamp.cpp:23)
| | |     | | ->07.40% (23,926,112B) 0x4B7605: ChatMessage::markAsSent(QDateTime const&) (chatmessage.cpp:202)
```
2. `Text::~Text` pushes `doc` onto `DocumentCache`, likely leading to the behaviour of memory usage not decreasing at all when clearing message history. I don't understand what the purpose of this DocumentCache is. It's used by Text::freeResources, but then just keeps it on the stack and doesn't clear it. It's used by ~Text which pushes it on the stack and keeps it.  When regenerating Text, which we do when loading history for each chat line, we pop a `doc` from the `DocumentCache` (which may not even be out own?) and then immediately overwrite all of its info, and that's the single place `pop` is called. This thing is a memory black hole.

3. Not overall usage but heavy re-allocation: when loading a message we mark it as unsent, then if it has been delivered (generally almost all messages from history), we create a new `Text` object and immediately replace the one `Text` we just made.

anthonybilinski:
[...]

PR #5081, opened 2018-04-14
perf(ui): cache stylesheets to reduce memory usage

anthonybilinski:
Issue #5078

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5081)

### DIFF
--- a/src/widget/style.cpp
+++ b/src/widget/style.cpp
@@ -92,15 +92,23 @@ QStringList Style::getThemeColorNames()
 QList<QColor> Style::themeColorColors = {QColor(), QColor("#004aa4"), QColor("#97ba00"),
                                          QColor("#c23716"), QColor("#4617b5")};
 
-QString Style::getStylesheet(const QString& filename, const QFont& baseFont)
-{
-    QFile file(filename);
-    if (!file.open(QFile::ReadOnly | QFile::Text)) {
-        qWarning() << "Stylesheet " << filename << " not found";
-        return QString();
-    }
+// stylesheet filename, font -> stylesheet
+// QString implicit sharing deduplicates stylesheets rather than constructing a new one each time
+std::map<std::pair<const QString, const QFont>, const QString> Style::stylesheetsCache;
 
-    return resolve(file.readAll(), baseFont);
+const QString Style::getStylesheet(const QString& filename, const QFont& baseFont)
+{
+    const std::pair<const QString, const QFont> cacheKey(filename, baseFont);
+    auto it = stylesheetsCache.find(cacheKey);
+    if (it != stylesheetsCache.end())
+    {
+        // cache hit
+        return it->second;
+    }
+    // cache miss, new styleSheet, read it from file and add to cache
+    const QString newStylesheet = resolve(filename, baseFont);
+    stylesheetsCache.insert(std::make_pair(cacheKey, newStylesheet));
+    return newStylesheet;
 }
 
 QColor Style::getColor(Style::ColorPalette entry)
@@ -128,8 +136,15 @@ QFont Style::getFont(Style::Font font)
     return fonts[font];
 }
 
-QString Style::resolve(QString qss, const QFont& baseFont)
+const QString Style::resolve(const QString& filename, const QFont& baseFont)
 {
+    QFile file{filename};
+    if (!file.open(QFile::ReadOnly | QFile::Text)) {
+        qWarning() << "Stylesheet " << filename << " not found";
+        return QString("");
+    }
+    QString qss = file.readAll();
+
     if (dict.isEmpty()) {
         dict = {// colors
                 {"@green", Style::getColor(Style::Green).name()},
@@ -162,7 +177,6 @@ QString Style::resolve(QString qss, const QFont& baseFont)
     for (const QString& key : dict.keys()) {
         qss.replace(QRegularExpression(QString("%1\\b").arg(key)), dict[key]);
     }
-
     return qss;
 }
 
--- a/src/widget/style.h
+++ b/src/widget/style.h
@@ -59,10 +59,10 @@ public:
     };
 
     static QStringList getThemeColorNames();
-    static QString getStylesheet(const QString& filename, const QFont& baseFont = QFont());
+    static const QString getStylesheet(const QString& filename, const QFont& baseFont = QFont());
     static QColor getColor(ColorPalette entry);
     static QFont getFont(Font font);
-    static QString resolve(QString qss, const QFont& baseFont = QFont());
+    static const QString resolve(const QString& filename, const QFont& baseFont = QFont());
     static void repolish(QWidget* w);
     static void setThemeColor(int color);
     static void setThemeColor(const QColor& color);
@@ -70,6 +70,7 @@ public:
     static QPixmap scaleSvgImage(const QString& path, uint32_t width, uint32_t height);
 
     static QList<QColor> themeColorColors;
+    static std::map<std::pair<const QString, const QFont>, const QString> stylesheetsCache;
 
 signals:
     void themeChanged();
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -143,8 +143,7 @@ void Widget::init()
     connect(actionQuit, &QAction::triggered, qApp, &QApplication::quit);
 
     layout()->setContentsMargins(0, 0, 0, 0);
-    ui->friendList->setStyleSheet(
-        Style::resolve(Style::getStylesheet(":/ui/friendList/friendList.css")));
+    ui->friendList->setStyleSheet(Style::getStylesheet(":/ui/friendList/friendList.css"));
 
     profilePicture = new MaskablePixmapWidget(this, QSize(40, 40), ":/img/avatar_mask.svg");
     profilePicture->setPixmap(QPixmap(":/img/contact_dark.svg"));


anthonybilinski:
Running my profile and loading all of my history in one of the chats, then clearing history in that chat. The loading is the increase in memory usage, and the clearing is not visible:
Benchmark off master 55d8922d69f035580dbea92285648d75645bfcf0
[Uploading Anthony_master_massif.log…]()
image

With stylesheet deduplication:
[Anthony_style_massif.log](https://github.com/qTox/qTox/files/1910731/Anthony_style_massif.log)
image

savings are basically 100% of stylesheet allocation, i.e. 80MB in my situation of lots of history loaded.

sudden6:
Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---

*[src/chatlog/content/filetransferwidget.cpp, line 185 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5081#-LA4Vs-5EEaRi0ZK47wx:-LA4Vs-5EEaRi0ZK47wy:bhx9v7a) ([raw file](https://github.com/qtox/qtox/blob/e8928acc37f2b542e34597638f7f8427574c02fa/src/chatlog/content/filetransferwidget.cpp#L185)):*
> ```cpp
> 
>     stylesheet = Style::getStylesheet(":/ui/fileTransferInstance/filetransferWidget.css");
>     setStyleSheet(*stylesheet.get());
> ```

I think you can directly dereference `std::shared_ptr` without having to call `get()` first, or is this considered better style?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081)*

sudden6:
Nice improvement :+1:

Talkless:
Interesting, though I am guessing that caching could work with "plain" QString, as it is implicitly shared data structure, although using shared pointer does provide kinda clear intent. Additionally, maybe worth using std::shared_ptr<*const* QString> to avoid any accidental deep copies? I imagine once it's cached, it's there forever unchanged?

---

Review status: 0 of 25 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA4ZjpI59G4khx519_7:b-fclhnu)*

Talkless:
Review status: 0 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/chatlog/content/filetransferwidget.cpp, line 185 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5081#-LA4Vs-5EEaRi0ZK47wx:-LA4_lFU9FXgMFSL1OoU:bbqg79q) ([raw file](https://github.com/qtox/qtox/blob/e8928acc37f2b542e34597638f7f8427574c02fa/src/chatlog/content/filetransferwidget.cpp#L185)):*

I think you can directly dereference `std::shared_ptr` without having to call `get()` first, or is this considered better style?
</blockquote></details>

Yeah, dereferencing pointer from get() looks kinda too verbose.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081)*

anthonybilinski:
Review status: 0 of 25 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/chatlog/content/filetransferwidget.cpp, line 185 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5081#-LA4Vs-5EEaRi0ZK47wx:-LA4bDGlES-UNtUKR1Ly:bmgzkxb) ([raw file](https://github.com/qtox/qtox/blob/e8928acc37f2b542e34597638f7f8427574c02fa/src/chatlog/content/filetransferwidget.cpp#L185)):*

Yeah, dereferencing pointer from get() looks kinda too verbose.
</blockquote></details>

Removed. Wasn't style, just a mistake

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081)*

anthonybilinski:
@talkless ah that's a good point. I didn't know about the implicit sharing. If the QString is then held onto by the Widget, I wouldn't have to add all of this lifetime management of the shared_ptr in each class. The cache with shared_ptr has the benefit that it's not there forever, because the cache itself only holds onto the data with a weak_ptr, so once the last user of the stylesheet destructs, it will be removed from cache as well. I don't think this would be handled from using raw QStrings, but that's pretty insignificant in terms of memory usage, maybe less than the thousands of shared_ptrs :P

---

Review status: 0 of 25 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA4cJid0g0_fhxgt-3V:b-55u26p)*

Talkless:
> so once the last user of the stylesheet destructs, it will be removed from cache as well

Yeah makes sense.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA4qoDj30bS5wht2EyY:b-a2di21)*

Talkless:
What's your take on std::shared_ptr<**const** QString> suggestion?

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA4r5_D1iCiQal_49QH:bi4tzbb)*

anthonybilinski:
Switched to const. Do you think the ability to empty the cache is worth the tradeoff of having every class hold onto shared_ptrs? By memory usage its basically the same, but there's lots of new member variables sprinkled around, and it makes it more complex to use the cache, plus we don't normally have every class using a specific stylesheet destruct unless we're quitting. I'm leaning towards just having things that enter the cache stay there forever to simplify.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA4zoyK25cNkXqt2cfd:bmsv8l6)*

Talkless:
> I'm leaning towards just having things that enter the cache stay there forever to simplify.

Uhm, so we go back to the question is shared pointer needed at all? I mean, if cache returns cached QString witch is simply passed into widget without calling any non-const members (and so no detaching, no deep copies performed), Qt's implicit-sharingness should be enough?

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA7YiFI5UfpNfiU86Wq:b-w5ff9p)*

anthonybilinski:
In the case of the cache never emptying, a shared_ptr is not required and the implicit copying of QString is enough. The only benefit we get from the shared_ptr is having the cache be empty-able. So basically we have to choose between 1) raw QString, no new class state, un-emptyable cache; or 2) shared_ptr/weak_ptr, new shared_ptr in every class, emptyable cache.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LA7vLp8-mqTtdrkVW4G:b2t06eb)*

Talkless:
Having cache emptyable looks like too much work and verbosity for a feature that is not sure if really needed in application of this scale and use case.

---

Review status: 0 of 25 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LAEAUxY7WvIEDWWGvol:b-v7mv74)*

anthonybilinski:
Re-benchmarked with change to plain `const QString`, and confirmed memory usage is just about as low as before. Difference could be a slight efficiency of shared_ptr vs QString's sharing implementation, but more likely it's just experimental noise.
[Anthony_QString_massif.log](https://github.com/qTox/qTox/files/1917969/Anthony_QString_massif.log)

image
On a side note I actually see memory going doing quite a lot after clearing history this time. Can't see how it would be related to this change, maybe there was just more time between clear and exit, but either way that investigation is still part of #5078

There's some stuff with `Style` that I want to refactor, but it's not pressing so I won't hold this PR up on it. I'll write it down here so I don't forget..
The behaviour of `dict` was to populates a few items on `setThemeColor` but in `resolve` it populates the map with all the default values only if `dict` is empty. This means that if `setThemeColor` is called before `loadNewStylesheet`, the map would only ever have the couple colour-related items in it. It also means that the first font passed into `loadNewStylesheet` is the only one that matters, since it's used to populate `dict` which only happens the first time. This might be a bug since we actually do pass in a non-default `QFont` in `Text::Text`. I haven't investigated that further. If we do want to just have a default QFont parsed into the `dict`, we can't do that on static init because `QFontInfo` (I think) requires `QGuiApplication` to be init, so the whole `Style` static class might need to be changed into a singleton :(

We might get some slight cache optimization by dropping the `QFont` argument too since it doesn't actually change uniqueness of the stylesheet the way it is now..

sudden6:
code :lgtm_strong: did you test it with switching themes and colors? could be that the cache must be cleared during a switch

---

Reviewed 13 of 25 files at r1, 12 of 12 files at r2, 3 of 25 files at r3.

---

*[src/widget/style.cpp, line 105 at r3](https://beta.reviewable.io/reviews/qtox/qtox/5081#-LAJkmZn-KtG_75ztyCm:-LAJkmZpDhTH7M8zkiY6:bb7ozr7) ([raw file](https://github.com/qtox/qtox/blob/13e5006d6ca87953092262bcf71ad6e4e32726f4/src/widget/style.cpp#L105)):*
> ```cpp
>     if (it != stylesheetsCache.end())
>     {
>     	// cache hit
> ```

tabs

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5081#-:-LAJl2op6EzWBpu2OULP:b-mopni1)*

tox-user:
Fonts and colors are defined in many different places, so those Style bugs might not affect anything in the UI. Themes need to be reimplemented and I'm planning to work on that.

anthonybilinski:
>did you test it with switching themes and colors

I hadn't, thanks for the suggestion. Tested now and behaviour looks identical before and after. From code inspection I don't expect risk related to themes/colours, as those don't alias the same stylesheets/fonts that are passed into the function that's caching.

anthonybilinski:
>did you test it with switching themes and colors? could be that the cache must be cleared during a switch

:1st_place_medal: https://github.com/qTox/qTox/issues/5092

Issue #4912, opened 2018-01-14
"Save chat log" function should save log with dates


yurivict:
Currently there are only timestamps, which is meaningless when the log covers more than one day.

sudden6:
Agreed.

isav83:
Is it issue fixed? I save chat log via "Export to file" and log saved with dates in format "[date] [time] [message]".

anthonybilinski:
Behaviour was changed in ed514d71667d87c30acd1f41213d8ec8c6041b04 to include datestamp. Thanks for pointing it out.

PR #3447, opened 2016-06-28
fix(profile): Fix for importing profile using Nautilus file manager

anoadragon453:
Importing a profile using Nautilus as the default file manager, which many users use, was causing a hang in Qt's  method. Setting Qt to use it's own file manager in this circumstance fixes the issue. Closes #3436, closes #3443.

PR #5098, opened 2018-04-23
fix: correct format log files

PKEv:
- correct format of "save chat log" #4912
 - add date like in chat
 - correct type dialog for export dialog

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5098)

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -1025,8 +1025,7 @@ void ChatForm::onExportChat()
     QDateTime now = QDateTime::currentDateTime();
     QList<History::HistMessage> msgs = history->getChatHistory(pk, epochStart, now);
 
-    QString path = QFileDialog::getSaveFileName(0, tr("Save chat log"), QString{}, QString{}, 0,
-                                                QFileDialog::DontUseNativeDialog);
+    QString path = QFileDialog::getSaveFileName(0, tr("Save chat log"));
     if (path.isEmpty()) {
         return;
     }
@@ -1038,11 +1037,12 @@ void ChatForm::onExportChat()
 
     QString buffer;
     for (const auto& it : msgs) {
-        QString timestamp = it.timestamp.toString();
+        QString timestamp = it.timestamp.time().toString("hh:mm:ss");
+        QString datestamp = it.timestamp.date().toString("yyyy-MM-dd");
         ToxPk authorPk(ToxId(it.sender).getPublicKey());
         QString author = getMsgAuthorDispName(authorPk, it.dispName);
 
-        QString line = QString("%1\t%2\t%3\n").arg(timestamp, author, it.message);
+        QString line = QString("%1\t%2\t%3\t%4\n").arg(datestamp, timestamp, author, it.message);
         buffer = buffer % line;
     }
     file.write(buffer.toUtf8());
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -124,7 +124,7 @@ ChatMessage::Ptr ChatMessage::createChatInfoMessage(const QString& rawMessage,
 
     msg->addColumn(new Image(QSize(18, 18), img),
                    ColumnFormat(NAME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
-    msg->addColumn(new Text("<b>" + text + "</b>", baseFont, false, ""),
+    msg->addColumn(new Text("<b>" + text + "</b>", baseFont, false, text),
                    ColumnFormat(1.0, ColumnFormat::VariableSize, ColumnFormat::Left));
     msg->addColumn(new Timestamp(date, Settings::getInstance().getTimestampFormat(), baseFont),
                    ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -487,17 +487,16 @@ void GenericChatForm::onSaveLogClicked()
     for (ChatLine::Ptr l : lines) {
         Timestamp* rightCol = qobject_cast<Timestamp*>(l->getContent(2));
 
-        if (!rightCol)
-            break;
-
         ChatLineContent* middleCol = l->getContent(1);
         ChatLineContent* leftCol = l->getContent(0);
 
-        QString timestamp = rightCol->getTime().isNull() ? tr("Not sent") : rightCol->getText();
-        QString nick = leftCol->getText();
+        QString nick = leftCol->getText().isNull() ? tr("[System message]") : leftCol->getText();
+
         QString msg = middleCol->getText();
 
-        plainText += QString("[%2] %1\n%3\n\n").arg(nick, timestamp, msg);
+        QString timestamp = (rightCol == nullptr) ? tr("Not sent") : rightCol->getText();
+
+        plainText += QString("%1\t%3\t%2\n").arg(nick, timestamp, msg);
     }
 
     file.write(plainText.toUtf8());


src/widget/form/chatform.cpp
```diff
@@ -1025,8 +1025,7 @@ void ChatForm::onExportChat()
     QDateTime now = QDateTime::currentDateTime();
     QList<History::HistMessage> msgs = history->getChatHistory(pk, epochStart, now);
 
-    QString path = QFileDialog::getSaveFileName(0, tr("Save chat log"), QString{}, QString{}, 0,
-                                                QFileDialog::DontUseNativeDialog);
```
Diadlo: Why do you remove `DontUseNativeDialog`? It was added in #3447 to fix issue with some file managers

PKEv:
@Diadlo 
I removed it as this is the only place where this kind of dialog was used

sudden6:
Reviewed 2 of 3 files at r1.
Review status: 2 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/widget/form/chatform.cpp, line 1040 at r1](https://beta.reviewable.io/reviews/qtox/qtox/5098#-LAn_q0T03aj4RCgxVzr:-LAn_q0T03aj4RCgxVzs:b-dxaa1f) ([raw file](https://github.com/qtox/qtox/blob/212ef3a5674a531da27b82cbb1bb16ed77cd9778/src/widget/form/chatform.cpp#L1040)):*
> ```cpp
>     QString buffer;
>     for (const auto& it : msgs) {
>         QString timestamp = it.timestamp.time().toString("hh:mm:ss");
> ```

I think it would be easier to use `Qt::ISODate` here, as this already implements the complete formatting.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/qtox/qtox/5098)*

PKEv:
it seems to me the display of the time zone is not necessary
```
 ISO 8601 extended specification:  yyyy-MM-ddTHH:mm:ss[Z|[+|-]HH:mm]
```

sudden6:
probably true, :lgtm_strong: 

---

Reviewed 1 of 3 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5098#-:-LAs_i774ftisag1mIxG:b63nte0)*

tox-user:
>        if (msg.isEmpty())
>        {

Move the bracket to the end of the line: ` if (msg.isEmpty()) {`

anthonybilinski:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/chatlog/chatmessage.cpp, line 127 at r1](https://reviewable.io/reviews/qtox/qtox/5098#-LAtVfZ8Ez1qFOm9iuPd:-LAtVfZ8Ez1qFOm9iuPe:bdefoc1) ([raw file](https://github.com/qtox/qtox/blob/212ef3a5674a531da27b82cbb1bb16ed77cd9778/src/chatlog/chatmessage.cpp#L127)):*
> ```cpp
>     msg->addColumn(new Image(QSize(18, 18), img),
>                    ColumnFormat(NAME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
>     msg->addColumn(new Text("<b>" + text + "</b>", baseFont, false, text),
> ```

nice! copying info lines works now :D

---

*[src/widget/form/chatform.cpp, line 1042 at r1](https://reviewable.io/reviews/qtox/qtox/5098#-LAtdT0Z8fd-hWzS9ELJ:-LAtdT0Z8fd-hWzS9ELK:bzgyedx) ([raw file](https://github.com/qtox/qtox/blob/212ef3a5674a531da27b82cbb1bb16ed77cd9778/src/widget/form/chatform.cpp#L1042)):*
> ```cpp
>         QString timestamp = it.timestamp.time().toString("hh:mm:ss");
>         QString datestamp = it.timestamp.date().toString("yyyy-MM-dd");
>         ToxPk authorPk(ToxId(it.sender).getPublicKey());
> ```

should we also add dateStamp to "Save chat log" in GenericChatForm::onSaveLogClicked? if so, can we de-duplicate and share parts of these functions to have the two export methods keep the same format in the future?

---

*[src/widget/form/genericchatform.cpp, line 491 at r1](https://reviewable.io/reviews/qtox/qtox/5098#-LAtePswD9ua2CDQA0P-:-LAtePswD9ua2CDQA0P0:b-1lu0zs) ([raw file](https://github.com/qtox/qtox/blob/02d6c63acaac0ae95fa8be3a1b9301657e6a4a94/src/widget/form/genericchatform.cpp#L491)):*
> ```cpp
> 
>         if (!rightCol)
>             break;
> ```

is it safe to remove this nullptr check? what's the reason for the change?

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5098)*

PKEv:
> should we also add dateStamp to "Save chat log" in GenericChatForm::onSaveLogClicked? if so, can we de-duplicate and share parts of these functions to have the two export methods keep the same format in the future?

now "chatform" does not have information about the date it appears, when the chat history  loading - date written in the message field and can not be filtered while saving

>is it safe to remove this nullptr check? what's the reason for the change?

I do not see a problem
If you do not clear the check, then the date string will not be displayed since it does not have the attribute of the departure time displayed in the right column
As in other and other service messages

PKEv:
@tox-user fix

anthonybilinski:
@PKEv from testing, removing the `if (!rightCol) {` check causes a seg fault when the right column has the Spinner for offline messages that haven't yet been delivered, since Spinner inherits from ChatLineContent but can't be qobject_cast to Timestamp. I think it should be changed to still check that the cast worked (that it's non-null) before dereferencing, but not necessarily skipping the whole line, maybe instead checking if it's a Spinner, and if so adding "Not delivered" instead of a timestamp and still adding to the log?

Also from testing
>If you do not clear the check, then the date string will not be displayed since it does not have the attribute of the departure time displayed in the right column

is not true, because the Timestamp class exists, just the time is null. This check is just to check whether Timestamp exists. The check farther down: `QString timestamp = rightCol->getTime().isNull() ? tr("Not sent")` handles the case of no time stamp present from info lines giving a log like:
```
[Not sent]
2018-04-24
```

While you're at it, it would be nice to get the "[Not sent]" message to be on the same line as the chat message, like timestamps are. Currently there's a new line character added from the string literal of "Not sent", but then again inserted into the plainText formatting string.

PKEv:
@anthonybilinski  You're right, I made a change.

tox-user:
Maybe replace "Not sent" with "Pending" since those messages will be sent at some point?

Talkless:
Review status: 2 of 3 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[src/widget/form/chatform.cpp, line 1040 at r1](https://reviewable.io/reviews/qtox/qtox/5098#-LAn_q0T03aj4RCgxVzr:-LDR58hbBSLtt965ApGb:b-k7jckb) ([raw file](https://github.com/qtox/qtox/blob/212ef3a5674a531da27b82cbb1bb16ed77cd9778/src/widget/form/chatform.cpp#L1040)):*

I think it would be easier to use `Qt::ISODate` here, as this already implements the complete formatting.
</blockquote></details>

If  Qt::ISODate will not be used, then QStringLiteral("hh....

---

*[src/widget/form/chatform.cpp, line 1045 at r2](https://reviewable.io/reviews/qtox/qtox/5098#-LDR5LgyAaAOaWjQYx54:-LDR5LgyAaAOaWjQYx55:b-vkmits) ([raw file](https://github.com/qtox/qtox/blob/0e6c665a344844589fa6274a06cf9221be7c937c/src/widget/form/chatform.cpp#L1045)):*
> ```cpp
>         QString author = getMsgAuthorDispName(authorPk, it.dispName);
> 
>         QString line = QString("%1\t%2\t%3\t%4\n").arg(datestamp, timestamp, author, it.message);
> ```

Maybe it would be worth to refactor to use QStringBuilder?
QString line{datestamp % '\t' % timestamp % '\t' % .... % '\n'};

that `line` variable could also be removed completely, concatinating directly to `buffer` ?

---

*[src/widget/form/genericchatform.cpp, line 499 at r2](https://reviewable.io/reviews/qtox/qtox/5098#-LDR63chDafXTm4tuQpm:-LDR63chDafXTm4tuQpn:b-u07u4) ([raw file](https://github.com/qtox/qtox/blob/0e6c665a344844589fa6274a06cf9221be7c937c/src/widget/form/genericchatform.cpp#L499)):*
> ```cpp
>         QString timestamp = (rightCol == nullptr) ? tr("Not sent") : rightCol->getText();
> 
>         plainText += QString("%1\t%3\t%2\n").arg(nick, timestamp, msg);
> ```

Consider using QStringBuilder too?

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5098)*

PKEv:
it seems to me not correct to go into the subtleties of implementing the standard library.
If there is a significant increase in performance (QStringBuilder) then it should be implemented inside QString.Format

Talkless:
> it seems to me not correct to go into the subtleties of implementing the standard library.

Sorry, do not follow that. How is using Qt library utility for concatenating strings comparable with re implementing standard library?

> If there is a significant increase in performance (QStringBuilder) then it should be implemented inside QString.Format

Perfromance is the only argument for using QStringBuilder's "magic". 

I believe there are works (or at least there where talks) about making faster alternative for `QString::arg()` i.e. parsing format string at compile time or something like that. Currently `arg()` has to process that format string every time.

At least `QString("%1\t%3\t%2\n")` could be changed into `QStringLiteral("%1...`, though of course all of these comments can be considered as out of scope for this PR.

anthonybilinski:
I've created https://github.com/qTox/qTox/pull/5267 where I rebased (keeping @PKEv's authorship), resolved minor merge conflict, and addressed the remaining review comments so we can get this change in.

PR #5137, opened 2018-05-11
chore(install): rename .desktop and .appdata.xml

sudden6:
The xdg-desktop specification strongly recommends to use a reverse
domain name, so follow that.

For the reason see https://github.com/flathub/flathub/pull/379#discussion_r187470760 and above.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5137)


cmake/Installation.cmake
```diff
@@ -32,9 +32,20 @@ if(APPLE)
...
+  set(QTOX_DESKTOP_DIR "${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications")
+  set(QTOX_DESKTOP_FILE "io.github.qtox.qTox.desktop")
   install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION "bin")
-  install(FILES "res/qTox.appdata.xml" DESTINATION "share/metainfo")
-  install(FILES "qtox.desktop" DESTINATION "share/applications")
+  install(FILES "res/io.github.qtox.qTox.appdata.xml" DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/metainfo")
+  install(FILES "${QTOX_DESKTOP_FILE}" DESTINATION "${QTOX_DESKTOP_DIR}")
+  
+  # try to set up the menu system
+  find_program(XDG-DESKTOP-MENU_EXECUTABLE xdg-desktop-menu)
```
TingPing: Don't use `xdg-*`, they are pretty bad tools. Just install it normally, use `desktop-file-validate` as part of your build system during `test`. Use `update-desktop-database` as a post-install step. (Handle neither being installed, in normal packages the update step isn't even used anyway)

cmake/Installation.cmake
```diff
@@ -32,9 +32,20 @@ if(APPLE)
...
   )
 else()
+  include( GNUInstallDirs )
+  # follow the xdg-desktop specification
+  set(QTOX_DESKTOP_DIR "${CMAKE_INSTALL_FULL_DATAROOTDIR}/applications")
+  set(QTOX_DESKTOP_FILE "io.github.qtox.qTox.desktop")
   install(TARGETS ${PROJECT_NAME} RUNTIME DESTINATION "bin")
-  install(FILES "res/qTox.appdata.xml" DESTINATION "share/metainfo")
-  install(FILES "qtox.desktop" DESTINATION "share/applications")
+  install(FILES "res/io.github.qtox.qTox.appdata.xml" DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/metainfo")
```
TingPing: Along the same lines (optionally) using `appstream-util validate-relax` during `test` is very useful.

cmake/Installation.cmake
sudden6: So should I just remove that line? If it's not strictly needed I'd prefer that.

TingPing: It is not needed. `update-desktop-database` is needed for non-packaged installs (end users building from source).

sudden6: ok, I removed that.

sudden6:
I'll add validation later, thanks for your help with these files.

res/io.github.qtox.qTox.appdata.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyleft 2016 SkyzohKey <skyzohkey@framasphere.org> -->
 <component type="desktop">
- <id>qtox.desktop</id>
+ <id>io.github.qtox.qTox</id>
  <metadata_license>CC-BY-SA-3.0</metadata_license>
```
TingPing: Need to add:
```xml
  <launchable type="desktop-id">io.github.qtox.qTox.desktop</launchable>
```

sudden6: Done

sudden6:
@TingPing can you comment with `:lgtm_strong:` via the reviewable button when you think this is now fixed?

TingPing:
:lgtm_strong:

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5137#-:-LCPv7fk3i-TcA_ZVl6E:b-jbsf8z)*

PR #5160, opened 2018-05-29
refactor(themes): load css and images from themes folder

tox-user:
The default theme is now placed in 'themes' folder in addition to being included
as a resource. That means:
- modifying css files doesn't require recompiling the program anymore
- users can now easily modify the existing theme
- whenever there is a style sheet or image file missing (for example deleted by user),
it will be loaded from the default theme that is included as a resource

This is a first of many future changes to make it possible to create themes.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)


sudden6:
Review status: 88 of 90 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LDmaTQc9YGyS_g6274b:b-1zw1gq) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*
> ```CMake
> 
> # Copy themes
> file(COPY ${CMAKE_SOURCE_DIR}/themes DESTINATION ${CMAKE_BINARY_DIR})
> ```

We should move this to Install.txt and make the path platform dependent, for linux `usr/share/` could probably be more correct. See the freedesktop.org site for specs.

---

*[src/widget/style.cpp, line 133 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgbnE8RQqAJ0pjeY2:-LDmgbnE8RQqAJ0pjeY3:bgmsqmk) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L133)):*
> ```cpp
>             qWarning() << "Stylesheet " << fullPath << " not found. Default will be used";
>             return fullPath;
>         } else {
> ```

this case should never happen, please add a warning

---

*[src/widget/style.cpp, line 176 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgtuw4VUXmDZDEjTJ:-LDmgtuw4VUXmDZDEjTK:bkh7u9k) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L176)):*
> ```cpp
>     } else {
>         fullPath = ":" + fullPath;
>         if (file.open(QFile::ReadOnly | QFile::Text)) {
> ```

file has not changed in the meantime, why try to open the same file again?

---

*[src/widget/style.cpp, line 233 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmhBL23sl-eY8AZPJW:-LDmhBL23sl-eY8AZPJX:bqxz7su) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L233)):*
> ```cpp
>         QFile file{fullImagePath};
> 
>         if (!file.open(QFile::ReadOnly)) {
> ```

Even though this code is probably not called often, it tries to open a lot of files, maybe introduce a list/cache of known existing and not existing files?

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
Review status: 88 of 90 files reviewed at latest revision, 4 unresolved discussions.

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LDoAOEJENGSC3njUphJ:b-3tfflj) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*

We should move this to Install.txt and make the path platform dependent, for linux `usr/share/` could probably be more correct. See the freedesktop.org site for specs.
</blockquote></details>

What do you mean with install.txt? Do you mean make install? Do you want themes to be somewhere else than in qTox's directory?

---

*[src/widget/style.cpp, line 176 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgtuw4VUXmDZDEjTJ:-LDoAFIR6KZ9UgWZgeJY:btxpf3z) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L176)):*

file has not changed in the meantime, why try to open the same file again?
</blockquote></details>

It needs to check if the file exists in themes/THEME_NAME. If it doesn't, try to check if it exists as a resorce (for that ":" is appended to the file name) in the default theme. If a user tries to link to "foo.css", or some other made up name, it won't exist as a resource. I have just now noticed that it uses getThemePath(), which means it will look for resource in currently selected theme instead of themes/default and that makes no sense.

---

*[src/widget/style.cpp, line 233 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmhBL23sl-eY8AZPJW:-LDo8S6Z3p4EbyohVOTl:b-vouybz) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L233)):*

Even though this code is probably not called often, it tries to open a lot of files, maybe introduce a list/cache of known existing and not existing files?
</blockquote></details>

There is a cache for style sheets in Style::getStylesheet, so if some style sheet is used in more than one place, it will only enter resolve function once. Do we really need another cache for images?

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

sudden6:
Review status: 88 of 90 files reviewed at latest revision, 4 unresolved discussions.

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LDpuUfwCXaJZdviSARc:b-ufq1ct) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

What do you mean with install.txt? Do you mean make install? Do you want themes to be somewhere else than in qTox's directory?
</blockquote></details>

Sorry, I gave you the completely wrong file. I meant to move the copying of themes to the destination folder should be platform dependent, since each platform has an own path for such resources. All this stuff is in https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake

---

*[src/widget/style.cpp, line 176 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgtuw4VUXmDZDEjTJ:-LDpuzzm-Xhw5pOzoe6V:b-8er4d0) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L176)):*
When this `if` a few lines above is not entered:
> ```
>  if (file.open(QFile::ReadOnly | QFile::Text)) {
> ```

You will never get into the if condition here, since you didn't modify the `file` object in any way. You only modify `fullPath`, but don't change `file`, you need to call the constructor of `file` again to let it point to a new file.

---

*[src/widget/style.cpp, line 233 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmhBL23sl-eY8AZPJW:-LDpvrhVAHZ-V569omMA:bife3gj) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L233)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

There is a cache for style sheets in Style::getStylesheet, so if some style sheet is used in more than one place, it will only enter resolve function once. Do we really need another cache for images?
</blockquote></details>

only local to this function. Because right now, if a stylesheet contains e.g. `buttons/redButton.svg` three times, you will open (and close) this file three times AFAICT. This is quite expensive and unnecessary.  Just keep track of which files exist and which don't in a hashmap or some similar structure.

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
> only local to this function. Because right now, if a stylesheet contains e.g. buttons/redButton.svg three times, you will open (and close) this file three times AFAICT. This is quite expensive and unnecessary. Just keep track of which files exist and which don't in a hashmap or some similar structure.

In that case shouldn't getImagePath function have a cache too? Some images are loaded by code.

tox-user:
>I meant to move the copying of themes to the destination folder should be platform dependent, since each platform has an own path for such resources. All this stuff is in https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake

Why can't we use the same directory as we use for profiles? On GNU/Linux it would be:
/home/user/.config/qtox/themes
/home/user/.config/qtox/plugins

On Windows:
C:/users/user/AppData/Roaming/qtox/...

sudden6:
> In that case shouldn't getImagePath function have a cache too?

probably

> Why can't we use the same directory as we use for profiles?

Also a good idea. In the future we could look for system wide installed themes too.

tox-user:
Review status: 86 of 90 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LEGuhWO4POVgrKFbd3k:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*

Sorry, I gave you the completely wrong file. I meant to move the copying of themes to the destination folder should be platform dependent, since each platform has an own path for such resources. All this stuff is in https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake
</blockquote></details>

Done.

---

*[src/widget/style.cpp, line 133 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgbnE8RQqAJ0pjeY2:-LEGv5BE-65KlfLkZr9W:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L133)):*

this case should never happen, please add a warning
</blockquote></details>

Done.

---

*[src/widget/style.cpp, line 176 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmgtuw4VUXmDZDEjTJ:-LEGv6ln1f71AZeX7o5j:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L176)):*

When this `if` a few lines above is not entered:
> ```
>  if (file.open(QFile::ReadOnly | QFile::Text)) {
> ```

You will never get into the if condition here, since you didn't modify the `file` object in any way. You only modify `fullPath`, but don't change `file`, you need to call the constructor of `file` again to let it point to a new file.
</blockquote></details>

Done.

---

*[src/widget/style.cpp, line 233 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmhBL23sl-eY8AZPJW:-LEGv8Nv2jtvLQ0-twZO:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L233)):*

only local to this function. Because right now, if a stylesheet contains e.g. `buttons/redButton.svg` three times, you will open (and close) this file three times AFAICT. This is quite expensive and unnecessary.  Just keep track of which files exist and which don't in a hashmap or some similar structure.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
@qTox/core @qTox/linux-team review and testing needed

sudden6:
Reviewed 2 of 4 files at r2.
Review status: 0 of 1 LGTMs obtained

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LFJOaIMARKcl8u3R1aB:b-4k8er) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Can you also move this to `Installation.cmake`?

---

*[src/widget/style.cpp, line 233 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmhBL23sl-eY8AZPJW:-LFJR6GVE6i2A4GETanM:b-otilcs) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/src/widget/style.cpp#L233)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

Done.
</blockquote></details>

See comment above.

---

*[src/widget/style.cpp, line 111 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFJOz-j8ZSKYLqVT5DB:-LFJOz-j8ZSKYLqVT5DC:bw2fotd) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L111)):*
> ```cpp
> #elif defined(Q_OS_WIN32)
>     const QString themePath = homePath + "/AppData/roaming/qtox/themes/" + themeName + "/";
> #endif
> ```

I think there's a Qt variable that maps to e.g. `AppData/roaming`, `.config/`, take a look at http://doc.qt.io/qt-5/qstandardpaths.html I think there's something more appropriate there.

---

*[src/widget/style.cpp, line 139 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFJQLv58rUuTcxwJLp8:-LFJQLv58rUuTcxwJLp9:bs9i5dc) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L139)):*
> ```cpp
> }
> 
> static QStringList existingImagesCache;
> ```

putting this here is most probably causing a resource leak, because the cache is never emptied
I'd recommend:
- put this in the class
- empty it on theme change

Additionally QStringList is the wrong data structure for this cache, because a list uses O(n) operations for lookup, try something hash based.

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

sudden6:
*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LFJRJ987A_GgxyfKtld:b-w1l4br) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*

Can you also move this to `Installation.cmake`?
</blockquote></details>

Also whitespace error

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
Review status: 0 of 1 LGTMs obtained

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LFK0KYX-IuuShMg9p12:b-yygs8f) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*
>Can you also move this to Installation.cmake

Remove it from CMakeLists and add to Installation.cmake or simply copy it there?

>Also whitespace error

What do you mean? Where?

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
>I think there's a Qt variable that maps to e.g. AppData/roaming, .config/, take a look at http://doc.qt.io/qt-5/qstandardpaths.html I think there's something more appropriate there.

Those variables don't help with what I'm trying to do. There isn't one that provides the right path on all of the OSes. Also:
>The paths above should not be relied upon, as they may change according to OS configuration, locale, or they may change in future Qt versions.

sudden6:
I don't really like hard coding such paths, because this can cause problems with Flatpak and other package formats, since they usually override these locations. For this path, `QStandardPaths::ConfigLocation` or `QStandardPaths::AppDataLocation` might be appropriate.

---

Review status: 0 of 1 LGTMs obtained

---

*[CMakeLists.txt, line 92 at r1](https://reviewable.io/reviews/qtox/qtox/5160#-LDmaTQc9YGyS_g6274a:-LFO7z0e1LZRCIbOWvsP:b-54l0pn) ([raw file](https://github.com/qtox/qtox/blob/51ee2439242a2bee2b2948fc2a3bc01c45f25b4b/CMakeLists.txt#L92)):*
<details><summary><i>Previously, tox-user wrote…</i></summary><blockquote>

>Can you also move this to Installation.cmake
Remove it from CMakeLists and add to Installation.cmake or simply copy it there?

>Also whitespace error
What do you mean? Where?
</blockquote></details>

Remove from CMakeLists.txt and add to Installation.cmak

You use tabs instead of spaces, reviewable shows this, GH doesn't unfortunately

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160#-:-LFO9sjX2HV8DAP4KDNR:bguggo6)*

Talkless:
Review status: 0 of 1 LGTMs obtained

---

*[src/widget/style.cpp, line 97 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO3XEOBkm47jwIcbVj:-LFO3XEOBkm47jwIcbVk:b-4iyg4p) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L97)):*
> ```cpp
>     //TODO: return name of the current theme
>     const QString themeName = "default";
>     return themeName;
> ```

It would be much better to return`return QStringLiteral("default")`. It's gonna be static baked-in QString.

---

*[src/widget/style.cpp, line 111 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFJOz-j8ZSKYLqVT5DB:-LFO3xD--42GIrsWktrP:b-455cxr) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L111)):*
> ```
> homePath + "/.config/qtox/themes/" + themeName + "/";
> ```

Also, please consider concatenating much more efficiently with QStringBuilder (with or without QStandardPaths):
```
homePath % QLatin1String("/.config/qtox/themes/") % themeName % '/';
```

https://github.com/qTox/qTox/blob/master/doc/coding_standards.md#strings

---

*[src/widget/style.cpp, line 151 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO57MaAij_n_wGjaQO:-LFO57MaAij_n_wGjaQP:b8nwsek) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L151)):*
> ```cpp
>     // if not in cache
>     QFile file{fullPath};
>     if (file.open(QFile::ReadOnly)) {
> ```

If the intention is to check for existence, consider using `QFileInfo::exists(const QString &file)`. No need to actually OS to open the file.

---

*[src/widget/style.cpp, line 157 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO68bA6HxYnNb9MWej:-LFO68bA6HxYnNb9MWek:b-3oet0t) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L157)):*
> ```cpp
>         qWarning() << "Image" << fullPath << "not found. Default will be used";
> 
>         fullPath = ":themes/default/" + filename;
> ```

`QStringLiteral(":themes/default/") +` or even better (with `#include <QStringBuilder>`):

`QLatin1String(":themes/default/") % filename`;

---

*[src/widget/style.cpp, line 160 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO65Qz77gzV6hTJ2Mv:-LFO65Qz77gzV6hTJ2Mw:b3rzqm8) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L160)):*
> ```cpp
>         QFile file{fullPath};
> 
>         if (file.open(QFile::ReadOnly)) {
> ```

Same as above, QFileInfo::exists()

---

*[src/widget/style.cpp, line 164 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO5h4f9tfiyejOCP94:-LFO5h4f9tfiyejOCP95:b-24xjoi) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L164)):*
> ```cpp
>         } else {
>             qWarning() << "Image" << fullPath << "default file not found";
>             return QString("");
> ```

just `return QString{}` or even `return {}`. `QString::QString(const char*)` is definitely overkill for empty string.

---

*[src/widget/style.cpp, line 198 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO6eRH6MCjUNZ9XvqX:-LFO6eRI80UMvPa_-fHz:b-sybb4p) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L198)):*
> ```cpp
>     QString themePath = getThemePath();
>     QString fullPath = themePath + filename;
>     QString qss = "";
> ```

just `QString qss;`, default constructor does the job.

---

*[src/widget/style.cpp, line 206 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO6mYP-DDjxunlEyDw:-LFO6mYP-DDjxunlEyDx:b-h3gqn0) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L206)):*
> ```cpp
>         qWarning() << "Stylesheet" << fullPath << "not found. Default will be used";
> 
>         fullPath = ":themes/default/" + filename;
> ```

Same as above, QStringLitereal or better QLatin1String + QStringBuilder 

---

*[src/widget/style.cpp, line 212 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO6xx4C3xwMAsdPK_A:-LFO6xx4C3xwMAsdPK_B:b-lat67v) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L212)):*
> ```cpp
>             qss = file.readAll();
>         } else {
>             qWarning() << "Stylesheet" << fullPath << "default file not found";
> ```

These errors (also above) could be more generic. Maybe it's permission denied or other I/O error for some reason?

How about `qWarning() << "Failed to open stylesheet" << fullPath << ":" << file.erorString();` or similar.

---

*[src/widget/style.cpp, line 213 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO8G360J3K6oKYZgde:-LFO8G360J3K6oKYZgdf:b6hgpf) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L213)):*
> ```cpp
>         } else {
>             qWarning() << "Stylesheet" << fullPath << "default file not found";
>             return QString("");
> ```

`return QStirng{}` or `return {};`

---

*[src/widget/style.cpp, line 247 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO8Mz888qYhXrTcnMk:-LFO8Mz888qYhXrTcnMl:bnz25mu) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L247)):*
> ```cpp
> 
>     for (const QString& key : dict.keys()) {
>         qss.replace(QRegularExpression(QString("%1\\b").arg(key)), dict[key]);
> ```

`QRegularExpression(arg % QLatin1String("\\b"));`

Because `.arg()` has to parse and search for placeholders. Is this really necessary for such simple small string concatenation? Especially when we are in the loop.

Also, are we sure that `key` will not have regex-special symbols in it? Maybe worth using `QRegularExpression::escape()`?

Also, could you explain what is tried to achieve here? (I don't know anything about qTox stylesheets).

---

*[src/widget/style.cpp, line 251 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFO9o9C1-HrxFLLMLoY:-LFO9o9C1-HrxFLLMLoZ:b-eacjgn) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L251)):*
> ```cpp
> 
>     // @getImagePath() function
>     QRegularExpression re("@getImagePath\\([^)\\s]*\\)");
> ```

Could use c++11 raw string literals to avoid escaping with `\\`, also QStringLiteral for created strings:
`const QRegularExpression re{QStringLiteral(R"(@getImagePath\([^)\s]*\))")}`

---

*[src/widget/style.cpp, line 259 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFOAP1t3dkrkitCAg5g:-LFOAP1t3dkrkitCAg5h:bkamrmb) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L259)):*
> ```cpp
>         const QString phrase = path;
> 
>         path.remove("@getImagePath(");
> ```

QStringLiteral()

---

*[src/widget/style.cpp, line 270 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFOAoRBCvRDUBRS_vPK:-LFOAoRBCvRDUBRS_vPL:bmq2fiq) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L270)):*
> ```cpp
>                 existingImagesCache << fullImagePath;
>             } else {
>                 qWarning() << "Image" << fullImagePath << "not found. Default will be used";
> ```

Same as above, could be more generic and useful error message.

---

*[src/widget/style.cpp, line 271 at r2](https://reviewable.io/reviews/qtox/qtox/5160#-LFOAv842ZfPwdL-E0lv:-LFOAv842ZfPwdL-E0lw:bivkiin) ([raw file](https://github.com/qtox/qtox/blob/3d62b01ea3b4f6463cd09f3eb1e0649668a1d8e8/src/widget/style.cpp#L271)):*
> ```cpp
>             } else {
>                 qWarning() << "Image" << fullImagePath << "not found. Default will be used";
>                 fullImagePath = ":themes/default/" + path;
> ```

QLatin1String(":themes/default/") % path;

---


*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5160)*

tox-user:
>For this path, QStandardPaths::ConfigLocation or QStandardPaths::AppDataLocation might be appropriate.

Those paths can not be relied on, because they might change in future versions of Qt. I've already hardcoded paths in cmake. What will happen when Qt changes those paths? That is just one issue.

We want our themes path to be similar to where Tox profiles and config files are located. Those paths are:
- ~/.config/tox (GNU/Linux)
- C:/user/%username%/AppData/roaming/tox (Windows)
- ~/Library/Application Support/tox (MacOS)

ConfigLocation that you mention is:
- ~/.config, /etc/xdg
- C:/Users/<USER>/AppData/Local/<APPNAME> and C:/ProgramData/<APPNAME>
- ~/Library/Preferences

So that won't work for us. AppDataLocation is:
- ~/.local/share/<APPNAME>, /usr/local/share/<APPNAME>, /usr/share/<APPNAME>
- C:/Users/<USER>/AppData/Roaming/<APPNAME>, C:/ProgramData/<APPNAME>, <APPDIR>, <APPDIR>/data, <APPDIR>/data/<APPNAME>
- ~/Library/Application Support/<APPNAME>, /Library/Application Support/<APPNAME>, <APPDIR>/../Resources

This one doesn't have all the paths we want either. I went through all of those variables and I didn't find one that does. Maybe we could mix them depending on the OS, but we already use hardcoded paths in cmake. There is a risk those variables will change in Qt and that would mess everything up.

Diadlo:
@tox-user 

> Those paths can not be relied on, because they might change in future versions of Qt
It doesn't mean that someone can come and just change a location :) Probably it can be changed if (for example) most of distributions starts store config in `~/.cfg` or something like that

> We want our themes path to be similar to where Tox profiles and config files are located.

Some time ago I thought, that current tox location for different OSes not the best. Probably we can place themes "right" using Qt location variables

tox-user:
What paths would be right in your opinion? If we place it in a similar path to Tox config files, that would be very convenient for users, because they already know that path. They wouldn't have to search for it.

To me Qt docs make it pretty clear that those variables shouldn't be used. [They say](https://doc.qt.io/qt-5/qstandardpaths.html):
>The paths above should not be relied upon, as they may change according to OS configuration, locale, or they may change in future Qt versions.

Diadlo:
@tox-user We can place `file://` link in UI, which will show theme path to user.
"should not be relied upon" means, that you should use Qt variable and don't think about real path in system. It doesn't really metter for you. You store all configs in `ConfigLocation` etc.

tox-user:
@Talkless you have a lot of good points about the error messages and QFileInfo::exists()
. You also suggest a lot of very small changes that are unimportant. You do that with a lot of my PRs. I really would prefer not to spend my time on those small things. I'm talking about stuff like:

>It would be much better to returnreturn QStringLiteral("default"). It's gonna be static baked-in QString.

>just return QString{} or even return {}. QString::QString(const char*) is definitely overkill for empty string.

I'm just returning an empty string, how is that an overkill? We don't need to watch out for every single bit of memory like something bad was going to happen when we lose a few bytes. In practice most of those things don't have any relevant impact. At least not on modern computers. Sorry, but I'm not going to make all of those changes.

>QRegularExpression(arg % QLatin1String("\\b"));

>Because .arg() has to parse and search for placeholders. Is this really necessary for such simple small string concatenation? Especially when we are in the loop.

I didn't write that. I'm not sure.

>Also, could you explain what is tried to achieve here? (I don't know anything about qTox stylesheets).

We have defined colors and fonts that we use in our UI: https://github.com/ItsDuke/Tox-UI/blob/master/UI%20GUIDELINES.md. Someone thought that it would be good idea to create variables with those colors and fonts and make them accessible from style sheets. So you can use variables like @red, @orange, @themeDark, @mediumBold in CSS files. In Style:resolve() we have a dictionary with those words. Every style sheet that we load will run the Style:resolve() function, which purpose is to parse all of those variables - find words like @green and @baseFont and replace them with a value (a color or a font size). Some style sheets are used in multiple places, so to make sure we don't parse them twice, we have a cache. You can see that cache used in Style::getStylesheet().

>Also, are we sure that key will not have regex-special symbols in it? Maybe worth using QRegularExpression::escape()?

Keys of that dictionary are those CSS variables that I described. They are just simple words defined by us. The only special character they contain is @.

tox-user:
@Diadlo it does matter because I hardcoded those paths in cmake. So if Qt changes them everything will get messed up. A link is a good idea.

tox-user:
@Diadlo configLocation on Windows is: C:/Users//AppData/Local/ and C:/ProgramData/. Qt says in its docs that instead of AppData/Local, AppData/roaming should be used on Windows:
>QStandardPaths::DataLocation Returns the same value as AppLocalDataLocation. This enumeration value is deprecated. Using AppDataLocation is preferable since on Windows, the roaming path is recommended.

sudden6:
>  it does matter because I hardcoded those paths in cmake

Good that you bring that up again, I missed it the first time reviewing...

This: `C:\\users\\%username%\\AppData\\roaming\\qtox)` is really bad. There's no guarantee the path to the "home" directory is `users\<username>\`, it could even not be on `C:\` and then this breaks a lot of stuff. We have to use environment variables for this too. Like `%appdata%`, that's what they are for.

Also I just now noticed, that for shipped themes to work, you have to update the windows installer scripts too.

tox-user:
%appdata% returns this path for me: C:\users\username\Application Data

Talkless:
> @tox-user You also suggest a lot of very small changes that are unimportant. You do that with a lot of my PRs.

I do my best to do for all PR's, not just yours. I'm suggesting to apply best practices. It's not hard.

> I'm just returning an empty string, how is that an overkill? 

[QString::QString()](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.h?h=v5.11.1#n1129) default constructor (which is inline)  just returns static object:
```
inline QString::QString() Q_DECL_NOTHROW : d(Data::sharedNull()) {}
```

Meanwhile, you are using [QString::QString(const char*)]( https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.h?h=v5.11.1#n693) constructor, that calls[QString::fromAscii_helper](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.cpp?h=v5.11.1#n5206), (which is out-of-line function BTW), and it takes new string from [fromUtf8()](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.h?h=v5.11.1#n567) helper, that calls [fromUtf8_helper()](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qstring.cpp?h=v5.11.1#n5320), ugh, that in the end calls [QUtf8::convertToUnicode](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/codecs/qutfcodec.cpp?h=v5.11.1#n307). Basically, it allocates new QString (that has shared state with reference count'n'stuff) with empty buffer. All that indirection just for "empty" string? Yes, it's overkill.

> We don't need to watch out for every single bit of memory like something bad was going to happen when we lose a few bytes.

We're writing in C++ for a reason, I believe.

>  In practice most of those things don't have any relevant impact.

There was blog post somewhere long time ago that measured QtCore (or similar library) startup after migrating to QStringLiteral. It was significant enough, though sorry but I cannot find that link now.

Anyway, I'm not suggesting to go to assembly level or similar, just use tools (macros, wrappers, overloads, etc) that there created for a reason to reduce memory allocations and unnecessary calls. Let's avoid premature pessimization. Again, were' writing in C++ for a reason.

> At least not on modern computers

There's not reason to waste cycles. _especially_ in our days when battery life is so important (lot's of mobile, laptop users). I'm not  saying these particular suggestions are significant, it's about mindset to _not write inefficient code by default_.

> Sorry, but I'm not going to make all of those changes.

That's unfortunate.

Diadlo:
@Talkless Did you talk about this [benchmark](https://catchchallenger.first-world.info/wiki/Benchmark_for_conception#String)?

@tox-user Maybe you will be interested in this [article](https://woboq.com/blog/qstringliteral.html)

tox-user:
@Talkless 
we are using Qt5 which is a heavy C++ framework. The performance impact of those changes is hardly visible. Modern computers can afford it with no problem. qTox is a desktop app, so it's not made for phones, but I'm sure modern phones would easily handle those few more function calls too. I don't think we even support any old devices (do we support Windows XP?). That makes me think you do this only because you are used to or you like to. I don't see why I should spend my time on this, but I don't mind you doing it with your own code. I can't see a benefit in spending time on performance improvements that don't have a visible impact.

@sudden6 
>Remove from CMakeLists.txt and add to Installation.cmake

That will cause qTox to throw errors about missing themes to everyone who runs qTox without doing make install. Are you sure that's the correct approach?

I was thinking of removing those errors, because when someone creates a theme that changes only some small thing in the default theme, they would have to copy all of the files from it. Instead it would be better if they could only include the modified files and the rest would be loaded from resource. What do you think?

Diadlo:
@tox-user Oh... I found a man who is to blame for the fact that everything freezes and consume a lot of memory :D
This is not that kind of chnage that requires algorithm rewriting. Just adding a couple of macros. I have launched ~10 user apps on my system. If each of them will consume extra 50MB of memory, together it will be 0.5GB which can be an issue on slow system.

> qTox is a desktop app

It can be misleading, because qTox can be launched on laptops, netbooks, etc.

> I don't see why I should spend my time on this

Why someone else should spend his money to upgrade computer just because you don't want to spend few minutes to make simple change? Write good code right away and you do not need to spend extra time ;)

sudden6:
> That will cause qTox to throw errors about missing themes to everyone who runs qTox without doing make install. Are you sure that's the correct approach?

I don't think so, at least I found no cmake code that would cause this.

tox-user:
@Diadlo
>Oh... I found a man who is to blame for the fact that everything freezes and consume a lot of memory :D

To my knowledge the thing that requires so much memory in qTox is displaying messages in the UI. The more messages for each contacts you have loaded, the more memory qTox will consume. You can create an empty profile and see for yourself. The difference will be tens of megabytes of RAM. If you have too many messages loaded it can take even a few seconds to load the chatlog. @anthonybilinski has done some improvements in that area.

>If each of them will consume extra 50MB of memory, together it will be 0.5GB which can be an issue on slow system.

That is true. If you can prove to me that me using QString instead of QStringLiteral or QLatin1String causes qTox to use 50MB of extra memory I will fix it as soon as possible. Even 5MB would be enough.

>Why someone else should spend his money to upgrade computer just because you don't want to spend few minutes to make simple change?

Upgrade their computer? For what :D? We are not doing any expensive computations :).


@sudden6
That's not what I meant. If we only install themes during "make install", people who don't do that step will get this error from qTox:
`qWarning() << "Stylesheet" << fullPath << "not found. Default will be used";`

Diadlo:
@tox-user All text above was just an example. Not directly a memory or qTox but a wastefulness.

I have not been an active developer of qTox for a long time. But if I were, I would insist on mentioned changes

Talkless:
@tox-user 
> That makes me think you do this only because you are used to or you like to.

I do this because more credible (Qt experts, for example) developers recommends to. I actually don't like `QStringlLiteral`'s or `QLatin1Stgring`'s verbosity (we might get "foo"_qsl user defined literal some day), or `%` hidden magic too much, but it's for the better.

For me it looks that you ignore them because _you don't like_ the said recommendations - the same argument in opposite direction :) ? 

Also, I do it for common sense, as usually we write `int x;` instead of `int* x = new int` by default, don't we? (it's C++, not Java, right?). In the same way there's no reason to write `QString("")` when it's better to `QString()`, `QString{}` or even `{}`, etc. It's not that hard. There's no reason to differentiate "is this desktop or mobile or embedded application; will it run on tiny ARM or huge Xeon?" - we should write better code _by default_.

> I don't see why I should spend my time on this

Yes, feel free to ignore any recommendations on your personal home project, but maybe public code could be handled a bit better, even it takes a bit more of precious personal time (for which I am grateful you and other contributors)?

> Did you talk about this benchmark?

No, but thanks, that's interesting source!

tox-user:
Yes, you are right that I don't like spending my time on something that doesn't make any difference. I don't care about what Qt experts say. What I care about is writing readable code that works well. We didn't care about this before, so I didn't realise qTox's development was going in the direction of doing useless "performance improvements" that don't have any impact. I didn't realise a few KB of memory lost was all we cared about. That is very sad and makes me not want to write code for this project anymore, because it starts to become impossible to post a PR without having to argue about a bunch of useless changes.

Take a look at the state of qTox. It has only a few active developers (probably still more than toxcore), which is not enough to do everything we need done (but we do more than toxcore) and it's barely enough to keep this project alive. Think of all the features that we haven't implemented yet, for example: spell check, push to talk, themes, plugins, resuming file transfers across restarts, system notifications. A ton of existing bugs: broken urls, broken emojis, bootstrapping from wrong nodes, not remembering the default profile, big memory consumption by displayed messages. Add to that lack of distribution packages, lack of reproducible builds and all the refactoring that is required. There are more important issues and I don't understand why you care so much that I use a convenient and universal QString instead of some specific type of string. That I return QString("") instead of {}. For the sake of invisible "performance improvements"? Why do we suddenly care about this now? What has changed in the last few months?

Talkless:
> What has changed in the last few months?

Nothing, really. You are overacting (you could just apply these changes instead of spending time writing long rant thought :) ).

Just ignore my suggestions, I'm not a maintainer to block any PR.

tox-user:
I'm still waiting for the decision on paths and error messages from @sudden6 and @Diadlo. In the meantime I can complain a bit.

tox-user:
I won't work on this anymore in the near future. Feel free to open a new PR and continue if you want to.

PR #5149, opened 2018-05-18
feat: Add spell checking

Diadlo:
Fix #1301

Spell checking implemented by KF5Sonnet.
Suggest currently don't add official support, because some peoples have troubles with it (and I don't know how to fix it).
But feature is really useful.

image

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5149)

PR #5168, opened 2018-06-05
feat: add spell check with QtSpell (WIP)

tox-user:
Implements #1301
Adds spell check in chat.

qtox spell check

qtox spell check2

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5168)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,7 +85,7 @@ endif()
 # Search for headers in current directory.
 include_directories(${CMAKE_BINARY_DIR})
 include_directories(${CMAKE_SOURCE_DIR})
-
+include_directories(SYSTEM /usr/local/include/QtSpell-qt5)
 include(Dependencies)
 
 ################################################################################
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -111,6 +111,7 @@ search_dependency(LIBSODIUM           PACKAGE libsodium)
 search_dependency(LIBSWSCALE          PACKAGE libswscale)
 search_dependency(SQLCIPHER           PACKAGE sqlcipher)
 search_dependency(VPX                 PACKAGE vpx)
+search_dependency(QTSPELL             PACKAGE QtSpell-qt5)
 
 # Try to find cmake toxcore libraries
 if(WIN32)
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -45,6 +45,7 @@
 #include <QFileDialog>
 #include <QKeyEvent>
 #include <QMessageBox>
+#include <QtSpell.hpp>
 
 /**
  * @class GenericChatForm
@@ -142,6 +143,8 @@ GenericChatForm::GenericChatForm(const Contact* contact, QWidget* parent)
     connect(&s, &Settings::chatMessageFontChanged, this, &GenericChatForm::onChatMessageFontChanged);
 
     msgEdit = new ChatTextEdit();
+    QtSpell::TextEditChecker* checker = new QtSpell::TextEditChecker(this);
+    checker->setTextEdit(msgEdit);
 
     sendButton = createButton("sendButton", this, &GenericChatForm::onSendTriggered);
     emoteButton = createButton("emoteButton", this, &GenericChatForm::onEmoteButtonClicked);


Diadlo:
What if I write text not in system language?

tox-user:
It's easy to add a setting for spell check language. From [QtSpell's readme](https://github.com/manisandro/qtspell):
```
// optional: set the language (by default, the system locale is used)
checker->setLanguage("de_CH");
```

System language is used only by default. Right now this PR doesn't work for me so I posted it hoping that maybe someone will test and help me figure out what is wrong.

tox-user:
Now it works. @qTox/core @qTox/linux-team please test on different operating systems.

Talkless:
Do we really want this kind of dependency? It's no available neither in Debian [0]  nor in openSUSE [1] as example.

(Or is this a excuse to layzy-out a test run? :) )

[0] https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=893913
[1] https://software.opensuse.org/search?utf8=%E2%9C%93&q=QtSpell

Talkless:
Spelling does work on Debian Sid (KDE) with hunspell package installed, but it would be nice if it would remember my language selection on popup menu. I have to select my native language every time qTox is started.

Also, it would be nice if it could match spelling language to the selected qTox's UI language by default?

sudden6:
Since there are now two open PRs with spell checking (#5149 is the other one) we have to decide which one to use as we can't use both libraries.

Currently I'm in favour of #5149 and I'll list my reasons here:
- Established packages available (Debian, OpenSUSE)
- Automatic language detection
- Seems to also support the Enchant backend, plus some more
- #5149 works on my system (OpenSUSE Tumbleweed), this seems to not work

Are there some things I'm maybe not aware in this PR that makes it better?

tox-user:
>Currently I'm in favour of #5149 and I'll list my reasons here
>Established packages available (Debian, OpenSUSE)

The version of sonnet shipped on many distros (like Debian stable) is broken and simply doesn't work. We need a library that works for everyone.

>Automatic language detection

Is it really a useful feature? The default would be system language or UI language anyway and with QtSpell you can change the language just for a particular conversation if you want.

>Seems to also support the Enchant backend, plus some more

This one supports multiple backends too.

>#5149 works on my system (OpenSUSE Tumbleweed), this seems to not work

It doesn't work on mine and many others. Compiling it is too difficult, so there is no easy way to get a working version.

So I think we should try to find out why it doesn't work on your OS. If it's a missing package that will be easy to fix. Otherwise we will have to find a new library (**but not sonnet since it doesn't work!**).

>Do we really want this kind of dependency? It's no available neither in Debian [0] nor in openSUSE [1] as example

We can use ANY spell check library that works. Sonnet didn't because versions that are available in many distros have a bug with dictionary paths. There is a workaround to that bug but we can't expect everyone to do it. I picked this library to see if it works. Spell check feature was requested in 2015. Not having it in a messenger is embarrassing. I just want it to be available already. If there is a working library that is as easy to implement we can try it instead of this one.

**So please pick a library that works on all systems. It doesn't have to be QtSpell and it can't be sonnet.**

>it would be nice if it would remember my language selection on popup menu

That comes with the library. I will need to add a proper setting for selecting spell check language. First we need to find out if it works on all OSes.

>Also, it would be nice if it could match spelling language to the selected qTox's UI language by default?

Should be doable.

Diadlo:
@tox-user 

> The version of sonnet shipped on many distros (like Debian stable) is broken 

"many" + one example. It's not cool. Event two (+ Ubuntu 18.04) not so many. As @Talkless said QtSpell "no available neither in Debian nor in openSUSE". So here is Sonnet is better :P

> Is it (Automatic language detection) really a useful feature?

Yes. I use two languages in Tox and it's very useful.

> you can change the language just for a particular conversation

An acceptable solution

> It (sonnet) doesn't work on mine

What distro do you use? Is there an ability to install new version of sonnet not from repo but from manually downloaded package?

tox-user:
>"many" + one example. It's not cool. 

We know for sure it doesn't work on Debian 9 and Ubuntu 16.04. We are unsure about Ubuntu 18.04. Other distros and OSes have not been tested yet.

>As @Talkless said QtSpell "no available neither in Debian nor in openSUSE". So here is Sonnet is better :P

QtSpell is easy to compile. It only requires enchant. We could easily do it from our simple_make.sh script. Sonnet is impossible to compile. If you all think this too much, we should then use another library that is not Sonnet and QtSpell. When we agree that QtSpell is a wrong choice, I will try to find something better.

>What distro do you use? Is there an ability to install new version of sonnet not from repo but from manually downloaded package?

Debian 9. There are binaries available for download on KDE's website, but I don't want to use binaries compiled by somebody else, unless they are reproducible.

Talkless:
> We are unsure about Ubuntu 18.04

I have bunch of VM's already, gonna install 18.10/daily too some day, I could try that.

> The version of sonnet shipped on many distros (like Debian stable) is broken and simply doesn't work. 

Interesting, are there distribution/upstream bug reports?

If the spell checker library we choose is not available in official distro repos, then that means once qTox enter officiall distros too (we now have libtoxcore [0] in Debian Buster), it's gonna be compiled without this feature. That's just means new bug reports :) . Unless someone packages QtSpell...

I'm currently not suggesting anything specific, though _if_ Sonnet now works in Debian Buster (current testing/unstable) and 18.04/18.10 correctly, this probably should be a valid choice. Older distribution users that compile themselves will not have this feature, though sometimes it's "normal" to wait for a new releases for fixed stuff...

[0] https://tracker.debian.org/pkg/libtoxcore

sudden6:
AFAIK sonnet 5.45 was working for me, maybe even 5.44 from ubuntu 17.10 works.
If we take 5.45 as baseline, there are more distros (more important ones) with working sonnet. See
https://pkgs.org/download/kf5sonnet and https://pkgs.org/download/qtspell

As for Debian 9, they have not fixed the libsqlcipher bug too, so qTox is already pretty broken there, IMO it's not worth it blocking this feature just because Debian has older software. Their users now it and want it this way.

> many distros (like Debian stable) is broken and simply doesn't work

Similar things can be said about QtSpell.

> This one supports multiple backends too.

The dependency structure is different for QtSpell (qtspell -> enchant -> (aspell | *)) vs Sonnet (sonnet -> (aspell | *)) but after thinking about it, it shouldn't concern us to much.

> So please pick a library that works on all systems.

This might be impossible, so we should try to get the library working on the most systems.

> So I think we should try to find out why it doesn't work on your OS. If it's a missing package that will be easy to fix. Otherwise we will have to find a new library (but not sonnet since it doesn't work!).

It might be the easier path to get sonnet working though.

> Debian 9. There are binaries available for download on KDE's website, but I don't want to use binaries compiled by somebody else, unless they are reproducible.

You have to agree that this problem is very specific to your use case, but most users will be fine with downloading the binaries from another repository.

tox-user:
>Interesting, are there distribution/upstream bug reports?

Yes, they were mentioned in the first PR that tried to use Sonnet: https://github.com/qTox/qTox/pull/4807. For example here: https://github.com/qTox/qTox/pull/4807#issuecomment-345433739

>If the spell checker library we choose is not available in official distro repos, then that means once qTox enter officiall distros too (we now have libtoxcore [0] in Debian Buster), it's gonna be compiled without this feature. That's just means new bug reports :) . Unless someone packages QtSpell...

According to this https://pkgs.org/download/qtspell only a few distros have it. So we can only use a spell checking library if it's available in all major distros?

>Older distribution users that compile themselves will not have this feature, though sometimes it's "normal" to wait for a new releases for fixed stuff...

People have waited 3 years for this feature. There must be a way to make it work for everyone?


>maybe even 5.44 from ubuntu 17.10 works

That's only an assumption. We can't know until we test it.

>As for Debian 9, they have not fixed the libsqlcipher bug too

It's Debian 9 and Ubuntu 16.04. There might be more distros. We don't know for sure in which version of sonnet this bug got fixed and it has been broken for a long time. What about Windows and MacOS?

>Similar things can be said about QtSpell.

You are assuming again. So far we only tested it on Debian stable, Debian testing and openSuse Tumbleweed. Only that last doesn't work. We don't know the reason yet. Maybe it's some missing package and not a bug in the library. It would be nice if you could report it in https://github.com/manisandro/qtspell. Maybe then we will find out faster.

>This might be impossible, so we should try to get the library working on the most systems.

So after 3 years of waiting for this feature we are going to implement something that is broken and doesn't work for many people? There must be a way to do this right. I'm pretty sure other clients figured it out. As I said I'm fine with using any library that works on all systems. We definitely know that Sonnet isn't one of them, so I don't understand why you want to use it.

>It might be the easier path to get sonnet working though.

If we can get it working, I don't mind using it, but right now it's broken. Here is one way of getting it to work: https://github.com/qTox/qTox/pull/4807#issuecomment-345433739

>You have to agree that this problem is very specific to your use case, but most users will be fine with downloading the binaries from another repository.

Maybe it is, but I think most people compile programs from source to not use binaries. Are we going to tell people to download binaries for other libraries we use too or only this one? The workaround I mentioned earlier is probably a better idea.

Diadlo:
@tox-user  Tested on Arch Linux. Works fine. But unfortunately can't check two languages in one messages.

> People have waited 3 years for this feature. There must be a way to make it work for everyone?

Maybe. But it can be really hard. There are problems with Linux distros: different packages/libraries/environments. Because of this you can find a lot of distro-specific bug reports.

> People have waited 3 years for this feature. There must be a way to make it work for everyone?

Waiting long time not related with working for everyone. All implemented things should work for everyone if it possible. If it isn't possible, it should work to as many people as possible. For example, qTox doesn't work on Debian oldstable, because it's very hard to maintain it even to minimal work.

> That's only an assumption. We can't know until we test it.

Same for QtSpell and any other library. We can't test it with any dirsto and any OS. So we assume, that X works for Y

>  There might be more distros

Or might not. Same as above.

What problem with compiling sonnet do you have? Maybe we can help with it?

tox-user:
Good to hear that it works on Arch. Can we try to find a library that works on the most systems? We can always go back to Sonnet if we don't find a better one. This is a very important feature and I want it to be done right. Perhaps if we find a library that works better than Sonnet we could use it as default, but still allow people to compile qTox to use Sonnet if they want to have this multi language feature?

The problem I have with compiling Sonnet is that it requires me to compile extra-cmake-modules. I think it was a lot of dependencies, but at the moment I can't find the list. Right now it's complaining about some file missing: Qt5QuickConfig.cmake or qt5quick-config.cmake.

sudden6:
@Diadlo can sonnet even check two languages in one message?

@tox-user I think the file you're missing can be found in the following package: https://packages.debian.org/search?searchon=contents&keywords=Qt5QuickConfig.cmake&mode=path&suite=stable&arch=any

The package content search of Debian is quite powerful to find missing dependencies.

Diadlo:
@sudden6 Yes. 
image

tox-user:
@sudden6 thanks, that solved that particular issue. Unfortunately Sonnet 5.45 requires Qt 5.8, which is newer than what Debian stable has. I would have to compile Qt 5.8 too.
```
CMake Error at CMakeLists.txt:23 (find_package):
  Could not find a configuration file for package "Qt5" that is compatible
  with requested version "5.8.0".
```

@Diadlo
That's a cool feature if it works correctly. QtSpell has one interesting feature too. You can right click on a misspelled word and it will show you spelling suggestions.

Diadlo:
@tox-user > You can right click on a misspelled word and it will show you spelling suggestions.

Hmm.. I thought it's "must have" feature. Of course sonnet has it too
image

Talkless:
@tox-user 
>  Unfortunately Sonnet 5.45 requires Qt 5.8

Just built Diadlo:feat/spellchecking on Ubuntu 18.04 LTS VM - it has Qt 5.9, and spell checking works (not sure how to change spelling language though). 

Yes it's unfortunate that current Debian stable cannot have it, but it will be fixed in upcoming Buster release, so since Ubuntu 18.04 and Debian Buster we will have spell checking while using default distribution packages, meaning easier deployment.

We can choose more "reckless" solution now, or wait for more cleaner approach later. Of course, it time we might get QtSpell in official repositories, but _it will not come into Ubuntu 18.04 LTS_ that will live for 5 years. The time is up.

tox-user:
>Just built Diadlo:feat/spellchecking on Ubuntu 18.04 LTS VM
>since Ubuntu 18.04 and Debian Buster we will have spell checking while using default distribution packages

I don't understand. Are we going to compile it or use distro's packages? Packages are broken on at least Ubuntu 16.04 and Debian Stretch. Ubuntu 18.04 comes with Sonnet 5.44, which has not been tested by us and we don't know if it works. I don't know about Ubuntu 17.10 either because it uses an old version 5.38 (https://packages.ubuntu.com/search?keywords=sonnet&searchon=names&suite=bionic&section=all, https://packages.ubuntu.com/search?keywords=sonnet&searchon=names&suite=artful&section=all).

So instead of trying to find a library that works for all you still want to use one that doesn't. Instead of having a working spell check you want it to not work for many people. I told you that we don't have to use QtSpell. I'm sure there exist more than 2 spell checking libraries. When qTox has missing or not working features that makes it hard to recommend it to anyone. To my knowledge Sonnet is only broken on some GNU/Linux distros, but I'm curious if you would use a spell check library that only works on Windows 10 and 8, but not on Windows 7.

If after this long discussion you still want to add a broken spell check then I give up. Go ahead and add a broken feature. I hope you realise that it means we won't get a proper spell check for many months. Only because instead of fixing the problem now, we will be waiting for someone to do it for us.

sudden6:
> Are we going to compile it or use distro's packages?

No, we will not compile a spellchecking library, we will use distro packages on every platform possible.

> Ubuntu 18.04 comes with Sonnet 5.44, which has not been tested by us and we don't know if it works.

If I understand @Talkless correctly, he used the distro packages from 18.04 and it worked.

> To my knowledge Sonnet is only broken on some GNU/Linux distros, but I'm curious if you would use a spell check library that only works on Windows 10 and 8, but not on Windows 7.

It seems sonnet itself can work on all platforms that qTox supports according to their sidebar https://api.kde.org/frameworks/sonnet/html/index.html

As for the Windows 7 question, yes I think it makes no sense to develop new features for Windows 7, because mainstream support ended 3 years ago. If it works as on 10 or 8 it's fine, but if a feature doesn't work on 7 we should just disable it.

Talkless:
>  Packages are broken on at least Ubuntu 16.04 and Debian Stretch.

That's sad but I could only suggest a "Wontfix" for these releases.

>  I don't know about Ubuntu 17.10...

Irrelevant, these intermediate Ubuntu versions are supported only 9 months...

> If I understand @Talkless correctly, he used the distro packages from 18.04 and it worked.

Yes, exactly. Test was really basic, but it worked. I could try that openSUSE Tumbleweed too.

sudden6:
> I could try that openSUSE Tumbleweed too.

No need, I tested there already.

tox-user:
This won't be merged, so closing. I recommend everyone running an unsupported OS like Debian to try using this code to get a working spell check.

PR #5175, opened 2018-06-11
Windows cross compile: version bumps

nurupo:
Needs someone to test ffmpeg video capture.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5175)


nurupo:
Looks like using static ffmpeg above 3.2 version still fails. It failed with 3.3, 3.4 and now with 4.0. I hoped it would somehow magically resolve on its own in some newer version, but no dice. We need this resolved eventually, we can't stay on 3.2 forever, so I will give a try fixing this linking failure.

nurupo:
Looks like we have to disable bsfs, because it requires ac3 parser which we have disabled. It builds successfully now.

Is it okay to use the newest ffmpeg 4.0 or should I downgrade to a release one below that -- 3.4.2? I'm always a bit iffy about using releases with no patch versions of a software that tends to get a lot of path versions.

Can someone with a Windows system check that this PR doesn't break anything? Especially the video/audio capture. I can run qTox in Wine to smoke-test QREncode and Qt updates, but for testing a/v you want an actual Windows system. I can upload build artifacts for testing if you don't want to build it yourself.

nurupo:
I have noticed that ffmpeg has `--enable-gpl` flag (there is also `--enable-nonfree`), which [enables some GPL-licensed optimizations](https://github.com/FFmpeg/FFmpeg/blob/master/LICENSE.md). By default ffmpeg is built under LGPL. Since qTox is GPL and we link ffmpeg statically anyway, we might as well pass `--enable-gpl` when building ffmpeg for those extra optimizations.

nurupo:
Here are the builds I have made if you want to test but can't build.

[qtox-i686-release.zip](https://github.com/qTox/qTox/files/2095717/qtox-i686-release.zip)
[qtox-x86_64-release.zip](https://github.com/qTox/qTox/files/2095719/qtox-x86_64-release.zip)

sudden6:
ping @qTox/windows-team

sudden6:
1. You need to rebase
2. I think disabling bsfs should be no problem AFAICT we don't use it
3.  Nobody seems to be interested in testing :/

nurupo:
@sudden6 rebased.

[Here is a mailing list thread for the FFmpeg build issue we had](http://ffmpeg.org/pipermail/libav-user/2018-June/011188.html), thought someone might be interested. You could also disable just the ac3 bsf, which is the part that was failing, or you could enable the ac3 parser, so that ac3 bsf doesn't fail to build anymore. Disabling all bsf sounds good to me, I wouldn't expect qTox to use [bitstream filters](https://ffmpeg.org/ffmpeg-bitstream-filters.html) at all.

nurupo:
Looks like FFmpeg has released 4.0.1 after my FFmpeg version 4.0 bump commit, so I have updated the commit to use 4.0.1 too.

sudden6:
:lgtm_strong:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5175#-:-LFO7G5B6IskfiV3HTqH:b-jbsf8z)*

PR #5192, opened 2018-06-25
fix: simple_make.sh script

tox-user:
- fix crash when run without parameters
- use cmake instad of automake for building toxcore

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5192)

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -93,23 +93,7 @@ install_toxcore() {
             "${BASE_DIR}/${TOXCORE_DIR}"
 
         pushd ${BASE_DIR}/${TOXCORE_DIR}
-        ./autogen.sh
-
-        # configure
-        if [[ $SYSTEM_WIDE = "false" ]]
-        then
-            ./configure --prefix=${BASE_DIR}
-        else
-            ./configure
-        fi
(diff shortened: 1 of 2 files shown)


tox-user:
I removed the condition that checked for av support in bootstrap.sh. If qTox fails to build without it anyway, it should be ok to remove it.

tox-user:
I've noticed in https://github.com/qTox/qTox/pull/5186 that we still list automake and autoconf as dependencies.

sudden6:
:lgtm_strong: tested successfully

yep, I think we need to clean that file up a bit.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5192#-:-LFuoBtgCP8l2KwAcf8a:bsahckw)*

sudden6:
went ahead and merged this one to publish the fix quickly.

tox-user:
Good decision!

ghost:
This commit seems to cause "./bootstrap.sh --local" to fail. See issue https://github.com/qTox/qTox/issues/5199

PR #5193, opened 2018-06-26
Refactor/options logger

sudden6:

@Diadlo Can you review if this is a good idea?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5193)

### DIFF
--- a/src/core/toxoptions.cpp
+++ b/src/core/toxoptions.cpp
@@ -1,0 +1,109 @@
+#include "toxoptions.h"
+
+#include "src/core/icoresettings.h"
+#include "src/core/toxlogger.h"
+
+#include <QByteArray>
+#include <QDebug>
+
+// TODO(sudden6): replace this constant with the function from toxcore 0.2.3
+static const int MAX_PROXY_ADDRESS_LENGTH = 255;
+
+/**
+ * @brief The ToxOptions class wraps the Tox_Options struct and the matching
+ *        proxy address data. This is needed to ensure both have equal lifetime and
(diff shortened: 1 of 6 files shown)


src/core/toxoptions.cpp
```diff
@@ -0,0 +1,109 @@
...
+        qWarning() << "Core starting with IPv6 disabled. LAN discovery may not work properly.";
+    }
+
+    Tox_Options* tox_opts = tox_options_new(nullptr);
+
+    if(!tox_opts) {
+        return {};
+    }
+
+    auto toxOptions = std::unique_ptr<ToxOptions>(new ToxOptions(tox_opts, proxyAddr.toUtf8()));
```
Diadlo, line 74: Why do you use pointer here? We can create `ToxOptions` on stack and use `std::move` at the end of the function. What do you think?

Diadlo (commented):
Mostly logic of change LGTM. Only one question (see below)

src/core/toxoptions.cpp
sudden6: Because the allocation of the internal Tox_Options struct can fail and `nullptr` is used to represent that.

Diadlo:
:lgtm_strong:

---

Reviewed 6 of 6 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/qtox/qtox/5193#-:-LFwL9jZBYyYj7JhWM9M:b-jbsf8z)*

PR #5244, opened 2018-07-15
fix(docs): update state of TCS compliance

### LINKED ISSUE #5243: Verify our compliance with the Tox Client Standard
Since the plan is to remove clients which don't comply to TCS from tox.chat, we should verify we do in fact comply with TCS.

#5241 adds a new document where we keep track of the current state.

tox-user:
Partially fixes #5243

Notes:
1.0.1 - do we use only encrypted connection when using toxme.io?
2.2.5 - we display avatars in different sizes in different places. Do we display it in full size as well?
3.5.2 - we don't meet this requirement because of markdown support
5.0.4 - does %APPDATA% variable ALWAYS point to AppData/Roaming on all supported Windows systems?
5.1.1, 5.1.2 - we don't store avatars in the root of client's current working directory. It would be a contradiction to 5.0.4. This part of TCS needs to be rewritten

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5244)


sudden6:
1.0.1: I took a look at the code and it's using `https` in the URL, no idea though if that protects against downgrade attacks to http

2.2.5 I think that means it must be square, not that we are not allowed to scale it.

3.5.2 MD support can be turned of or into hybrid mode, not sure if that would be ok though. We should extend TCS to specify which MD flavour or formatting style should be used.

5.1.3 Is a bad idea for privacy, since you expose all PKs of friends by following it.

5.1.2 Makes no sense, maybe they meant to store it in OS specific directory?

tox-user:
>1.0.1: I took a look at the code and it's using https in the URL, no idea though if that protects against downgrade attacks to http

I think it depends on the server but maybe we can somehow make sure to never allow http connection?

>5.1.2 Makes no sense, maybe they meant to store it in OS specific directory?

I guess it's just outdated.

>5.1.3 Is a bad idea for privacy, since you expose all PKs of friends by following it.

Very good point.

anthonybilinski:
If we're connecting via HTTPS, we can't be downgraded to HTTP (our HTTPS protocol version may be downgraded, though, depending on both our's and the server's lowest support HTTPS protocol version). https://moxie.org/software/sslstrip/ works by intercepting HTTP, preventing it from starting to connect to HTTPS:
>It will transparently hijack HTTP traffic on a network, watch for HTTPS links and redirects, then map those links into either look-alike HTTP links or homograph-similar HTTPS links.

If our client is connecting by HTTPS explicitly, if it's tampered with or intercepted we we will get certificate errors, whether or not the server is using HSTS.

tox-user:
What now?

sudden6:
sorry, totally forgot about this.

Seems `1.0.1` can be changed to `Y` after the comment from @anthonybilinski 

I'm strongly against implementing 5.1.3

For 3.5.2 and 5.1.2 we'd need a clarification in the standard, IDK what to do about this one?

sudden6:
Also we should consider the possibility that TCS is effectively dead, I got no responses to https://github.com/Tox/Tox-Client-Standard/pull/35

tox-user:
I expected it might be dead, but that's good to know. We will need to fork it then, give access to the repo to all currently active developers, propose new modifications and make a decision together (by vote?). Maybe we should close this PR and come back to it after we fix the standard?

PR #5261, opened 2018-07-20
Move platform ifdefs to .cpp files and remove `#error`.

iphydf:
This allows us to always compile all .cpp files, making the build system
simpler while not increasing complexity in the source, i.e. this is a net
reduction of complexity. I can see some weak arguments for wanting strict
checking, but I don't see what bugs could be prevented by that
strictness.

This PR does not itself simplify the CMake build system, but opens the
door to doing so.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5261)

### DIFF
--- a/src/platform/camera/directshow.cpp
+++ b/src/platform/camera/directshow.cpp
@@ -21,6 +21,8 @@
 
 #include "directshow.h"
 
+#ifdef Q_OS_WIN
+
 // Because of replacing to incorrect order, which leads to building failing,
 // this region is ignored for clang-format
 // clang-format off
@@ -255,3 +257,5 @@ QVector<VideoMode> DirectShow::getDeviceModes(QString devName)
 
     return modes;
 }
+
+#endif  // Q_OS_WIN
--- a/src/platform/camera/directshow.h
+++ b/src/platform/camera/directshow.h
@@ -26,10 +26,6 @@
 #include <QString>
 #include <QVector>
 
-#ifndef Q_OS_WIN
-#error "This file is only meant to be compiled for Windows targets"
-#endif
-
 namespace DirectShow {
 QVector<QPair<QString, QString>> getDeviceList();
 QVector<VideoMode> getDeviceModes(QString devName);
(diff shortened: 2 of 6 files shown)


anthonybilinski:
IMO it seems easier to manage it in source but less correct. Compiling empty Win specific source files on Linux seems like a bit of a hack, and managing platforms and which parts to compile seems directly the job of the build system.

I'll do a bit of research and update tomorrow.

EDIT: After some research I still feel like cmake is the correct place to manage the OS-specific implementations, and also leads to clear compile-time errors rather than more confusing link-time errors. I'm not convinced this change is an improvement.

sudden6:
After some discussion, I agree with @anthonybilinski so I'll close this.

iphydf:
Can you summarise the discussion here?

sudden6:
In our discussion the main points against this change were:

- deciding which file should be compiled is the job of the build system
- moving this to the C files doesn't really change anything from the software architecture PoV

in favor of the change:
- all files would be visible in the project overview of qtcreator

iphydf:
Additional in favour of the change:
* Simpler build system.
* Ability to run static analysis on `$(find . -name "*.cpp")` (preprocessor error = bail out).
* More uniformity across system-dependent implementation files.

Instead of arguments against the change ("it's no change from some point of view" sounds pretty neutral to me), could you list some arguments that would be in favour of going the other direction? What if this PR were the presumed current way, why would you want to change it to the actual current way?

Diadlo:
@iphydf Not doing useless changes (maybe changes of current PR is not useless, word used just for explain) is one of favor. So "it's no change from some point of view" can be used as argument against something.

**Added:**
* "Simpler build system". I agree with @anthonybilinski and @sudden6, that platform specific logic should be moved at build system, if it's possible.
* "static analysis". Good point, but probably `find *.cpp` is not a right way to select source files. Since here can be `hpp` files, platform specific files (like in current implementation), etc.

iphydf:
Agreed, useless changes waste time.

So, can you add this rule to the coding conventions, with all the pros and cons and a decision based on these? Also, any idea how to enforce the rule? The change this pr introduces is enforcing itself, so that's trivial. You'll want to add something to enforce your wish for additional complexity.

PR #5270, opened 2018-07-25
refactor(core): cleanup thread and timer destruction in Core and CoreAV

sudden6:

@hkarel can you test?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5270)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -52,21 +52,17 @@ Core::Core(QThread* coreThread)
 {
     toxTimer.setSingleShot(true);
     connect(&this->toxTimer, &QTimer::timeout, this, &Core::process);
+    connect(coreThread, &QThread::finished, &toxTimer, &QTimer::stop);
 }
 
 Core::~Core()
 {
-    if (QThread::currentThread() == coreThread.get()) {
-        killTimers();
-    } else {
-        // ensure the timer is stopped, even if not called from this thread
-        QMetaObject::invokeMethod(this, "killTimers", Qt::BlockingQueuedConnection);
-    }
-
-    coreThread->exit(0);
-
     // need to reset av first, because it uses tox
     av.reset();
+
+    coreThread->exit(0);
+    coreThread->wait();
+
     tox.reset();
 }
 
@@ -308,10 +304,6 @@ void Core::process()
     QMutexLocker ml{coreLoopLock.get()};
 
     ASSERT_CORE_THREAD;
-    if (!isReady()) {
-        av->stop();
-        return;
-    }
 
     static int tolerance = CORE_DISCONNECT_TOLERANCE;
     tox_iterate(tox.get(), this);
@@ -1481,16 +1473,4 @@ void Core::setNospam(uint32_t nospam)
 
     tox_self_set_nospam(tox.get(), nospam);
     emit idSet(getSelfId());
-}
-
-/**
- * @brief Stops the AV thread and the timer here
- */
-void Core::killTimers()
-{
-    ASSERT_CORE_THREAD;
-    if (av) {
-        av->stop();
-    }
-    toxTimer.stop();
 }
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -257,7 +257,6 @@ private:
     static void registerCallbacks(Tox* tox);
 
 private slots:
-    void killTimers();
     void process();
     void onStarted();
 
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -91,6 +91,7 @@ CoreAV::CoreAV(Tox* tox)
 
     iterateTimer->setSingleShot(true);
     connect(iterateTimer.get(), &QTimer::timeout, this, &CoreAV::process);
+    connect(coreavThread.get(), &QThread::finished, iterateTimer.get(), &QTimer::stop);
 
     toxav = toxav_new(tox, nullptr);
 
@@ -113,13 +114,10 @@ CoreAV::~CoreAV()
     for (const auto& call : calls) {
         cancelCall(call.first);
     }
-    killTimerFromThread();
-    toxav_kill(toxav);
+
     coreavThread->exit(0);
-    while (coreavThread->isRunning()) {
-        qApp->processEvents();
-        coreavThread->wait(100);
-    }
+    coreavThread->wait();
+    toxav_kill(toxav);
 }
 
 const ToxAV* CoreAV::getToxAv() const
@@ -147,18 +145,6 @@ void CoreAV::stop()
     if (QThread::currentThread() != coreavThread.get())
         return (void)QMetaObject::invokeMethod(this, "stop", Qt::BlockingQueuedConnection);
     iterateTimer->stop();
-}
-
-/**
- * @brief Calls itself blocking queued on the coreav thread
- */
-void CoreAV::killTimerFromThread()
-{
-    // Timers can only be touched from their own thread
-    if (QThread::currentThread() != coreavThread.get())
-        return (void)QMetaObject::invokeMethod(this, "killTimerFromThread",
-                                               Qt::BlockingQueuedConnection);
-    iterateTimer.release();
 }
 
 void CoreAV::process()
--- a/src/core/coreav.h
+++ b/src/core/coreav.h
@@ -108,7 +108,6 @@ private slots:
                                 void* self);
     static void audioBitrateCallback(ToxAV* toxAV, uint32_t friendNum, uint32_t rate, void* self);
     static void videoBitrateCallback(ToxAV* toxAV, uint32_t friendNum, uint32_t rate, void* self);
-    void killTimerFromThread();
 
 private:
     void process();


hkarel:
@sudden6, I took this PR for testing. Now need wait when the mistake is appear, or not appear :)

hkarel:
The program crashed when the OS was restarted
[all-bt.txt](https://github.com/qTox/qTox/files/2232705/all-bt.txt)

sudden6:
It seems to crash in CoreFile, do you have any filetransfers or avatar transfers running?

hkarel:
I'm currently experimenting with the transfer of avatars in my client, so most likely the transfer of avatars was.

sudden6:
Ok, that's a good hint, I'll take a closer look at that part of the code then

sudden6:
@anthonybilinski can you review this please?

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5270#-:-LNA72aRD84MMVlrTV5G:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

sphaerophoria:
I think I'm seeing some assertions due to this commit. I'm seeing the following error

```
[05:27:36.981 UTC] kernel/qcoreapplication.cpp:576 : Fatal: ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to objects owned by a different thread. Current thread 0x0x3b17d3d0. Receiver '' (of type 'CoreAV') was created in thread 0x0x3c962ca0", file kernel/qcoreapplication.cpp, line 576
```

With the following backtrace

```
1  QMessageLogger::fatal                                                                                                                    qlogging.cpp             872  0x5c1fdca  
2  qt_assert_x                                                                                                                              qlogging.h               91   0x5c1ac43  
3  QCoreApplicationPrivate::checkReceiverThread                                                                                             qarraydata.h             206  0x5da861b  
4  QApplication::notify                                                                                                                     qapplication.cpp         2947 0x2b3df884 
5  QCoreApplication::notifyInternal2                                                                                                        qcoreapplication.cpp     1048 0x5da8b02  
6  QCoreApplication::sendEvent                                                                                                              qcoreapplication.h       234  0x5dd8285  
7  QObjectPrivate::setParent_helper                                                                                                         qobject.cpp              2023 0x5dd8285  
8  QObject::~QObject                                                                                                                        qobject.cpp              1034 0x5dd8a76  
9  QTimer::~QTimer                                                                                                                          qtimer.cpp               161  0x5ddadb2  
10 QTimer::~QTimer                                                                                                                          qtimer.cpp               159  0x5ddadcd  
11 std::default_delete<QTimer>::operator()                                                                                                  unique_ptr.h             81   0x59823e   
12 std::unique_ptr<QTimer, std::default_delete<QTimer>>::~unique_ptr                                                                        unique_ptr.h             274  0x5a2393   
13 CoreAV::~CoreAV                                                                                                                          coreav.cpp               112  0x46a25a   
14 CoreAV::~CoreAV                                                                                                                          coreav.cpp               121  0x46a297   
15 std::default_delete<CoreAV>::operator()                                                                                                  unique_ptr.h             81   0x5981be   
16 std::unique_ptr<CoreAV, std::default_delete<CoreAV>>::reset                                                                              unique_ptr.h             382  0x5a1ffc   
17 Core::~Core                                                                                                                              core.cpp                 61   0x46fc7b   
18 Core::~Core                                                                                                                              core.cpp                 67   0x46fd63   
19 std::default_delete<Core>::operator()                                                                                                    unique_ptr.h             81   0x59813e   
20 std::unique_ptr<Core, std::default_delete<Core>>::~unique_ptr                                                                            unique_ptr.h             274  0x5a1e33   
21 Profile::~Profile                                                                                                                        profile.cpp              228  0x4091b7   
22 Profile::~Profile                                                                                                                        profile.cpp              241  0x4091e5   
23 Nexus::~Nexus                                                                                                                            nexus.cpp                70   0x406363   
24 Nexus::~Nexus                                                                                                                            nexus.cpp                76   0x4063a9   
25 Nexus::destroyInstance                                                                                                                   nexus.cpp                264  0x407b4d   
26 cleanup                                                                                                                                  main.cpp                 71   0x4019da   
27 QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void ( *)()>::call(void ( *&)(), void * *)                     qobjectdefs_impl.h       128  0x56a819   
28 QtPrivate::FunctionPointer<void ( *)()>::call<QtPrivate::List<>, void>(void ( *)(), void *, void * *)                                    qobjectdefs_impl.h       191  0x57f134   
29 QtPrivate::QStaticSlotObject<void ( *)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase *, QObject *, void * *, bool *) qobject_impl.h           90   0x57fad1   
30 QtPrivate::QSlotObjectBase::call                                                                                                         qobjectdefs_impl.h       376  0x5dd1690  
31 QMetaObject::activate                                                                                                                    qobject.cpp              3754 0x5dd1690  
32 QMetaObject::activate                                                                                                                    qobject.cpp              3633 0x5dd1acc  
33 QCoreApplication::aboutToQuit                                                                                                            moc_qcoreapplication.cpp 249  0x5da7b26  
34 QCoreApplicationPrivate::execCleanup                                                                                                     qcoreapplication_p.h     76   0x5daeca5  
35 QCoreApplication::exec                                                                                                                   qscopedpointer.h         138  0x5db057b  
36 QGuiApplication::exec                                                                                                                    qguiapplication.cpp      1761 0x165dee0d 
37 QApplication::exec                                                                                                                       qapplication.cpp         2901 0x2b3d828e 
38 main                                                                                                                                     main.cpp                 377  0x403aba   
```

Looks a lot like the destruction of the timer is turning it off from the wrong thread, if I revert a139a9933b0defbacb9c25c3489fd0b58107026f I stop seeing the assertion

sudden6:
Hmm, then I wonder how to correctly destroy a QObject derived class that has a timer in it. Do you have any ideas how to solve this for the multiple classes that do this in qTox?

sphaerophoria:
Pretty new to Qt so I may be off base, but I believe the following works (assuming the rest of the code is post-refactor). 

```
CoreAV::CoreAV(Tox* tox)
    : coreavThread{new QThread}
    , threadSwitchLock{false}
{
    coreavThread->setObjectName("qTox CoreAV");
    moveToThread(coreavThread.get());
    iterateTimer.reset(new QTimer{this});
```

Thought process here is that we construct the timer on a thread we never use it on ever again since we immediately move ourselves to the CoreAV thread, instead we can defer construction of the timer until we're on the thread we plan to use it from.

What is tripping me up is that the Qt docs claim that running our moveToThread function should change the thread affinity for us *and* all our children. Maybe QTimers don't move threads while running for some reason? I'm not really sure.

sphaerophoria:
Interestingly if I remove the unique_ptr and replace it with a raw pointer the problem goes away. Thought process here is that we're calling the destructor from the wrong thread because of unique_ptr. When we call this->moveToThread we must not be actually moving our object, just our object from Qt's ownership model perspective. I guess on destruction of CoreAV the member variable destructor for the unique_ptr<QTimer> runs before the QObject destructor who manages pointers from Qt's perspective. This destructs the timer from the thread that CoreAV was created on, not the thread it was moved to causing the assertion.

Or maybe I'm just leaking memory.

Please note the above has not been fact checked I'm just guessing.

Thanks @anthonybilinski  for the suggestion.

sphaerophoria:
Found some data to support previous theory..

> In a non-delegating constructor, initialization proceeds in the following order:
> — First, and only for the constructor of the most derived class (1.8), virtual base classes are initialized in
> the order they appear on a depth-first left-to-right traversal of the directed acyclic graph of base classes,
> where “left-to-right” is the order of appearance of the base classes in the derived class
> base-specifier-list
> .
> — Then, direct base classes are initialized in declaration order as they appear in the
> base-specifier-list
> (regardless of the order of the
> mem-initializers
> ).
> — Then, non-static data members are initialized in the order they were declared in the class definition
> (again regardless of the order of the
> mem-initializers
> ).
> — Finally, the
> compound-statement
> of the constructor body is executed

Source http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf

So we destruct our members before we destruct our base class. Making a couple assumptions that

1. When moving an object to a QThread it doesn't actually move the object, just the context in which Qt calls it from. This means that the destructor will still run on the thread it was created on since the object lives on that thread's stack
2. QObject will properly destruct from the correct thread.
3. Destructor order is reverse of initialization order

sudden6:
I think you're correct here with the thesis that `QTimer` does only move in the context of Qt's thread model, but the `unique_ptr` does not move on `moveToThread`.

I just noticed, that due to `QTimer` being a Qt object we don't even need the `unique_ptr` and we can make it a member of the `CoreAV` class directly since the QObject destructor will clean it up.

I also found this important note:
_Note: A QObject's member variables do not automatically become its children. The parent-child relationship must be set by either passing a pointer to the child's constructor, or by calling setParent(). Without this step, the object's member variables will remain in the old thread when moveToThread() is called._

Here http://doc.qt.io/qt-5/qobject.html and I guess this type of bug is present many times in qTox :(

sudden6:
I made a PR trying to fix this https://github.com/qTox/qTox/pull/5378

PR #5305, opened 2018-08-25
fix(audio): specify format for sounds and make sounds follow it

sudden6:
This should reduce the problem that some sounds are very silent and some are loud.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5305)


anthonybilinski:
I like the idea of standardizing format and volume but it seems like at least your notification.pcm is off. After the change, it sounds kind of muted, as if under water.

Looking at the wave forms of:
1) original notification.pcm (converted to wav) `ffmpeg -f s16le -acodec pcm_s16le -ac 1 -ar 44.1k -i before.pcm before.wav`
2) original wav converted to 48K `ffmpeg -i before.wav -acodec pcm_s16le -ac 1 -ar 48k newafter.wav`
3) 48K amplified to -1dB peak using audacity (effects -> amplify -> new peak amplitude)
4) your notification converted to wav `ffmpeg -f s16le -acodec pcm_s16le -ac 1 -ar 48k -i after.pcm after.wav`
image

the first two sound identical, the third sounds the same but louder, and the fourth sounds wrong. I'm not sure what's wrong with it from the waveform, the only thing I notice visually is that the silence at the start and finish is removed, but it certainly sounds wrong.

As for -1dB, I don't really understand audio levels, this is quite a bit louder than the old notification.pcm, but if I set my headphones to a comfortable volume with my browser mixed to full volume and youtube's volume at full volume, then the new notification amplified to -1dB is at a good volume relative to random videos, and the old one is too quiet, so I guess this is a good level ¯\\\_(ツ)\_/¯

sudden6:
Thanks for taking a closer look at the audio files. I'll check the notification sound again.

The -1dB level is to make best use of the available 16bit range, if we limit to a lower level we would waste more and more dynamic range, but if the limit is too high it could cause distortion when amplified and not converted to float or a format with more bits. I choose -1dB as a trade-off, but I have to say I have no experience with audio mixing so if you know any best practices/rules they are probably better :P

sudden6:
@anthonybilinski re-did notification sound. I applied some filter to it in an attempt to make it sound better, but testing again with a headset it made it worse actually...

anthonybilinski (commented):
> Reviewed 12 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5305)* status: 0 of 1 LGTMs obtained

anthonybilinski:
Sounds good now. I agree with the -1dB choice. I also noticed (and heard) that you chopped off this nasty jaggy bit at the start of ToxOutgoing and ToxEndCall:
image
which use to cause an annoying crack a the start of the files. Good job cleaning those up, I'm guessing they're WAV file headers that were converted directly into PCM or something.

I'm not sure it makes sense to include the original/ dir with the WAV versions since there's no lossy conversion or anything, so it's basically just duplicating the info, plus the files are in git history before the amplification. Do you think there are benefits?

sudden6:
Yes, we actually played back the WAV headers...

I'm not sure that resampling is a lossless conversion, so I left the originals. Also as an example for future stuff, that may be higher quality but downsampled for qTox.

anthonybilinski (approved):
Fair enough, :lgtm_strong: 

> Reviewed 4 of 16 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5305#-:-LMpj6GY4mZLhHc3llOO:b-xs5np3)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5313, opened 2018-08-29
fix(settings): set default proxy type to None, enabling UDP by default

### LINKED ISSUE #5311: UDP and lan discovery greyed out by default
##### Brief Description

OS: Linux, Debian 9
qTox version: v1.16.3.
Commit hash:  8eed684c37bd711de2ca4bd863f5f0509edcdc2d
toxcore:   0.2.3
Qt:  5.11.1
Hardware:  pretty standard

qTox was installed using flatpak io.github.qtox.qTox since Debian does not offer any other (obvious) way to do it anymore.

Reproducible: Always

##### Steps to reproduce

1. Install qTox via flatpak
2. Run qTox
3. check advanced config

##### Observed Behavior

Enable UDP is disabled and greyed out.
LAN discovery is disabled and greyed out.

##### Expected Behavior

Enable UDP is not greyed out.
LAN discovery is not greyed out.
And if they are greyed out, I would expect to see some info as to why (missing libs ?)

##### Additional Info

Unfortunately, I had to install qTox via flatpak. It is my only package ever installed via flatpak.
tox_udp_greyedout

anthonybilinski:
Fix #5311

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5313)

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -400,8 +400,7 @@ void Settings::loadPersonal(Profile* profile)
 
     ps.beginGroup("Proxy");
     {
-        int type = ps.value("proxyType", static_cast<int>(getProxyType())).toInt();
-        setProxyType(static_cast<ProxyType>(type));
+        proxyType = static_cast<ProxyType>(ps.value("proxyType", 0 /* ProxyType::None */).toInt());
         proxyAddr = ps.value("proxyAddr", proxyAddr).toString();
         proxyPort = static_cast<quint16>(ps.value("proxyPort", proxyPort).toUInt());
     }


anthonybilinski:
Before it was using the uninitialized member variable as default proxy type, resulting in blank "Proxy Type" from drop down (int was not one of 3 valid numbers for enum). Could repro on master in with a clean ~/.config/tox. Tested fix and proxy type is correctly none, and is still preserved when changed.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5313#-:-LLPVCNIBhoxM2qYhzAe:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5319, opened 2018-08-30
chore(build): build windows spellcheck deps

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5319)


anthonybilinski:
This should now be working and loaded by qTox. There's the problem of needing dictionaries for each language supported, I'm not sure what the best way to support this on Windows is. On Linux, there are system-wide dictionaries available for each language, and any installed will be available to applications. On Windows, I guess we just pick the most popular languages and bundle those with the install, and if someone's personal language isn't there they have to manually install a file a drop it in the directory?

anthonybilinski:
I've got an English only dictionary packaged now and Sonnet working well on Windows IFF Hunspell is built with debug symbols, if I build Hunspell without debug symbols, it crashes when attempting to load the dict. I've created an issue with Hunspell to see if I can get some help figuring out what's going on: https://github.com/hunspell/hunspell/issues/627

anthonybilinski:
Hunspell has since released version 1.7.0. I updated this PR to include that version - I'll check soon if this fixes the crash on dictionary load.

anthonybilinski:
Sonnet doesn't find hunspell anymore once hunspell's updated to 1.7.0, even after sonnet it upgraded to the latest version available at https://download.kde.org/stable/frameworks, 5.62 (5.63 gives a http 503). The reason it can't find it is probably because of https://github.com/KDE/sonnet/blob/852f24888d09af0b80770f4f80e06b559369e6d3/cmake/FindHUNSPELL.cmake#L43, and hunspell is only installing `libhunspell-1.7`, not a generic name. I'll try patching sonnet to find the newer hunspell. I'm not sure if that's all that's required, or if they're incompatible in other ways.

anthonybilinski:
Patching sonnet to find hunspell 1.7.0 works, and it continues to run correctly when hunspell is built with debug flags, but continues to crash on dictionary load when build without debug flags :(

Basically no change from a year ago, and no response on the hunspell ticket. I guess I'll work on building aspell as the backend for sonnet.

PR #5346, opened 2018-09-19
feat(ui): add update notification enabled with -DUPDATE_CHECK

### LINKED ISSUE #5335: Proposal: "update available" notification based on latest github release or bootstrap versions
Without adding any centralized server polling, it's still possible for qTox to check if it's an older version than other nodes in the network.

Bootstrap nodes return a version number that correspond to toxcore release versions. If qTox collected that information while already bootstrapping, and there is at least a qTox patch release for every toxcore release (not done currently, but it seems like a good idea to me), then anytime qTox bootstraps and sees a node with a newer version of toxcore than what is running, we could show a notifier than there's probably a new version available and that the user should go check.

This is obviously not as complete as the old window auto updater or linux package based updating, but it would be totally cross platform (including source builds, appImage, etc.) and contained within qTox, plus not leak any metadata.

anthonybilinski:
Fix #5335

If up to date:
image

If not up to date:
settings
hover
selected

About
no hover effect like other tabs.
selected looks like other selected tabs, not green.
update button
hover a little lighter, click a little darker, default amounts.
Button opens latest release page, like https://github.com/qTox/qTox/releases/tag/v1.16.3.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)


sudden6 (changes requested):
> Reviewed 5 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/settingswidget.h, line 54 at r1](https://reviewable.io/reviews/qtox/qtox/5346#-LMmKE0bEXjVpavOGhv5:-LMmKE0c9ALvUBPZZ8qx:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/09394ffdb618c7f785d7d3405c926192645972fa/src/widget/form/settingswidget.h#L54)):*
> ```objc
> signals:
>     void updateAvailable(void);
>     
> ```

whitespace error
___
*[src/widget/form/settings/aboutform.cpp, line 164 at r1](https://reviewable.io/reviews/qtox/qtox/5346#-LMmKIAE5FRjJZ7wQPZJ:-LMmKIAE5FRjJZ7wQPZK:b-nshkqb) ([raw file](https://github.com/qtox/qtox/blob/09394ffdb618c7f785d7d3405c926192645972fa/src/widget/form/settings/aboutform.cpp#L164)):*
> ```cpp
> }
> 
> void AboutForm::checkForUpdate()
> ```

I'd be happier if the logic is in it's own class and not in the UI class

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/settingswidget.h, line 54 at r1](https://reviewable.io/reviews/qtox/qtox/5346#-LMmKE0bEXjVpavOGhv5:-LMoR_ju8EqYpMuaejMq:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/09394ffdb618c7f785d7d3405c926192645972fa/src/widget/form/settingswidget.h#L54)):*

whitespace error
</blockquote></details>

Done.
___
*[src/widget/form/settings/aboutform.cpp, line 164 at r1](https://reviewable.io/reviews/qtox/qtox/5346#-LMmKIAE5FRjJZ7wQPZJ:-LMoRalz7kSV8cnpvRLv:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/09394ffdb618c7f785d7d3405c926192645972fa/src/widget/form/settings/aboutform.cpp#L164)):*

I'd be happier if the logic is in it's own class and not in the UI class
</blockquote></details>

Done.

sudden6 (changes requested):
> Reviewed 2 of 10 files at r1, 5 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/widget.cpp, line 375 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpdsZeBSx0W4nEZYPm:-LMpdsZf-A5mtT0uiGM7:bpfxxvm) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/widget.cpp#L375)):*
> ```cpp
> #endif
> ```

useless whitespace
___
*[src/widget/form/settings/aboutform.h, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeyDd2WEviCcUgNev:-LMpeyDd2WEviCcUgNew:b-p7uzoh) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.h#L69)):*
> ```objc
>     Ui::AboutSettings* bodyUI;
>     QTimer* progressTimer;
>     UpdateCheck* updateCheck{nullptr};
> ```

What do you think about moving this to main.cpp or some other higher level file and passing it to AboutSettings via pointer?
___
*[src/widget/form/settings/aboutform.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeImm5f0eWDlGFWsH:-LMpeImm5f0eWDlGFWsI:b-2ff05k) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L99)):*
> ```cpp
>     connect(updateCheck, &UpdateCheck::upToDate, this, &AboutForm::onUpToDate);
>     connect(updateCheck, &UpdateCheck::updateCheckFailed, this, &AboutForm::onUpdateCheckFailed);
>     updateCheck->checkForUpdate();
> ```

AFAICT with this we check for updates on every load of the AboutForm right? I think it would make more sense to make this timer controlled. For example, check on every start of qTox + once a day afterwards. What do you think? Also IMO there should be a setting to disable checking for updates.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/settings/aboutform.h, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeyDd2WEviCcUgNev:-LMpp8UG1krsAl9v_cdw:bpgp9ov) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.h#L69)):*

What do you think about moving this to main.cpp or some other higher level file and passing it to AboutSettings via pointer?
</blockquote></details>

singleton Widget -> SettingsWidget -> AboutFormCurrently, so I can't pass it from any higher than Widget. I can do that now to make the intent clear that UpdateCheck's construction should be moved up further if/when Widget becomes a non-singleton class. Currently AboutForm deletes UpdateCheck after checking for update since it has ownership, but since I'll be adding a timer might as well keep UpdateCheck alive for the lifetime of its owner (Widget for now).
___
*[src/widget/form/settings/aboutform.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeImm5f0eWDlGFWsH:-LMpmnE00qc9HSGh1SvX:bjnos0k) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L99)):*

AFAICT with this we check for updates on every load of the AboutForm right? I think it would make more sense to make this timer controlled. For example, check on every start of qTox + once a day afterwards. What do you think? Also IMO there should be a setting to disable checking for updates.
</blockquote></details>

Currently it only checks once on GUI construction. Adding a timer to also check once a day would help with very long running instances, good idea. Will also add a setting to disable, but would like to default to true because I don't see a very valid disable reason since Tox user -> IP is already visible in the DHT.

sudden6 (changes requested):
> Reviewed 2 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/net/updatecheck.h, line 38 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrSbBQ4Yy5p7quOr8m:-LMrSbBQ4Yy5p7quOr8n:b-ewkmbo) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/net/updatecheck.h#L38)):*
> ```objc
> 
> private:
>     std::unique_ptr<QNetworkAccessManager> manager;
> ```

I think it makes sense to remove `manager` from the class, since it will be used only once in a while so the construction overhead is probably way less than the runtime memory overhead.
___
*[src/net/updatecheck.cpp, line 36 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrUDNbEooeqznVBTmM:-LMrUDNbEooeqznVBTmN:b-jadaar) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/net/updatecheck.cpp#L36)):*
> ```cpp
>     auto reply = manager->get(request);
> 
>     QObject::connect(reply, &QNetworkReply::finished, [this, reply](){
> ```

IMO it would be better to avoid a lambda here, in case the network query needs longer than the lifetime of the object. Because AFAIK the connection is not destroyed when the lambda is destroyed and this could lead to a use-after-free, which burned me multiple times already.
___
*[src/widget/form/settings/aboutform.h, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeyDd2WEviCcUgNev:-LMrSKDZ3DOEqSj-c-pS:bkoeh6v) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.h#L69)):*

singleton Widget -> SettingsWidget -> AboutFormCurrently, so I can't pass it from any higher than Widget. I can do that now to make the intent clear that UpdateCheck's construction should be moved up further if/when Widget becomes a non-singleton class. Currently AboutForm deletes UpdateCheck after checking for update since it has ownership, but since I'll be adding a timer might as well keep UpdateCheck alive for the lifetime of its owner (Widget for now).
</blockquote></details>

ok, the singleton thing makes sense

When you add a timer it probably makes sense to keep `updateCheck` around, but make the class itself hold no ressources besides the timer, see also the comment on `updatecheck.h`
___
*[src/widget/form/settings/aboutform.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeImm5f0eWDlGFWsH:-LMrTEtgDdIEkquT9LE2:bshyolw) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L99)):*

Currently it only checks once on GUI construction. Adding a timer to also check once a day would help with very long running instances, good idea. Will also add a setting to disable, but would like to default to true because I don't see a very valid disable reason since Tox user -> IP is already visible in the DHT.
</blockquote></details>

Yeah, I for example restart qTox not very often, so it would make sense to check once every few days or every week.

Since we also focus on privacy I think it's only fair to provide a disable switch, I also see it as "insurance" in case something goes wrong with this update thing ;) I fully agree with default on though.
___
*[src/widget/form/settings/aboutform.cpp, line 183 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrTpYL5--OxMUHPZKd:-LMrTpYL5--OxMUHPZKe:b7gqy4c) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L183)):*
> ```cpp
>         delete oldLayout;
>     }
>     else {
> ```

change to `} else {`?
___
*[src/widget/form/settings/aboutform.cpp, line 198 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrTw1wAu4TZbMx-X6M:-LMrTw1wAu4TZbMx-X6N:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L198)):*
> ```cpp
>         delete oldLayout;
>     }
>     else {
> ```

same

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/widget.cpp, line 375 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpdsZeBSx0W4nEZYPm:-LTjo68U9BKHwbpVzVJb:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/widget.cpp#L375)):*

useless whitespace
</blockquote></details>

Done.
___
*[src/widget/form/settings/aboutform.h, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeyDd2WEviCcUgNev:-LTp4bxGDXfIg7mf26KO:b-tlxzwm) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.h#L69)):*

ok, the singleton thing makes sense

When you add a timer it probably makes sense to keep `updateCheck` around, but make the class itself hold no ressources besides the timer, see also the comment on `updatecheck.h`
</blockquote></details>

not completely addressed since QNetworkManager is held, but I think this is right for the reason in the UpdateCheck.h comment.
___
*[src/widget/form/settings/aboutform.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMpeImm5f0eWDlGFWsH:-LTp3lod-4hni4DBXq0B:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L99)):*

Yeah, I for example restart qTox not very often, so it would make sense to check once every few days or every week.

Since we also focus on privacy I think it's only fair to provide a disable switch, I also see it as "insurance" in case something goes wrong with this update thing ;) I fully agree with default on though.
</blockquote></details>

Done.
___
*[src/widget/form/settings/aboutform.cpp, line 183 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrTpYL5--OxMUHPZKd:-LTp3kg7D9LqgieL-vOy:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L183)):*

change to `} else {`?
</blockquote></details>

Done.
___
*[src/widget/form/settings/aboutform.cpp, line 198 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrTw1wAu4TZbMx-X6M:-LTp3kF04XaqzaRqBZLg:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/widget/form/settings/aboutform.cpp#L198)):*

same
</blockquote></details>

Done.
___
*[src/net/updatecheck.h, line 38 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrSbBQ4Yy5p7quOr8m:-LTjmOtPDOJRP7ED-SIo:b-5syxv3) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/net/updatecheck.h#L38)):*

I think it makes sense to remove `manager` from the class, since it will be used only once in a while so the construction overhead is probably way less than the runtime memory overhead.
</blockquote></details>

manager's lifetime needs to outlast the request, so I think it's cleanest to just keep it as a member (instead of the lambda I had before). In the future I think it would be best to have only a single QNetworkmanager per Profile so that proxy settings only need to be set once and TCP connections can be reused. I'm not sure how to check the memory cost of the class, but I think all of the request state is stored in just the QNetworkReply and QNetworkRequest, so it should be pretty small when there are no outstanding requests.
___
*[src/net/updatecheck.cpp, line 36 at r2](https://reviewable.io/reviews/qtox/qtox/5346#-LMrUDNbEooeqznVBTmM:-LTjnvGbCSAatLb8LzPx:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8f149f7d7093e063abe9b6da4428aced34f970c7/src/net/updatecheck.cpp#L36)):*

IMO it would be better to avoid a lambda here, in case the network query needs longer than the lifetime of the object. Because AFAIK the connection is not destroyed when the lambda is destroyed and this could lead to a use-after-free, which burned me multiple times already.
</blockquote></details>

Done.

sudden6 (changes requested):
> Reviewed 12 of 23 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/core/icoresettings.h, line 21 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhb7j96Tv5f4C5T45:-LTqhb7k4sRzGUaXE2GN:bh7b8mb) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/core/icoresettings.h#L21)):*
> ```objc
>     };
> 
>     virtual bool getEnableUpdateCheck() const = 0;
> ```

why is this in the core settings interface?
___
*[src/net/updatecheck.cpp, line 40 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqiRRH4WaO3evTwduF:-LTqiRRH4WaO3evTwduG:bpid3n0) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L40)):*
> ```cpp
> void UpdateCheck::checkForUpdate()
> {
>     auto& s = Settings::getInstance();
> ```

Pass reference to settings via constructor to avoid `getInstance()`?
___
*[src/net/updatecheck.cpp, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqiBcJ4I6nIFxHI9EV:-LTqiBcK2lHyR9rfPf0p:b-efjvxw) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L46)):*
> ```cpp
>     }
>     manager.setProxy(s.getProxy());
>     QNetworkRequest request{QStringLiteral("https://api.github.com/repos/qTox/qTox/releases/latest")};
> ```

extract URL to constant 
___
*[src/net/updatecheck.cpp, line 50 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhx533U1OP6zqZs9U:-LTqhx533U1OP6zqZs9V:bvrgx48) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L50)):*
> ```cpp
> }
> 
> void UpdateCheck::handleResponse(QNetworkReply *reply)
> ```

Change `reply` to reference or add `nullptr`check.
___
*[src/net/updatecheck.cpp, line 83 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhrai9Y2R7pGBt7HP:-LTqhrai9Y2R7pGBt7HQ:beypj4a) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L83)):*
> ```cpp
>     }
>     reply->deleteLater();
> }
> ```

strange line ending
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LTqj36w4CzO3Uhlvfj9:bg3tmns) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*
> ```cpp
> {
>     if (!settingsWidget) {
>         updateCheck = std::unique_ptr<UpdateCheck>(new UpdateCheck());
> ```

This seems problematic, because it only constructs the update checker when the user opens the settings. What if the user never opens the settings?
I think construction of `UpdateCheck` should be moved to `Widget` constructor or even better to `main.cpp`
___
*[src/widget/form/settingswidget.h, line 44 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqjc-UDRD0mU6yxxJ0:-LTqjc-UDRD0mU6yxxJ1:bkata1r) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/form/settingswidget.h#L44)):*
> ```objc
>     Q_OBJECT
> public:
>     SettingsWidget(UpdateCheck* updateCheck, QWidget* parent = nullptr);
> ```

make `updateCheck` a reference?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 12 at r4](https://reviewable.io/reviews/qtox/qtox/5346#-LTy3CLL9OCufw_JE0cf:-LTy3CLL9OCufw_JE0cg:b-mgtnp4) ([raw file](https://github.com/qtox/qtox/blob/d287a0bd8b46760ef2579af6cbdbbcece4f74009/CMakeLists.txt#L12)):*
> ```CMake
> option(PLATFORM_EXTENSIONS "Enable platform specific extensions, requires extra dependencies" ON)
> option(USE_FILTERAUDIO "Enable the echo canceling backend" ON)
> option(UPDATE_CHECK "Enable automatic update check" ON)
> ```

note that I've changed UPDATE_CHECK to default to ON despite our earlier conversation. I think the argument that distro packages won't want it on is true, but I think package maintainers will take more time to go over build args than users compiling for themselves, who would benefit from UPDATE_CHECK. I can add a comment to install.md just for package maintainers of this is still a major concern?
___
*[src/net/updatecheck.cpp, line 40 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqiRRH4WaO3evTwduF:-LTtSwb50_64QVvuDEJz:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L40)):*

Pass reference to settings via constructor to avoid `getInstance()`?
</blockquote></details>

Done.
___
*[src/net/updatecheck.cpp, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqiBcJ4I6nIFxHI9EV:-LTtSxCdDu100CI76kBC:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L46)):*

extract URL to constant 
</blockquote></details>

Done.
___
*[src/net/updatecheck.cpp, line 50 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhx533U1OP6zqZs9U:-LTtULPJ2nBirRfApxdC:buhqvc9) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L50)):*

Change `reply` to reference or add `nullptr`check.
</blockquote></details>

Can't change to reference singe it comes directly from QNetworkManager. I think it's guaranteed to be non-null from API, but since we can't guarantee that at compile time, nullptr check makes sense.
___
*[src/net/updatecheck.cpp, line 83 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhrai9Y2R7pGBt7HP:-LTtYesLBkL420o0mya2:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/net/updatecheck.cpp#L83)):*

strange line ending
</blockquote></details>

Done.
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LTv9t9fCSQNeHNTCFgD:btu5bg1) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

This seems problematic, because it only constructs the update checker when the user opens the settings. What if the user never opens the settings?
I think construction of `UpdateCheck` should be moved to `Widget` constructor or even better to `main.cpp`
</blockquote></details>

UpdateCheck was actually being constructed as soon as qTox opened for some reason (with update notification showing right away), but you're right, construction moved to ::init. Initial checkForUpdate call is still after showSettings so that slots will be connected.
___
*[src/widget/form/settingswidget.h, line 44 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqjc-UDRD0mU6yxxJ0:-LTvB7YJDhj8qghLqC7t:bnv03sz) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/form/settingswidget.h#L44)):*

make `updateCheck` a reference?
</blockquote></details>

updateCheck can be nullptr if UPDATE_CHECK_ENABLED is not set, kept as pointer with nullptr checks at all uses.
___
*[src/core/icoresettings.h, line 21 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqhb7j96Tv5f4C5T45:-LTtSWYyEWv2RfT3DAXs:b6ttlnr) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/core/icoresettings.h#L21)):*

why is this in the core settings interface?
</blockquote></details>

Bug. Instead of fixing, I've removed it from settings/UI to reuse the autoupdate setting, since this is essentially a feature reduced version of the same thing.

sudden6 (changes requested):
> Reviewed 12 of 21 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 12 at r4](https://reviewable.io/reviews/qtox/qtox/5346#-LTy3CLL9OCufw_JE0cf:-LTyT4Pb2N_o3Xs8F6sY:beic8we) ([raw file](https://github.com/qtox/qtox/blob/d287a0bd8b46760ef2579af6cbdbbcece4f74009/CMakeLists.txt#L12)):*

note that I've changed UPDATE_CHECK to default to ON despite our earlier conversation. I think the argument that distro packages won't want it on is true, but I think package maintainers will take more time to go over build args than users compiling for themselves, who would benefit from UPDATE_CHECK. I can add a comment to install.md just for package maintainers of this is still a major concern?
</blockquote></details>

will the update check work correctly for builds from master? Does it signal an update even though the user is already further in the future due to building from master? If that's no problem I'm ok with it.
___
*[src/net/updatecheck.cpp, line 32 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyTnkg8tny1ELoTLAc:-LTyTnkg8tny1ELoTLAd:b-xqgrth) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/net/updatecheck.cpp#L32)):*
> ```cpp
> #include <cassert>
> 
> static const QString versionUrl{QStringLiteral("https://api.github.com/repos/qTox/qTox/releases/latest")};
> ```

Instead of static it seems we use an anonymous namespace usually, adopt this here too? (maybe this should go into coding style too)
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LTyU_J83eMAvemn1SnV:b-isj55g) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

UpdateCheck was actually being constructed as soon as qTox opened for some reason (with update notification showing right away), but you're right, construction moved to ::init. Initial checkForUpdate call is still after showSettings so that slots will be connected.
</blockquote></details>

So the check for updates is still only performed when `SettingsWidget` is constructed? Seems like a bug to me? Or will the periodic update check execute correctly in all cases?
___
*[src/widget/widget.cpp, line 845 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyV2ea3KroQ4YeVlb-:-LTyV2ea3KroQ4YeVlb0:bsr7duy) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/widget/widget.cpp#L845)):*
> ```cpp
> void Widget::onShowSettings()
> {
>     if (!settingsWidget) {       
> ```

white space problems
___
*[src/widget/form/settingswidget.cpp, line 69 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyVykV21cjD1zRDBa9:-LTyVykV21cjD1zRDBaA:b-80du72) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/widget/form/settingswidget.cpp#L69)):*
> ```cpp
>     std::unique_ptr<AboutForm> abtfrm(new AboutForm(updateCheck));
> 
> #if UPDATE_CHECK_ENABLED && !AUTOUPDATE_ENABLED
> ```

why the `!AUTOPUDAT_ENABLED` condition? I thought this was completely removed?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 12 at r4](https://reviewable.io/reviews/qtox/qtox/5346#-LTy3CLL9OCufw_JE0cf:-LTyY7SVEmW2XkremQCD:bf404rl) ([raw file](https://github.com/qtox/qtox/blob/d287a0bd8b46760ef2579af6cbdbbcece4f74009/CMakeLists.txt#L12)):*

will the update check work correctly for builds from master? Does it signal an update even though the user is already further in the future due to building from master? If that's no problem I'm ok with it.
</blockquote></details>

It works purely off tag version number, so it will consider "v1.16.3-205-gc589462b" == "v1.16.3" and say that qTox is up to date. If qTox v1.17.0 is released, it will consider "v1.16.3-anything" to be out of date. If the user is past v1.17.0 but without the tag, it will assume an update is available.
___
*[src/net/updatecheck.cpp, line 32 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyTnkg8tny1ELoTLAc:-LTyhHVN5lSt5oU00Dd0:bh1myyt) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/net/updatecheck.cpp#L32)):*

Instead of static it seems we use an anonymous namespace usually, adopt this here too? (maybe this should go into coding style too)
</blockquote></details>

I think static (variable) would deduplicate the variable across instantiations of the class, whereas anoynmous namespace wouldn't. This seems like a really minor benefit and this class should only be instantiated once. Changed to anonymous.
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LTyaAJAEuYgBv3ky8VM:bg5tyar) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

So the check for updates is still only performed when `SettingsWidget` is constructed? Seems like a bug to me? Or will the periodic update check execute correctly in all cases?
</blockquote></details>

It's true that it only performs when SettingsWidget is constructed. I do this to make sure that the update available signal is caught by AboutForm so that the notification shows properly. I'm not sure of a better way to have a delayed signal like that, I guess AboutForm could check saved update state from UpdateCheck whenever it constructs, and have UpdateCheck just cache the result?
___
*[src/widget/widget.cpp, line 845 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyV2ea3KroQ4YeVlb-:-LTya7zA7AWp4xIN205y:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/widget/widget.cpp#L845)):*

white space problems
</blockquote></details>

Done.
___
*[src/widget/form/settingswidget.h, line 44 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqjc-UDRD0mU6yxxJ0:-LTyaZ8pAy_nuObUppFI:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/form/settingswidget.h#L44)):*

updateCheck can be nullptr if UPDATE_CHECK_ENABLED is not set, kept as pointer with nullptr checks at all uses.
</blockquote></details>

Done.
___
*[src/widget/form/settingswidget.cpp, line 69 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyVykV21cjD1zRDBa9:-LTya_btAfZlmNQnWhZ7:b-2881pd) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/widget/form/settingswidget.cpp#L69)):*

why the `!AUTOPUDAT_ENABLED` condition? I thought this was completely removed?
</blockquote></details>

oops, removed. My second commit has both together as a template for how autoupdate can be reintroduced, third commit missed removing this.
___
*[src/widget/form/settings/aboutform.cpp, line 96 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyakSr5KcGTNb9jbFe:-LTyakSr5KcGTNb9jbFf:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/widget/form/settings/aboutform.cpp#L96)):*
> ```cpp
>     bodyUI->youAreUsing->setText(tr("You are using qTox version %1.").arg(QString(GIT_DESCRIBE)));
> 
> #if UPDATE_CHECK_ENABLED && !AUTOUPDATE_ENABLED
> ```

same

sudden6 (changes requested):
> Reviewed 1 of 23 files at r3, 2 of 21 files at r4, 4 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 12 at r4](https://reviewable.io/reviews/qtox/qtox/5346#-LTy3CLL9OCufw_JE0cf:-LU-xvHp5elLK9NsQH1M:b32wtd) ([raw file](https://github.com/qtox/qtox/blob/d287a0bd8b46760ef2579af6cbdbbcece4f74009/CMakeLists.txt#L12)):*

It works purely off tag version number, so it will consider "v1.16.3-205-gc589462b" == "v1.16.3" and say that qTox is up to date. If qTox v1.17.0 is released, it will consider "v1.16.3-anything" to be out of date. If the user is past v1.17.0 but without the tag, it will assume an update is available.
</blockquote></details>

ok
___
*[src/net/updatecheck.cpp, line 32 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyTnkg8tny1ELoTLAc:-LU-wnVF4zHm9ZXHfy-r:b69ayp1) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/net/updatecheck.cpp#L32)):*

I think static (variable) would deduplicate the variable across instantiations of the class, whereas anoynmous namespace wouldn't. This seems like a really minor benefit and this class should only be instantiated once. Changed to anonymous.
</blockquote></details>

ok, but IIRC `const` does that already at file scope
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LU-y9zn6FW6G8RxiROW:b-93u3el) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

It's true that it only performs when SettingsWidget is constructed. I do this to make sure that the update available signal is caught by AboutForm so that the notification shows properly. I'm not sure of a better way to have a delayed signal like that, I guess AboutForm could check saved update state from UpdateCheck whenever it constructs, and have UpdateCheck just cache the result?
</blockquote></details>

Isn't it possible to make the settings button green without the full settings widget being constructed? Then if the user clicks it, all the forms are constructed and they will load all the info from `UpdateCheck`?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/net/updatecheck.cpp, line 32 at r5](https://reviewable.io/reviews/qtox/qtox/5346#-LTyTnkg8tny1ELoTLAc:-LU2Ld74BvahNWE7EdG7:b-ecel17) ([raw file](https://github.com/qtox/qtox/blob/c589462b84dcdf54ceaf31959339db81c960a61f/src/net/updatecheck.cpp#L32)):*

ok, but IIRC `const` does that already at file scope
</blockquote></details>

I change my mind, I think for `static` member variables it deduplicates between classes, but `static` namespace scope variable like this just limits external linkage, the same as anonymous namespace. 
___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LUIt_HE7Z85sQC9jm8h:bcev4cs) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

Isn't it possible to make the settings button green without the full settings widget being constructed? Then if the user clicks it, all the forms are constructed and they will load all the info from `UpdateCheck`?
</blockquote></details>

(don't do a thorough review on this yet, this is just for discussion purposes). Do you mean something like this, where UpdateCheck::checkForUpdate is called before AboutForm is constructed, UpdateCheck emits the signals which are missed by AboutForm, but UpdateCheck caches the response which AboutForm can then query once constructed? If so, I feel like this is a bit messy since it basically duplicates the logic between signals/caching and slots/querying the data from UpdateCheck, plus adds new state to UpdateCheck. I'm not sure if there's a good reason to delay construction of SettingsWidget in the first place like we do, if SettingsWidget was constructed on Widget construction, UpdateCheck::checkForUpdate could be called early with all the signals connected down to AboutForm, without having to add this caching and querying of update state.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/widget.cpp, line 851 at r3](https://reviewable.io/reviews/qtox/qtox/5346#-LTqj36w4CzO3Uhlvfj8:-LVbuCeH0Cvc1gTba_wO:b-4p2eyd) ([raw file](https://github.com/qtox/qtox/blob/95e10491f2a1a740a27a2b91e701034b15f79c60/src/widget/widget.cpp#L851)):*

(don't do a thorough review on this yet, this is just for discussion purposes). Do you mean something like this, where UpdateCheck::checkForUpdate is called before AboutForm is constructed, UpdateCheck emits the signals which are missed by AboutForm, but UpdateCheck caches the response which AboutForm can then query once constructed? If so, I feel like this is a bit messy since it basically duplicates the logic between signals/caching and slots/querying the data from UpdateCheck, plus adds new state to UpdateCheck. I'm not sure if there's a good reason to delay construction of SettingsWidget in the first place like we do, if SettingsWidget was constructed on Widget construction, UpdateCheck::checkForUpdate could be called early with all the signals connected down to AboutForm, without having to add this caching and querying of update state.
</blockquote></details>

I've removed all this duplication work and just constructed SettingsWidget before the settings button is clicked, like we already do with some other widgets.

anthonybilinski:
@sudden6 FYI review comments should be finally addressed and ready to be re-reviwed whenever you have time

sudden6 (changes requested):
> Reviewed 5 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/settings/aboutform.cpp, line 33 at r7](https://reviewable.io/reviews/qtox/qtox/5346#-LViDXnF1kWHHP99ZOAo:-LViDXnGEul_qZrf2XNU:b-swrgqj) ([raw file](https://github.com/qtox/qtox/blob/f06443133d13f4cc59d88ea60a08d63e8e7b94d2/src/widget/form/settings/aboutform.cpp#L33)):*
> ```cpp
> #include "src/persistence/settings.h"
> 
> #include <tox/tox.h>
> ```

it's already included above

sudden6 (commented):
:lgtm_strong: to me besides the small include issue

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346#-:-LViEb8sApGuM4ZCsan_:b-9ys6f8)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5346)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/widget/form/settings/aboutform.cpp, line 33 at r7](https://reviewable.io/reviews/qtox/qtox/5346#-LViDXnF1kWHHP99ZOAo:-LViQFImCEyu011OZ8bx:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/f06443133d13f4cc59d88ea60a08d63e8e7b94d2/src/widget/form/settings/aboutform.cpp#L33)):*

it's already included above
</blockquote></details>

Done.

PR #5349, opened 2018-09-21
[WIP] refactor(themes): load css and images from themes folder

Chiitoo:
Original work by @tox-user (https://github.com/qTox/qTox/pull/5160).

This feature seems so handy, that I would really like for it to not be forgotten.

From the original pull request:

> The default theme is now placed in 'themes' folder in addition to being included
> as a resource. That means:
> 
> - modifying css files doesn't require recompiling the program anymore
> - users can now easily modify the existing theme
> - whenever there is a style sheet or image file missing (for example deleted by user),
> it will be loaded from the default theme that is included as a resource
> 
> This is a first of many future changes to make it possible to create themes.

Changes I have made:

- Rebase.
- Implemented the changes requested for in the original request (I think!).
- Moved the new files introduced with the search feature (did I catch them all?).
- Warning messages are different, and not final.  Still playing around with them to see what looks the best.

The path handling question seems to still be open.

Now there is at least one problem with this: it does not work.  :]  Well, it builds, and runs, but for some reason the changes I make to the theme files do not seem to stick.  It seems as if the defaults are used regardless.

I can see the warnings for when the files are not found, if I remove them, and they do go away if the files are put back.  Hopefully I just missed something simple somewhere.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5349)


Chiitoo:
Forgot to note that I did try also with only the required changes to rebase, and had the same experience, so it should not be about those (but rather me doing something wrong otherwise).

sudden6 (changes requested):
> Reviewed 95 of 96 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5349)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 100 at r1](https://reviewable.io/reviews/qtox/qtox/5349#-LN0-v5k9RnFR3mVBA8V:-LN0-v5lDgZd2h49rhno:b-tcm8b8) ([raw file](https://github.com/qtox/qtox/blob/8185e3bc8a8679806ac47ecb6bbe1420638a4e8e/CMakeLists.txt#L100)):*
> ```CMake
> # Copy themes
> if (UNIX AND NOT APPLE)
> 	file(COPY ${CMAKE_SOURCE_DIR}/themes DESTINATION $ENV{HOME}/.config/qtox)
> ```

I'm not sure if it was that before, but you seem to use tabs here and the rest of the file uses spaces

But anyway, this block of code should be moved to the `https://github.com/qTox/qTox/blob/master/cmake/Installation.cmake` file

sudden6:
Thanks for taking this up again :D

As for changes not appearing in the UI, did you check the files that you changed are actually copied to the correct location? Maybe it would make sense to pass the paths from cmake per define to the *.cpp files?

nurupo:
There was a discussion on IRC related to this PR that I'd like to reply to, but sudden6 is not on IRC, so posting it here seems appropriate.

>\<sudden6\> well, we could install the default theme to something system wide and let the user (or qTox) create the `.config/themes` directory to override themes

You don't put things like themes into `~/.config`, you should put them into `$XDG_DATA_HOME`, which is usually `$HOME/.local/share`. Please follow [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which you can easily do by using [`QStandardPaths`](http://doc.qt.io/qt-5/qstandardpaths.html), which will also conform with Windows and macOS standard paths.


>\<Chiitoo\> And yeah, I'm trying to think of how it would be packaged.  One thing that comes to mind, is if the themes actually installed into a system-wide directory, then the user could copy them over themselves if they wanted to

Your package should install default themes into `/usr/share`. User-installed themes would go into `~/.local/share`. qTox should look into both of those directories. It's also a good idea to additionally look into `/usr/local/share`. qTox can get a list of all those paths, and not just for Linux but for other platforms too, by checking the appropriate `*DataLocation` path from [`QStandardPaths`](http://doc.qt.io/qt-5/qstandardpaths.html).

nurupo:
If you want to support usb-stick-like portable mode, you might also want to check some subdirectory next to qtox executable and/or provide env variable for overriding/extending the directory list qTox will look at.

Chiitoo:
Yeah, as I was playing with this more, and especially while thinking of packaging for, say, Gentoo Linux, installing things during cmake-time into a user directory wouldn't work (as far as I can tell).  I don't have /too/ much experience with what different software does with CMake though, so I wasn't sure.

Indeed, I was thinking that this would probably go under the XDG specs.

Thanks for the comments!

I'll see if I can figure it all out.

sudden6:
@nurupo thanks, I think this is the way to go then.

Chiitoo:
Closing in favour of: https://github.com/qTox/qTox/pull/5397

PR #5354, opened 2018-09-26
Add file transfers to history

sphaerophoria:
* **Careful testing these changes**: I suggest only testing on a new profile, this is so that if db table changes are required you can just smoke the db for that profile
* db stuff is quite hairy, suggestions are welcome
    * file transfer history unfortunately can not use the same logic as the peer/sender ids since we could have a ton of file transfers compared to peers/senders so keeping that data in memory just to generate an id (which the db can do for us better) is a little silly.
    * Had to shared_ptr some things to avoid passing "this" into some callbacks that would have ended up potentially referencing a destructed history class
    * If we want to keep the asynchronous nature of inserts into the db (which I heavily suggest), it's hard to insert both into the file_transfers table and the sender_alias table and preserve row ids, there's a chance that I missed some way to work around this though
* Sorry for the whitespace changes, still getting familiar with the coding style so I turned on clang-format on save. Unfortunately I didn't notice the whitespace changes this was generating until later. If it's a PITA to review with the whitespace changes in I can invest some time into ripping them out but the changes look sane, so I figured might as well submit them.
* Db upgrades should be supported, but this does introduce a downgrade barrier, if downgrading after running this changeset you will no longer have history, not sure what qTox policy on that is (it seems like there have been no db changes for a while)

Tests run
* All file transfer states and closing
    * Sending -> loaded from history as canceled with preview
    * Receiving -> loaded from history as canceled without preview
    * Sent -> loaded from history as finished with preview
    * Received -> loaded from history as finished with preview
    * Waiting for accept (either from other chat participant or oureslves) -> loaded from history as canceled, image preview available if we are sender
* As well as
    * other user closing chat as transferring

* Extra bugs to submit
    * If one user pauses and the other unpauses the transfer widget gets into a bad state, both users spamming pause/unpause seems to eventually fix it
        * This was a problem pre and post refactor

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,6 +249,8 @@ set(${PROJECT_NAME}_SOURCES
   src/chatlog/documentcache.h
   src/chatlog/pixmapcache.cpp
   src/chatlog/pixmapcache.h
+  src/chatlog/toxfileprogress.cpp
+  src/chatlog/toxfileprogress.h
   src/chatlog/textformatter.cpp
   src/chatlog/textformatter.h
   src/core/coreav.cpp
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -41,6 +41,7 @@
 
 #include <math.h>
 
+
 // The leftButton is used to accept, pause, or resume a file transfer, as well as to open a
 // received file.
 // The rightButton is used to cancel a file transfer, or to open the directory a file was
@@ -50,7 +51,6 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
     : QWidget(parent)
     , ui(new Ui::FileTransferWidget)
     , fileInfo(file)
-    , lastTick(QTime::currentTime())
     , backgroundColor(Style::getColor(Style::LightGrey))
     , buttonColor(Style::getColor(Style::Yellow))
     , buttonBackgroundColor(Style::getColor(Style::White))
@@ -84,8 +84,6 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
         update();
     });
 
-    setBackgroundColor(Style::getColor(Style::LightGrey), false);
-
     connect(Core::getInstance(), &Core::fileTransferInfo, this,
             &FileTransferWidget::onFileTransferInfo);
     connect(Core::getInstance(), &Core::fileTransferAccepted, this,
@@ -105,16 +103,11 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
     connect(ui->previewButton, &QPushButton::clicked, this,
             &FileTransferWidget::onPreviewButtonClicked);
 
-    setupButtons();
+    // Set lastStatus to anything but the file's current value, this forces an update
+    lastStatus = file.status == ToxFile::FINISHED ? ToxFile::INITIALIZING : ToxFile::FINISHED;
+    updateWidget(file);
 
     // preview
-    if (fileInfo.direction == ToxFile::SENDING) {
-        showPreview(fileInfo.filePath);
-        ui->progressLabel->setText(tr("Waiting to send...", "file transfer widget"));
-    } else {
-        ui->progressLabel->setText(tr("Accept to receive this file", "file transfer widget"));
-    }
-
     setFixedHeight(64);
 }
 
@@ -267,154 +260,32 @@ void FileTransferWidget::paintEvent(QPaintEvent*)
 
 void FileTransferWidget::onFileTransferInfo(ToxFile file)
 {
-    QTime now = QTime::currentTime();
-    qint64 dt = lastTick.msecsTo(now); // ms
-
-    if (fileInfo != file || dt < 1000)
-        return;
-
-    fileInfo = file;
-
-    if (fileInfo.status == ToxFile::TRANSMITTING) {
-        // update progress
-        qreal progress = static_cast<qreal>(file.bytesSent) / static_cast<qreal>(file.filesize);
-        ui->progressBar->setValue(static_cast<int>(progress * 100.0));
-
-        // ETA, speed
-        qreal deltaSecs = dt / 1000.0;
-
-        // (can't use ::abs or ::max on unsigned types substraction, they'd just overflow)
-        quint64 deltaBytes = file.bytesSent > lastBytesSent ? file.bytesSent - lastBytesSent
-                                                            : lastBytesSent - file.bytesSent;
-        qreal bytesPerSec = static_cast<int>(static_cast<qreal>(deltaBytes) / deltaSecs);
-
-        // calculate mean
-        meanIndex = meanIndex % TRANSFER_ROLLING_AVG_COUNT;
-        meanData[meanIndex++] = bytesPerSec;
-
-        qreal meanBytesPerSec = 0.0;
-        for (size_t i = 0; i < TRANSFER_ROLLING_AVG_COUNT; ++i)
-            meanBytesPerSec += meanData[i];
-
-        meanBytesPerSec /= static_cast<qreal>(TRANSFER_ROLLING_AVG_COUNT);
-
-        // update UI
-        if (meanBytesPerSec > 0) {
-            // ETA
-            QTime toGo = QTime(0, 0).addSecs((file.filesize - file.bytesSent) / meanBytesPerSec);
-            QString format = toGo.hour() > 0 ? "hh:mm:ss" : "mm:ss";
-            ui->etaLabel->setText(toGo.toString(format));
-        } else {
-            ui->etaLabel->setText("");
-        }
-
-        ui->progressLabel->setText(getHumanReadableSize(meanBytesPerSec) + "/s");
-
-        lastBytesSent = file.bytesSent;
-    }
-
-    lastTick = now;
-
-    // trigger repaint
-    update();
+    updateWidget(file);
 }
 
 void FileTransferWidget::onFileTransferAccepted(ToxFile file)
 {
-    if (fileInfo != file)
-        return;
-
-    fileInfo = file;
-
-    setBackgroundColor(Style::getColor(Style::LightGrey), false);
-
-    setupButtons();
+    updateWidget(file);
 }
 
 void FileTransferWidget::onFileTransferCancelled(ToxFile file)
 {
-    if (fileInfo != file)
-        return;
-
-    fileInfo = file;
-    active = false;
-
-    setBackgroundColor(Style::getColor(Style::Red), true);
-
-    setupButtons();
-    hideWidgets();
-
-    disconnect(Core::getInstance(), nullptr, this, nullptr);
+    updateWidget(file);
 }
 
 void FileTransferWidget::onFileTransferPaused(ToxFile file)
 {
-    if (fileInfo != file)
-        return;
-
-    fileInfo = file;
-
-    ui->etaLabel->setText("");
-    ui->progressLabel->setText(tr("Paused", "file transfer widget"));
-
-    // reset mean
-    meanIndex = 0;
-    for (size_t i = 0; i < TRANSFER_ROLLING_AVG_COUNT; ++i)
-        meanData[i] = 0.0;
-
-    setBackgroundColor(Style::getColor(Style::LightGrey), false);
-
-    setupButtons();
+    updateWidget(file);
 }
 
 void FileTransferWidget::onFileTransferResumed(ToxFile file)
 {
-    if (fileInfo != file)
-        return;
-
-    fileInfo = file;
-
-    ui->etaLabel->setText("");
-    ui->progressLabel->setText(tr("Resuming...", "file transfer widget"));
-
-    // reset mean
-    meanIndex = 0;
-    for (size_t i = 0; i < TRANSFER_ROLLING_AVG_COUNT; ++i)
-        meanData[i] = 0.0;
-
-    setBackgroundColor(Style::getColor(Style::LightGrey), false);
-
-    setupButtons();
+    updateWidget(file);
 }
 
 void FileTransferWidget::onFileTransferFinished(ToxFile file)
 {
-    if (fileInfo != file)
-        return;
-
-    fileInfo = file;
-    active = false;
-
-    setBackgroundColor(Style::getColor(Style::Green), true);
-
-    setupButtons();
-    hideWidgets();
-
-    ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/yes.svg")));
-    ui->leftButton->setObjectName("ok");
-    ui->leftButton->setToolTip(tr("Open file"));
-    ui->leftButton->show();
-
-    ui->rightButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/dir.svg")));
-    ui->rightButton->setObjectName("dir");
-    ui->rightButton->setToolTip(tr("Open file directory"));
-    ui->rightButton->show();
-
-    // preview
-    if (fileInfo.direction == ToxFile::RECEIVING)
-        showPreview(fileInfo.filePath);
-
-    disconnect(Core::getInstance(), nullptr, this, nullptr);
+    updateWidget(file);
 }
 
 void FileTransferWidget::fileTransferRemotePausedUnpaused(ToxFile file, bool paused)
@@ -443,18 +314,146 @@ QString FileTransferWidget::getHumanReadableSize(qint64 size)
     return QString().setNum(size / pow(1024, exp), 'f', exp > 1 ? 2 : 0).append(suffix[exp]);
 }
 
-void FileTransferWidget::hideWidgets()
+void FileTransferWidget::updateWidgetColor(ToxFile const& file)
 {
-    ui->leftButton->hide();
-    ui->rightButton->hide();
-    ui->progressBar->hide();
-    ui->progressLabel->hide();
-    ui->etaLabel->hide();
+    if (lastStatus == file.status)
+        return;
+
+    switch (file.status) {
+    case ToxFile::INITIALIZING:
+    case ToxFile::PAUSED:
+    case ToxFile::TRANSMITTING:
+        setBackgroundColor(Style::getColor(Style::LightGrey), false);
+        break;
+    case ToxFile::BROKEN:
+    case ToxFile::CANCELED:
+        setBackgroundColor(Style::getColor(Style::Red), true);
+        break;
+    case ToxFile::FINISHED:
+        setBackgroundColor(Style::getColor(Style::Green), true);
+        break;
+    }
 }
 
-void FileTransferWidget::setupButtons()
+void FileTransferWidget::updateWidgetText(ToxFile const& file)
 {
-    switch (fileInfo.status) {
+    if (lastStatus == file.status)
+        return;
+
+    switch (file.status) {
+    case ToxFile::INITIALIZING:
+        if (file.direction == ToxFile::SENDING) {
+            ui->progressLabel->setText(tr("Waiting to send...", "file transfer widget"));
+        } else {
+            ui->progressLabel->setText(tr("Accept to receive this file", "file transfer widget"));
+        }
+        break;
+    case ToxFile::PAUSED:
+        ui->etaLabel->setText("");
+        ui->progressLabel->setText(tr("Paused", "file transfer widget"));
+        break;
+    case ToxFile::TRANSMITTING:
+        ui->etaLabel->setText("");
+        ui->progressLabel->setText(tr("Resuming...", "file transfer widget"));
+        break;
+    case ToxFile::BROKEN:
+    case ToxFile::CANCELED:
+        break;
+    case ToxFile::FINISHED:
+        break;
+    }
+}
+
+void FileTransferWidget::updatePreview(ToxFile const& file)
+{
+    if (lastStatus == file.status)
+        return;
+
+    switch (file.status) {
+        case ToxFile::INITIALIZING:
+        case ToxFile::PAUSED:
+        case ToxFile::TRANSMITTING:
+        case ToxFile::BROKEN:
+        case ToxFile::CANCELED:
+            if (file.direction == ToxFile::SENDING) {
+                showPreview(file.filePath);
+            }
+            break;
+        case ToxFile::FINISHED:
+            showPreview(file.filePath);
+            break;
+    }
+
+}
+
+void FileTransferWidget::updateFileProgress(ToxFile const& file)
+{
+    switch (file.status) {
+    case ToxFile::INITIALIZING:
+        break;
+    case ToxFile::PAUSED:
+        fileProgress.resetSpeed();
+        break;
+    case ToxFile::TRANSMITTING: {
+        if (!fileProgress.needsUpdate())
+            break;
+
+        fileProgress.addSample(file);
+        auto speed = fileProgress.getSpeed();
+        auto progress = fileProgress.getProgress();
+        auto remainingTime = fileProgress.getTimeLeftSeconds();
+
+        ui->progressBar->setValue(static_cast<int>(progress * 100.0));
+
+        // update UI
+        if (speed > 0) {
+            // ETA
+            QTime toGo = QTime(0, 0).addSecs(remainingTime);
+            QString format = toGo.hour() > 0 ? "hh:mm:ss" : "mm:ss";
+            ui->etaLabel->setText(toGo.toString(format));
+        } else {
+            ui->etaLabel->setText("");
+        }
+
+        ui->progressLabel->setText(getHumanReadableSize(speed) + "/s");
+        break;
+    }
+    case ToxFile::BROKEN:
+    case ToxFile::CANCELED:
+    case ToxFile::FINISHED: {
+        ui->progressBar->hide();
+        ui->progressLabel->hide();
+        ui->etaLabel->hide();
+        break;
+    }
+    }
+}
+
+void FileTransferWidget::updateSignals(ToxFile const& file)
+{
+    if (lastStatus == file.status)
+        return;
+
+    switch (file.status) {
+    case ToxFile::CANCELED:
+    case ToxFile::BROKEN:
+    case ToxFile::FINISHED:
+        active = false;
+        disconnect(Core::getInstance(), nullptr, this, nullptr);
+        break;
+    case ToxFile::INITIALIZING:
+    case ToxFile::PAUSED:
+    case ToxFile::TRANSMITTING:
+        break;
+    }
+}
+
+void FileTransferWidget::setupButtons(ToxFile const& file)
+{
+    if (lastStatus == file.status)
+        return;
+
+    switch (file.status) {
     case ToxFile::TRANSMITTING:
         ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/pause.svg")));
         ui->leftButton->setObjectName("pause");
@@ -479,13 +478,12 @@ void FileTransferWidget::setupButtons()
         setButtonColor(Style::getColor(Style::LightGrey));
         break;
 
-    case ToxFile::STOPPED:
-    case ToxFile::BROKEN:
+    case ToxFile::INITIALIZING:
         ui->rightButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/no.svg")));
         ui->rightButton->setObjectName("cancel");
         ui->rightButton->setToolTip(tr("Cancel transfer"));
 
-        if (fileInfo.direction == ToxFile::SENDING) {
+        if (file.direction == ToxFile::SENDING) {
             ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/pause.svg")));
             ui->leftButton->setObjectName("pause");
             ui->leftButton->setToolTip(tr("Pause transfer"));
@@ -494,6 +492,23 @@ void FileTransferWidget::setupButtons()
             ui->leftButton->setObjectName("accept");
             ui->leftButton->setToolTip(tr("Accept transfer"));
         }
+        break;
+    case ToxFile::CANCELED:
+    case ToxFile::BROKEN:
+        ui->leftButton->hide();
+        ui->rightButton->hide();
+        break;
+    case ToxFile::FINISHED:
+        ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/yes.svg")));
+        ui->leftButton->setObjectName("ok");
+        ui->leftButton->setToolTip(tr("Open file"));
+        ui->leftButton->show();
+
+        ui->rightButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/dir.svg")));
+        ui->rightButton->setObjectName("dir");
+        ui->rightButton->setToolTip(tr("Open file directory"));
+        ui->rightButton->show();
+
         break;
     }
 }
@@ -549,7 +564,8 @@ void FileTransferWidget::showPreview(const QString& filename)
         }
         const QByteArray imageFileData = imageFile.readAll();
         QImage image = QImage::fromData(imageFileData);
-        const int exifOrientation = getExifOrientation(imageFileData.constData(), imageFileData.size());
+        const int exifOrientation =
+            getExifOrientation(imageFileData.constData(), imageFileData.size());
         if (exifOrientation) {
             applyTransformation(exifOrientation, image);
         }
@@ -562,13 +578,12 @@ void FileTransferWidget::showPreview(const QString& filename)
         // Show mouseover preview, but make sure it's not larger than 50% of the screen width/height
         const QRect desktopSize = QApplication::desktop()->screenGeometry();
         const int maxPreviewWidth{desktopSize.width() / 2};
-        const int maxPreviewHeight{desktopSize.height() /2};
+        const int maxPreviewHeight{desktopSize.height() / 2};
         const QImage previewImage = [&image, maxPreviewWidth, maxPreviewHeight]() {
             if (image.width() > maxPreviewWidth || image.height() > maxPreviewHeight) {
-                return image.scaled(maxPreviewWidth, maxPreviewHeight,
-                                                     Qt::KeepAspectRatio, Qt::SmoothTransformation);
-            }
-            else {
+                return image.scaled(maxPreviewWidth, maxPreviewHeight, Qt::KeepAspectRatio,
+                                    Qt::SmoothTransformation);
+            } else {
                 return image;
             }
         }();
@@ -578,8 +593,7 @@ void FileTransferWidget::showPreview(const QString& filename)
         buffer.open(QIODevice::WriteOnly);
         previewImage.save(&buffer, "PNG");
         buffer.close();
-        ui->previewButton->setToolTip("<img src=data:image/png;base64," + imageData.toBase64()
-                                      + "/>");
+        ui->previewButton->setToolTip("<img src=data:image/png;base64," + imageData.toBase64() + "/>");
     }
 }
 
@@ -641,12 +655,11 @@ int FileTransferWidget::getExifOrientation(const char* data, const int size)
 void FileTransferWidget::applyTransformation(const int orientation, QImage& image)
 {
     QTransform exifTransform;
-    switch(static_cast<ExifOrientation>(orientation))
-    {
+    switch (static_cast<ExifOrientation>(orientation)) {
     case ExifOrientation::TopLeft:
         break;
     case ExifOrientation::TopRight:
-        image = image.mirrored(1,0);
+        image = image.mirrored(1, 0);
         break;
     case ExifOrientation::BottomRight:
         exifTransform.rotate(180);
@@ -672,4 +685,34 @@ void FileTransferWidget::applyTransformation(const int orientation, QImage& imag
         qWarning() << "Invalid exif orientation passed to applyTransformation!";
     }
     image = image.transformed(exifTransform);
+}
+
+void FileTransferWidget::updateWidget(ToxFile const& file)
+{
+    if (fileInfo != file)
+        return;
+
+    fileInfo = file;
+
+    // If we repainted on every packet our gui would be *very* slow
+    bool bTransmitNeedsUpdate = fileProgress.needsUpdate();
+
+    updatePreview(file);
+    updateFileProgress(file);
+    updateWidgetText(file);
+    updateWidgetColor(file);
+    setupButtons(file);
+    updateSignals(file);
+
+    lastStatus = file.status;
+
+    // trigger repaint
+    switch (file.status) {
+    case ToxFile::TRANSMITTING:
+        if (!bTransmitNeedsUpdate)
+            break;
+        // fallthrough
+    default:
+        update();
+    }
 }
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -39,6 +39,7 @@
 #include <QPainter>
 #include <QVariantAnimation>
 
+#include <cassert>
 #include <math.h>
 
 
@@ -148,10 +149,11 @@ void FileTransferWidget::autoAcceptTransfer(const QString& path)
 
     // Do not automatically accept the file-transfer if the path is not writable.
     // The user can still accept it manually.
-    if (tryRemoveFile(filepath))
+    if (tryRemoveFile(filepath)) {
         Core::getInstance()->acceptFileRecvRequest(fileInfo.friendId, fileInfo.fileNum, filepath);
-    else
+    } else {
         qWarning() << "Cannot write to " << filepath;
+    }
 }
 
 bool FileTransferWidget::isActive() const
@@ -161,8 +163,9 @@ bool FileTransferWidget::isActive() const
 
 void FileTransferWidget::acceptTransfer(const QString& filepath)
 {
-    if (filepath.isEmpty())
+    if (filepath.isEmpty()) {
         return;
+    }
 
     // test if writable
     if (!tryRemoveFile(filepath)) {
@@ -290,17 +293,19 @@ void FileTransferWidget::onFileTransferFinished(ToxFile file)
 
 void FileTransferWidget::fileTransferRemotePausedUnpaused(ToxFile file, bool paused)
 {
-    if (paused)
+    if (paused) {
         onFileTransferPaused(file);
-    else
+    } else {
         onFileTransferResumed(file);
+    }
 }
 
 void FileTransferWidget::fileTransferBrokenUnbroken(ToxFile file, bool broken)
 {
     // TODO: Handle broken transfer differently once we have resuming code
-    if (broken)
+    if (broken) {
         onFileTransferCancelled(file);
+    }
 }
 
 QString FileTransferWidget::getHumanReadableSize(qint64 size)
@@ -308,16 +313,18 @@ QString FileTransferWidget::getHumanReadableSize(qint64 size)
     static const char* suffix[] = {"B", "kiB", "MiB", "GiB", "TiB"};
     int exp = 0;
 
-    if (size > 0)
+    if (size > 0) {
         exp = std::min((int)(log(size) / log(1024)), (int)(sizeof(suffix) / sizeof(suffix[0]) - 1));
+    }
 
     return QString().setNum(size / pow(1024, exp), 'f', exp > 1 ? 2 : 0).append(suffix[exp]);
 }
 
 void FileTransferWidget::updateWidgetColor(ToxFile const& file)
 {
-    if (lastStatus == file.status)
+    if (lastStatus == file.status) {
         return;
+    }
 
     switch (file.status) {
     case ToxFile::INITIALIZING:
@@ -332,13 +339,16 @@ void FileTransferWidget::updateWidgetColor(ToxFile const& file)
     case ToxFile::FINISHED:
         setBackgroundColor(Style::getColor(Style::Green), true);
         break;
+    default:
+        assert(false);
     }
 }
 
 void FileTransferWidget::updateWidgetText(ToxFile const& file)
 {
-    if (lastStatus == file.status)
+    if (lastStatus == file.status) {
         return;
+    }
 
     switch (file.status) {
     case ToxFile::INITIALIZING:
@@ -361,29 +371,33 @@ void FileTransferWidget::updateWidgetText(ToxFile const& file)
         break;
     case ToxFile::FINISHED:
         break;
+    default:
+        assert(false);
     }
 }
 
 void FileTransferWidget::updatePreview(ToxFile const& file)
 {
-    if (lastStatus == file.status)
+    if (lastStatus == file.status) {
         return;
-
-    switch (file.status) {
-        case ToxFile::INITIALIZING:
-        case ToxFile::PAUSED:
-        case ToxFile::TRANSMITTING:
-        case ToxFile::BROKEN:
-        case ToxFile::CANCELED:
-            if (file.direction == ToxFile::SENDING) {
-                showPreview(file.filePath);
-            }
-            break;
-        case ToxFile::FINISHED:
-            showPreview(file.filePath);
-            break;
     }
 
+    switch (file.status) {
+    case ToxFile::INITIALIZING:
+    case ToxFile::PAUSED:
+    case ToxFile::TRANSMITTING:
+    case ToxFile::BROKEN:
+    case ToxFile::CANCELED:
+        if (file.direction == ToxFile::SENDING) {
+            showPreview(file.filePath);
+        }
+        break;
+    case ToxFile::FINISHED:
+        showPreview(file.filePath);
+        break;
+    default:
+        assert(false);
+    }
 }
 
 void FileTransferWidget::updateFileProgress(ToxFile const& file)
@@ -395,8 +409,9 @@ void FileTransferWidget::updateFileProgress(ToxFile const& file)
         fileProgress.resetSpeed();
         break;
     case ToxFile::TRANSMITTING: {
-        if (!fileProgress.needsUpdate())
+        if (!fileProgress.needsUpdate()) {
             break;
+        }
 
         fileProgress.addSample(file);
         auto speed = fileProgress.getSpeed();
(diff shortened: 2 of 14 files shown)


sudden6 (changes requested):
Mostly cosmetic issues, will do some testing later. Very good work! Also for formatting you can use `tools/format-code.sh`, it applies most of our coding style, but unfortunately misses the enforcment of curly brackets.

> Reviewed 2 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LNKv7e90UI4_XjWrol_:b-wqdc4l)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.h, line 1 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKuVtMEiNurwONY0rl:-LNKuVtN9yH8o1GA-2sD:b-ckof31) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.h#L1)):*
> ```objc
> #ifndef TOXFILEPROGRESS_H
> ```

Please add the qTox License and Copyright header
___
*[src/chatlog/toxfileprogress.cpp, line 1 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKuZPU3TAaNiOIPXIt:-LNKuZPU3TAaNiOIPXIu:b-gtzv9c) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.cpp#L1)):*
> ```cpp
> #include "toxfileprogress.h"
> ```

missing header too
___
*[src/chatlog/toxfileprogress.cpp, line 10 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKpr_m3Dqf0Oz3g3UG:-LNKpr_m3Dqf0Oz3g3UH:b68wasw) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.cpp#L10)):*
> ```cpp
>     qint64 dt = lastTick.msecsTo(now); // ms
> 
>     if (dt < 1000)
> ```

our coding style mandates curly braces everywhere
___
*[src/core/toxfile.h, line 13 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKr2rn53hjW-tizRBN:-LNKr2rn53hjW-tizRBO:b-qqdaxm) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/core/toxfile.h#L13)):*
> ```objc
> {
>     // Note do not change order, these are directly inserted into the DB in their
>     // current form, changing order would mess up database state!
> ```

order is not important, the values are, right?
___
*[src/core/toxfile.h, line 25 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKrA_a4DeTbTt3I6tC:-LNKrA_b421mCz86bjgt:b-7jl26k) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/core/toxfile.h#L25)):*
> ```objc
> 
>     // Note do not change order, these are directly inserted into the DB in their
>     // current form (can add fields though as db representation is an int)
> ```

same?
___
*[src/persistence/history.cpp, line 256 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKsdvd50vTNCAGuKM3:-LNKsdvd50vTNCAGuKM4:b-dpkgbu) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/persistence/history.cpp#L256)):*
> ```cpp
> {
>     auto file_state = success ? ToxFile::FINISHED : ToxFile::CANCELED;
>     if (filePath.length()) {
> ```

can a filetransfer without a path even happen?
___
*[src/persistence/history.cpp, line 645 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKtM-r3IUxJvcVjk2E:-LNKtM-r3IUxJvcVjk2F:bc0zw5m) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/persistence/history.cpp#L645)):*
> ```cpp
> }
> 
> void History::dbSchemaUpgrade()
> ```

Very good idea to handle the database upgrades this way, could you add maybe some doxygen style documentation that explains how to extend it for future schema upgrades?

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 256 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKsdvd50vTNCAGuKM3:-LNLQaxW9vzt94qRw7dJ:b-tfqbdq) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/persistence/history.cpp#L256)):*

can a filetransfer without a path even happen?
</blockquote></details>

Yeah, when we receive a file transfer request from a peer we add to history (as canceled) before we've accepted.  We then populate the filepath when we mark it as finished. This workflow will have to change if we ever support resumable file transfers.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 645 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKtM-r3IUxJvcVjk2E:-LNLTh0FE4rxHq7YVX2L:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/persistence/history.cpp#L645)):*

Very good idea to handle the database upgrades this way, could you add maybe some doxygen style documentation that explains how to extend it for future schema upgrades?
</blockquote></details>

Done.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.h, line 1 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKuVtMEiNurwONY0rl:-LNLTm895mbJsC_u4PoX:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.h#L1)):*

Please add the qTox License and Copyright header
</blockquote></details>

Done.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.cpp, line 10 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKpr_m3Dqf0Oz3g3UG:-LNLTn_66Z5e1Mgnxzpn:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.cpp#L10)):*

our coding style mandates curly braces everywhere
</blockquote></details>

Done.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.cpp, line 1 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKuZPU3TAaNiOIPXIt:-LNLTofpEApcgFkjbGTs:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/chatlog/toxfileprogress.cpp#L1)):*

missing header too
</blockquote></details>

Done.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 13 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKr2rn53hjW-tizRBN:-LNLTqQ4EY-RhLqWdT-u:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/core/toxfile.h#L13)):*

order is not important, the values are, right?
</blockquote></details>

Done.
___
*[src/core/toxfile.h, line 25 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKrA_a4DeTbTt3I6tC:-LNLTqm67Hfh0DOKQV0t:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/core/toxfile.h#L25)):*

same?
</blockquote></details>

Done.

sphaerophoria:
Looks like CI failed due to a network error, is it possible to kick it off again?

sudden6 (changes requested):
> Reviewed 6 of 12 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.cpp, line 58 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMQ_J4-HCzpaC6O-cG:-LNMQ_J4-HCzpaC6O-cH:bij8mwt) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/toxfileprogress.cpp#L58)):*
> ```cpp
> 
>     double meanBytesPerSec = 0.0;
>     for (size_t i = 0; i < TRANSFER_ROLLING_AVG_COUNT; ++i)
> ```

missing curly brackets
___
*[src/chatlog/content/filetransferwidget.cpp, line 679 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRB4FEz4dqkQiTiT0:-LNMRB4FEz4dqkQiTiT1:b-zf7034) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/content/filetransferwidget.cpp#L679)):*
> ```cpp
> void FileTransferWidget::updateWidget(ToxFile const& file)
> {
>     if (fileInfo != file)
> ```

brackets
___
*[src/chatlog/content/filetransferwidget.cpp, line 699 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRD6DEKasmYEgO-dW:-LNMRD6DEKasmYEgO-dX:b-zf7034) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/content/filetransferwidget.cpp#L699)):*
> ```cpp
>     switch (file.status) {
>     case ToxFile::TRANSMITTING:
>         if (!bTransmitNeedsUpdate)
> ```

brackets
___
*[src/persistence/history.cpp, line 256 at r1](https://reviewable.io/reviews/qtox/qtox/5354#-LNKsdvd50vTNCAGuKM3:-LNMRaYT1kGUulox1S7q:b-2o9dou) ([raw file](https://github.com/qtox/qtox/blob/de2c301479142ed15223b334c1005c1f6d9c3f0b/src/persistence/history.cpp#L256)):*

Yeah, when we receive a file transfer request from a peer we add to history (as canceled) before we've accepted.  We then populate the filepath when we mark it as finished. This workflow will have to change if we ever support resumable file transfers.
</blockquote></details>

ok, then
___
*[src/persistence/history.cpp, line 39 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRo8y-Dq21Hg04MxE:-LNMRo8y-Dq21Hg04MxF:b7glm7g) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/persistence/history.cpp#L39)):*
> ```cpp
> static constexpr int NUM_MESSAGES_DEFAULT =
>     100; // arbitrary number of messages loaded when not loading by date
> static constexpr int SCHEMA_VERSION = 1;
> ```

could we maybe namespace this variable?

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 39 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRo8y-Dq21Hg04MxE:-LNMiYNu2PVTlyZAA_ER:bnntx2c) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/persistence/history.cpp#L39)):*

could we maybe namespace this variable?
</blockquote></details>

hm, i didnt dee much use of namespaces when i was browsing the rest of the codebase. What were you thinking? A namespace just for the variable? For what purpose? It wont leak anywhere else since its static constexpr in the translation unit.

sudden6 (commented):
> Reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 39 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRo8y-Dq21Hg04MxE:-LNMn-D7AN816tOJ0BGA:b-croy3v) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/persistence/history.cpp#L39)):*

hm, i didnt dee much use of namespaces when i was browsing the rest of the codebase. What were you thinking? A namespace just for the variable? For what purpose? It wont leak anywhere else since its static constexpr in the translation unit.
</blockquote></details>

ok, true, doesn't make to much sense in this instance.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/toxfileprogress.cpp, line 58 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMQ_J4-HCzpaC6O-cG:-LNN2A9T8vQwWXMIp15F:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/toxfileprogress.cpp#L58)):*

missing curly brackets
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 679 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRB4FEz4dqkQiTiT0:-LNN2CeuCDtqTmwCyLiV:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/content/filetransferwidget.cpp#L679)):*

brackets
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 699 at r2](https://reviewable.io/reviews/qtox/qtox/5354#-LNMRD6DEKasmYEgO-dW:-LNN2DhUCxYXZQsuMRSG:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b5b6830766b132702ca08c047eb226af7b8a138b/src/chatlog/content/filetransferwidget.cpp#L699)):*

brackets
</blockquote></details>

Done.

mschilli87:
@sphaerophoria 
> Looks like CI failed due to a network error, is it possible to kick it off again?

You can close and re-open the PR to restart CI.

sudden6:
@mschilli87 IIRC PRs can not be reopened, but you can instead make a meaningless change and back and force push to the branch

sudden6 (commented):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained

mschilli87:
@sudden6
> IRC PRs can not be reopened, but you can instead make a meaningless change and back and force push to the branch

Could be a setting or a change in github but [I have done this](https://github.com/BIMSBbioinfo/ciRcus/pull/40#event-1077813210). :wink:

sudden6:
@mschilli87 Ok, maybe I remember wrong :) I would still prefer a force push to re-trigger builds, because else it causes confusion why PRs are closed and reopened all the time and I get spammed with notifications^^

sudden6 (approved):
:lgtm_strong: tested it and works as described :) waiting for a review from @anthonybilinski

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LNWEr9UD-EAEHqtgj_v:br31b5f)* status: :shipit: complete! 1 of 1 LGTMs obtained

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/persistence/history.cpp, line 317 at r3](https://reviewable.io/reviews/qtox/qtox/5354#-LNXukMo1SvJCG8iNTSl:-LNXukMo1SvJCG8iNTSm:b-wlhjql) ([raw file](https://github.com/qtox/qtox/blob/adac2e8b132abdd3be4071dd1242e20c232dfb92/src/persistence/history.cpp#L317)):*
> ```cpp
>                                {fileId.toUtf8(), filePath.toUtf8(), fileName},
>                                [fileInfosPtr, fileId, dbPtr](int64_t id) {
>                                    auto& fileInfo = (*fileInfosPtr)[fileId];
> ```

Whoops, found a multi-threading bug here. Somehow I didn't register that the database calls come in from another thread. In the case where we have a slow disk and a fast file transfer this path can be triggered at the same time as setFileFinished. This causes finished file transfers to load as failed on next boot.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/persistence/history.cpp, line 317 at r3](https://reviewable.io/reviews/qtox/qtox/5354#-LNXukMo1SvJCG8iNTSl:-LNZflewBtoYUhXq6RWI:biti1u6) ([raw file](https://github.com/qtox/qtox/blob/adac2e8b132abdd3be4071dd1242e20c232dfb92/src/persistence/history.cpp#L317)):*

Whoops, found a multi-threading bug here. Somehow I didn't register that the database calls come in from another thread. In the case where we have a slow disk and a fast file transfer this path can be triggered at the same time as setFileFinished. This causes finished file transfers to load as failed on next boot.
</blockquote></details>

This should be dealt with now, substantially changed the history.cpp file to use signals/slots instead of callbacks from the db thread. This should hopefully ensure no threading issues. I had to pass this into the callbacks in the db thread to use their signals though, so I changed History to be a shared_ptr instead of unique and used shared_from_this to get weak pointers to avoid circular dependencies and still ensure we don't lose scope.

zetok (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/persistence/history.cpp, line 452 at r4](https://reviewable.io/reviews/qtox/qtox/5354#-LNg3H_3CNNLNGOEYGsS:-LNg3H_3CNNLNGOEYGsT:b-cyvitt) ([raw file](https://github.com/qtox/qtox/blob/39dc6dacb87c50060a4d19bac05e615f457b11a3/src/persistence/history.cpp#L452)):*
> ```cpp
>     auto rowCallback = [&messages](const QVector<QVariant>& row) {
>         // dispName and message could have null bytes, QString::fromUtf8
>         // truncates on null bytes so we strip them
> ```

Why is this comment removed? Its absence could lead to a situation where future refactoring would remove the null stripping, which then could result in a situation where nicks ending with null or their associated messages would be mishandled. Generally, you want more comments, not less, especially in the cases like this one where it's not clear why something is in place.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/persistence/history.cpp, line 452 at r4](https://reviewable.io/reviews/qtox/qtox/5354#-LNg3H_3CNNLNGOEYGsS:-LNgQRsy5yxHkZvus-kO:bsdjc7) ([raw file](https://github.com/qtox/qtox/blob/39dc6dacb87c50060a4d19bac05e615f457b11a3/src/persistence/history.cpp#L452)):*
<details><summary><i>Previously, zetok wrote…</i></summary><blockquote>

Why is this comment removed? Its absence could lead to a situation where future refactoring would remove the null stripping, which then could result in a situation where nicks ending with null or their associated messages would be mishandled. Generally, you want more comments, not less, especially in the cases like this one where it's not clear why something is in place.
</blockquote></details>

That was a mistake. Thanks for the comment :)

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/persistence/history.cpp, line 452 at r4](https://reviewable.io/reviews/qtox/qtox/5354#-LNg3H_3CNNLNGOEYGsS:-LNgUw9G8VMotx6zCyES:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/39dc6dacb87c50060a4d19bac05e615f457b11a3/src/persistence/history.cpp#L452)):*

That was a mistake. Thanks for the comment :)
</blockquote></details>

Done.

zetok (approved):

sudden6:
`src/chatlog/content/filetransferwidget.cpp` is missing `#include <cassert>` so the CI fails

sphaerophoria:
Huh, weird that it compiled with mingw on my machine but not ubuntu's GCC. I would have expected this to fail on my compiler too.

jakieoymj-github:
I like the feature, how to try it? clone from sphaerophoria/qTox:file_history_pr?

sudden6:
@jakieoymj-github you can use the `test-pr.sh` script in the root of the repository for this: `./test-pr.sh 5354` and then compile as usual.

anthonybilinski:
I've added the file hash to the db since we're planning on using that in the near future and it saves on level of db schema update. I've got those changes on my fork at the same branch, not sure how to handle the merge. Maybe I'll PR @sphaerophoria's fork then he can update his PR to qTox.

Something I don't understand about the original implementation is:
```
RawDatabase::Query History::generateFileFinished(int64_t id, bool success, const QString& filePath)
{
    auto file_state = success ? ToxFile::FINISHED : ToxFile::CANCELED;
    if (filePath.length()) {
        return RawDatabase::Query(QStringLiteral("UPDATE file_transfers "
                                                 "SET file_state = %1, file_path = ?"
                                                 "WHERE id = %2")
                                      .arg(file_state)
                                      .arg(id),
                                  {filePath.toUtf8()});
    } else {
        return RawDatabase::Query(QStringLiteral("UPDATE file_transfers "
                                                 "SET finished = %1 "
                                                 "WHERE id = %2")
                                      .arg(file_state)
                                      .arg(id));
    }
}
```
how the file could ever not have a name by this point. In that case it seems like the path will never be set in the db, as well as the has in my impl. I noticed it's possible to send a Tox file without a name (though not from qTox) and even though our accept dialog forces a name, our auto-accept doesn't, but even in that case I end up with a file named "/tmp/ (1).". I feel like we should force a name on files, and find a way to remove this branch so that the file history never has completed entries without paths.

anthonybilinski:
File hash improved (data added to hash as sent/received, reducing memory usage and eliminating reading file back off disk), tested sending/receiving. The file finished without a length is from cancelled transfers that haven't started, so not having a hash in those cases is fine and not a bug.

I've pushed my change onto @sphaerophoria's branch (muhaha) so it's now in this PR ready to be reviewed.

sudden6:
changes from @anthonybilinski lgtm

waiting for rebase by @sphaerophoria

sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 4 files at r4, 3 of 6 files at r6, 5 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LSZhY5S0defF0Q-t1jY:b-jbsf8z)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

sphaerophoria:
Alright cool I'll free up some time this weekend to move it up and review @anthonybilinski's changes. I believe this will free up #5366 as well.

sudden6:
@sphaerophoria Did you finish the rebase? Just asking, because you didn't leave a message, but I see a force-push.

sphaerophoria:
Ah right sorry, I forgot to leave a comment. It should be rebased but I didn't get a chance to double check functionality. Reviewed @anthonybilinski's changes and they look reasonable to me.

anthonybilinski:
I've done some testing on the merge:
* db state looks correct for cancelled/received/sent
* hash saved correctly on sent/received, blank on cancelled as expected
* preview looks good, file open looks good.
I noticed for files that have been removed:
```
[00:24:33.015 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.017 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.019 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.022 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.025 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.027 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
[00:24:33.030 UTC] chatlog/content/filetransferwidget.cpp:586 : Critical: Failed to open file for preview
```
this should definitely be lower than critical (maybe no log at all to avoid spam?). Priority is probably left over from when this was only for just received files which expect to always be present.

anthonybilinski (changes requested):
> Reviewed 1 of 12 files at r1, 1 of 4 files at r2, 1 of 6 files at r6, 6 of 11 files at r8, 5 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/chatlog/content/filetransferwidget.cpp, line 111 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqxiF066nFIxH-Ne2y:-LSqxiF066nFIxH-Ne2z:b-9pb0m4) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L111)):*
> ```cpp
>     updateWidget(file);
> 
>     // preview
> ```

comment no longer relevant?
___
*[src/chatlog/content/filetransferwidget.cpp, line 343 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqy08WC1X0lVK9hvRe:-LSqy08WC1X0lVK9hvRf:b-n5dzyr) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L343)):*
> ```cpp
>         break;
>     default:
>         assert(false);
> ```

error log as well for release builds?
___
*[src/chatlog/content/filetransferwidget.cpp, line 399 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyHUj7LlBv9HauYEE:-LSqyHUj7LlBv9HauYEF:b-vxjl2) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L399)):*
> ```cpp
>         break;
>     default:
>         assert(false);
> ```

same, log?
___
*[src/chatlog/content/filetransferwidget.cpp, line 445 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyKV217ccO7RsnMFn:-LSqyKV217ccO7RsnMFo:b-2bimrf) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L445)):*
> ```cpp
>     }
>     default:
>         assert(false);
> ```

same, log
___
*[src/chatlog/content/filetransferwidget.cpp, line 467 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyLSS2RK_De3J8o4R:-LSqyLSS2RK_De3J8o4S:b-2bimrf) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L467)):*
> ```cpp
>         break;
>     default:
>         assert(false);
> ```

same, log
___
*[src/chatlog/content/filetransferwidget.cpp, line 746 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqz2RY7iAXZGcNyIYO:-LSqz2RY7iAXZGcNyIYP:b-q43h88) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L746)):*
> ```cpp
>     // fallthrough
>     default:
>         update();
> ```

why do we need to update if not transmitting and the status hasn't changed?
___
*[src/persistence/history.h, line 66 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqzyYF-YR-7YqDEIvK:-LSqzyYG3OTpw1IFeD8D:b-22s8h9) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.h#L66)):*
> ```objc
>     QString& asMessage()
>     {
>         assert(type == HistMessageContentType::message);
> ```

like asserts above, we should probably do something here on release, even if just log error and abort
___
*[src/persistence/history.cpp, line 652 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr0qlL0kQr4UUyFDpJ:-LSr0qlL0kQr4UUyFDpK:b-9uf535) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L652)):*
> ```
> for some reason
> ```
can you expand on this? Testing quickly myself, it seems to work fine as just `file.filePath = row[8].toByteArray()`
Reading docs it seems like the only difference from converting to utf8 then back to bytes is that the conversion to utf8 would strip/replace invalid characters, but it seems like a problem if we have invalid characters in our file path column in the first place.
___
*[src/persistence/history.cpp, line 706 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr1Tga2L7dSKF8c9n3:-LSr1Tga2L7dSKF8c9n4:bhvcyn8) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L706)):*
> ```cpp
>         qWarning() << "Database version is newer than we currently support. Please upgrade qTox";
>         // We don't know what future versions have done, we have to disable db access until we re-upgrade
>         db.reset();
> ```

won't this immediately segfault in the default case of the next switch? if so, IMO logging critical and abort seems better.
___
*[src/widget/form/chatform.cpp, line 916 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr25jB0nSYh0yT_05w:-LSr25jB0nSYh0yT_05x:b-auucv) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/widget/form/chatform.cpp#L916)):*
> ```cpp
>     }
>     default:
>         assert(false);
> ```

release case?
___
*[src/widget/form/chatform.cpp, line 1209 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr2IA7EYYrO41uZF64:-LSr2IA7EYYrO41uZF65:b-ei4msa) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/widget/form/chatform.cpp#L1209)):*
> ```cpp
>         QString author = getMsgAuthorDispName(authorPk, it.dispName);
> 
>         if (it.content.getType() == HistMessageContentType::message) {
> ```

isn't this already guaranteed to == HistMessageContentType::message by the check at the start of the for loop?

anthonybilinski:
Creating a new profile, I see:
```
[19:28:10.892 UTC] core/toxoptions.cpp:91 : Debug: Core starting with IPv6 enabled                                                                                                                                 
[19:28:10.892 UTC] persistence/settings.cpp:463 : Debug: "Saving global settings at /home/abilinski/.config/tox/qtox.ini"                                                                                          
[19:28:11.010 UTC] persistence/history.cpp:723 : Debug: Database upgrade finished (databaseSchemaVersion  0  ->  1 )                                                                                               
[19:28:11.016 UTC] persistence/settings.cpp:338 : Debug: Loading personal settings from "/home/abilinski/.config/tox/testnewnew.ini"
[19:28:11.025 UTC] nexus.cpp:83 : Debug: Starting up
[19:28:11.027 UTC] persistence/db/rawdatabase.cpp:574 : Warning: Failed to prepare statement "ALTER TABLE history ADD file_id INTEGER;" with error 1
```
I guess it's accidentally trying to upgrade a newly created db that doesn't need updating.

sphaerophoria (commented):
> this should definitely be lower than critical (maybe no log at all to avoid spam?). Priority is probably left over from when this was only for just received files which expect to always be present.

I've removed the log, you're right it's a reasonable case now

> I guess it's accidentally trying to upgrade a newly created db that doesn't need updating.

I think I did this intentionally, looking up if the db existed is too hard without blocking so I figured I'd eat the warning once on new profile creation. I've added a comment in code to indicate this is the case.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LT0wGJi30Kbesn0ViXG:b-tfbctc)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 111 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqxiF066nFIxH-Ne2y:-LSwobLF4JYdV8W55ogm:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L111)):*

comment no longer relevant?
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 343 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqy08WC1X0lVK9hvRe:-LT0pXl83Ac8Oy8ddiad:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L343)):*

error log as well for release builds?
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 399 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyHUj7LlBv9HauYEE:-LT0pY3C62Sh0wDnLMn3:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L399)):*

same, log?
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 445 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyKV217ccO7RsnMFn:-LT0pYY_7IdDxLwb9ce7:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L445)):*

same, log
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 467 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqyLSS2RK_De3J8o4R:-LT0pYugBhrrLcNn9Wk-:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L467)):*

same, log
</blockquote></details>

Done.
___
*[src/chatlog/content/filetransferwidget.cpp, line 746 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqz2RY7iAXZGcNyIYO:-LSwqF8w64W3iTP7n0w8:b646br2) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L746)):*
> and the status hasn't changed

Where are you seeing that? The status may have changed in this case.
___
*[src/persistence/history.h, line 66 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqzyYF-YR-7YqDEIvK:-LT0p_2J00NgmeX7H6S8:b-zguv7n) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.h#L66)):*

like asserts above, we should probably do something here on release, even if just log error and abort
</blockquote></details>

I'm not really sure on this one. This is a more low level class and there isn't really reasonable behavior for it to do in the wrong case. These asserts are here more to prevent debug logic failures. Since all callers are supposed to get the type of the stored object first it seems reasonable that a logic bug would quickly show itself in debug.
___
*[src/persistence/history.cpp, line 652 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr0qlL0kQr4UUyFDpJ:-LT0oiYXBpn6VrX1x4ng:b-vdl2f0) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L652)):*

> ```
> for some reason
> ```
can you expand on this? Testing quickly myself, it seems to work fine as just `file.filePath = row[8].toByteArray()`
Reading docs it seems like the only difference from converting to utf8 then back to bytes is that the conversion to utf8 would strip/replace invalid characters, but it seems like a problem if we have invalid characters in our file path column in the first place.
</blockquote></details>

If I remember correctly on windows I was getting corrupted filenames on the output. It could be that the utf16 file paths on windows are causing some extra layer of WTF
___
*[src/persistence/history.cpp, line 706 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr1Tga2L7dSKF8c9n3:-LT0ouck6RMHSJo4gOTg:bgrc1mo) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L706)):*

won't this immediately segfault in the default case of the next switch? if so, IMO logging critical and abort seems better.
</blockquote></details>

Yeah that's a good point, we should return here so that the above layer can exit early and act as if we have no history
___
*[src/widget/form/chatform.cpp, line 916 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr25jB0nSYh0yT_05w:-LT0po2lD5Ttt_0d40j2:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/widget/form/chatform.cpp#L916)):*

release case?
</blockquote></details>

Done.
___
*[src/widget/form/chatform.cpp, line 1209 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr2IA7EYYrO41uZF64:-LT0vqgFBjIYCn8aGWo5:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/widget/form/chatform.cpp#L1209)):*

isn't this already guaranteed to == HistMessageContentType::message by the check at the start of the for loop?
</blockquote></details>

Done.

anthonybilinski (changes requested):
> Reviewed 3 of 3 files at r10.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 746 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSqz2RY7iAXZGcNyIYO:-LT1Ilr66uuV_9KQHj6m:b-uhfd88) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/chatlog/content/filetransferwidget.cpp#L746)):*

> and the status hasn't changed

Where are you seeing that? The status may have changed in this case.
</blockquote></details>

oops, yeah you're right
___
*[src/chatlog/content/filetransferwidget.cpp, line 728 at r10](https://reviewable.io/reviews/qtox/qtox/5354#-LT1JG-P0VQqSk4ikHK_:-LT1JG-P0VQqSk4ikHKa:b-hvtp2v) ([raw file](https://github.com/qtox/qtox/blob/25005c5c19c80a3cbd2d966a2bb6bfeaa20eab36/src/chatlog/content/filetransferwidget.cpp#L728)):*
> ```cpp
>     }
> 
>     fileInfo = file;
> ```

impossible case?

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 728 at r10](https://reviewable.io/reviews/qtox/qtox/5354#-LT1JG-P0VQqSk4ikHK_:-LT1piZ12Q20QacxuKFC:biaf1h2) ([raw file](https://github.com/qtox/qtox/blob/25005c5c19c80a3cbd2d966a2bb6bfeaa20eab36/src/chatlog/content/filetransferwidget.cpp#L728)):*

impossible case?
</blockquote></details>

nevermind, equal just means same ID, not actually equal.
___
*[src/persistence/history.cpp, line 652 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr0qlL0kQr4UUyFDpJ:-LT1q2LR2Rrvu4mJXug6:b67b9xa) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L652)):*

If I remember correctly on windows I was getting corrupted filenames on the output. It could be that the utf16 file paths on windows are causing some extra layer of WTF
</blockquote></details>

Without UTF8 conversion on windows I get screwed up filenames in the widgets with the first 4 characters being squares, and all my widgets have the same filename (latest file). The DB still has the full path and filenames stored correctly, and if I click either of the icons on the widget the file and location open correctly. Also if I clear chatlog and then load history, the first 4 characters of broken symbols change, but stay the same between the multiple widgets. This is pretty weird o.O

anthonybilinski:
Ok I think I've finally found out why we're getting these corrupt filenames if it's not converted to and from utf8:
When we read back a response from the DB, `RawDatabase::extractData` uses `QByteArray::fromRawData` to avoid deep-copying the data, so the `QVariant` passed into the `rowCallback`'s backing data becomes invalid after the callback exits. In addition to this, since `QByteArray` has implicit sharing, copying the `QVariant into file.fileName` in the callback again doesn't do a deep copy, so our `fileName` points to the stack `QVariant` in `RawDatabase::process`. This explains why we see parts of the same last filename appear for each file transfer in history. The fromUtf8/toUtf8 was just forcing a deep copy, saving out data, without actually doing any special conversion.

In `extractData` in the case of ints, we copy, in the cast of text, we call `QString::fromUtf8` which although not explicitly stated in docs: https://doc.qt.io/qt-5/qstring.html#fromUtf8, almost definitely does deep copy since https://doc.qt.io/qt-5/qstring.html#fromUtf16 explicitly states it deep copies and https://doc.qt.io/qt-5/qstring.html#fromRawData exists which explicitly shallow copies.

Having `extractData` deep copy the data that it passes to the callback function seems safest since QByteArray is a hassle to explicitly deep copy, plus callbacks may not be aware of the restriction. There would be a performance overhead if the callback copies a second time, which would effect most of our queries since all messages are stored as `BLOB` and we convert them to `toString()`. Maybe just fixing it in this one callback and having some comment-only warning about making sure to copy data out of the callback, even for implicitly shared types is a better option.

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.h, line 101 at r10](https://reviewable.io/reviews/qtox/qtox/5354#-LTLbhzM2CJqjw5hx5Pv:-LTLbhzM2CJqjw5hx5Pw:b-12enyp) ([raw file](https://github.com/qtox/qtox/blob/25005c5c19c80a3cbd2d966a2bb6bfeaa20eab36/src/persistence/history.h#L101)):*
> ```objc
>     QString friendPk;
>     QString fileId;
>     QByteArray fileName;
> ```

any reason `fileName` is a QByteArray but `filePath` is a `QString`? Seems like both should be `QString`s.

sudden6 (approved):
> Reviewed 3 of 11 files at r8, 3 of 5 files at r9.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.h, line 101 at r10](https://reviewable.io/reviews/qtox/qtox/5354#-LTLbhzM2CJqjw5hx5Pv:-LTOcwLI3pygksQ0WdCe:bo0oavn) ([raw file](https://github.com/qtox/qtox/blob/25005c5c19c80a3cbd2d966a2bb6bfeaa20eab36/src/persistence/history.h#L101)):*

any reason `fileName` is a QByteArray but `filePath` is a `QString`? Seems like both should be `QString`s.
</blockquote></details>

Could be because toxcore returns it as bytes and nobody bothered to use `ToxString`. I see no real reason not to use `QString`.

anthonybilinski (approved):
:lgtm_strong:

@sudden6 if you could review my last minor type change, this is ready to merge. I've tested on Windows, and with the change to QString and the toString() call on the variant, the data is copied and fileNames are no longer corrupted.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LTYowWNE3CFHWKOg1Fl:b-aprt9r)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 652 at r9](https://reviewable.io/reviews/qtox/qtox/5354#-LSr0qlL0kQr4UUyFDpJ:-LTYnfh6E2BTEQ66WUBI:bkyoqjo) ([raw file](https://github.com/qtox/qtox/blob/8427be6678eae2ef151b704feb6bf408038ccdd3/src/persistence/history.cpp#L652)):*

Without UTF8 conversion on windows I get screwed up filenames in the widgets with the first 4 characters being squares, and all my widgets have the same filename (latest file). The DB still has the full path and filenames stored correctly, and if I click either of the icons on the widget the file and location open correctly. Also if I clear chatlog and then load history, the first 4 characters of broken symbols change, but stay the same between the multiple widgets. This is pretty weird o.O
</blockquote></details>

I've created #5460 to address the root cause of this. It's not blocking this change since toString() already forces a deep copy.

sudden6 (changes requested):
> Reviewed 4 of 5 files at r11.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/core/corefile.cpp, line 111 at r11](https://reviewable.io/reviews/qtox/qtox/5354#-LTaFXP13iouwv87_cXN:-LTaFXP13iouwv87_cXO:b-in0bx8) ([raw file](https://github.com/qtox/qtox/blob/f685b92fc76481c984cc86efcfc7843b26e39a0c/src/core/corefile.cpp#L111)):*
> ```cpp
>     ToxFile file{fileNum, friendId, "", "", ToxFile::SENDING};
>     file.filesize = filesize;
>     file.fileName = QString::fromUtf8((char*)avatarHash, TOX_HASH_LENGTH);
> ```

This seems incorrect, because `avatarHash` is binary data and not all binary data can be represented as UTF-8. The previous solution seems to have been correct, because http://doc.qt.io/qt-5/qbytearray.html#QByteArray-1 says the constructor makes a deep copy of the data already.
___
*[src/core/corefile.cpp, line 125 at r11](https://reviewable.io/reviews/qtox/qtox/5354#-LTaGCYbEwC10dx6EDXk:-LTaGCYbEwC10dx6EDXl:b6h32od) ([raw file](https://github.com/qtox/qtox/blob/f685b92fc76481c984cc86efcfc7843b26e39a0c/src/core/corefile.cpp#L125)):*
> ```cpp
>     QMutexLocker mlocker(&fileSendMutex);
> 
>     QByteArray fileNameBytes = filename.toUtf8();
> ```

use `ToxString` here to convert?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/core/corefile.cpp, line 111 at r11](https://reviewable.io/reviews/qtox/qtox/5354#-LTaFXP13iouwv87_cXN:-LTfqgVG337hh9MJPm-w:bvndbay) ([raw file](https://github.com/qtox/qtox/blob/f685b92fc76481c984cc86efcfc7843b26e39a0c/src/core/corefile.cpp#L111)):*

This seems incorrect, because `avatarHash` is binary data and not all binary data can be represented as UTF-8. The previous solution seems to have been correct, because http://doc.qt.io/qt-5/qbytearray.html#QByteArray-1 says the constructor makes a deep copy of the data already.
</blockquote></details>

Good catch. When looking into how this was used before since we converted the QByteArray into QString when used, I saw that we never actually read the avatar's filename. In profile, we convert the hash toHex() to get the filename, so I've just removed it completely here, keeping the empty string used in constructor two lines above.
___
*[src/core/corefile.cpp, line 125 at r11](https://reviewable.io/reviews/qtox/qtox/5354#-LTaGCYbEwC10dx6EDXk:-LTfs3q25ow4-3zgXcy8:bb6un1j) ([raw file](https://github.com/qtox/qtox/blob/f685b92fc76481c984cc86efcfc7843b26e39a0c/src/core/corefile.cpp#L125)):*

use `ToxString` here to convert?
</blockquote></details>

Done

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5354#-:-LTfvIWM3dKtZ-f6QAeR:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

PR #5356, opened 2018-09-28
chore: make {merge,test}-pr.sh scripts accept hash-preceded PR number

zetok:
Previously one needed to provide only digits, (e.g. `123`), otherwise
scripts would fail to recognise supplied argument as an issue number.

Github almost everywhere precedes PR numbers with hash sign, so it might
be easier/faster to copypaste PR number with along with preceding hash.

With the change scripts will ignore the first preceding hash in the
supplied PR number.

---
Although, shell will just treat the characters preceded with a hash as a comment, unless enclosed in quotes. Which makes me wonder if this is an actual improvement, or whether I'm adding a pitfall here..

```bash
# pitfall
./test-pr.sh #123
# works
./test-pr.sh '#123'
```

Merge only if you think that this is an actual improvement.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5356)

### DIFF
--- a/merge-pr.sh
+++ b/merge-pr.sh
@@ -31,7 +31,7 @@
 
 set -e -o pipefail
 
-readonly PR=$1
+readonly PR="${1###}"
 
 # make sure to add newlines to the message, otherwise merge message
 # will not look well
--- a/test-pr.sh
+++ b/test-pr.sh
@@ -30,7 +30,7 @@
 
 set -e -o pipefail
 
-readonly PR=$1
+readonly PR="${1###}"
 
 # make sure to add newlines to the message, otherwise merge message
 # will not look well


sudden6:
I like it, two questions though, does this still accept  the number only? and does it also accept stuff >9999?

zetok:
>I like it, two questions though, does this still accept the number only? and does it also accept stuff >9999?

Yep, `$PR` is still checked against [`exit_if_not_pr`](https://github.com/qTox/qTox/blob/0796434fdf3bc9609c33a35ef106ff03800e1e99/merge-pr.sh#L54), which is just [matching it against a regexp](https://github.com/qTox/qTox/blob/0796434fdf3bc9609c33a35ef106ff03800e1e99/tools/lib/PR_bash.source#L25,L34), and the regexp in question accepts any string that contains only 1 or more digits in it. There's no length check whatsoever.

The 3 hash characters are perhaps a bit misleading, and if we get to the bottom of things more proper parameter expansion to remove a single preceding hash would have been `"${1#\#}"`, but since both end up doing the same in this particular case, I went for the more "visually pleasing" variant.

```bash
FOO="#12345667723452352345234532452365745674567456734562353245234523452345234"
echo "${FOO#\#}"
echo "${FOO###}"
````

sudden6 (approved):
:lgtm_strong: thanks for the explanation 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5356#-:-LN_zNOw6TzAqit-kz_O:b6uorgx)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5366, opened 2018-10-04
Pause state fix

sphaerophoria:
Toxcore has a 3 state pause, us, them, or both. Currently our UI messes up if both parties pause. This changeset changes our UI behavior to show whether we're paused, or if we are waiting on the remote to
unpause.

*Note*: This fix is on top of my other PR #5354. This is because I touched this area pretty heavily and don't want to have to fix the UI twice.

Demo of post-fix behavior.

tox_file_transfer

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)

### DIFF
--- a/src/core/toxfilepause.h
+++ b/src/core/toxfilepause.h
@@ -1,0 +1,76 @@
+/*
+    Copyright © 2018 by The qTox Project Contributors
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef TOX_FILE_PAUSE_H
+#define TOX_FILE_PAUSE_H
+
+class ToxFilePause
+{
+public:
+    void localPause()
+    {
+        localPauseState = true;
+    }
+
+    void localResume()
+    {
+        localPauseState = false;
+    }
+
+    void localPauseToggle()
+    {
+        localPauseState = !localPauseState;
+    }
+
+    void remotePause()
+    {
+        remotePauseState = true;
+    }
+
+    void remoteResume()
+    {
+        remotePauseState = false;
+    }
+
+    void remotePauseToggle()
+    {
+        remotePauseState = !remotePauseState;
+    }
+
+    bool localPaused() const
+    {
+        return localPauseState;
+    }
+
+    bool remotePaused() const
+    {
+        return remotePauseState;
+    }
+
+    bool paused() const
+    {
+        return localPauseState || remotePauseState;
+    }
+
+private:
+    bool localPauseState = false;
+    bool remotePauseState = false;
+};
+
+#endif // TOX_FILE_PAUSE_H
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,6 +268,7 @@ set(${PROJECT_NAME}_SOURCES
   src/core/toxencrypt.h
   src/core/toxfile.cpp
   src/core/toxfile.h
+  src/core/toxfilepause.h
   src/core/toxid.cpp
   src/core/toxid.h
   src/core/toxlogger.cpp
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -557,18 +557,18 @@ void FileTransferWidget::handleButton(QPushButton* btn)
         if (btn->objectName() == "cancel") {
             Core::getInstance()->cancelFileSend(fileInfo.friendId, fileInfo.fileNum);
         } else if (btn->objectName() == "pause") {
-            Core::getInstance()->pauseResumeFileSend(fileInfo.friendId, fileInfo.fileNum);
+            Core::getInstance()->pauseResumeFile(fileInfo.friendId, fileInfo.fileNum);
         } else if (btn->objectName() == "resume") {
-            Core::getInstance()->pauseResumeFileSend(fileInfo.friendId, fileInfo.fileNum);
+            Core::getInstance()->pauseResumeFile(fileInfo.friendId, fileInfo.fileNum);
         }
     } else // receiving or paused
     {
         if (btn->objectName() == "cancel") {
             Core::getInstance()->cancelFileRecv(fileInfo.friendId, fileInfo.fileNum);
         } else if (btn->objectName() == "pause") {
-            Core::getInstance()->pauseResumeFileRecv(fileInfo.friendId, fileInfo.fileNum);
+            Core::getInstance()->pauseResumeFile(fileInfo.friendId, fileInfo.fileNum);
         } else if (btn->objectName() == "resume") {
-            Core::getInstance()->pauseResumeFileRecv(fileInfo.friendId, fileInfo.fileNum);
+            Core::getInstance()->pauseResumeFile(fileInfo.friendId, fileInfo.fileNum);
         } else if (btn->objectName() == "accept") {
             QString path =
                 QFileDialog::getSaveFileName(Q_NULLPTR,
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -346,7 +346,7 @@ void FileTransferWidget::updateWidgetColor(ToxFile const& file)
 
 void FileTransferWidget::updateWidgetText(ToxFile const& file)
 {
-    if (lastStatus == file.status) {
+    if (lastStatus == file.status && file.status != ToxFile::PAUSED) {
         return;
     }
 
@@ -360,7 +360,11 @@ void FileTransferWidget::updateWidgetText(ToxFile const& file)
         break;
     case ToxFile::PAUSED:
         ui->etaLabel->setText("");
-        ui->progressLabel->setText(tr("Paused", "file transfer widget"));
+        if (file.pauseStatus.localPaused()) {
+            ui->progressLabel->setText(tr("Paused", "file transfer widget"));
+        } else {
+            ui->progressLabel->setText(tr("Remote Paused", "file transfer widget"));
+        }
         break;
     case ToxFile::TRANSMITTING:
         ui->etaLabel->setText("");
@@ -474,7 +478,7 @@ void FileTransferWidget::updateSignals(ToxFile const& file)
 
 void FileTransferWidget::setupButtons(ToxFile const& file)
 {
-    if (lastStatus == file.status) {
+    if (lastStatus == file.status && file.status != ToxFile::PAUSED) {
         return;
     }
 
@@ -492,9 +496,15 @@ void FileTransferWidget::setupButtons(ToxFile const& file)
         break;
 
     case ToxFile::PAUSED:
-        ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/arrow_white.svg")));
-        ui->leftButton->setObjectName("resume");
-        ui->leftButton->setToolTip(tr("Resume transfer"));
+        if (file.pauseStatus.localPaused()) {
+            ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/arrow_white.svg")));
+            ui->leftButton->setObjectName("resume");
+            ui->leftButton->setToolTip(tr("Resume transfer"));
+        } else {
+            ui->leftButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/pause.svg")));
+            ui->leftButton->setObjectName("pause");
+            ui->leftButton->setToolTip(tr("Pause transfer"));
+        }
 
         ui->rightButton->setIcon(QIcon(Style::getImagePath("fileTransferInstance/no.svg")));
         ui->rightButton->setObjectName("cancel");
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -715,18 +715,11 @@ void Core::sendAvatarFile(uint32_t friendId, const QByteArray& data)
     CoreFile::sendAvatarFile(this, friendId, data);
 }
 
-void Core::pauseResumeFileSend(uint32_t friendId, uint32_t fileNum)
+void Core::pauseResumeFile(uint32_t friendId, uint32_t fileNum)
 {
     QMutexLocker ml{coreLoopLock.get()};
 
-    CoreFile::pauseResumeFileSend(this, friendId, fileNum);
-}
-
-void Core::pauseResumeFileRecv(uint32_t friendId, uint32_t fileNum)
-{
-    QMutexLocker ml{coreLoopLock.get()};
-
-    CoreFile::pauseResumeFileRecv(this, friendId, fileNum);
+    CoreFile::pauseResumeFile(this, friendId, fileNum);
 }
 
 void Core::cancelFileSend(uint32_t friendId, uint32_t fileNum)
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -130,8 +130,7 @@ public slots:
     void cancelFileRecv(uint32_t friendId, uint32_t fileNum);
     void rejectFileRecvRequest(uint32_t friendId, uint32_t fileNum);
     void acceptFileRecvRequest(uint32_t friendId, uint32_t fileNum, QString path);
-    void pauseResumeFileSend(uint32_t friendId, uint32_t fileNum);
-    void pauseResumeFileRecv(uint32_t friendId, uint32_t fileNum);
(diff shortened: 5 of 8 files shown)


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/5366#-LNy0cnS4u91dR7KnSwA:-LNy0cnT4j7Qsq6_5chE:b-m9smcv) ([raw file](https://github.com/qtox/qtox/blob/e8c92c943b401cb20fe580e033963bad23ca7a11/src/core/toxfile.h#L32)):*
> ```objc
>     };
> 
>     enum PauseStatus {
> ```

I would prefer to use an `enum class` here to get the type safety of C++ enums.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/5366#-LNy0cnS4u91dR7KnSwA:-LNy6IRa8H2CFWRP7SxW:b-bi8kug) ([raw file](https://github.com/qtox/qtox/blob/e8c92c943b401cb20fe580e033963bad23ca7a11/src/core/toxfile.h#L32)):*

I would prefer to use an `enum class` here to get the type safety of C++ enums.
</blockquote></details>

Enum classes don't really play well with bitfields I find. I guess we could model this differently but I think an enum class would make implementation harder with little benefit (if we use it like I have in the current implementation it would just be the same things with more casts).

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/5366#-LNy0cnS4u91dR7KnSwA:-LNy6kiW7iRs1OJVyfi_:b-wt4pyz) ([raw file](https://github.com/qtox/qtox/blob/e8c92c943b401cb20fe580e033963bad23ca7a11/src/core/toxfile.h#L32)):*

Enum classes don't really play well with bitfields I find. I guess we could model this differently but I think an enum class would make implementation harder with little benefit (if we use it like I have in the current implementation it would just be the same things with more casts).
</blockquote></details>

I guess there's no reason we couldn't implement this with 2 bools, I just thought this was cleaner. Let me know what you think :).

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/5366#-LNy0cnS4u91dR7KnSwA:-LNyFx0a90sBKfRTMwPI:bxzu7cn) ([raw file](https://github.com/qtox/qtox/blob/e8c92c943b401cb20fe580e033963bad23ca7a11/src/core/toxfile.h#L32)):*

I guess there's no reason we couldn't implement this with 2 bools again, I just thought this was cleaner. Let me know what you think :).
</blockquote></details>

I think it wouldn't make the code much harder to read if we don't use the enum as bitfield, but if you really want to use a bitfield, maybe you can take a look at http://doc.qt.io/archives/qt-4.8/qflags.html

Personally I would just change the bitfield logic to handle both cases.

sphaerophoria (commented):
> Reviewed 17 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfile.h, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/5366#-LNy0cnS4u91dR7KnSwA:-LO1rU57CluBjNfxspyO:b-srof3) ([raw file](https://github.com/qtox/qtox/blob/e8c92c943b401cb20fe580e033963bad23ca7a11/src/core/toxfile.h#L32)):*

I think it wouldn't make the code much harder to read if we don't use the enum as bitfield, but if you really want to use a bitfield, maybe you can take a look at http://doc.qt.io/archives/qt-4.8/qflags.html

Personally I would just change the bitfield logic to handle both cases.
</blockquote></details>

I opted to just write a simple class to wrap the concept with a couple bools. This ended up being cleaner anyways. You could argue that just exposing the two bools is nicer but I found that I'm less prone to errors this way :).

src/core/toxfilepause.h
```diff
@@ -0,0 +1,76 @@
...
+    {
+        localPauseState = true;
+    }
+
+    void localResume()
+    {
+        localPauseState = false;
+    }
+
+    void localPauseUnpause()
```
Diadlo: Name looks a bit strange, isn't it?)
What do you think about `localPauseToggle`?

sphaerophoria: I thought I was following the naming scheme from above... turns out I was just introducing more dumb naming (above was pauseResume). I like pause toggle.

sudden6 (changes requested):
> Reviewed 7 of 17 files at r1, 5 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LO200-N0lwbymMEVvQu:b64hoq8) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*
> ```cpp
> 
>     switch (file.status) {
>     case ToxFile::INITIALIZING:
> ```

add comments that the fallthrough is intentional? add a `default` case that hits an assertion?
___
*[src/chatlog/content/filetransferwidget.cpp, line 420 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO20DA084bym2fbtedR:-LO20DA084bym2fbtedS:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L420)):*
> ```cpp
>         break;
>     }
>     case ToxFile::BROKEN:
> ```

same
___
*[src/widget/form/chatform.cpp, line 337 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO212HDD7NGRmxbEedJ:-LO212HE3gv1hwvqWZw-:bb1dbrs) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/widget/form/chatform.cpp#L337)):*
> ```cpp
> void ChatForm::onFileTransferBrokenUnbroken(ToxFile file, bool broken)
> {
>     if (broken)
> ```

curly braces
___
*[src/widget/form/chatform.cpp, line 833 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO216FmDJPZ09Q2B8jA:-LO216FmDJPZ09Q2B8jB:bb1dbrs) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/widget/form/chatform.cpp#L833)):*
> ```cpp
>         auto msg = chatMessageFromHistMessage(histMessage, metadata);
> 
>         if (!msg)
> ```

curly braces

sphaerophoria (commented):
> Reviewed 5 of 5 files at r2, 8 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LO2Dj0WBv4fuvwe2x1P:bh8mvpf) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*

add comments that the fallthrough is intentional? add a `default` case that hits an assertion?
</blockquote></details>

Thanks for the comments, just in case it wasn't clear... you've already reviewed this part of the changeset in my previous PR, I just had the work ready to go for an easy merge after that one. I hope that wasn't too confusing.

Hm, I don't think I've ever seen that. I've always considered multiple tags like this a single jump point which is where multiple tags go. Clang warns on implicit fallthrough for cases where you put code in between but not on this. If you don't feel strongly about it I'd prefer to leave it as is in terms of fallthrough comments.

As for default assert I've always been 50/50 on it. If we add states the default assert is more likely to be hit in a valid case, but if we have memory corruption we'll catch it in debug mode. It's a toughy. Anyways it seems like we use the assert(false) pattern so I'll update these and add the changes to PR #5354 as well.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LO2JKGr4I62c2uPwGT1:b-ymqsb9) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*

Thanks for the comments, just in case it wasn't clear... you've already reviewed this part of the changeset in my previous PR, I just had the work ready to go for an easy merge after that one. I hope that wasn't too confusing.

Hm, I don't think I've ever seen that. I've always considered multiple tags like this a single jump point which is where multiple tags go. Clang warns on implicit fallthrough for cases where you put code in between but not on this. If you don't feel strongly about it I'd prefer to leave it as is in terms of fallthrough comments.

As for default assert I've always been 50/50 on it. If we add states the default assert is more likely to be hit in a valid case, but if we have memory corruption we'll catch it in debug mode. It's a toughy. Anyways it seems like we use the assert(false) pattern so I'll update these and add the changes to PR #5354 as well.
</blockquote></details>

Done assuming no fallthrough okay
___
*[src/chatlog/content/filetransferwidget.cpp, line 420 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO20DA084bym2fbtedR:-LO2JO4D1FzUaE_MvLYL:bnj7hl0) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L420)):*

same
</blockquote></details>

same as above
___
*[src/widget/form/chatform.cpp, line 337 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO212HDD7NGRmxbEedJ:-LO2JR0g6uyJupuF_BrK:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/widget/form/chatform.cpp#L337)):*

curly braces
</blockquote></details>

Done.
___
*[src/widget/form/chatform.cpp, line 833 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO216FmDJPZ09Q2B8jA:-LO2JRtnD_ty2hLQS4RN:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/widget/form/chatform.cpp#L833)):*

curly braces
</blockquote></details>

Done.

sudden6 (changes requested):
> Reviewed 3 of 13 files at r6.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LO3WCzSAdVN3S6x3DlO:blt6hyf) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*

Done assuming no fallthrough okay
</blockquote></details>

Right, I overlooked that this part belongs to a different PR... anyway I sometimes notice stuff on later review and want to point it out still, I hope this is not too annoying.

I think you're right that the warning happens only when code is in between, didn't know this and I think this is ok then.

For the assert I think it's a nice and easy way to ensure that things break in debug mode when you add additional cases/enum values and I don't really see a drawback.

You bring up quite a few new points questions which are not really handled by our coding guidelines, do you think we should add them there too?
___
*[src/core/toxfilepause.h, line 2 at r6](https://reviewable.io/reviews/qtox/qtox/5366#-LO3X_HO8CPWhgI0WR3E:-LO3X_HO8CPWhgI0WR3F:b-6826g3) ([raw file](https://github.com/qtox/qtox/blob/93f4a872dfa00eaf8469e2807b95dead474a40b2/src/core/toxfilepause.h#L2)):*
> ```objc
> /*
>     Copyright © 2017-2018 by The qTox Project Contributors
> ```

Not 100% sure, but I think the copyright should start with the current year.

sudden6:
Somewhere you forgot to include `<cassert>` so the build is failing currently.

sphaerophoria (commented):
> Reviewed 13 of 13 files at r6.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LO4YFrn6PxiJ8B3J6CB:b-n2izd8) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*

Right, I overlooked that this part belongs to a different PR... anyway I sometimes notice stuff on later review and want to point it out still, I hope this is not too annoying.

I think you're right that the warning happens only when code is in between, didn't know this and I think this is ok then.

For the assert I think it's a nice and easy way to ensure that things break in debug mode when you add additional cases/enum values and I don't really see a drawback.

You bring up quite a few new points questions which are not really handled by our coding guidelines, do you think we should add them there too?
</blockquote></details>

Just to flush out the other side a little more, the main benefit to non-default is if you add a value to the enum the compiler will warn you that you didn't handle here. If you add default: assert(false) then you might not notice until you hit the assertion at runtime.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfilepause.h, line 2 at r6](https://reviewable.io/reviews/qtox/qtox/5366#-LO3X_HO8CPWhgI0WR3E:-LO4Z09u3QBhmCzYwaAl:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/93f4a872dfa00eaf8469e2807b95dead474a40b2/src/core/toxfilepause.h#L2)):*

Not 100% sure, but I think the copyright should start with the current year.
</blockquote></details>

Done.

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 17 files at r1, 8 of 13 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366#-:-LODl5jd2G9-qyqTAMQN:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
Assigned to myself so we don't forget to merge #5354 first.

anthonybilinski:
#5354 is now merged. @sphaerophoria could you rebase when you have time so that this can be merged too?

sphaerophoria:
Rebased and ran quick sanity test to make sure stuff was still working as I intended

anthonybilinski (approved):
> Reviewed 1 of 13 files at r6, 1 of 2 files at r7, 6 of 14 files at r8.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366)* status: 0 of 1 LGTMs obtained (waiting on @anthonybilinski)

___
*[src/chatlog/content/filetransferwidget.cpp, line 372 at r4](https://reviewable.io/reviews/qtox/qtox/5366#-LO200-N0lwbymMEVvQt:-LTtC3QT8daN703JmE2r:bmo0uhh) ([raw file](https://github.com/qtox/qtox/blob/dc72a5d901324b667cad3d45e9bb30483751fce8/src/chatlog/content/filetransferwidget.cpp#L372)):*

Just to flush out the other side a little more, the main benefit to non-default is if you add a value to the enum the compiler will warn you that you didn't handle here. If you add default: assert(false) then you might not notice until you hit the assertion at runtime.
</blockquote></details>

Just to chime in, I think the no-default case is safer than default: assert(false) due to the compile time check that all enum values are handled instead of the runtime check. I'd be open to this being added to the coding guideline.

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5366#-:-LTtCdCRBeVIHS3JAOmM:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
> Just to chime in, I think the no-default case is safer than default: assert(false) due to the compile time check that all enum values are handled instead of the runtime check. I'd be open to this being added to the coding guideline.

I'd be ok with that, even better if we can integrate it into our CI so it causes an error.

PR #5378, opened 2018-10-07
fix(core): ensure QTimers are moved with the objects they belong to

sudden6:
We use the Qt parent/child model instead of unique_ptr to achieve this.

@sphaerophoria can you please review this?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5378)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -47,12 +47,14 @@ const QString Core::TOX_EXT = ".tox";
 Core::Core(QThread* coreThread)
     : tox(nullptr)
     , av(nullptr)
+    , toxTimer{new QTimer{this}}
     , coreLoopLock(new QMutex(QMutex::Recursive))
     , coreThread(coreThread)
 {
-    toxTimer.setSingleShot(true);
-    connect(&this->toxTimer, &QTimer::timeout, this, &Core::process);
-    connect(coreThread, &QThread::finished, &toxTimer, &QTimer::stop);
+    assert(toxTimer);
+    toxTimer->setSingleShot(true);
+    connect(toxTimer, &QTimer::timeout, this, &Core::process);
+    connect(coreThread, &QThread::finished, toxTimer, &QTimer::stop);
 }
 
 Core::~Core()
@@ -225,8 +227,6 @@ ToxCorePtr Core::makeToxCore(const QByteArray& savedata, const ICoreSettings* co
     // connect the thread with the Core
     connect(thread, &QThread::started, core.get(), &Core::onStarted);
     core->moveToThread(thread);
-    // since this is allocated in the constructor move it to the other thread too
-    core->toxTimer.moveToThread(thread);
 
     // when leaving this function 'core' should be ready for it's start() action or
     // a nullptr
@@ -323,7 +323,7 @@ void Core::process()
 
     unsigned sleeptime =
         qMin(tox_iteration_interval(tox.get()), CoreFile::corefileIterationInterval());
-    toxTimer.start(sleeptime);
+    toxTimer->start(sleeptime);
 }
 
 bool Core::checkConnection()
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -273,7 +273,7 @@ private:
     ToxPtr tox;
 
     std::unique_ptr<CoreAV> av;
-    QTimer toxTimer;
+    QTimer* toxTimer = nullptr;
     // recursive, since we might call our own functions
     // pointer so we can circumvent const functions
     std::unique_ptr<QMutex> coreLoopLock = nullptr;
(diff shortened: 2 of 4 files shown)


sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5378)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/5378#-LOEZ9wZ0s7-CHB5kCcU:-LOEZ9wZ0s7-CHB5kCcV:b-wokysl) (no related file):*
I don't think this fixes the problem. I just pulled and built and ended up with the same issue. I think your options are

1. Heap allocate the QTimer and set this as it's parent, keep a raw pointer locally and let qt manage it
2. Stack allocate the QTimer, do not set a parent, and moveToThread on it manually.

---
___
*[src/core/core.h, line 276 at r1](https://reviewable.io/reviews/qtox/qtox/5378#-LOEXA81-O4mya9Meo7j:-LOEXA81-O4mya9Meo7k:b80aw6a) ([raw file](https://github.com/qtox/qtox/blob/691aea19317d245599be0bc5a85e5f79431b309a/src/core/core.h#L276)):*
> ```objc
> 
>     std::unique_ptr<CoreAV> av;
>     QTimer toxTimer{this};
> ```

Is this even legal? As far as my understanding of the Qt ownership model goes, Qt will call delete on the QTimer even though it's stack allocated. I believe if you stack allocate you want to call "moveToThread" on the object directly?

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5378)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.h, line 276 at r1](https://reviewable.io/reviews/qtox/qtox/5378#-LOEXA81-O4mya9Meo7j:-LOEffTp4ynj7ToCKOg5:b7pjq5k) ([raw file](https://github.com/qtox/qtox/blob/691aea19317d245599be0bc5a85e5f79431b309a/src/core/core.h#L276)):*

Is this even legal? As far as my understanding of the Qt ownership model goes, Qt will call delete on the QTimer even though it's stack allocated. I believe if you stack allocate you want to call "moveToThread" on the object directly?
</blockquote></details>

I guess this won't end up with an extra delete since the QTimer destructor will destruct itself before the QObject destructor runs.

sudden6:
I settled on using heap allocation and letting Qt handle the destruction. For some reason I'm not hitting the assert you hit, are you using a special build of Qt?

Also feel free to contact me at `3D464D08D5B556221A4E5FFC35657BBD2B90028C949A8C56C878925D65ABC506A6BBF3F166A1` or the ToxID mentioned in my GH profile.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5378)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.h, line 276 at r1](https://reviewable.io/reviews/qtox/qtox/5378#-LOEXA81-O4mya9Meo7j:-LOEiasL7TSdl7cZ66R_:b47j4wj) ([raw file](https://github.com/qtox/qtox/blob/691aea19317d245599be0bc5a85e5f79431b309a/src/core/core.h#L276)):*

I guess this won't end up with an extra delete since the QTimer destructor will destruct itself before the QObject destructor runs.
</blockquote></details>

After reading through http://doc.qt.io/qt-5/objecttrees.html and http://doc.qt.io/qt-5/qobject.html#moveToThread and playing around with it a little myself I've come up with a guess as to why this doesn't work. I think moveToThread *only* touches state of the parentless class.

The documentation for moveToThread notes that you cannot call this function on a class with a parent. Why would this be the case? Well it would make sense if the parent is managing calls to the child, not the child itself. This includes the "delete" call. When the QTimer is created on the stack it calls it's own destructor first. Since the moveToThread function was called on the parent, the QTimer internal thread state hasn't been changed so it just destructs on it's current thread. All other calls have been made on the other thread so this fails.

sphaerophoria (commented):
I'm using msys2 on windows to manage all dependencies, it's possible that my versions are not lining up with everyone elses.

```
$ pacman -Qi mingw-w64-x86_64-qt5
Name            : mingw-w64-x86_64-qt5
Version         : 5.11.2-2
```

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5378#-:-LOEjr3m6ED7dExwDwwq:b-9b0y8)* status: 0 of 1 LGTMs obtained

sphaerophoria (commented):
:lgtm_strong:

This looks reasonable to me and no longer asserts on shutdown, I'll pass it assuming you've double checked that the timers are still firing :).

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5378#-:-LOEl10K7sw0otvqqpK4:b6anpk1)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5397, opened 2018-10-15
Chiitoo theming continued

sudden6:

I implemented theme loading from standard paths as discussed in #5349

This PR does not introduce new themes or installs themes. It can also only load a theme from `themes/default` for now, because the searching for themes and the UI to select them are not yet implemented.

I still think we should merge this now, so people not experienced with programming can start to modify the default theme without having to recompile qTox.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5397)


Chiitoo:
Thanks for looking into this more than I seem to be able to at this time.  :]

I'll hopefully get to test this soon.

sphaerophoria:
> refactor: remove unused includes and don't set Tox as organization
> 
> Having "Tox" as Organization results in all standard paths being "Tox/
> qTox" instead of "qTox". Since there is no such thing as "Tox"
> organization I removed it.

I have a few questions about this

1. Will this lose all my settings because we're looking in .config/qTox instead of .config/tox?
2. Was there a shared consensus between clients before that the toxcore data would be stored in the .config/tox folder? If not it seems reasonable
3. Why do we change this but not any of the path resolvers in Settings::get<>Dir?

sudden6:
> Will this lose all my settings because we're looking in .config/qTox instead of .config/tox?

I saw no other place where we use QStandardPaths and on my system no files exist in `.config/Tox/qTox`.

> Was there a shared consensus between clients before that the toxcore data would be stored in the .config/tox folder? If not it seems reasonable

As far as I'm aware the consensus goes only for the `.tox` profile file. IMO it also doesn't make sense to store client specific data in a shared directory, because no other client can do anything with it.

https://tox.gitbooks.io/tox-client-standard/content/data_storage/data_storage.html also says nothing about client specific data, so I guess we're fine.

> Why do we change this but not any of the path resolvers in Settings::get<>Dir?

Because they don't use any of the standard paths at the moment anyway. We need to upgrade them at some point though.

Chiitoo:
It seems like there is a '/' that gets lost somewhere:

```
widget/style.cpp:163 : Warning: Failed to open file (using defaults): 
"[...].local/share/qTox/themes/defaultchatArea/spinner.svg"
```

So for testing, I did this:

```
diff --git a/src/widget/style.cpp b/src/widget/style.cpp
index c56534d4..05e75119 100644
--- a/src/widget/style.cpp
+++ b/src/widget/style.cpp
@@ -148,7 +148,7 @@ const QString Style::getStylesheet(const QString& filename, const QFont& baseFon
 static QStringList existingImagesCache;
 const QString Style::getImagePath(const QString& filename)
 {
-    QString fullPath = getThemePath() + filename;
+    QString fullPath = getThemePath() + "/" + filename;

     // search for image in cache
     if (existingImagesCache.contains(fullPath)) {
@@ -201,7 +201,7 @@ QFont Style::getFont(Style::Font font)
 const QString Style::resolve(const QString& filename, const QFont& baseFont)
 {
     QString themePath = getThemePath();
-    QString fullPath = themePath + filename;
+    QString fullPath = themePath + "/" + filename;
     QString qss;

     QFile file{fullPath};
@@ -266,7 +266,7 @@ const QString Style::resolve(const QString& filename, const QFont& baseFont)
         path.remove(QStringLiteral("@getImagePath("));
         path.chop(1);

-        QString fullImagePath = getThemePath() + path;
+        QString fullImagePath = getThemePath() + "/" + path;
         // image not in cache
         if (!existingImagesCache.contains(fullPath)) {
             if (QFileInfo::exists(fullImagePath)) {
```

After that, things seemed better again (looked pretty funny without any theming by the by).  I'm still unable to have my edits show up, but I'm obviously doing something weirdly wrong since it works for you.  :]

As a sidey-note, how about adding a 'theme' to the warning messages?  That is, something like “Failed to open theme file" instead of “Failed to open file”, just to make it more apparent what files we're talking about.

sudden6:
@Chiitoo fixed the path thing

Also atm qTox only attempts to load the default theme, so only `themes/default` makes sense to overwrite.

I tested by setting the background color to green in  `.local/share/qTox/themes/default/settings/mainHead.css`

Chiitoo:
Yeah, the default is what I had been testing with.

I had been changing colours in `.local/share/qTox/themes/default/chatArea/chatArea.css` and `.local/share/qTox/themes/default/sendButton/sendButton.css` to name a couple.

Trying your example now, it works!  Hmmm.

Anyblue, yeah, while there are things that require work (like making it not install anything at build-time), since it's behind an option now, I believe it's okay to merge so that people can start tinkering with it (and that it doesn't keep other things back, or fall behind too much and all that).

sudden6:
> like making it not install anything at build-time

Do you think we should remove that in general maybe? Right now it's only useful for devs and I suspect they can copy the theme to the correct path them self.

sphaerophoria (commented):
> Reviewed 93 of 96 files at r1, 4 of 4 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5397)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/genericchatform.cpp, line 61 at r3](https://reviewable.io/reviews/qtox/qtox/5397#-LPN0q4REM2OloEHnBK2:-LPN0q4REM2OloEHnBK3:b-uvilfp) ([raw file](https://github.com/qtox/qtox/blob/4cbab426ab9b6e3128ea9805a4920553d449a24d/src/widget/form/genericchatform.cpp#L61)):*
> ```cpp
>  */
> 
> #define SET_STYLESHEET(x) (x)->setStyleSheet(Style::getStylesheet(":/themes/default/" #x "/" #x ".css"))
> ```

how come we hardcode :/themes/default here? Everywhere else we let Style::getStylesheet handle that.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5397)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/genericchatform.cpp, line 61 at r3](https://reviewable.io/reviews/qtox/qtox/5397#-LPN0q4REM2OloEHnBK2:-LPN3VDx5St5Xn7Yrg7P:bycisbc) ([raw file](https://github.com/qtox/qtox/blob/4cbab426ab9b6e3128ea9805a4920553d449a24d/src/widget/form/genericchatform.cpp#L61)):*

how come we hardcode :/themes/default here? Everywhere else we let Style::getStylesheet handle that.
</blockquote></details>

That's probably a bug caused by search replace, since this macros is not used I'll remove it.

Chiitoo:
> > like making it not install anything at build-time
> 
> Do you think we should remove that in general maybe? Right now it's only useful for devs and I suspect they can copy the theme to the correct path them self.

I have been thinking of suggesting that, actually.

Like you say, it's not exactly needed as we can just copy them over manually, and since I suspect the behaviour would be removed (or rather changed to install phase or so), there's probably little point in having it in at all.  :]

sudden6:
@Chiitoo done

sphaerophoria (commented):
:lgtm_strong:

> Reviewed 7 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5397#-:-LPNECMsCY47XwuoeRR8:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

Chiitoo:
Seems good to me too.

One note about the changes suggested for the original, that I have been forgetting to mention...

From https://github.com/qTox/qTox/pull/5160#issuecomment-398489305 I did not touch `qss.replace(QRegularExpression(QString("%1\\b").arg(key)), dict[key]);` because I didn't quite get it to work (or/and I misunderstood the suggestion).

sudden6:
@Chiitoo implemented the more efficient string concatenation. LGTM now?

Chiitoo:
Looks good to me.  :]

PR #5401, opened 2018-10-18
fix(core): remove pgc code

sudden6:
This code was needed for the pgc implementation, but is not needed for
minipgc

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5401)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -460,21 +460,11 @@ void Core::onGroupInvite(Tox* tox, uint32_t friendId, Tox_Conference_Type type,
     switch (type) {
     case TOX_CONFERENCE_TYPE_TEXT:
         qDebug() << QString("Text group invite by %1").arg(friendId);
-        if (friendId == UINT32_MAX) {
-            // Rejoining existing (persistent) conference after disconnect and reconnect.
-            tox_conference_join(tox, friendId, cookie, length, nullptr);
-            return;
-        }
         emit core->groupInviteReceived(inviteInfo);
         break;
 
     case TOX_CONFERENCE_TYPE_AV:
         qDebug() << QString("AV group invite by %1").arg(friendId);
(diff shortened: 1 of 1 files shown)


sphaerophoria:
Mind flushing out the rational here a little? I currently don't have enough context to review because I don't know the difference between pgc and minipgc and why their specs are different. I can't verify that this isn't needed anymore with the amount of information I currently have.

sudden6:
PGC was the first implementation for making persistent group chats in c-toxcore. It never made it into master. minipgc is the second implementation for making persistent group chats, this time it was merged. We had the code for PGC already in qTox because it doesn't interfere with the c-toxcore/master code, but for minipgc the API changed slightly compared to PGC and we need to adapt to this changes.

sphaerophoria:
Alright had an offline discussion with @sudden6 and figured out what's going on here.

A good start point to read on this would be https://github.com/TokTok/c-toxcore/blob/master/docs/minpgc.md

Basically there was an implementation of persistent group chats that never made it into master. As part of this PR (826) this change was made to tox.h

```
 *  if friend_number == UINT32_MAX then conference automatically joined.
 *  On auto-join client must call tox_conference_join or toxav_join_av_groupchat
 *  immediately in callback. For api compatibility reason, if client don't
 *  call one these functions, conference will be deleted and toxcore
 *  totally forget this conference.
 *
```

We preemptively accepted this change into qTox thinking that the PR would get merged. This was not the case. We can back out the change since it's unlikely persistent group chats will ever get merged in that form.

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5401#-:-LPMxOmC3prLilUB6tou:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5406, opened 2018-10-23
Added feature: ability to set custom audio files for qtox events

Monsterovich:

Sometimes you want to change notification sound but they were stored inside qtox binary. Audio files must be put in %APPDATA%/qtox/audio otherwise the default sound is used.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5406)

### DIFF
--- a/src/audio/audio.cpp
+++ b/src/audio/audio.cpp
@@ -25,6 +25,8 @@
 #include "src/persistence/settings.h"
 
 #include <QDebug>
+#include <QFile>
+#include <QDir>
 
 #include <cassert>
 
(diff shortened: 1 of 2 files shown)


sudden6 (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5406)* status: 0 of 1 LGTMs obtained

___
*[src/audio/audio.cpp, line 203 at r1](https://reviewable.io/reviews/qtox/qtox/5406#-LPaCdph9CcGU1Fl0UQK:-LPaCdpj7nbq08gg0nDz:b-ryd7wk) ([raw file](https://github.com/qtox/qtox/blob/a28f1902366aca9e162d0694d761021a9984d436/src/audio/audio.cpp#L203)):*
> ```cpp
> }
> 
> QString Audio::getSoundPath(QString fn)
> ```

can you align this code for the path structure with the one from https://github.com/qTox/qTox/blob/master/src/widget/style.cpp

so that we have the possibility to have different sound themes. I think `<AppDataPath/audio/default/<...>` would be a good structure.

Chiitoo:
Just wanted to say thanks for this.  It's one of those things I've been wanting to do for years... but never got to.  :]

Definitely agree with the path.  Currently on Linux, it would be `"$HOME"/.config/tox/audio/`, which is certainly not optimal, and this would definitely go well with what the themes will be/are already doing.

Inari-Whitebear:
So this is pretty old now, any idea if this will still go anywhere? Would be a nice feature.

If not, what are the rules for picking up dead PRs? Do I use the existing PR and modify it? Do I make my own code that does the same change?

anthonybilinski:
Making changes on top of the existing branch and creating a new PR would be good. It'll need to be rebased up to tip since there are some conflicts.

PR #5415, opened 2018-10-30
fix:(chatform): Issue 5115, use QSharedPointer for groupChatForms

PTroester:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5415)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -525,10 +525,6 @@ Widget::~Widget()
         delete form;
     }
 
-    for (auto form : groupChatForms) {
-        delete form;
-    }
-
     delete icon;
     delete profileForm;
     delete profileInfo;
@@ -1155,7 +1151,7 @@ void Widget::openDialog(GenericChatroomWidget* widget, bool newWindow)
         form = chatForms[id];
     } else {
         id = group->getId();
-        form = groupChatForms[id];
+        form = groupChatForms[id].data();
     }
 
     bool chatFormIsSet;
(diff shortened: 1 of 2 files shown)


anthonybilinski:
Is there a reason here to use a shared pointer over a unique pointer? Unique pointers generally have a clearer ownership model and IMO should be preferred in cases where either can be used.

PTroester:
> Is there a reason here to use a shared pointer over a unique pointer? Unique pointers generally have a clearer ownership model and IMO should be preferred in cases where either can be used.

QMap requires a shared pointer, because it needs a copy-constructor.
[Issue 5115](https://github.com/qTox/qTox/issues/5115)

anthonybilinski:
Good point.

In general I think that
* If types are being used with Qt API, keep Qt types to avoid conversion
* If types are different than std equivalent, e.g. QString, use the more suited one (probably QString)
* otherwise favour std

So in because of that I think a std::map with a std::unique_ptr is better than a QMap with a QSharedPointer, but I'm not sure this is a project-wide position.

@sudden6 thoughts?

sudden6:
So far there is no project wide decision that I'm aware of, however I think in general the code in its current state prefers Qt types over std. At least for the UI focused components.

In this special case I see no advantage of using unique_ptr, since I don't think it should be possible to only have one pointer to the Form.

anthonybilinski:
>At least for the UI focused components

this seems reasonably for API compatibility reasons.

>no advantage of using unique_ptr, since I don't think it should be possible to only have one pointer to the Form

You can still have multiple pointers to a single element with `unique_ptr`, by passing out non-ownership raw pointers. Since these forms are `Widget` members and destructed by `Widget`, no one should have a reference to the forms after `Widget` has destructed, making `Widget` the sole owner and `unique_ptr` a good option. IMO `shared_ptr`s should rarely be used, only when you have true shared ownership of elements that anyone could have the last reference to in an unknown destruction order.

sudden6:
Yeah, I think Qt doesn't really have an answer to this specific problem. Even if the QSharedPointer maybe isn't the optimum in this case, it's still way better than a raw pointer.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5415#-:-LQB9JzO3hFb7hg0X1vV:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5417, opened 2018-10-30
feat(core): prepare qTox for groupchat saving

sudden6:
This change creates groups on startup of Core. We need this once
https://github.com/TokTok/c-toxcore/pull/1156 is merged to load existing
groups.

Marked as WIP, because I didn't have a chance to test it yet.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5417)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -250,6 +250,7 @@ void Core::onStarted()
     emit idSet(id);
 
     loadFriends();
+    loadGroups();
 
     process(); // starts its own timer
     av->start();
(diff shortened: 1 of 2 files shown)


sudden6:
I have tested this now in combination with https://github.com/TokTok/c-toxcore/pull/1156 and it seems to work as expected.

Ready for review.

src/core/core.cpp
```diff
@@ -1012,6 +1012,23 @@ void Core::loadFriends()
     delete[] ids;
 }
 
+void Core::loadGroups()
+{
+    QMutexLocker ml{coreLoopLock.get()};
+
+    const size_t groupCnt = tox_conference_get_chatlist_size(tox.get());
```
Diadlo: Come on. `Cnt` instead of `Count`. It's not the best idea :)

src/core/core.cpp
```diff
@@ -1012,6 +1012,23 @@ void Core::loadFriends()
...
+    if (groupCnt == 0) {
+        return;
+    }
+
+    uint32_t* groupIds = new uint32_t[groupCnt];
```
Diadlo: I think vector can be used.
```
std::vector<uint32_t> groupIds(groupCount);
tox_conference_get_chatlist(tox.get(), groupIds.data());
for (auto id : groupIds) {
    emit ...
}
```

Why not a raw array? Because you already forgot to deallocate memory :P

Diadlo (changes requested):

src/core/core.cpp
Diadlo: Or just delete `groupIds`

sudden6: Crap, force pushed the wrong stuff, I already had that fixed^^

src/core/core.cpp
sudden6: Done

Diadlo (approved):
:lgtm_strong:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5417#-:-LRuJ1OK0WUYePdXYgi6:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5410, opened 2018-10-24
Document file system layout for qTox


sudden6:
##### Observed Behavior

Currently qTox uses different methods to compute the paths for the `profile.ini` `profile.db` `qTox.ini` `profile.tox` and the themes folder path.

##### Expected Behavior

- qTox will use `QStandardPaths` for it's own directories and implement a class to abstract the paths for files defined in the Tox Client Standard.

- qTox will follow the freedesktop.org specifications on Linux and store data/config files in the appropriate locations

sudden6:
After some offline discussion with @sphaerophoria I propose the following filesystem layout changes for qTox:

- move the global `qtox.ini` to `QStandardPaths::AppConfigLocation`
- move `<profile>.db` and `<profile>.ini` to `QStandardPaths::AppDataLocation`
- default filetransfer directory will be `QStandardPaths::AppDataLocation/transfers/<PubKey>/`

- the `<profile>.tox` file and the avatars will stay at `~/.config/share/tox/` because this location is given by the TCS in https://tox.gitbooks.io/tox-client-standard/content/data_storage/data_storage.html and https://tox.gitbooks.io/tox-client-standard/content/data_storage/export_format.html

on Linux this would result in the following paths:
- `~/.config/qTox/qtox.ini`
- `~/.local/share/qTox/transfers/<PubKey>`
- `~/.local/share/qTox/<profile>.db`
- `~/.local/share/qTox/<profile>.ini`

The main reason for this change is that currently there are some hardcoded paths for Linux that don't respect https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html. Additionally with that change I want to prevent a mess because there are slightly different implementations inside qTox on how to access these paths and centralize it.

sudden6:
@anthonybilinski @Diadlo @sphaerophoria do you agree with that change?

sphaerophoria:
that sounds reasonable, only possible change I would make is put transfers in AppLocalDataLocation. I'm not sure what the normal policy is for storing lots of files in the roaming folder on windows is.

*However* I'm not entirely convinced that the value for moving the files is worth the pain of having to support old locations forever. We could also follow freedesktop spec using tox as our appname instead of qTox just to avoid having to move anything and support the old locations for the rest of time.

sudden6:
> that sounds reasonable, only possible change I would make is put transfers in AppLocalDataLocation. I'm not sure what the normal policy is for storing lots of files in the roaming folder on windows is.

[...]

PR #5431, opened 2018-11-11
add upgrade code for persistence layer upgrades

sudden6:

This is the rest of the changes for finishing #5410

These are not yet final so don't waste your time reviewing it.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5431)


sudden6:
Based on #5440 this integrates the upgrade code to move the persistent data to the new places.

Once merged, this fixes #5410

src/persistence/upgrader/persistenceupgrader.cpp
```diff
@@ -0,0 +1,194 @@
...
+    s.setIniCodec("UTF-8");
+
+    return s.value("Version", "1.16.3").toString();
+}
+
+bool PersistenceUpgrader::doUpgrade()
+{
+    assert(!curVersion.isEmpty());
+
+    if(curVersion < "1.17.0" && !upgradeFrom_1_16_3()) {
```
Diadlo: Version strings can be compared wrong. `"10.1" < "1.2"`

Diadlo:
I suggest instead of hardcored `if(curVersion < "1.17.0" && !upgradeFrom_1_16_3())` + `if(curVersion < "1.18.0" && !upgradeFrom_1_17_0())` etc. use something like this
```
struct Version {
    int major, minor, patch;
    bool operator<(const Version& other);
};

class IVersionUpdater {
    Version getVersion() virtual = 0;
    bool apply() virtual = 0;
};

class VersionUpdater_1_16_3 : IVersionUpdater {
    Version getVersion() { return {1, 16, 3}; }
    bool apply() { /* something */ }
};

std::vector<IVersionUpdater> PersistenceUpgrader::updates = {
    VersionUpdater_1_16_3(),
    VersionUpdater_1_17(), 
    /* ... */
};

bool PersistenceUpgrader::doUpdate() {
    for (auto update : updates) {
        if (update.getVersion() < currentVersion) {
            if (!update.apply()) {
                return false;
            }
        }
    }
}
```

**ADDED:** Instead of abstract class `IVersionUpdater` with extra memory allocation etc., you can use
```
struct VersionUpdater {
    Version version;
    std::function<void()> apply;
};


std::vector<VersionUpdater> PersistenceUpgrader::updates = {
    { {1, 16, 3}, apply_from_1_16_3 },
    { {1, 17, 0}, apply_from_1_17_0 },
    /* ... */
};
```

Diadlo:
Or even better use not version, but incremental number. Otherwise you can't do two persistence update in one release. In this case you just need to apply `PersistenceUpgrader::updates` from `currentId` until the end

src/persistence/upgrader/persistenceupgrader.cpp
sudden6: I suspected this solution is to simple to be good -_-

sudden6:
Thanks for the feedback, I'll rework this and then update it.

PR #5448, opened 2018-12-02
feat(core): set group title for loaded groups

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -44,6 +44,26 @@ const QString Core::TOX_EXT = ".tox";
 
 #define ASSERT_CORE_THREAD assert(QThread::currentThread() == coreThread.get())
 
+namespace {
+    bool LogConferenceTitleError(TOX_ERR_CONFERENCE_TITLE error)
+    {
+        switch(error)
+        {
+        case TOX_ERR_CONFERENCE_TITLE_OK:
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND:
+            qWarning() << "Conference title not found";
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_INVALID_LENGTH:
+            qWarning() << "Invalid conference title length";
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_FAIL_SEND:
+            qWarning() << "Failed to send title packet";
+        }
+        return error;
+    }
+} // namespace
+
 Core::Core(QThread* coreThread)
     : tox(nullptr)
     , av(nullptr)
@@ -451,8 +471,7 @@ void Core::onGroupInvite(Tox* tox, uint32_t friendId, Tox_Conference_Type type,
                          const uint8_t* cookie, size_t length, void* vCore)
 {
     Core* core = static_cast<Core*>(vCore);
-    // static_cast is used twice to replace using unsafe reinterpret_cast
-    const QByteArray data(static_cast<const char*>(static_cast<const void*>(cookie)), length);
+    const QByteArray data(reinterpret_cast<const char*>(cookie), length);
     const GroupInvite inviteInfo(friendId, type, data);
     switch (type) {
     case TOX_CONFERENCE_TYPE_TEXT:
@@ -1025,7 +1044,18 @@ void Core::loadGroups()
     tox_conference_get_chatlist(tox.get(), groupIds);
 
     for(size_t i = 0; i < groupCount; ++i) {
-        emit emptyGroupCreated(static_cast<int>(groupIds[i]));
+        TOX_ERR_CONFERENCE_TITLE error;
+        size_t titleSize = tox_conference_get_title_size(tox.get(), groupIds[i], &error);
+        if (LogConferenceTitleError(error)) {
+            continue;
+        }
+
+        QByteArray name(titleSize, Qt::Uninitialized);
+        if (!tox_conference_get_title(tox.get(), groupIds[i], reinterpret_cast<uint8_t*>(name.data()), &error))
+        if (LogConferenceTitleError(error)) {
+            continue;
+        }
+        emit emptyGroupCreated(static_cast<int>(groupIds[i]), ToxString(name).getQString());
     }
 
     delete[] groupIds;
@@ -1114,7 +1144,7 @@ QString Core::getGroupPeerName(int groupId, int peerId) const
     }
 
     QByteArray name(length, Qt::Uninitialized);
-    uint8_t* namePtr = static_cast<uint8_t*>(static_cast<void*>(name.data()));
+    uint8_t* namePtr = reinterpret_cast<uint8_t*>(name.data());
     bool success = tox_conference_peer_get_name(tox.get(), groupId, peerId, namePtr, &error);
     if (!parsePeerQueryError(error) || !success) {
         qWarning() << "getGroupPeerName: Unknown error";
@@ -1179,7 +1209,7 @@ QStringList Core::getGroupPeerNames(int groupId) const
         }
 
         QByteArray name(length, Qt::Uninitialized);
-        uint8_t* namePtr = static_cast<uint8_t*>(static_cast<void*>(name.data()));
+        uint8_t* namePtr = reinterpret_cast<uint8_t*>(name.data());
         bool ok = tox_conference_peer_get_name(tox.get(), groupId, i, namePtr, &error);
         if (ok && parsePeerQueryError(error)) {
             names.append(ToxString(name).getQString());
@@ -1267,7 +1297,7 @@ uint32_t Core::joinGroupchat(const GroupInvite& inviteInfo) const
     const uint32_t friendId = inviteInfo.getFriendId();
     const uint8_t confType = inviteInfo.getType();
     const QByteArray invite = inviteInfo.getInvite();
-    const uint8_t* const cookie = static_cast<const uint8_t*>(static_cast<const void*>(invite.data()));
+    const uint8_t* const cookie = reinterpret_cast<const uint8_t*>(invite.data());
     const size_t cookieLength = invite.length();
     switch (confType) {
     case TOX_CONFERENCE_TYPE_TEXT: {
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -190,7 +190,7 @@ signals:
     void friendRemoved(uint32_t friendId);
     void friendLastSeenChanged(uint32_t friendId, const QDateTime& dateTime);
 
-    void emptyGroupCreated(int groupnumber);
+    void emptyGroupCreated(int groupnumber, const QString& title = QString());
     void groupInviteReceived(const GroupInvite& inviteInfo);
     void groupMessageReceived(int groupnumber, int peernumber, const QString& message, bool isAction);
     void groupNamelistChanged(int groupnumber, int peernumber, uint8_t change);
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -1907,16 +1907,19 @@ Group* Widget::createGroup(int groupId)
     return newgroup;
 }
 
-void Widget::onEmptyGroupCreated(int groupId)
+void Widget::onEmptyGroupCreated(int groupId, const QString& title)
 {
     Group* group = createGroup(groupId);
     if (!group) {
         return;
     }
-
-    // Only rename group if groups are visible.
-    if (Widget::getInstance()->groupsVisible()) {
-        groupWidgets[groupId]->editName();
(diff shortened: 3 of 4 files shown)


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.h, line 194 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpoJnD2QJDEIW1uJK1:-LSpoJnD2QJDEIW1uJK2:bc9up70) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.h#L194)):*
> ```objc
> 
>     void emptyGroupCreated(int groupnumber);
>     void groupLoaded(int groupnumber, const QString& title);
> ```

`groupLoaded` seems to be quite similar to `emptyGroupCreated`, maybe extend it and use a default parameter for `title`?
___
*[src/core/core.cpp, line 1051 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpoptZ8xh8nqV8cCrP:-LSpopt_8Dmnz-746kAb:b-skfgnv) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1051)):*
> ```cpp
>         size_t titleSize = tox_conference_get_title_size(tox.get(), groupIds[i], &error);
>         if (LogConferenceTitleError(error))
>             continue;
> ```

braces
___
*[src/core/core.cpp, line 1053 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpp43JEbi3YK7HcLQg:-LSpp43JEbi3YK7HcLQh:b-fnvefu) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1053)):*
> ```cpp
>             continue;
> 
>         std::vector<uint8_t> name(titleSize);
> ```

I think `ToxString` would be better here
___
*[src/core/core.cpp, line 1055 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSporP87CPCJgfH1Iz2:-LSporP87CPCJgfH1Iz3:b-skfgnv) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1055)):*
> ```cpp
>         std::vector<uint8_t> name(titleSize);
>         if (!tox_conference_get_title(tox.get(), groupIds[i], name.data(), &error))
>         if (LogConferenceTitleError(error))
> ```

braces
___
*[src/widget/widget.cpp, line 1923 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSppK63BATEGCBY2RnC:-LSppK63BATEGCBY2RnD:b5kuqgq) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/widget/widget.cpp#L1923)):*
> ```cpp
> }
> 
> void Widget::onGroupLoaded(int groupId, const QString& title)
> ```

same as above, possibly merge with `onEmptyGroupCreated`

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.h, line 194 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpoJnD2QJDEIW1uJK1:-LSqSZY0-4xCsyMHYo-v:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.h#L194)):*

`groupLoaded` seems to be quite similar to `emptyGroupCreated`, maybe extend it and use a default parameter for `title`?
</blockquote></details>

Done.
___
*[src/core/core.cpp, line 1051 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpoptZ8xh8nqV8cCrP:-LSqS_ISBfapzgxoKgc2:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1051)):*

braces
</blockquote></details>

Done.
___
*[src/core/core.cpp, line 1053 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpp43JEbi3YK7HcLQg:-LSqS_eA-5zunONeeHcD:b-kijqxx) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1053)):*

I think `ToxString` would be better here
</blockquote></details>

Done, constructed ToxString to get QString from byte array. Added constructor for std::vector<uint8_t> instead of using QByteArray with the existing constructor due to my preference of using std types when not interfacing with Qt API.
___
*[src/core/core.cpp, line 1055 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSporP87CPCJgfH1Iz2:-LSqSl3K5D5z2muTtAY9:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1055)):*

braces
</blockquote></details>

Done.
___
*[src/widget/widget.cpp, line 1923 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSppK63BATEGCBY2RnC:-LSqSngQCCq_HxqanciW:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/widget/widget.cpp#L1923)):*

same as above, possibly merge with `onEmptyGroupCreated`
</blockquote></details>

Done.

sudden6 (changes requested):
> Reviewed 2 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1053 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpp43JEbi3YK7HcLQg:-LSq_e0yAq_BA4Fc5jrx:bbbqidx) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1053)):*

Done, constructed ToxString to get QString from byte array. Added constructor for std::vector<uint8_t> instead of using QByteArray with the existing constructor due to my preference of using std types when not interfacing with Qt API.
</blockquote></details>

I see no advantage of using the std type here, also the rest of the code in this file mostly uses `QByteArray` so IMO it's better to stick to it.
___
*[src/core/toxstring.cpp, line 56 at r2](https://reviewable.io/reviews/qtox/qtox/5448#-LSq_IkT8Rr58RJ3wfU0:-LSq_IkT8Rr58RJ3wfU1:b6y6t6g) ([raw file](https://github.com/qtox/qtox/blob/04fd60695a28c1d3d3ab6367e8ebecbe639dd9a5/src/core/toxstring.cpp#L56)):*
> ```cpp
>  * @param text Input text bytes.
>  */
> ToxString::ToxString(const std::vector<uint8_t>& text)
> ```

If you persist with adding this constructor, please also extend the tests covering this class.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1053 at r1](https://reviewable.io/reviews/qtox/qtox/5448#-LSpp43JEbi3YK7HcLQg:-LSqtgKA9qgF4jXTfQJA:bwub2xj) ([raw file](https://github.com/qtox/qtox/blob/404cb84a100600a308eb19bdf6dc0fbbdd5e2bdc/src/core/core.cpp#L1053)):*

I see no advantage of using the std type here, also the rest of the code in this file mostly uses `QByteArray` so IMO it's better to stick to it.
</blockquote></details>

My logic was that std was a better default and either would function well. You pointed out (in chat) that the conversion from QByteArray -> QString is already handled and tested, plus the addition of a new constructor API adds complexity, which I agree with. Changed to QByteArray. You also made some good counters to my view of defaulting to std where possible which I'll think about some more.
___
*[src/core/core.cpp, line 455 at r3](https://reviewable.io/reviews/qtox/qtox/5448#-LSquDPy7tk8ee5wOVY3:-LSquDPz4n9MXphRh3vj:b9bx597) ([raw file](https://github.com/qtox/qtox/blob/f188409b8ce85515fc3b45a2aedb3af79af2b009/src/core/core.cpp#L455)):*
> ```cpp
>     Core* core = static_cast<Core*>(vCore);
>     // static_cast is used twice to replace using unsafe reinterpret_cast
>     const QByteArray data(static_cast<const char*>(static_cast<const void*>(cookie)), length);
> ```

I don't think this logic makes any sense. Double static cast is as unsafe in that the underlying bytes lose meaning in the new type and accomplishes the same thing. Changed here and across core.cpp where double static cast of pointers was used.
___
*[src/core/toxstring.cpp, line 56 at r2](https://reviewable.io/reviews/qtox/qtox/5448#-LSq_IkT8Rr58RJ3wfU0:-LSqtzgc6LVeDrBTrYxo:b-lhbv3a) ([raw file](https://github.com/qtox/qtox/blob/04fd60695a28c1d3d3ab6367e8ebecbe639dd9a5/src/core/toxstring.cpp#L56)):*

If you persist with adding this constructor, please also extend the tests covering this class.
</blockquote></details>

Removed so no longer relevant.

Diadlo:
@anthonybilinski I prefer not to use `reinterpret_cast` because of this:
```
$ cat static_cast.cpp
int main()
{
  unsigned char* uptr = nullptr;
  char* ptr = static_cast<char*>(static_cast<void*>(uptr));
  int i = static_cast<int>(uptr);
}
```
```
$ g++ -m32 static_cast.cpp
static_cast.cpp: In function ‘int main()’:
static_cast.cpp:5:32: error: invalid static_cast from type ‘unsigned char*’ to type ‘int’
   int i = static_cast<int>(uptr);
                                ^
```
```
$ cat reiterpret_cast.cpp 
int main()
{
  unsigned char* uptr = nullptr;
  char* ptr = reinterpret_cast<char*>(uptr);
  int i = reinterpret_cast<int>(uptr);
}
```
```
$ g++ -m32 reiterpret_cast.cpp
# No error
```

anthonybilinski:
casting a pointer to an int is meaningless because the data in incompatible, so you get an error with static_cast (useful, good in normal cases). `reinterpret_cast` is saying to ignore if the data is incomparable and just use the data. This is exactly what we're trying to do when we double `static_cast`. The whole reason we have two is to trick the compiler into not realizing that we're converting `unsigned char` to `char` by first converting to `void*` so that it can't tell what the conversion is. If the data was compatible we would only need a single `static_cast` and that would be preferable, but doing that here causes the same warning that `unsigned char*` and `char*` have incompatible data. Either with `reinterpret_cast` or double `static_cast` we're forcing it to use the data anyway, but `reinterpret_cast` is explicitly saying that that's what we want, but double static_cast is just working around the warning.

Diadlo:
@anthonybilinski You can't cast pointer to non-pointer with two `static_cast`

sudden6 (commented):
> Reviewed 1 of 7 files at r2, 1 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448)* status: 0 of 1 LGTMs obtained

anthonybilinski:
@Diadlo true, but the fact remains that we're _trying_ to reinterpret the data here, which is the textbook use case of `reinterpret_cast`, we're not trying to implicitly cast data between compatible types using `static_cast`. The whole reason we need two `static_cast`s in a row is because the data is not implicitly castable.

sudden6:
@Diadlo I think @anthonybilinski is correct here, that we're trying to reinterpret data and we should express that using `reinterpret_cast`.

anthonybilinski:
Is this pending changes, or good the way it is? It would be nice to get this in before v1.17.0.

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5448#-:-LTqhBgSDr9twbvmIhTt:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5454, opened 2018-12-07
WIP: Group history

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5454)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ qt5_wrap_ui(${PROJECT_NAME}_FORMS
   src/widget/about/aboutfriendform.ui
   src/widget/form/loadhistorydialog.ui
   src/widget/form/profileform.ui
-  src/widget/form/removefrienddialog.ui
+  src/widget/form/removecontactdialog.ui
   src/widget/form/searchsettingsform.ui
   src/widget/form/setpassworddialog.ui
   src/widget/form/settings/aboutsettings.ui
@@ -476,8 +476,8 @@ set(${PROJECT_NAME}_SOURCES
   src/widget/tool/profileimporter.h
   src/widget/tool/recursivesignalblocker.cpp
   src/widget/tool/recursivesignalblocker.h
-  src/widget/tool/removefrienddialog.cpp
-  src/widget/tool/removefrienddialog.h
+  src/widget/tool/removecontactdialog.cpp
+  src/widget/tool/removecontactdialog.h
   src/widget/tool/screengrabberchooserrectitem.cpp
   src/widget/tool/screengrabberchooserrectitem.h
   src/widget/tool/screengrabberoverlayitem.cpp
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -44,6 +44,26 @@ const QString Core::TOX_EXT = ".tox";
 
 #define ASSERT_CORE_THREAD assert(QThread::currentThread() == coreThread.get())
 
+namespace {
+    bool LogConferenceTitleError(TOX_ERR_CONFERENCE_TITLE error)
+    {
+        switch(error)
+        {
+        case TOX_ERR_CONFERENCE_TITLE_OK:
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND:
+            qWarning() << "Conference title not found";
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_INVALID_LENGTH:
+            qWarning() << "Invalid conference title length";
+            break;
+        case TOX_ERR_CONFERENCE_TITLE_FAIL_SEND:
+            qWarning() << "Failed to send title packet";
+        }
+        return error;
+    }
+} // namespace
+
 Core::Core(QThread* coreThread)
     : tox(nullptr)
     , av(nullptr)
@@ -451,8 +471,7 @@ void Core::onGroupInvite(Tox* tox, uint32_t friendId, Tox_Conference_Type type,
                          const uint8_t* cookie, size_t length, void* vCore)
 {
     Core* core = static_cast<Core*>(vCore);
-    // static_cast is used twice to replace using unsafe reinterpret_cast
-    const QByteArray data(static_cast<const char*>(static_cast<const void*>(cookie)), length);
+    const QByteArray data(reinterpret_cast<const char*>(cookie), length);
     const GroupInvite inviteInfo(friendId, type, data);
     switch (type) {
     case TOX_CONFERENCE_TYPE_TEXT:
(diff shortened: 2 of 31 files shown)


anthonybilinski:
This is ready to be reviewed, but shouldn't be merged until https://github.com/TokTok/c-toxcore/pull/1267 is merged, since current Toxcore persistent groups can corrupt Tox profiles. We should probably also bump minimum supported Toxcore version once persistent groups are released, or else users will have inaccessible group history saved for each group they join.

sudden6 (changes requested):
Some of your files have a strange symbol as last line, it's visible on reviewable.

> Reviewed 1 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5454#-:-LTHdqcI6vnlNa0AxOy7:btu4kel)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxpk.h, line 35 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHbJXM2ozVpw2UqDDb:-LTHbJXM2ozVpw2UqDDc:b3ku1y3) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/core/toxpk.h#L35)):*
> ```objc
> inline uint qHash(const ToxPk& key)
> {
>     return qHash(key.toString());
> ```

`qHash` can work on the `QByteArray` directly, no need to to convert to `QString`. Please also extend the test cases for this class to cover the new function.
___
*[src/model/contact.h, line 37 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHc0-F0oqigVpRiIVG:-LTHc0-F0oqigVpRiIVH:b-dti5ws) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/model/contact.h#L37)):*
> ```objc
>     virtual QString getDisplayedName() const = 0;
>     virtual uint32_t getId() const = 0;
>     virtual const ToxPk getPersistentId() const = 0;
> ```

Since technically the group persistent id is not a ToxPk, what do you think about introducing some sort of unique id interface and implementing it in `ToxPk` as well as an additional type for the group persistent id?
___
*[src/widget/widget.cpp, line 2525 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHcuJIDZGbjVyqn15Q:-LTHcuJIDZGbjVyqn15R:b-jbbis9) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/widget/widget.cpp#L2525)):*
> ```cpp
> }
> 
> void Widget::removeContactHistory(Contact* c)
> ```

pass `c` by reference? or else include some `nullptr` check

anthonybilinski (commented):
> Some of your files have a strange symbol as last line, it's visible on reviewable.

I accidentally  removed the trailing newline  on a couple files, should be fixed now.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5454#-:-LTI9Z__-9Gpb6iLzc3e:b-gfno4p)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxpk.h, line 35 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHbJXM2ozVpw2UqDDb:-LTI26h27S_NJzIQdIAl:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/core/toxpk.h#L35)):*

`qHash` can work on the `QByteArray` directly, no need to to convert to `QString`. Please also extend the test cases for this class to cover the new function.
</blockquote></details>

Done.
___
*[src/model/contact.h, line 37 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHc0-F0oqigVpRiIVG:-LTI27yNAGqMlGZjcb7g:b-nqi5tx) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/model/contact.h#L37)):*

Since technically the group persistent id is not a ToxPk, what do you think about introducing some sort of unique id interface and implementing it in `ToxPk` as well as an additional type for the group persistent id?
</blockquote></details>

I think it makes sense to not use the term Pk with groups. I'm not sure a second class is needed instead of just renaming ToxPk to something more generic like "ContactId" which could still be used for both friends and groups. With naming, I'm having a hard time distinguishing the 32-byte "ID"s for friends and groups from the ToxID's (including nospam and checksum) from the friend/group ephemeral int IDs from toxcore. Toxcore uses "ID" to mean persistent 32-byte ID and "number" to mean instance friend/group int identifier. Maybe we should adopt that naming across qTox?
___
*[src/widget/widget.cpp, line 2525 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHcuJIDZGbjVyqn15Q:-LTI2_HL-R_-AhAOQijx:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/widget/widget.cpp#L2525)):*

pass `c` by reference? or else include some `nullptr` check
</blockquote></details>

Done.

sudden6 (changes requested):
> Reviewed 4 of 26 files at r1, 3 of 13 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5454)* status: 0 of 1 LGTMs obtained

___
*[src/model/contact.h, line 37 at r1](https://reviewable.io/reviews/qtox/qtox/5454#-LTHc0-F0oqigVpRiIVG:-LTIIrXB3MnmQxb9IgHS:bo35jz1) ([raw file](https://github.com/qtox/qtox/blob/737ad973aa82e7f592bbad1edfb93f9b70e808be/src/model/contact.h#L37)):*

I think it makes sense to not use the term Pk with groups. I'm not sure a second class is needed instead of just renaming ToxPk to something more generic like "ContactId" which could still be used for both friends and groups. With naming, I'm having a hard time distinguishing the 32-byte "ID"s for friends and groups from the ToxID's (including nospam and checksum) from the friend/group ephemeral int IDs from toxcore. Toxcore uses "ID" to mean persistent 32-byte ID and "number" to mean instance friend/group int identifier. Maybe we should adopt that naming across qTox?
</blockquote></details>

I'm not sure what the persistent id actually is, but `ToxPk` was originally only meant to be the public key part of a `ToxID`. I think this is actually a very sharp definition and should not be mixed with other stuff that is some other 32 bytes data by chance.

Another reason I don't really like reusing ToxPk here is, that we pull in a hard dependency on `src/core/toxpk.h` throughout the code. If some behavior of it ever changes (e.g. Tox uses a different key lenght) we have a major problem.

Yeah, toxcore is not very consistent with naming in that regard. I'd be happiest if we didn't have to deal with the friend/group int identifier outside Core at all.

anthonybilinski:
I'm working on some refactoring here and scope is starting to grow quite a bit. I've changed use of ToxPk to have both ToxPk and GroupId which both inherit from a ContactId interface like you suggested. ContactId has a getType which History can use, to put GroupIds and ToxPks into their own tables. It shouldn't be used outside of History (as it is an interface that should be treated generically by all users), so I'm considering using the passkey idiom to prevent access.

New DB changes since current state are that there is a `groups` table to match the `peers` table but with a "group id" instead of "tox pk" field, and then a new `contacts` table to manage the unique ID required between `peers` and `groups` so that the union of both doesn't result in duplicate IDs used by `history`. Since this is also effecting DB schema, I'll block it on @sphaerophoria's DB changes, to avoid creating more conflicts for that, and since this PR is already blocked on Toxcore's persistent groupchats.

Monsterovich:
Are we going to merge this already?

keyandthegate:
What would it take for a new contributor to pick this up?

anthonybilinski:
Replaced by https://github.com/qTox/qTox/pull/6561, which is ready for review.

PR #5457, opened 2018-12-09
refactor: edit default theme

TriKriSta:

The current theme is badly displayed on my pc.
So I edited it.

before:
screenshot_20181209_162450
screenshot_20181209_162556

after:
screenshot_20181209_174908
screenshot_20181209_174958

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5457)


Chiitoo:
What version of qTox, and which OS is that on?

I can't seem to be able to see what you see on Linux (current git master).

Makes me think there are theme files under the user paths, that are being used instead of defaults.  Else it definitely seems broken, though I can't reproduce.

TriKriSta:
I use qTox version v1.16.3-189-ge80dbe2d.
Arch Linux, KDE, theme Breeze Dark

sudden6:
I see a similar problem with Breeze Dark.

@TriKriSta the problem I see with your theme is, that it seems to only work around the problem. ie. that qTox doesn't integrate with the desktop color scheme.

TriKriSta:
I think, qTox doesn't integrate with the desktop color scheme, because qTox use self scheme, but this scheme override not all colors and because of this, problems arise

sudden6:
Hmm, maybe it is the better way to do set this as in your PR until we can select different themes.

@anthonybilinski what do you think?

TriKriSta:
Oh, LoginScreen looks bad in the light scheme. I'll think about how to do it better.

sudden6 (changes requested):
> Reviewed 2 of 4 files at r2, 2 of 5 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5457)* status: 0 of 1 LGTMs obtained

___
*[res.qrc, line 48 at r3](https://reviewable.io/reviews/qtox/qtox/5457#-LTqlEg379_UH4DDeC1G:-LTqlEg379_UH4DDeC1H:b-jrld4r) ([raw file](https://github.com/qtox/qtox/blob/147cec3312088d6d8c6407c8e5d0c40ab681f4e4/res.qrc#L48)):*
> ```
>         <file>img/transfer.svg</file>
>         <file>themes/default/fileTransferWidget/fileDone.svg</file>
> 	<file>themes/default/centralWidget/centralWidget.css</file>
> ```

use spaces here

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5457#-:-LTsUxEJ5F0FDuJR2BLW:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5466, opened 2018-12-21
fix(widget): always force show window on activate event

### LINKED ISSUE #5459: qTox on Windows and FreeBSD doesn't bring up the running instance
When one qTox process already runs, launching qtox.exe quits immediately and doesn't do anything.

It should find the running instance and bring it up.

sudden6:
fixes #5459

I tested this on KDE, where it still works without any regression.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5466)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -81,10 +81,12 @@
 bool toxActivateEventHandler(const QByteArray&)
 {
     Widget* widget = Nexus::getDesktopGUI();
-    if (!widget)
+    if (!widget) {
         return true;
-    if (!widget->isActiveWindow())
-        widget->forceShow();
+    }
+
+    qDebug() << "Handling [activate] event from other instance";
+    widget->forceShow();
 
     return true;
 }


sudden6:
@anthonybilinski can you review?

anthonybilinski (commented):
Do you have a root cause on this? I guess what's happening is that on FreeBSD, the window is active even though it's hidden, skipping the forceShow which both activates and shows the window, but on other OSes, the window isn't considered active when hidden, hitting our check that then both shows and activates it.  I don't see anything in docs or online indicating this would be different for FreeBSD, but if my understand sounds reasonable, :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5466#-:-LVrHUlb1o5Nb2INPQpo:b-ssqpes)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
Yeah, I guessed that this condition could behave differently on different window systems from the descriptions in the linked bug report. After I removed the condition I saw no regression on a system that worked before, so I removed it unconditionally.

anthonybilinski:
@sudden6 please rebase onto master to fix CI + merge when you can. Might as well get this into v1.17.0.

PR #5480, opened 2019-01-04
fix(core): this should resolve message handling in persistent groups

### LINKED ISSUE #5478: "Peer has joined group" message misses one peer
##### Brief Description
OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash:  aa83edf7599acafcfb546eb02d18834aeb6efa20
toxcore:  v0.2.8
Qt:  5.9.5

Reproducible: Sometimes

##### Steps to reproduce

1. Load persistent group with multiple members

Happened on 3/3 of the groups I reconnected to, two with 3 members and one with 6 members. Each time I was short exactly one name.

##### Observed Behavior
1. "A new user joined" message for each peer, but missing one name message.
image

##### Expected Behavior
1. Anytime there is a "A new user joined" message, it should be followed by a name message.

@Monsterovich in case you have something in mind.

Monsterovich:
This fixes: https://github.com/qTox/qTox/issues/5478

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5480)

### DIFF
--- a/src/model/group.cpp
+++ b/src/model/group.cpp
@@ -104,19 +104,27 @@ void Group::regeneratePeerList()
     for (int i = 0; i < nPeers; ++i) {
         const auto pk = core->getGroupPeerPk(groupId, i);
 
-        toxpks[pk] = peers[i];
-        if (toxpks[pk].isEmpty()) {
-            toxpks[pk] =
-                tr("<Empty>", "Placeholder when someone's name in a group chat is empty");
-        }
-
         Friend* f = FriendList::findFriend(pk);
         if (f != nullptr && f->hasAlias()) {
             toxpks[pk] = f->getDisplayedName();
+            empty_nick[pk] = false;
+            continue;
+        }
+
+        empty_nick[pk] = peers[i].isEmpty();
+        if (empty_nick[pk]) {
+            toxpks[pk] = tr("<Empty>", "Placeholder when someone's name in a group chat is empty");
+        } else {
+            toxpks[pk] = peers[i];
         }
     }
 
     emit userListChanged(groupId, toxpks);
+}
+
+bool Group::peerHasNickname(ToxPk pk)
+{
+    return !empty_nick[pk];
 }
 
 void Group::updateUsername(ToxPk pk, const QString newName)
--- a/src/model/group.h
+++ b/src/model/group.h
@@ -39,6 +39,7 @@ public:
     int getPeersCount() const;
     void regeneratePeerList();
     const QMap<ToxPk, QString>& getPeerList() const;
+    bool peerHasNickname(ToxPk pk);
 
     void setEventFlag(bool f) override;
     bool getEventFlag() const override;
@@ -67,6 +68,7 @@ private:
     QString selfName;
     QString title;
     QMap<ToxPk, QString> toxpks;
+    QMap<ToxPk, bool> empty_nick;
     bool hasNewMessages;
     bool userWasMentioned;
     int groupId;
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -377,7 +377,8 @@ ChatMessage::Ptr GenericChatForm::createMessage(const ToxPk& author, const QStri
 {
     const Core* core = Core::getInstance();
     bool isSelf = author == core->getSelfId().getPublicKey();
-    QString authorStr = isSelf ? core->getUsername() : resolveToxPk(author);
+    QString myNickName = core->getUsername().isEmpty() ? author.toString() : core->getUsername();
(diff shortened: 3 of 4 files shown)


sudden6 (changes requested):
IMO we should display the public key of the peer instead of `<Empty>`, like we do it in the friendslist. Maybe in a different color so nobody can fake a public key. What do you think?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5480#-:-LVSmqUr8CjPGD2FV-Z7:bscwni0)* status: 0 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/5480#-LVSmfxY32ZFu50AF455:-LVSmfxY32ZFu50AF456:bm3sv55) ([raw file](https://github.com/qtox/qtox/blob/bb8ff23a306abd800de1bc3d52fcbc9e55af9e3b/src/model/group.cpp#L124)):*
> ```cpp
> bool Group::peerHasNickname(ToxPk pk)
> {
>     return toxpks[pk] != emptyName;
> ```

I'm pretty sure this will cause problems if someone actually sets their nick to `<Empty>`, which is valid in Tox

Monsterovich:
>IMO we should display the public key of the peer instead of ""
>I'm pretty sure this will cause problems if someone actually sets their nick to <Empty>, which is valid in Tox

Done.

Monsterovich:
When a user connects to the group after you:
default
When you connect to the group after qtox restart:
default
If a user doesn't have a nick, it should show PK instead:
default
There was a minor issue: if your nick is empty, you just don't see it, I fixed it too:
qtox_image_2019-01-05 23-20-43 804

Monsterovich:
Can you review this PR?

sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 2 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5480#-:-LVZYiZiBZcGawsgIc17:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/5480#-LVSmfxY32ZFu50AF455:-LVZYaEQD4qRlU2wT3as:b-fw6rlo) ([raw file](https://github.com/qtox/qtox/blob/bb8ff23a306abd800de1bc3d52fcbc9e55af9e3b/src/model/group.cpp#L124)):*

I'm pretty sure this will cause problems if someone actually sets their nick to `<Empty>`, which is valid in Tox
</blockquote></details>

ok, this seems better

sudden6:
@Monsterovich I don't really have much time at the moment so it can take 1-2 days for me to react to changes.

Also to speed things up, please answer to my comments on reviewable, so I know what you changed and why.

sudden6:
@anthonybilinski IIRC you're doing some persistent groupchat testing, can you include this PR and merge it afterwards?

anthonybilinski (approved):
I like the PK being used in chat when the peer's name is empty, but it does make the group header and "is now known as" messages inconsistent since they both still use "\<Empty\>" instead of the PK like in chat. This doesn't seem blocking for this PR.

> Reviewed 2 of 2 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5480#-:-LVbonCP1lvDY3vkTHRd:b-nwg2gg)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/widget/form/groupchatform.cpp, line 337 at r3](https://reviewable.io/reviews/qtox/qtox/5480#-LVbn__X3bbnrRF5Ld8s:-LVbn__X3bbnrRF5Ld8t:b-6qf1pm) ([raw file](https://github.com/qtox/qtox/blob/72bcc6acaf22ef78c2db852a5cabf45f42e5af90/src/widget/form/groupchatform.cpp#L337)):*
> ```cpp
>                     continue;
>                 }
>                 addSystemInfoMessage(tr("A new user has connected to the group"), ChatMessage::INFO, QDateTime::currentDateTime());
> ```

If we were guaranteed to get the new member's name (currently we're not due to this bug: https://github.com/TokTok/c-toxcore/issues/865) I'd probably be in favour of removing the "new user" message since the "name has joined" will always follow and be more informative, but for the time being this is required for at least that reason.

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5480#-:-LVbpA4W9VBOBSSFM_rS:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

PR #5493, opened 2019-01-14
Perf/smileys

### LINKED ISSUE #5486: Emoji parsing uses too much CPU
##### Brief Description

OS: Arch Linux Arm, Linux 3.10.18
qTox version: v1.16.3+  
Commit hash: f2e8567b21b140ad1b459a2ff7dfc4e61ce66369
toxcore: 0.2.8
Qt: 5.12.0
Hardware: ARM Chromebook, armv7l 

Reproducible: Always

##### Steps to reproduce

1. Enable settings -> user interface -> use emoticons
2. Have about 50 friends, most with at least 100 messages (the default loaded on start)
3. Restart qTox

##### Observed Behavior
After entering password, about 3m20s of 100% CPU usage before qTox main window opens.

##### Expected Behavior
qTox to open within a couple seconds.

##### Additional Info
With perf, I've identified that almost all of the startup CPU usage is in libpcre2, used by Qt's regex. Disabling emoticons makes the problem go away almost completely, meaning the problem is isolated to `SmileyPack::smileyfied`, where we iterate over all of our emoticon keys (~100) for every possible start position in the passed in string, causing ~1000 regex expressions to be constructed for each message passed in.

Regex should probably be avoided completely here if we're going to try to match 100s of keys to every character spot of every message, even if it makes the start/end whitespace constraints easier.

sudden6:

fixes #5486

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5493)

### DIFF
--- a/src/persistence/smileypack.cpp
+++ b/src/persistence/smileypack.cpp
@@ -245,8 +245,36 @@ bool SmileyPack::load(const QString& filename)
         emoticons.append(emoticonList);
     }
 
+    constructRegex();
+
     loadingMutex.unlock();
     return true;
+}
+
+/**
+ * @brief Creates the regex for replacing emoticons with the path to their pictures
+ */
+void SmileyPack::constructRegex()
+{
+    QString allPattern = QStringLiteral("(");
+
+    // construct one big regex that matches on every emoticon
+    for (const QString& emote : emoticonToPath.keys()) {
+        if (emote.toUcs4().length() == 1) {
+            // UTF-8 emoji
+            allPattern = allPattern % emote;
+        } else {
+            // patterns like ":)" or ":smile:", don't match inside a word or else will hit punctuation and html tags
+            allPattern = allPattern % QStringLiteral(R"((?<=^|\s))") % QRegularExpression::escape(emote) % QStringLiteral(R"((?=$|\s))");
+        }
+        allPattern = allPattern % QStringLiteral("|");
+    }
+
+    allPattern[allPattern.size() - 1] = QChar(')');
+
+    // compile and optimize regex
+    smilify.setPattern(allPattern);
+    smilify.optimize();
 }
 
 /**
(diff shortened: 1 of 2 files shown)


sudden6:
@anthonybilinski I didn't actually setup perf for this, but I'm quite sure it's now faster than before. Maybe you can redo your tests with these patches.

anthonybilinski:
Unfortunately my slow laptop is won't start its display manager at the moment after an update, but testing on my desktop and with perf:
before change, consistently 10s from "login" to window
after change, consistently 5s from "login" to window
before change, emojis disabled, consistently 5s  from "login" to window

I'm missing symbols in my libs, but here's the perf run from before change, with emojis. The emoji parsing is the wide stack starting at _ZN14QMe.
image

After change, with emojis enabled, where you can see that call stack is much thinner:
image

Since on my ARM laptop before the problem went away completely disabling smileys, I'm guessing I'll see 3m40 -> 5s instead of just a halving of time on my desktop, I guess our other start-up times are less CPU bound.

Nice fix!

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5493#-:-LWHF_5n5p9G3LHTNIji:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/smileypack.h, line 48 at r1](https://reviewable.io/reviews/qtox/qtox/5493#-LWHCcBO06BnHeKRlxM3:-LWHCcBP5rOQtfkU9feY:b-6vbvoa) ([raw file](https://github.com/qtox/qtox/blob/9cd013895ad05e00ac47875aff1b53e9fda06bf1/src/persistence/smileypack.h#L48)):*
> ```objc
>     void onSmileyPackChanged();
>     void cleanupIconsCache();
>     void constructRegex();
> ```

why is this a slot rather than a private member function?

sudden6:
Thanks for the tests, with that reduction I'm now pretty confident it will do for your ARM Chromebook too. Else we can still implement the other things I mentioned in the Issue.

> why is this a slot rather than a private member function?

was a mistake, fixed it

PR #5516, opened 2019-01-24
fix(groups): don't invalidate all audio sources when peer list changes

### LINKED ISSUE #5508: Group audio call goes silent if someone joins or leaves the group
##### Brief Description

OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash:  f5aa07f5694a9172bb0bb6c4870cf5486a3460a7
toxcore:  v0.2.8
Qt:  5.11.1
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. A and B are in an audio group, both in call.
2. C joins the group (but not the call).

##### Observed Behavior
1. The audio call goes silent for A and B, but their clients still show them as in the call.
2. A and B leaving and rejoining the call makes audio functional again.
3. The same thing occurs if C leaves the group, even if not in the call.

##### Additional Info
This was originally reported to me as occurring when someone in the group crashes, so that case should be tested when this issue is fixed. It's possible the crash was just triggering the same leave group bug, or that the bug has gotten worse recently, also covering clean group leaves.

anthonybilinski:
Fix #5508

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5516)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -557,10 +557,6 @@ void Core::onGroupMessage(Tox*, uint32_t groupId, uint32_t peerId, Tox_Message_T
 void Core::onGroupPeerListChange(Tox*, uint32_t groupId, void* vCore)
 {
     const auto core = static_cast<Core*>(vCore);
-    if (core->getGroupAvEnabled(groupId)) {
-        CoreAV::invalidateGroupCallSources(groupId);
-    }
-
     qDebug() << QString("Group %1 peerlist changed").arg(groupId);
     emit core->groupPeerlistChanged(groupId);
 }
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -501,11 +501,11 @@ void CoreAV::groupCallCallback(void* tox, uint32_t group, uint32_t peer, const i
     }
 
     Audio& audio = Audio::getInstance();
-    if(!call.havePeer(peer)) {
-        call.addPeer(peer);
+    if(!call.havePeer(peerPk)) {
+        call.addPeer(peerPk);
     }
 
-    audio.playAudioBuffer(call.getAlSource(peer), data, samples, channels, sample_rate);
+    audio.playAudioBuffer(call.getAlSource(peerPk), data, samples, channels, sample_rate);
 }
 
 /**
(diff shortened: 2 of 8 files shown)


sudden6 (changes requested):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5516)* status: 0 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 208 at r1](https://reviewable.io/reviews/qtox/qtox/5516#-LWyma683qvIJAL8TWOt:-LWyma683qvIJAL8TWOu:blexmrf) ([raw file](https://github.com/qtox/qtox/blob/66d427893ff717957f2fcf59448b5ede36849418/src/model/group.cpp#L208)):*
> ```cpp
> {
>     auto departedPks = oldPks;
>     for (auto tIt = newPks.begin(); tIt != newPks.end(); ++tIt) {
> ```

I'd suggest to simplify this as follows (example):

```
for(const auto& pk: oldPks) {
    if(!newPks.contains(pk)) {
          core->getAv()->invalidateGroupCallPeerSource(groupId, pk);
    }
}
```

Also has the advantage of using the O(log(n)) lookup property of `QMap` instead of linear searching the list

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5516)* status: 0 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 208 at r1](https://reviewable.io/reviews/qtox/qtox/5516#-LWyma683qvIJAL8TWOt:-LWz8EeIC-lmV6uVMFS9:b-w6g2kl) ([raw file](https://github.com/qtox/qtox/blob/66d427893ff717957f2fcf59448b5ede36849418/src/model/group.cpp#L208)):*

I'd suggest to simplify this as follows (example):

```
for(const auto& pk: oldPks) {
    if(!newPks.contains(pk)) {
          core->getAv()->invalidateGroupCallPeerSource(groupId, pk);
    }
}
```

Also has the advantage of using the O(log(n)) lookup property of `QMap` instead of linear searching the list
         
</blockquote></details>

That's a lot nicer. Done.

sudden6 (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5516)* status: 0 of 1 LGTMs obtained

anthonybilinski:
@sudden6 is the review on this complete? (I see "approved" but not lgtm)

sudden6 (commented):
:lgtm_strong: yes, thanks for reminding me

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5516#-:-LXKYnmq2aHGVJgseSMn:bfszmbl)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5240, opened 2018-07-14
Refactor/coreav

sudden6:
The goal is to remove `Audio::getInstance`

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5240)

PR #5533, opened 2019-01-27
refactor: make ToxCall non-moveable

sudden6:
We don't need move functionality and the code for it is complex and
error prone.

This is a fraction of  #5240 and a bit reworked.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5533)

### DIFF
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -73,13 +73,13 @@
  * @brief Maps friend IDs to ToxFriendCall.
  * @note Need to use STL container here, because Qt containers need a copy constructor.
  */
-std::map<uint32_t, ToxFriendCall> CoreAV::calls;
+std::map<uint32_t, CoreAV::ToxFriendCallPtr> CoreAV::calls;
 
 /**
  * @brief Maps group IDs to ToxGroupCalls.
  * @note Need to use STL container here, because Qt containers need a copy constructor.
  */
-std::map<int, ToxGroupCall> CoreAV::groupCalls;
+std::map<int, CoreAV::ToxGroupCallPtr> CoreAV::groupCalls;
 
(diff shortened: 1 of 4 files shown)


anthonybilinski:
1. The current state of ToxCall handling a bunch of class state as well as custom registration/deregistration behaviour means that a move constructor must be custom, and also that a move constructor is difficult. We've had bugs before specifically related to the move constructor.

2. Deleting the move constructor seems bad, and somewhat anti-C++. Having the class always on the heap doesn't really cause problems for us, and the class is not performance critical, but it still seems wrong. This obviously isn't very strong justification, so between those two options, deleting the move constructor is probably best.

I think better than both is to move the custom registration/deregistration behaviour into subclasses that have no other state, say ScopedAudioInSubscription, which has an explicit move constructor handling just that behaviour. Then ToxCall holds these movable subclasses as well as its other class state. Then because all of ToxCall/ToxFriendCall/ToxGroupCall's members are movable (either POD or have explicit move constructors) move will be well formed by default, so no large explicit move constructors will be required, and destruction also won't have to have special behaviour to avoid double-unsubscribing.

What do you think of that?

sudden6:
I think this matches well with the rest of the audio refactoring, but I think it would be easier/better to first reduce the complexity with this PR and then reintroduce this as a feature.

anthonybilinski (approved):
I agree this is simpler, leaving making it more "correct" for later makes sense. :lgtm_strong: 

P.S. I rebased your conflicting branch since I've blocked review for over a month.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5533#-:-L_C7dkh3ePu2i1HxF8U:bpkx80j)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5545, opened 2019-02-02
Interface signals without inheriting from QObject

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5545)

### DIFF
--- a/src/audio/backend/alsink.h
+++ b/src/audio/backend/alsink.h
@@ -23,11 +23,12 @@
 #include <QMutex>
 #include <QObject>
 
+#include "src/model/interface.h"
 #include "src/audio/iaudiosink.h"
 
 class OpenAL;
 class QMutex;
-class AlSink : public IAudioSink
+class AlSink : public QObject, public IAudioSink
 {
     Q_OBJECT
 public:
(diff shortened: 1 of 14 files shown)


Diadlo:
Why did you remove headers from CMakeLists.txt?

anthonybilinski:
AFAIK the headers listed serve no purpose. The list of sources is only used for the [add_library](https://cmake.org/cmake/help/v3.5/command/add_library.html#id2) call to create our static lib, which "Adds a library target called <name> to be built from the **source** files listed in the command invocation" according to docs. Removing them seems to have no negative effect.

sudden6:
One difference I notice with that PR is, that qtcreator now shows the headers in a different location in the project view. Not sure if that matters (doesn't for me) or changes anything else.

Diadlo:
It isn't true. Some IDEs (at least Visual Studio) show project files only described in CMakeLists.txt. So if you remove it users will not able to open headers with VS

anthonybilinski:
Ok good point. I've updated the PR to not remove headers from the sources list, but the other work I did to enable removing the headers, namely using our `interface.h` macros for interface signals without inheriting from QObject, still seems correct, so I've updated the PR to only contain that part.

This would allow us to compile without the headers in the sources list, if wanted.

anthonybilinski:
Closed and re-opened PR to regenerate merge for travis check, so fixed travis config on the target branch can be used.

sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5545#-:-M-S5j5G2x1fTuL2uQF4:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6 (approved):
> Reviewed 12 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5545)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5552, opened 2019-02-12
[WIP] refactor: use non-deprecated functions to fix build with Qt 5.13

Chiitoo:
With the current build options, the deprecated warnings lead to a build failure, which this commit fixes (should the commits be split?).

I'm not sure everything is correct, but I haven't bumped into any major issues (yet).

The 'drawRoundedRect' works a bit differently compared to 'drawRoundRect', so that's probably something to look into.  It makes the filetransferwidget look a bit... funny, which actually is the main reason for why I didn't use the new functions already back when I first touched those parts a while ago. [1]

One major thing to note here though, is that some of the functions I've used are definitely not available in Qt 5.5.0, which is currently listed as the minimum Qt version for qTox.  Some of them were implemented as late as 5.10.  Actually, I think one might be only in 5.13... so this will definitely need more attention (I mainly used what the compiler suggested to be used, without, at first, really considering when the suggested things first appeared).

How ancient Qt should qTox support anyblue?  I'm not too familiar with what distributions other than Gentoo provide.

Should there be conditional code for different versions, instead of outright 'upgrading' some of these things at this time?

1. https://github.com/qTox/qTox/commit/9322f29e

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5552)

### DIFF
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -225,12 +225,12 @@ void FileTransferWidget::paintEvent(QPaintEvent*)
     // Draw the widget background:
     painter.setClipRect(QRect(0, 0, width(), height()));
     painter.setBrush(QBrush(backgroundColor));
-    painter.drawRoundRect(geometry(), r * ratio, r);
+    painter.drawRoundedRect(geometry(), r * ratio, r);
 
     if (drawButtonAreaNeeded()) {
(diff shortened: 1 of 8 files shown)


anthonybilinski:
#5745 is a superset of this PR and was merged. Closing. Thanks for your initial work - it was useful for completing the task.

PR #5555, opened 2019-02-15
feat(ui): Added more preferences for chat window

Monsterovich:
default

That's why I added this: 
Reason 1: default
Reason 2: Coding like "#define YOUCANTCHANGETHISSETTINGLOL 100.0" is a bad style.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5555)


sudden6:
I don't really like the idea of introducing an additional setting for the width of these columns. I think the better strategy would be to automatically resize the columns based on their content, up to a defined maximum width (maybe bigger than now).

I think the following strategy would be appropriate: Assign a maximum percentage of the available width to each column like: nickname 14%, message 66%, date 10%, if any column needs less, the message column takes the free space.

> Reason 2: Coding like "#define YOUCANTCHANGETHISSETTINGLOL 100.0" is a bad style.

You're right that this define is bad style, but only because C++ offers `const` and anonymous namespaces which would be a better way to store this constants in a typesafe way.

Monsterovich:
>I don't really like the idea of introducing an additional setting for the width of these columns. I think the better strategy would be to automatically resize the columns based on their content, up to a defined maximum width (maybe bigger than now).

What if someone would take this nickname: "lolmynicknameisgoddamnlongijustbrokeyouruienjoyandhaveaniceday" or just leave it empty and the nickname will be replaced by a long tox peer key/public key, whatever. You can't decide what preferences would be better for a user. 

By the way, I wanted to implement an ability to move those columns with the mouse, but it seems very difficult because qTox uses custom widgets for chat. Changing two digits was easier than this.

If you don't want to have these variables in user settings, we can remove them and make them accessible only from qTox.ini. Anything is better than making them unchangeable.

sudden6:
> What if someone would take this nickname: "lolmynicknameisgoddamnlongijustbrokeyouruienjoyandhaveaniceday" or just leave it empty and the nickname will be replaced by a long tox peer key/public key, whatever. You can't decide what preferences would be better for a user.

I think it would be a very bad idea to dedicate so much space in the chat view to the username, it's not the most important thing in this view, the message is.

Another idea I just had would be to replace the user name with an icon, displaying "incoming message" and "outgoing message" for 1:1 chat.

PR #5636, opened 2019-04-23
Move Status out of Widget, make asset path not dependent on translation

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5636)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,6 +323,8 @@ set(${PROJECT_NAME}_SOURCES
   src/model/groupinvite.h
   src/model/group.cpp
   src/model/group.h
+  src/model/status.cpp
+  src/model/status.h
   src/model/interface.h
   src/model/profile/iprofileinfo.h
   src/model/profile/profileinfo.cpp
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -27,6 +27,7 @@
 #include "src/core/toxoptions.h"
 #include "src/core/toxstring.h"
 #include "src/model/groupinvite.h"
+#include "src/model/status.h"
 #include "src/nexus.h"
 #include "src/net/bootstrapnodeupdater.h"
 #include "src/persistence/profile.h"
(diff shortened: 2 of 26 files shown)


sudden6 (changes requested):
> Reviewed 25 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5636)* status: 0 of 1 LGTMs obtained

___
*[src/model/status.h, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/5636#-Ld9FAmF7xJnDK2ZDblc:-Ld9FAmF7xJnDK2ZDbld:b-kqsv05) ([raw file](https://github.com/qtox/qtox/blob/aa172b97ac18d84519c9fa458ddb3bd543ffcb48/src/model/status.h#L29)):*
> ```objc
> {
>     // Status::Status is weird, but Status is a fitting name for both the namespace and enum class..
>     enum class Status
> ```

I think explicitly making the base type `int` would save a lot of casting

Diadlo:
@sudden6 Just my opinion: explicit is better than implicit. There is not so much status casting (just 4 places)

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5636)* status: 0 of 1 LGTMs obtained

___
*[src/model/status.h, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/5636#-Ld9FAmF7xJnDK2ZDblc:-LdF343p09VOPKvGPNTE:b7y9cgn) ([raw file](https://github.com/qtox/qtox/blob/aa172b97ac18d84519c9fa458ddb3bd543ffcb48/src/model/status.h#L29)):*

I think explicitly making the base type `int` would save a lot of casting
</blockquote></details>

Changing the underlying type to `int` still fails at converting a QVariant, and doesn't allow for implicit casting. I also think the strong typing of `enum class` is useful for Status in every case but this one, so although changing to `enum` would allow it to implicitly cast during serialization/deserialization, I think it would be less safe as a type in all other cases.

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 26 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5636#-:-LdFHbOo8mTcKBtUkEYB:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/model/status.h, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/5636#-Ld9FAmF7xJnDK2ZDblc:-LdFHVWdBeyvE-ZqisU6:b7g79ah) ([raw file](https://github.com/qtox/qtox/blob/aa172b97ac18d84519c9fa458ddb3bd543ffcb48/src/model/status.h#L29)):*

Changing the underlying type to `int` still fails at converting a QVariant, and doesn't allow for implicit casting. I also think the strong typing of `enum class` is useful for Status in every case but this one, so although changing to `enum` would allow it to implicitly cast during serialization/deserialization, I think it would be less safe as a type in all other cases.
</blockquote></details>

ok then, since you and @Diadlo seem to agree it's better otherwise.

PR #5670, opened 2019-05-17
[Refactor] Dialogs

Diadlo:
Move more logic from the `FriendWidget` and `GroupWidget` into the corresponding model `Chatroom` classes.
Reduce the usage of `ContentDialogManager` singleton.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,6 +336,7 @@ set(${PROJECT_NAME}_SOURCES
   src/model/profile/iprofileinfo.h
   src/model/profile/profileinfo.cpp
   src/model/profile/profileinfo.h
+  src/model/dialogs/idialogs.h
   src/net/bootstrapnodeupdater.cpp
   src/net/bootstrapnodeupdater.h
   src/net/avatarbroadcaster.cpp
--- a/src/model/chatroom/friendchatroom.cpp
+++ b/src/model/chatroom/friendchatroom.cpp
@@ -1,5 +1,6 @@
 #include "src/grouplist.h"
 #include "src/model/chatroom/friendchatroom.h"
+#include "src/model/dialogs/idialogsmanager.h"
 #include "src/model/friend.h"
 #include "src/model/group.h"
 #include "src/persistence/settings.h"
@@ -21,8 +22,9 @@ QString getShortName(const QString& name)
 
 }
 
-FriendChatroom::FriendChatroom(Friend* frnd)
+FriendChatroom::FriendChatroom(Friend* frnd, IDialogsManager* dialogsManager)
     : frnd{frnd}
+    , dialogsManager{dialogsManager}
 {
 }
 
@@ -140,4 +142,32 @@ QVector<CircleToDisplay> FriendChatroom::getOtherCircles() const
 void FriendChatroom::resetEventFlags()
 {
     frnd->setEventFlag(false);
+}
+
+bool FriendChatroom::possibleToOpenInNewWindow() const
+{
+    const auto friendPk = frnd->getPublicKey();
+    const auto dialogs = dialogsManager->getFriendDialogs(friendPk);
+    return !dialogs || dialogs->chatroomCount() > 1;
+}
+
+bool FriendChatroom::canBeRemovedFromWindow() const
+{
+    const auto friendPk = frnd->getPublicKey();
+    const auto dialogs = dialogsManager->getFriendDialogs(friendPk);
+    return dialogs && dialogs->hasContact(friendPk);
+}
+
+bool FriendChatroom::friendCanBeRemoved() const
+{
+    const auto friendPk = frnd->getPublicKey();
+    const auto dialogs = dialogsManager->getFriendDialogs(friendPk);
+    return !dialogs || !dialogs->hasContact(friendPk);
+}
+
+void FriendChatroom::removeFriendFromDialogs()
+{
+    const auto friendPk = frnd->getPublicKey();
+    auto dialogs = dialogsManager->getFriendDialogs(friendPk);
+    dialogs->removeFriend(friendPk);
 }
--- a/src/model/chatroom/friendchatroom.h
+++ b/src/model/chatroom/friendchatroom.h
@@ -26,6 +26,7 @@
 #include <QString>
 #include <QVector>
 
+class IDialogsManager;
 class Friend;
 class Group;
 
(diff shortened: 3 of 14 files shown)


sudden6 (changes requested):
> Reviewed 7 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)* status: 0 of 1 LGTMs obtained

___
*[src/model/dialogs/idialogs.h, line 33 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQj6hg2luh2uqoyqGX:-LfQj6hh8V-940sWU7Wm:b-cgbmyg) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogs.h#L33)):*
> ```objc
>     virtual bool isContactActive(const ContactId& contactId) const = 0;
> 
>     virtual void removeFriend(const ToxPk& friendPk) = 0;
> ```

Would it be possible to do this in one method `removeContact(const ContactId& contact)` using `qobject_cast`? I think this would reduce the complexity
___
*[src/model/dialogs/idialogsmanager.h, line 31 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQjO9N4CHUyY_sRCON:-LfQjO9ODzPg6tIJ58kR:bs2cfuh) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogsmanager.h#L31)):*
> ```objc
> {
> public:
>     virtual IDialogs* getFriendDialogs(const ToxPk& friendPk) const = 0;
> ```

Same as above, should be possible to have one method.

Diadlo (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)* status: 0 of 1 LGTMs obtained

___
*[src/model/dialogs/idialogs.h, line 33 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQj6hg2luh2uqoyqGX:-LfYKbCT7n12tKBUXajd:byfqkon) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogs.h#L33)):*

Would it be possible to do this in one method `removeContact(const ContactId& contact)` using `qobject_cast`? I think this would reduce the complexity
</blockquote></details>

Why? I don't see any reason to do it, it's just overcomplicating code.

While the caller knows the type, it can select the needed method (I can use the same name to create overload). One method can be used when (if) `ContentDialog` will store contacts as a list of `Contact`.
___
*[src/model/dialogs/idialogsmanager.h, line 31 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQjO9N4CHUyY_sRCON:-LfYM2VX3wLhw8euYR77:bcgtfmn) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogsmanager.h#L31)):*

Same as above, should be possible to have one method.
</blockquote></details>

For what?

Diadlo:
@sudden6 ping

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)* status: 0 of 1 LGTMs obtained

___
*[src/model/dialogs/idialogs.h, line 33 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQj6hg2luh2uqoyqGX:-LfzhML8EpfzQKblfeLI:b-pwvalb) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogs.h#L33)):*

Why? I don't see any reason to do it, it's just overcomplicating code.

While the caller knows the type, it can select the needed method (I can use the same name to create overload). One method can be used when (if) `ContentDialog` will store contacts as a list of `Contact`.
</blockquote></details>

I just don't see a reason why we would invent a common base class `ContactId` and then never use it. AFAICT both methods should do the same, just the type is different, so why not eliminate the difference?

Diadlo (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)* status: 0 of 1 LGTMs obtained

___
*[src/model/dialogs/idialogs.h, line 33 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQj6hg2luh2uqoyqGX:-Lg1Ee_R6e72OmkEJMKW:b-jauiwn) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogs.h#L33)):*

I just don't see a reason why we would invent a common base class `ContactId` and then never use it. AFAICT both methods should do the same, just the type is different, so why not eliminate the difference?
</blockquote></details>

Remove two methods, with two different logics. Instead create one method, which accepts parent class and checks subclass and does the same two different logics? Is it correct? If so, don't you think this is strange?
I think until we can't implement common logic for all `Contact` subclasses, there is no reason to merge these methods. I will be absolutely in favor when we can

Diadlo:
@sudden6 ping

sudden6 (approved):
> Reviewed 7 of 14 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670)* status: 0 of 1 LGTMs obtained

___
*[src/model/dialogs/idialogs.h, line 33 at r1](https://reviewable.io/reviews/qtox/qtox/5670#-LfQj6hg2luh2uqoyqGX:-Lhj1aHZBfksEkPkMIrd:by08ztt) ([raw file](https://github.com/qtox/qtox/blob/667afd7fde603a777edb90bc7f2c244efa710f21/src/model/dialogs/idialogs.h#L33)):*

Remove two methods, with two different logics. Instead create one method, which accepts parent class and checks subclass and does the same two different logics? Is it correct? If so, don't you think this is strange?
I think until we can't implement common logic for all `Contact` subclasses, there is no reason to merge these methods. I will be absolutely in favor when we can
</blockquote></details>

ok, if you are sure this is the better way I won't object

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5670#-:-Lhj1loU1ulWSaHJCjde:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5687, opened 2019-05-24
fix(settings): apply toxcore settings for encrypted profiles

### LINKED ISSUE #5682: Loading an encrypted doesn't apply core settings
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
toxcore:  v0.2.9
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Load an encrypted profile.

##### Observed Behavior
1. During initialization, `persistence/settingsserializer.cpp:333 : Critical: The settings file is encrypted, but we don't have a passkey!`
2. Toxcore saved settings, such as proxy, are not applied. Setting an invalid proxy in settings saves and shows in the settings page, but Toxcore still connects to the DHT.

##### Expected Behavior
1. Settings are restored properly.

anthonybilinski:
Fix #5682

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5687)

### DIFF
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -107,7 +107,7 @@ Profile::Profile(QString name, const QString& password, bool isNewProfile,
     Settings& s = Settings::getInstance();
     s.setCurrentProfile(name);
     s.saveGlobal();
-    s.loadPersonal(name, passkey.get());
+    s.loadPersonal(name, this->passkey.get());
     initCore(toxsave, s, isNewProfile);
 
     const ToxId& selfId = core->getSelfId();


sudden6 (approved):
:lgtm_strong: Additionally, I think we should change our coding style to forbid function/contructor parameters with the same name as member variables. Any idea for a good style on this?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5687#-:-LfkdpU5B0AzY4bIITNz:b-6osol5)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
I merged as is for now to get it in sooner rather than later since it has proxy applications.

I agree that it would be good to prevent this from happening in the future. We could use -Wshadow  to catch this at compile time.

As for naming I guess we should have some standardized suffix or prefix on arguments or class members so that the programmer doesn't have to choose new names for each constructor. I don't really have strong opinions of which to chose, but would prefer something that's present in a style guide or is generally common so that it's more familiar to more people.

PR #5692, opened 2019-06-02
reafctor(strongtype): Add property types to strong types

sphaerophoria:
Add strongtype parameters to toggle on/off operators. Will ease use of more strong typed vars in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5692)

### DIFF
--- a/src/core/receiptnum.h
+++ b/src/core/receiptnum.h
@@ -25,7 +25,7 @@
 #include <QMetaType>
 #include <cstdint>
 
-using ReceiptNum = NamedType<uint32_t, struct ReceiptNumTag>;
+using ReceiptNum = NamedType<uint32_t, struct ReceiptNumTag, Orderable>;
 Q_DECLARE_METATYPE(ReceiptNum);
 
 #endif /* RECEIPT_NUM_H */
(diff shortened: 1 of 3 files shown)


sudden6:
Is this required for some further work? Else I'd like to apply the YAGNI (https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle here

sphaerophoria:
I was going to use the concept fairly heavily in my upcoming PR. The rational is that the strongtypes provide a lot more value if you can define what types of operations are valid on them. If you force people to open up to the underlying type every time it's very easy for a type to get used in a way that it wasn't intended to be in the future.

PR'd on it's own to reduce scope of a future pull request I was going to use it in.

I'd argue that this is one of those things that no one ever really _needs_ but once it's in it improves ability to reason about types and type conversions resulting in cleaner and easier to review code.

sudden6 (approved):
:lgtm_strong: then, but needs a rebase because of conflicts

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5692#-:-LhgG6FE6_5Vr9z7EnTr:b-ajx8if)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5694, opened 2019-06-10
Crash when logging back in after logging out while owning a group chat


Kribylet:
##### Brief Description

OS: Linux Mint 19.1
qTox version: master
Commit hash: 8568a14 
toxcore:  v0.2.9
Qt:  5.9.5
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Create new profiles A and B
2. B adds A as friend
3. A requests a groupchat with B through right-click on B in friend list
4. B accepts and joins group chat
5. A enters profile page by clicking own display picture, selects logout
6. A logs back in, crashes if/when B is found online

##### Observed Behavior
Crashing upon logging in if detecting participant of previous group chat online

##### Additional Info

[backtrace.txt](https://github.com/qTox/qTox/files/3272102/backtrace.txt)

Kribylet:
Probably related to https://github.com/qTox/qTox/issues/5542

anthonybilinski:
To expand a bit for posterity, #5542 causes groups to not be loaded when they have no name, so the group is never added to our GroupList. But the group still exists in toxcore, so we get callbacks for it, which the assert catches. The assert here is a good check that caught an issue - the core problem is that we 1) save groups with no length and 2) don't handle 0-length names that are saved, as discussed in #5542.

anthonybilinski:
Fixed by https://github.com/qTox/qTox/pull/5695.

PR #5695, opened 2019-06-12
fix(group): set default group chat title when provided title is invalid

### LINKED ISSUE #5542: Groups with default names fail to load on next start
##### Brief Description
qTox version: v1.16.3+
Commit hash:  d2508e7eac26d731c746333b5d74e9a051d3f5b7
toxcore:  v0.2.9

Reproducible: Always

##### Steps to reproduce

1. Create a group.
2. Don't change from the default, e.g. "Groupchat #0"
3. Restart qTox.

##### Observed Behavior
1. The group does not appear in the contacts list on load.
2. ```core/core.cpp:58 : Warning: Invalid conference title length``` is logged.

##### Expected Behavior
1. The group appears in the contacts list with the same name "Groupchat #0" as was seen on creation.

##### Additional Info
We likely don't call down to toxcore to set group title and the "Groupchat #0" title is only used in UI. Either we should set "Groupchat #0" in toxcore on group creation, or we should check for 0-length titles in toxcore and instead of fetching their name, send them to qTox and again generate a default title.

IMO setting the name in toxcore would be better so that peers using other clients would see the same name, and so that the name wouldn't change numbers between starts.

Kribylet:
Add handling logic for when a group exists but has an invalid title.
Title is created using the first 8 hexadecimals of the group chat id.

Fixes #5542 and #5694

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5695)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -28,8 +28,8 @@
 #include "src/core/toxstring.h"
 #include "src/model/groupinvite.h"
 #include "src/model/status.h"
-#include "src/nexus.h"
 #include "src/net/bootstrapnodeupdater.h"
+#include "src/nexus.h"
 #include "src/persistence/profile.h"
 #include "src/util/strongtype.h"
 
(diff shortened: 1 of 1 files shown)


sudden6 (changes requested):
Please reformat according to https://github.com/qTox/qTox/blob/master/doc/coding_standards.md, you can use the script linked in this file.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5695#-:-LhJX4hN0UfnOtrDlQey:bw9s30i)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1036 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhJXbYX7uYRFJFrNz_k:-LhJXbYX7uYRFJFrNz_l:b-8l3t9w) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1036)):*
> ```cpp
>         QByteArray nameByteArray;
>         QString name;
>         bool invalidTitle;
> ```

directly initialize this variable with the function call a few lines below?
___
*[src/core/core.cpp, line 1040 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhJY4LUAmdkTcBdi_h2:-LhJY4LUAmdkTcBdi_h3:bavlasm) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1040)):*
> ```cpp
>         size_t titleSize = tox_conference_get_title_size(tox.get(), groupNumber, &error);
>         invalidTitle = LogConferenceTitleError(error);
>         if (!invalidTitle)
> ```

merge this with the `if (invalidTitle)`and use local variables to make it easier to read?

Kribylet:
> > ```c++
> >         QByteArray nameByteArray;
> >         QString name;
> >         bool invalidTitle;
> > ```
> 
> directly initialize this variable with the function call a few lines below?
> 
> >         size_t titleSize = tox_conference_get_title_size(tox.get(), groupNumber, &error);
> >         invalidTitle = LogConferenceTitleError(error);
> >         if (!invalidTitle)
> > ```
> 
> merge this with the `if (invalidTitle)`and use local variables to make it easier to read?

I will reformat. I really tried to condense the boolean logic, but if we entertain the possibility that we can get an error in between retrieving the title size and the title name, there will either be code duplication or this kind of awkward bool variable. I'm new to the qTox project so if you tell me which you consider the lesser evil I will adjust accordingly!

sudden6 (changes requested):
I just now realized that you re-use the `invalidTitle` variable for the second error check. I think you are correct in that there can be errors between those two function calls, but it may be enough to guard against that using an `assert(error == TOX_ERR_CONFERENCE_TITLE_OK)`.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5695#-:-LhMZcC_9YiqHUq2kJHk:buj8prt)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1046 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhMZAkBBvcDwZLwmcbT:-LhMZAkBBvcDwZLwmcbU:b-5jfby4) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1046)):*
> ```cpp
>             invalidTitle = LogConferenceTitleError(error);
>         }
>         if (error == TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND)
> ```

after further reviewing, I'm not 100% sure what you intend to do with this line. Why should all error cases, besides `TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND` be ignored? How could this error code even come up, when we iterate over the groups that c-toxcore provides to us? Did you observer that? (Then it's probably also a c-toxcore bug :) )

Kribylet:
When I wrote the code I studied the toxcore functions being called, and they both have three possible outcomes for "error":
TOX_ERR_CONFERENCE_TITLE_INVALID_LENGTH,
TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND and
TOX_ERR_CONFERENCE_TITLE_OK

It's possible that we could never run into CONFERENCE_NOT_FOUND. My reasoning was that the previous code did react to this possibility, and that my change should attempt to respect the previous behavior outside of my changes. If a title has invalid length, we simply substitute our own title, so all cases are covered at that point. (Invalid title is assigned a title by us, a valid title is retrieved from tox)

And finally to argue my preference, I think we should respect the possibility of a group ceasing to exist after we've retrieved the group list on the basis that qTox is multi-threaded. My understanding of what is running concurrently isn't perfect, but if we consider it possible that we can retrieve a valid title size and then get an invalid title length error in the following lines of code, it doesn't seem far-fetched with the same reasoning to suppose a group could be closed, too.

sudden6:
> I think we should respect the possibility of a group ceasing to exist after we've retrieved the group list on the basis that qTox is multi-threaded. My understanding of what is running concurrently isn't perfect, but if we consider it possible that we can retrieve a valid title size and then get an invalid title length error in the following lines of code, it doesn't seem far-fetched with the same reasoning to suppose a group could be closed, too.

Multi-threading will not be an issue, since in https://github.com/qTox/qTox/pull/5695/files#diff-5db0ca19c12b31af1e97555eeadc9eeaR1022 we acquire a lock that (should) guard against that. So there should be no creations/destructions of groups possible. If it still happens c-toxcore is breaking some promises and we have bigger problems ;)

Kribylet:
Sorry about the messy pushes, I retracted the commit when I saw I had missed condensing the QByteArray initialization.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5695#-:-Lhg-WK08OXhxFBJAtWO:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1036 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhJXbYX7uYRFJFrNz_k:-Lhg-J4L6ZWeiX8k8P3p:b32wtd) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1036)):*

directly initialize this variable with the function call a few lines below?
</blockquote></details>

ok
___
*[src/core/core.cpp, line 1040 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhJY4LUAmdkTcBdi_h2:-Lhg-SoR1-esqCAnyrgQ:b32wtd) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1040)):*

merge this with the `if (invalidTitle)`and use local variables to make it easier to read?
</blockquote></details>

ok
___
*[src/core/core.cpp, line 1046 at r1](https://reviewable.io/reviews/qtox/qtox/5695#-LhMZAkBBvcDwZLwmcbT:-Lhg-TjZ4VZAPrWe3loW:b32wtd) ([raw file](https://github.com/qtox/qtox/blob/a7d1a5189f399a72bfda50dfbfb82f1022212c5e/src/core/core.cpp#L1046)):*

after further reviewing, I'm not 100% sure what you intend to do with this line. Why should all error cases, besides `TOX_ERR_CONFERENCE_TITLE_CONFERENCE_NOT_FOUND` be ignored? How could this error code even come up, when we iterate over the groups that c-toxcore provides to us? Did you observer that? (Then it's probably also a c-toxcore bug :) )
</blockquote></details>

ok

PR #5698, opened 2019-06-14
fix(group): always retain own name when updating group peer list

### LINKED ISSUE #5686: Self name change notification from PK can be seen in groups
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
toxcore:  v0.2.9
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Rarely

##### Steps to reproduce

I haven't reproduced this myself, but have heard it reported by others.
1. Accept group invite.

##### Observed Behavior
1. See status message of <self PK is now known as self name>

##### Expected Behavior
1. We always know our own name.

##### Additional Info
This seems like likely a toxcore oddity of it giving us a peerlist changed with us in the peerlist, but not yet known the "peer"'s name. We could add a short delay on peer list changed of < 1s to allow toxcore's names to usually become consistent - but still show PK after that if no name is set, and we can also just always show our own name for our own PK.

Very likely introduced by b7bd7c621557d598d8ab7a21346b83c436063ee9.

Kribylet:
Adds special case for updating our own PK in regeneratePeerList
Format according to format-code.sh

Fixes #5686

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5698)

### DIFF
--- a/src/model/group.cpp
+++ b/src/model/group.cpp
@@ -19,12 +19,12 @@
 
 #include "group.h"
 #include "friend.h"
-#include "src/friendlist.h"
+#include "src/core/contactid.h"
 #include "src/core/core.h"
 #include "src/core/coreav.h"
-#include "src/core/contactid.h"
 #include "src/core/groupid.h"
 #include "src/core/toxpk.h"
+#include "src/friendlist.h"
 #include "src/persistence/settings.h"
 #include "src/widget/form/groupchatform.h"
 #include "src/widget/groupwidget.h"
@@ -32,7 +32,8 @@
 
 static const int MAX_GROUP_TITLE_LENGTH = 128;
 
-Group::Group(int groupId, const GroupId persistentGroupId, const QString& name, bool isAvGroupchat, const QString& selfName)
+Group::Group(int groupId, const GroupId persistentGroupId, const QString& name, bool isAvGroupchat,
+             const QString& selfName)
     : selfName{selfName}
     , title{name}
     , toxGroupNum(groupId)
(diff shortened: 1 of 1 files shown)


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5698)* status: 0 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 98 at r1](https://reviewable.io/reviews/qtox/qtox/5698#-LhN21dg6jbLkFbFWqW_:-LhN21dhBfYG7hpIUZ6s:b-b85nqj) ([raw file](https://github.com/qtox/qtox/blob/c772db3baa3f9e4694fa7f956ffdfb9ffab36cef/src/model/group.cpp#L98)):*
> ```cpp
>     for (int i = 0; i < nPeers; ++i) {
>         const auto pk = core->getGroupPeerPk(toxGroupNum, i);
>         if (pk == core->getSelfPublicKey())
> ```

braces are mandatory in our coding style, unfortunately clang-format can't insert them...

Also shouldn't there be an `else` here to not call `decideNickname(...)` for our own PK?

Kribylet:
Absolutely, my bad.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5698#-:-LhNGULQ5zeuzJSDZggo:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/model/group.cpp, line 98 at r1](https://reviewable.io/reviews/qtox/qtox/5698#-LhN21dg6jbLkFbFWqW_:-LhNGQue9BV_ZONp_9dP:b4hwzln) ([raw file](https://github.com/qtox/qtox/blob/c772db3baa3f9e4694fa7f956ffdfb9ffab36cef/src/model/group.cpp#L98)):*

braces are mandatory in our coding style, unfortunately clang-format can't insert them...

Also shouldn't there be an `else` here to not call `decideNickname(...)` for our own PK?
</blockquote></details>

ok, better now

PR #5711, opened 2019-06-21
refactor(startup): decouple loginscreen from model elements

### LINKED ISSUE #5707: Logging out after autologin quits immediately
##### Brief Description

Clicking logout when starting qTox with -p <profile name> quits immediately. 

OS: Linux Mint 19.1
qTox version: nightly
Commit hash:  c2bdcdf
toxcore:  v0.2.9
Qt:  5.9.5
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Start qTox by commandline: "qTox -p myProfile"
2. Log out by clicking Identicon and selecting Logout

##### Observed Behavior

qTox quits immediately.

##### Expected Behavior

Login screen appears.

Kribylet:
LoginScreen was directly modifying Profile, and the overall startup process of qTox was confusing. This commit aims to simplify the code and reduce overall singleton dependency during startup. The scope of the commit could not be allowed to grow too large; as such there are awkward compromises in places due to the structure of singletons that could not be changed at this time. These should not be any worse than what was already in place.

Moved audio instance into Nexus instance. This was a necessary evil to concentrate related logic and facilitate a future shifting of the Nexus object responsibilities elsewhere.

Stop treating singleton instances of Nexus, Profile and Settings as singletons during main.cpp and related processes as much as possible. This does not break Singleton functionality elsewhere, but starts the process of replacing Singleton reliance with a central model object:

- If getInstance() for a singleton is used, keep and use that pointer during function call. This enables passing the object by reference easily later.

-  If getInstance() calls are used to manipulate singleton variables, prefer an emit-to-slot solution whenever possible. Due to deep nesting preventing connect() calls or heavily singleton dependent existing solution this may be non-trivial, but this should become simpler as changes propagate through the project elsewhere.

Remove code duplication related to initial LoginScreen vs. returning to LoginScreen later. This favored the Nexus solution because it is called in code elsewhere, but as above this concentration of logic should simplify moving out of the Nexus design pattern in the long term.

Made Logout action select logged out profile by default.

Profile creation and loading is now performed using signal emits rather than immediate singleton access. This means LoginScreen no longer modifies Profile instance directly.

Made AutoLogin checkbox during LoginScreen use signals instead.

Lifted makeToxPortable initialization logic into separate utility function for future decoupling of global settings vs. profile (eg. model specific) settings. This decoupling would simplify startup behavior as a Settings instance would not be required for initial file reading.

Tests performed:
 Login (default)
 Login (-l)
 Login (-p <profile>)
 Login (create new profile)
 Login (load profile)
 Login (no existing profiles)
 Login (import profile)
 Login (autologin)
 Login (autologin checkbox)
 Login (locale translation)
 Login (quit)

 MainGUI (connect)
 MainGUI (add friend)
 MainGUI (remove friend)
 MainGUI (create group)
 MainGUI (leave group)
 MainGUI (chat)
 MainGUI (separate friend chat window)
 MainGUI (logout)
 MainGUI (quit)
 MainGUI (change profile settings)
 MainGUI (change global settings)

Concerns:
 Mac OSX compatability due to moving Mac specific logic out of Nexus and changing call order. I have not been able to test those changes.

Fixes #5707

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5711)

### DIFF
--- a/src/nexus.cpp
+++ b/src/nexus.cpp
@@ -209,7 +209,8 @@ void Nexus::connectLoginScreen(const LoginScreen& loginScreen)
     QObject::connect(&loginScreen, &LoginScreen::createNewProfile, this, &Nexus::onCreateNewProfile);
     QObject::connect(&loginScreen, &LoginScreen::loadProfile, this, &Nexus::onLoadProfile);
     // LoginScreen -> Settings
-    QObject::connect(&loginScreen, &LoginScreen::autoLoginChanged, settings, &Settings::onSetAutoLogin);
+    QObject::connect(&loginScreen, &LoginScreen::autoLoginChanged, settings, &Settings::setAutoLogin);
+    QObject::connect(&loginScreen, &LoginScreen::autoLoginChanged, settings, &Settings::saveGlobal);
     // Settings -> LoginScreen
     QObject::connect(settings, &Settings::autoLoginChanged, &loginScreen,
                      &LoginScreen::onAutoLoginChanged);
--- a/src/nexus.cpp
+++ b/src/nexus.cpp
@@ -35,6 +35,7 @@
 #include <QThread>
 #include <cassert>
 #include <vpx/vpx_image.h>
+#include <src/audio/audio.h>
 
 #ifdef Q_OS_MAC
 #include <QActionGroup>
@@ -67,7 +68,7 @@ Nexus::~Nexus()
     widget = nullptr;
     delete profile;
     profile = nullptr;
-    Settings::getInstance().saveGlobal();
+    emit saveGlobal();
 #ifdef Q_OS_MAC
     delete globalMenuBar;
 #endif
@@ -148,7 +149,7 @@ void Nexus::start()
 /**
  * @brief Hides the main GUI, delete the profile, and shows the login screen
  */
-void Nexus::showLogin()
+int Nexus::showLogin(const QString& profileName)
 {
     delete widget;
     widget = nullptr;
@@ -156,26 +157,71 @@ void Nexus::showLogin()
     delete profile;
     profile = nullptr;
 
-    LoginScreen loginScreen;
-    loginScreen.exec();
+    LoginScreen loginScreen{profileName};
+    connectLoginScreen(loginScreen);
 
-    profile = loginScreen.getProfile();
+    // TODO(kriby): Move core out of profile
+    // This is awkward because the core is in the profile
+    // The connection order ensures profile will be ready for bootstrap for now
+    connect(this, &Nexus::currentProfileChanged, this, &Nexus::bootstrapWithProfile);
+    int returnval = loginScreen.exec();
+    disconnect(this, &Nexus::currentProfileChanged, this, &Nexus::bootstrapWithProfile);
+    return returnval;
+}
+
+void Nexus::bootstrapWithProfile(Profile *p)
+{
+    // kriby: This is a hack until a proper controller is written
+
+    profile = p;
 
     if (profile) {
-        Nexus::getInstance().setProfile(profile);
-        Settings::getInstance().setCurrentProfile(profile->getName());
-        showMainGUI();
-    } else {
-        qApp->quit();
+        audioControl = std::unique_ptr<IAudioControl>(Audio::makeAudio(*settings));
+        assert(audioControl != nullptr);
+        profile->getCore()->getAv()->setAudio(*audioControl);
+        start();
     }
+}
+
+void Nexus::setSettings(Settings* settings)
+{
+    if (this->settings) {
+        QObject::disconnect(this, &Nexus::currentProfileChanged, this->settings,
+                            &Settings::updateProfileData);
+        QObject::disconnect(this, &Nexus::saveGlobal, this->settings, &Settings::saveGlobal);
+    }
+    this->settings = settings;
+    if (this->settings) {
+        QObject::connect(this, &Nexus::currentProfileChanged, this->settings,
+                         &Settings::updateProfileData);
+        QObject::connect(this, &Nexus::saveGlobal, this->settings, &Settings::saveGlobal);
+    }
+}
+
+void Nexus::connectLoginScreen(const LoginScreen& loginScreen)
+{
+    // TODO(kriby): Move connect sequences to a controller class object instead
+
+    // Nexus -> LoginScreen
+    QObject::connect(this, &Nexus::profileLoaded, &loginScreen, &LoginScreen::onProfileLoaded);
+    QObject::connect(this, &Nexus::profileLoadFailed, &loginScreen, &LoginScreen::onProfileLoadFailed);
+    // LoginScreen -> Nexus
+    QObject::connect(&loginScreen, &LoginScreen::createNewProfile, this, &Nexus::onCreateNewProfile);
+    QObject::connect(&loginScreen, &LoginScreen::loadProfile, this, &Nexus::onLoadProfile);
+    // LoginScreen -> Settings
+    QObject::connect(&loginScreen, &LoginScreen::autoLoginChanged, settings, &Settings::onSetAutoLogin);
+    // Settings -> LoginScreen
+    QObject::connect(settings, &Settings::autoLoginChanged, &loginScreen,
+                     &LoginScreen::onAutoLoginChanged);
 }
 
 void Nexus::showMainGUI()
 {
+    // TODO(kriby): Rewrite as view-model connect sequence only, add to a controller class object
     assert(profile);
 
     // Create GUI
-    widget = Widget::getInstance(audio);
+    widget = Widget::getInstance(audioControl.get());
 
     // Start GUI
     widget->init();
@@ -270,14 +316,36 @@ Profile* Nexus::getProfile()
 }
 
 /**
- * @brief Unload the current profile, if any, and replaces it.
- * @param profile Profile to set.
+ * @brief Creates a new profile and replaces the current one.
+ * @param name New username
+ * @param pass New password
  */
-void Nexus::setProfile(Profile* profile)
+void Nexus::onCreateNewProfile(const QString& name, const QString& pass)
 {
-    getInstance().profile = profile;
-    if (profile)
-        Settings::getInstance().loadPersonal(profile->getName(), profile->getPasskey());
+    setProfile(Profile::createProfile(name, pass));
+}
+
+/**
+ * Loads an existing profile and replaces the current one.
+ */
+void Nexus::onLoadProfile(const QString& name, const QString& pass)
+{
+    setProfile(Profile::loadProfile(name, pass));
+}
+/**
+ * Changes the loaded profile and notifies listeners.
+ * @param p
+ */
+void Nexus::setProfile(Profile* p) {
+    if (!p) {
+        emit profileLoadFailed();
+        // Warnings are issued during respective createNew/load calls
+        return;
+    } else {
+        emit profileLoaded();
+    }
+
+    emit currentProfileChanged(p);
 }
 
 /**
--- a/src/nexus.h
+++ b/src/nexus.h
@@ -27,6 +27,8 @@
 
 class Widget;
 class Profile;
+class Settings;
+class LoginScreen;
 class Core;
 
 #ifdef Q_OS_MAC
@@ -44,16 +46,13 @@ class Nexus : public QObject
 public:
     void start();
     void showMainGUI();
-
+    void setSettings(Settings* settings);
     static Nexus& getInstance();
     static void destroyInstance();
     static Core* getCore();
     static Profile* getProfile();
-    static void setProfile(Profile* profile);
     static Widget* getDesktopGUI();
 
-public slots:
-    void showLogin();
 
 #ifdef Q_OS_MAC
 public:
@@ -81,17 +80,29 @@ private:
     QSignalMapper* windowMapper;
     QActionGroup* windowActions = nullptr;
 #endif
-public:
-    // TODO(sudden6): hack to pass the audio instance
-    IAudioControl* audio = nullptr;
+signals:
+    void currentProfileChanged(Profile* Profile);
+    void profileLoaded();
+    void profileLoadFailed();
+    void saveGlobal();
+
+public slots:
+    void onCreateNewProfile(const QString& name, const QString& pass);
+    void onLoadProfile(const QString& name, const QString& pass);
+    int showLogin(const QString& profileName = QString());
+    void bootstrapWithProfile(Profile *p);
 
 private:
     explicit Nexus(QObject* parent = nullptr);
+    void connectLoginScreen(const LoginScreen& loginScreen);
+    void setProfile(Profile* p);
     ~Nexus();
 
 private:
     Profile* profile;
+    Settings* settings;
     Widget* widget;
+    std::unique_ptr<IAudioControl> audioControl;
 };
 
 #endif // NEXUS_H
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -105,9 +105,6 @@ Profile::Profile(QString name, const QString& password, bool isNewProfile,
     , encrypted{this->passkey != nullptr}
 {
     Settings& s = Settings::getInstance();
-    s.setCurrentProfile(name);
-    s.saveGlobal();
-    s.loadPersonal(name, this->passkey.get());
     initCore(toxsave, s, isNewProfile);
 
     const ToxId& selfId = core->getSelfId();
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -269,21 +269,17 @@ QStringList Profile::getFilesByExt(QString extension)
  * @brief Scan for profile, automatically importing them if needed.
  * @warning NOT thread-safe.
  */
-void Profile::scanProfiles()
+const QStringList Profile::getAllProfileNames()
 {
     profiles.clear();
     QStringList toxfiles = getFilesByExt("tox"), inifiles = getFilesByExt("ini");
-    for (QString toxfile : toxfiles) {
+    for (const QString& toxfile : toxfiles) {
         if (!inifiles.contains(toxfile)) {
             Settings::getInstance().createPersonal(toxfile);
         }
 
         profiles.append(toxfile);
     }
-}
-
-QStringList Profile::getProfiles()
-{
     return profiles;
 }
 
--- a/src/persistence/profile.cpp
+++ b/src/persistence/profile.cpp
@@ -105,10 +105,11 @@ Profile::Profile(QString name, const QString& password, bool isNewProfile,
     , encrypted{this->passkey != nullptr}
 {
     Settings& s = Settings::getInstance();
+    // TODO(kriby): Move/refactor core initialization to remove settings dependency
+    //  note to self: use slots/signals for this?
     initCore(toxsave, s, isNewProfile);
 
-    const ToxId& selfId = core->getSelfId();
-    loadDatabase(selfId, password);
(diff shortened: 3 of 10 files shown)


Diadlo:
@Kribylet Thanks for your refactoring. Is it possible to split your commit on smaller? Once all changes are messed up in one commit it's **very** hard to review

Also, Travis build [failed](https://travis-ci.org/qTox/qTox/jobs/548646379#L7241) on mac os

Kribylet:
I am splitting the commit right now. I see what failed on the Mac build, and I will attempt to fix that to the extent that I can without a system to review it on. If Travis builds from then I think this PR would still have to be tested on Mac to be sure it works.

src/presenter/presenter.h
```diff
@@ -0,0 +1,69 @@
+//
```
sudden6: Please use the qTox copyright notice: https://github.com/qTox/qTox/blob/3fd4ce5952211d39a5c7561ff8c9144bf00f8d29/src/model/message.cpp#L1

src/presenter/presenter.cpp
```diff
@@ -0,0 +1,215 @@
+//
```
sudden6: Please use the qTox Copyright notice: https://github.com/qTox/qTox/blob/3fd4ce5952211d39a5c7561ff8c9144bf00f8d29/src/model/message.cpp#L1

src/presenter/presenter.h
```diff
@@ -0,0 +1,69 @@
...
+    QSignalMapper* windowMapper;
+    QActionGroup* windowActions = nullptr;
+#endif
+
+public:
+    Presenter();
+
+    void setupEnvironment();
+
+public slots:
```
sudden6: there are no slots here, so this is not needed

src/widget/loginscreen.cpp
```diff
@@ -72,45 +73,56 @@ LoginScreen::~LoginScreen()
     delete ui;
 }
 
+void LoginScreen::closeEvent(QCloseEvent* event)
+{
+
+    // If we are in the bootstrap, returning -1 will give us something to exit with in main.cpp
+    this->setResult(-1);
+    // If we are in application exec, we can quit by closing it, instead.
+    qApp->quit();
```
sudden6, line 81: I don't think it's a good idea to kill the application from here, why not use the return value for this case too?

sudden6 (changes requested):
Did a quick review and found some minor things, I'm leaving the architecture review to @Diadlo

src/nexus.cpp
```diff
@@ -67,115 +58,90 @@ Nexus::~Nexus()
...
-{
     delete widget;
     widget = nullptr;
 
     delete profile;
     profile = nullptr;
 
-    LoginScreen loginScreen;
-    loginScreen.exec();
+    std::unique_ptr<LoginScreen> loginScreen(new LoginScreen{profileName});
```
Diadlo: Why instance replaced on pointer here?

src/widget/loginscreen.cpp
Kribylet: Because I do not believe it is possible at this time :/ 
In the first case we have called loginScreen as part of the initial startup, in which case we are about to call a->exec() (in main.cpp) and enter an empty application running indefinitely. The return value in that case will allow us to return early instead.

In the other case we have logged in (and thus are operating from the context of QApplication "a" in main.cpp) and then logged out. A closeEvent of LoginScreen is not sufficient at that stage to halt execution of the application, and nothing is listening to the return value either since showLogin was called by InvokeMethod from profileinfo.

I tried a few approaches to resolve this issue (Changing qApp->setQuitOnLastWindowClosed(false); to true, for one) but this was the only one I found that worked. I don't mean to say there couldn't be a better solution, but I could not find a better one.

Thank you for your other points, I will fix them ASAP. I also think I have a solution for the widget related mac OSX issue, again, at least to successfully build with some hands on testing recommended. Commits forthcoming, I have struggled to escape the event horizon of swedish midsummer celebrations.

src/presenter/presenter.h
```diff
@@ -0,0 +1,69 @@
...
+ * The Presenter class should be responsible for linking view interface elements with appropriate
+ * model altering methods, and linking model changes to appropriate slots in views. It should not
+ * contain model information, nor should it instantiate views (eg. mainGUI, audioNotifications..)
+ */
+class Presenter : public QObject
+{
+    Q_OBJECT
+
+#ifdef Q_OS_MAC
+public:
```
Diadlo: Why these fields are public?

src/nexus.cpp
```diff
@@ -235,20 +235,34 @@ Profile* Nexus::getProfile()
...
 
 /**
- * @brief Unload the current profile, if any, and replaces it.
- * @param profile Profile to set.
+ * @brief Create a new profile and replace the current one.
+ * @param name New username
+ * @param pass New password
  */
-void Nexus::setProfile(Profile* profile)
+void Nexus::onCreateNewProfile(QString name, const QString& pass)
```
Diadlo: const reference for `name`

src/nexus.cpp
```diff
@@ -235,20 +235,34 @@ Profile* Nexus::getProfile()
...
+        return;
+    }
 
+    emit currentProfileChanged(p);
+}
 
+void Nexus::onLoadProfile(QString name, const QString& pass)
+{
+    Profile* p = Profile::loadProfile(name, pass);
+    emit profileLoaded(bool(p));
```
Diadlo: I think this signal can be split on two: `profileLoadFailed` and `profileLoaded`. It's a bit strange to have `profileLoaded(false)` because it means that profile not loaded. Also on practice slot have two branches. One to handle success and second to handle failure. 

But feel free to leave.

src/nexus.cpp
```diff
@@ -235,20 +235,34 @@ Profile* Nexus::getProfile()
...
 
+    if (!p) {
+        // Warnings are issued during Profile::createProfile
+        return;
+    }
 
+    emit currentProfileChanged(p);
+}
 
+void Nexus::onLoadProfile(QString name, const QString& pass)
```
Diadlo: const reference for `name`

src/nexus.h
```diff
@@ -44,29 +46,34 @@ class Nexus : public QObject
...
+    void profileLoaded(bool success);
+    void saveGlobal();
 
 public slots:
-    // TODO(sudden6): hack to pass the audio instance
-    IAudioControl* audio = nullptr;
+    void onCreateNewProfile(QString name, const QString& pass);
+    void onLoadProfile(QString name, const QString& pass);
+    int showLogin(const QString& profileName = QString());
+    void bootStrapWithProfile(Profile* p);
```
Diadlo: AFAIK "bootstrap" it's one word, so there shouldn't be uppercase "S"

src/persistence/profile.cpp
```diff
@@ -105,9 +105,8 @@ Profile::Profile(QString name, const QString& password, bool isNewProfile,
     , encrypted{this->passkey != nullptr}
 {
     Settings& s = Settings::getInstance();
-    s.setCurrentProfile(name);
-    s.saveGlobal();
-    s.loadPersonal(name, this->passkey.get());
+    // TODO(kriby): Move/refactor core initialization to remove settings dependency
```
Diadlo, line 108: Why settings dependency should be removed? I suppose `Settings` should be reduced to `IProfileSettings`, but not fully removed.

src/persistence/settings.cpp
```diff
@@ -2318,12 +2331,13 @@ bool Settings::getAutoLogin() const
     return autoLogin;
 }
 
-void Settings::setAutoLogin(bool state)
+void Settings::onSetAutoLogin(bool state)
```
Diadlo: Why the name was changed? It's one of the settings. It has a `get/set` pair

src/persistence/settings.cpp
```diff
@@ -2318,12 +2331,13 @@ bool Settings::getAutoLogin() const
...
 {
     QMutexLocker locker{&bigLock};
 
     if (state != autoLogin) {
         autoLogin = state;
+        saveGlobal();
```
Diadlo: Why did this add?

src/widget/loginscreen.cpp
```diff
@@ -72,45 +73,56 @@ LoginScreen::~LoginScreen()
...
+}
+
 /**
  * @brief Resets the UI, clears all fields.
  */
-void LoginScreen::reset(QString initialProfile)
+void LoginScreen::reset(QString initialProfileName)
```
Diadlo: const reference

src/widget/loginscreen.cpp
```diff
@@ -72,45 +73,56 @@ LoginScreen::~LoginScreen()
...
-        initialProfile = Settings::getInstance().getCurrentProfile();
-    }
-    QStringList profiles = Profile::getProfiles();
-    for (QString profile : profiles) {
-        ui->loginUsernames->addItem(profile);
-        if (profile == initialProfile) {
-            ui->loginUsernames->setCurrentIndex(ui->loginUsernames->count() - 1);
-        }
-    }
+    QStringList allProfileNames = Profile::getAllProfileNames(); // static utility function
```
Diadlo: I think this comment is redundant. Feel free to leave

Also looks like the result can be const.

src/widget/loginscreen.cpp
```diff
@@ -266,7 +246,7 @@ void LoginScreen::retranslateUi()
 
 void LoginScreen::onImportProfile()
 {
-    ProfileImporter* pi = new ProfileImporter(this);
+    auto pi = new ProfileImporter(this);
```
Diadlo: Does anyone know, why `new` used here? Variable deleted on the end of the function. I think the instance can be used. What do you think? Or at least `unique_ptr`

src/widget/loginscreen.h
```diff
@@ -36,22 +36,27 @@ class LoginScreen : public QDialog
...
-    void reset(QString selectedProfile = QString());
-    Profile* getProfile() const;
-
     bool event(QEvent* event) final override;
 
 signals:
+
     void windowStateChanged(Qt::WindowStates states);
-    void closed();
+    // Connect to Settings instance
```
Diadlo: It can be used not only by settings. I think this comment is redundant

src/widget/loginscreen.h
```diff
@@ -36,22 +36,27 @@ class LoginScreen : public QDialog
...
+    void setAutoLogin(bool state);
```
Diadlo: Also, I prefer the name of signals with `ed` suffix. I.e. `autoLoginChanged`,`newProfileCreated`, `profileLoaded`

src/persistence/settings.cpp
```diff
@@ -273,6 +263,22 @@ void Settings::loadGlobal()
     loaded = true;
 }
 
+bool Settings::isToxPortable()
+{
+    QString localSettingsPath = qApp->applicationDirPath() + QDir::separator() + globalSettingsFile;
+    bool result;
+
+    if (QFile(localSettingsPath).exists()) {
```
Diadlo: I prefer early return here. Feel free to ignore.
```
if (!QFile(localSettingsPath).exists()) {
    return false;
}
```

src/persistence/settings.cpp
```diff
@@ -143,9 +143,13 @@ void Settings::loadGlobal()
         }
         autoAwayTime = s.value("autoAwayTime", 10).toInt();
         checkUpdates = s.value("checkUpdates", true).toBool();
-        notifySound = s.value("notifySound", true).toBool(); // note: notifySound and busySound UI elements are now under UI settings
+        notifySound =
+            s.value("notifySound", true)
+                .toBool(); // note: notifySound and busySound UI elements are now under UI settings
```
Diadlo: Probably better to move comment on the line above but save code formatting

Diadlo (changes requested):

src/presenter/presenter.h
Kribylet: fixed

src/presenter/presenter.cpp
Kribylet: fixed

src/presenter/presenter.h
Kribylet: fixed

src/persistence/profile.cpp
Kribylet: My argument is as follows; In the long term, there is nothing about a user profile that implies it should hold elements of the tox connection itself. As a direct consequence of  Profile holding Core right now, the startup sequence code complexity goes up significantly. I imagine code complexity elsewhere in the project could be affected as well, at a minimum the tight coupling makes refactoring efforts to reduce singleton dependency in the project overall harder.

In the long term, I would argue that it would healthier for qTox as a project to separate class concerns concisely and ween off use of singleton access so that we can minimize unexpected adverse effects of code changes throughout the project. I added this todo in the spirit of that goal, but I admit it may take a while to reach a point where we can safely resolve it.

src/persistence/settings.cpp
Kribylet: This was changed to update and read changes through signals rather than direct object access. Now it should emit a signal when the checkbox is clicked which prompts settings to update the value internally. When the value is updated in Settings, another signal is sent back to the LoginScreen notifying it that the autologin value has changed, which updates the visual of the checkbox. This means that LoginScreen does not have to know anything about Settings to function, which makes the code more concise and modular.

A somewhat unlikely utility of this would be if the autologin property was changed elsewhere. The previous implementation would have no way of knowing a change had taken place, while this one would send a signal indicating the update. The usefulness of this implementation is much more obvious in the case of multiple windows representing the same group chat, for example.

src/persistence/settings.cpp
Kribylet: Because previously clicking the autologin checkbox called saveGlobal directly through Settings::getInstance(). The emit solution I wrote above could emit saveGlobal to preserve that behavior, but since all instances of autologin changing currently would behave the same, I opted to simply call the relevant function in its own class.

src/widget/loginscreen.cpp
Kribylet: Changed and tested this, PR update will use an instance as done elsewhere in code.

src/nexus.cpp
Kribylet: No good reason, fixed.

src/presenter/presenter.h
Kribylet: Unknown. Reverting to old Nexus implementation of Mac OSX hooks for this PR after discussion with sudden6. If you wish me to modify this code please indicate as such and I will make a separate PR.

src/nexus.cpp
Kribylet: fixed

src/nexus.cpp
Kribylet: Changed as indicated.

src/nexus.cpp
Kribylet: fixed

src/widget/loginscreen.h
Kribylet: Removed.

src/widget/loginscreen.h
Kribylet: Changed as indicated

src/persistence/settings.cpp
Kribylet: Changed as indicated

src/persistence/settings.cpp
Kribylet: Changed as indicated

src/widget/loginscreen.cpp
Kribylet: fixed

src/presenter/presenter.h
Diadlo: Ok. I did not notice, that it was moved

src/persistence/settings.cpp
Diadlo: The meaning of the early return is to get rid of the `else` branch and its indent. I.e.
```
if (!QFile(localSettingsPath).exists()) {
    return false;
}

QSettings ps(localSettingsPath, QSettings::IniFormat);
ps.setIniCodec("UTF-8");
ps.beginGroup("Advanced");
bool result = ps.value("makeToxPortable", false).toBool();
ps.endGroup();
return result;
```

src/persistence/profile.cpp
Diadlo: > nothing about a user profile that implies it should hold elements of the tox connection itself

Yes, but we talk about settings.

I agree with all your thesis. In particular, that Settings singleton should be replaced on Dependency Injection. But in your comment, you said that settings dependency should be completely removed

src/persistence/settings.cpp
Diadlo: Again. You misunderstood my point. I'm not against using a signal/slot mechanism to interact between classes. I talk about method name. I think `Settings` shouldn't know is its methods used as an endpoint of the connection or not. The method should be `onXXX` only if they connected inside the class itself.
I.e.
```
void MyClass::init()
{
    connect(view, &MyView::autoLoginChanged, this, &MyClass::onAutoLoginChanged);
}
```
but
```
void MyClass::init()
{
    connect(view, &MyView::autoLoginChanged, service, &MyService::setAutoLogin);
}
```

src/persistence/settings.cpp
Diadlo: Suppose to connect `autoLoginChanged` to `saveGlobal` and remove from here

Diadlo:
A few small notes. Otherwise LGTM

src/widget/loginscreen.cpp
sudden6: ok, this is an acceptable workaround then

src/widget/loginscreen.cpp
```diff
@@ -245,10 +245,7 @@ void LoginScreen::retranslateUi()
 
 void LoginScreen::onImportProfile()
 {
-    ProfileImporter* pi = new ProfileImporter(this);
-
-    if (pi->importProfile())
+    ProfileImporter pi(this);
+    if (pi.importProfile())
```
sudden6: braces required

sudden6 (commented):
Not much to add to the review of @Diadlo

src/persistence/settings.cpp
Kribylet: Fixed

src/widget/loginscreen.cpp
Kribylet: fixed

sudden6 (approved):
:lgtm_strong: from my side, @Diadlo please merge if also from your side

> Reviewed 1 of 13 files at r1, 4 of 11 files at r2, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5711#-:-LiAVR1H1Hx2RNmQtnys:bus6ktt)* status: :shipit: complete! 1 of 1 LGTMs obtained

src/persistence/profile.cpp
Kribylet: It was mostly meant as a reminder to myself to attempt to refactor that code to the effect I described at some point in the near future. I will remove it.

src/persistence/settings.cpp
Kribylet: Absolutely. Changed as indicated.

src/persistence/settings.cpp
Kribylet: Yes, I see what you mean now. Sorry this took so long to respond to, I didn't realize Github would hide unresolved discussions when condensing multiple. Changed as indicated.

Kribylet: I renamed onCurrentProfileChanged to updateProfileData well to match this line of thinking.

Diadlo (approved):

PR #5715, opened 2019-06-23
fix: use currentDateTime for get num messages for friend (fix #5714)

### LINKED ISSUE #5714: history doesn't load
##### Brief Description

OS: Arch Linux
qTox version: v1.16.3-476-g033f28e6
Commit hash: 033f28e67ecd61cd0fdb6cc0e8de8a2bc920e1d7
toxcore: 0.2.9
Qt: 5.12.4

Reproducible: Always

##### Steps to reproduce

1. start the program
2. go to friend chat and see empty chat without history

TriKriSta:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5715)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -523,10 +523,7 @@ void History::setFileFinished(const QString& fileId, bool success, const QString
 
 size_t History::getNumMessagesForFriend(const ToxPk& friendPk)
 {
-    return getNumMessagesForFriendBeforeDate(friendPk,
-                                             // Maximum possible time
-                                             QDateTime::fromMSecsSinceEpoch(
-                                                 std::numeric_limits<int64_t>::max()));
+    return getNumMessagesForFriendBeforeDate(friendPk, QDateTime());
 }
 
 size_t History::getNumMessagesForFriendBeforeDate(const ToxPk& friendPk, const QDateTime& date)
@@ -534,10 +531,14 @@ size_t History::getNumMessagesForFriendBeforeDate(const ToxPk& friendPk, const Q
     QString queryText = QString("SELECT COUNT(history.id) "
                                 "FROM history "
                                 "JOIN peers chat ON chat_id = chat.id "
-                                "WHERE chat.public_key='%1'"
-                                "AND timestamp < %2;")
-                            .arg(friendPk.toString())
-                            .arg(date.toMSecsSinceEpoch());
+                                "WHERE chat.public_key='%1'")
+                            .arg(friendPk.toString());
+
+    if (date.isNull()) {
+        queryText += ";";
+    } else {
+        queryText += QString(" AND timestamp < %1;").arg(date.toMSecsSinceEpoch());
+    }
 
     size_t numMessages = 0;
     auto rowCallback = [&numMessages](const QVector<QVariant>& row) {


Diadlo (approved):

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5715)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 526 at r1](https://reviewable.io/reviews/qtox/qtox/5715#-LiEHZ612Gk0yToVshvO:-LiEHZ612Gk0yToVshvP:bgya8to) ([raw file](https://github.com/qtox/qtox/blob/1c42efeebbe377792e0b857048241366bef7ae3a/src/persistence/history.cpp#L526)):*
> ```cpp
> size_t History::getNumMessagesForFriend(const ToxPk& friendPk)
> {
>     return getNumMessagesForFriendBeforeDate(friendPk, QDateTime::currentDateTime());
> ```

I think this is incorrect. If the timezone changes we may miss some messages. I think what I want to do here is use QDateTime() and then in getNumMessagesForFriendBeforeDate only add the timestamp restriction when !date.isNull()

TriKriSta:
for example, I can create a function getFinishDateChatHistory which will take the date of the last message. And this date use for load history

sphaerophoria:
keep in mind database access can have significant user noticable latency. If youre going to make a new call I'd prefer one that just does what we want, but I think its fairly easy to allow invalid dates. I dont mind doing the change if you dont want to btw. I introduced the bug 🤷

sphaerophoria:
Just for completeness the root cause of the bug is that QDateTime has undefined behavior if the stored time goes out of range. I couldn't find docs on what the valid range is, just that it can change based on timezone and timespec


From the Qt docs

The range of valid values able to be stored in QDateTime is dependent on the internal storage implementation. QDateTime is currently stored in a qint64 as a serial msecs value encoding the date and time. This restricts the date range to about +/- 292 million years, compared to the QDate range of +/- 2 billion years. Care must be taken when creating a QDateTime with extreme values that you do not overflow the storage. The exact range of supported values varies depending on the Qt::TimeSpec and time zone.

TriKriSta:
And if use it QDateTime::currentDateTime().addDays(1) ?

sphaerophoria:
What happens if someone changes their system time? I think it's fundamentally incorrect to include the timestamp filter when we don't want a timestamp filter. I think it's better to remove the filter by dynamically adding it based on a valid QDateTime.

TriKriSta:
> I think this is incorrect. If the timezone changes we may miss some messages. I think what I want to do here is use QDateTime() and then in getNumMessagesForFriendBeforeDate only add the timestamp restriction when !date.isNull()

Did you mean, If getNumMessagesForFriendBeforeDate gets QDateTime.isNull this function will return all num messages?

TriKriSta:
this decision breaks search, but search breaks anyway. So search can fix on another pull request

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5715#-:-LiF5YGzEYlZ6zpHk8p2:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5721, opened 2019-06-24
fix(core): correctly relink ui/core when core is changed

### LINKED ISSUE #5710: Reconnect button doesn't reconnect
qTox only goes offline, and never comes back online.

Enable UDP - on
Enable LAN discovery - on
No proxy

version 1.16.3.467 (rev.678fc51c) from a package.
FreeBSD 12

Kribylet:
Widget was only connecting the Core to itself during initialization, but the Core instance could change during a restartCore call. This commit will make Widget link the Core to itself when it changes rather than only on initialization.

Fixes #5710

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5721)

### DIFF
--- a/src/widget/widget.cpp
+++ b/src/widget/widget.cpp
@@ -219,9 +219,9 @@ void Widget::init()
     filterDisplayActivity->setCheckable(true);
     filterDisplayGroup->addAction(filterDisplayActivity);
     filterMenu->addAction(filterDisplayActivity);
-    settings.getFriendSortingMode() == FriendListWidget::SortingMode::Name ?
-        filterDisplayName->setChecked(true) :
-        filterDisplayActivity->setChecked(true);
+    settings.getFriendSortingMode() == FriendListWidget::SortingMode::Name
+        ? filterDisplayName->setChecked(true)
+        : filterDisplayActivity->setChecked(true);
     filterMenu->addSeparator();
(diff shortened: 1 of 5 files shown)


Kribylet:
Changed to use reference instead of pointer

src/widget/widget.cpp
```diff
@@ -675,6 +675,37 @@ void Widget::onSelfAvatarLoaded(const QPixmap& pic)
     profilePicture->setPixmap(pic);
 }
 
+void Widget::onCoreChanged(Core& core) {
```
sudden6: brace should be on the next line, was this formatted by clang-format this way?

sudden6 (commented):
lgtm, small mistake found

src/widget/widget.cpp
Kribylet: It wasn't, I forgot to run format. Fixed.

sudden6 (approved):
:lgtm_strong:

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5721#-:-LiI7yc92lUpwBlA0Hfh:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5735, opened 2019-07-01
fix(offlinemsg): Fix offline message dispatching on history load

### LINKED ISSUE #5734: Loaded faux offline messages are sometimes not delivered since refactor
##### Brief Description
qTox version: v1.16.3+
Commit hash:  e7e6bff318a995cf4fe80b7660a67a4148053046
toxcore:  v0.2.10
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Create a new profile
2. Add another profile, don't accept the friend invite
3. Send an offline message
4. Restart qTox

##### Observed Behavior
1. Before restarting, the message is pending. Spinner is spinning, message is in both `history` table and `faux_offline_pending`.
2. After restarting, the message is not pending. There is no spinner, only  a timestamp. The message is not delivered to the friend once we see them come online. The message is still correctly present in the `faux_offline_pending` table.

##### Expected Behavior
1. Faux offline messages are displayer as pending in chat, and sent when we see the friend come online

##### Additional Info
This bug has been hit during normal use, but I couldn't repro reliably until this case. Bug was only seen after @sphaerophoria's chat refactor.

In ChatHistory. I see this in gdb, which looks like an issue:
```
Thread 1 "qtox" hit Breakpoint 2, ChatHistory::loadHistoryIntoSessionChatLog (this=0x5555560f7da0, start=...) at /home/abilinski/builds/qTox/src/model/chathistory.cpp:365                                 
365                 if (isComplete) {                                                                                    
(gdb) p message.isSent                                                                                                             
$15 = false                                                                                                              
(gdb) p isComplete                                                                                                                                                                                         
$16 = true                                                                                                                                                                                                 
```

sphaerophoria:
Fixes #5734 

I apologize, this path had several bugs. I remember testing this case but I either touched the code more since then or just remembered incorrectly.

Fixes are
* Fix callback hookup order in ChatHistory
* Add correct call to SessionChatLog to insert an unfinished message
* Fix incorrect logic in parsing History database response

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5735)

### DIFF
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -94,9 +94,10 @@ ChatHistory::ChatHistory(Friend& f_, History* history_, const ICoreIdHandler& co
     , settings(settings_)
     , coreIdHandler(coreIdHandler)
 {
-    connect(&messageDispatcher, &IMessageDispatcher::messageSent, this, &ChatHistory::onMessageSent);
     connect(&messageDispatcher, &IMessageDispatcher::messageComplete, this,
             &ChatHistory::onMessageComplete);
+    connect(&messageDispatcher, &IMessageDispatcher::messageReceived, this,
+            &ChatHistory::onMessageReceived);
 
     if (canUseHistory()) {
         // Defer messageSent callback until we finish firing off all our unsent messages.
@@ -105,8 +106,7 @@ ChatHistory::ChatHistory(Friend& f_, History* history_, const ICoreIdHandler& co
     }
 
     // Now that we've fired off our unsent messages we can connect the message
-    connect(&messageDispatcher, &IMessageDispatcher::messageReceived, this,
-            &ChatHistory::onMessageReceived);
+    connect(&messageDispatcher, &IMessageDispatcher::messageSent, this, &ChatHistory::onMessageSent);
 
     // NOTE: this has to be done _after_ sending all sent messages since initial
     // state of the message has to be marked according to our dispatch state
@@ -383,11 +383,12 @@ void ChatHistory::loadHistoryIntoSessionChatLog(ChatLogIdx start) const
 
             if (isComplete) {
                 auto chatLogMessage = ChatLogMessage{true, processedMessage};
-                sessionChatLog.insertMessageAtIdx(currentIdx, sender, message.dispName, chatLogMessage);
+                sessionChatLog.insertCompleteMessageAtIdx(currentIdx, sender, message.dispName,
+                                                          chatLogMessage);
             } else {
-                // If the message is incomplete we have to pretend we sent it to ensure
-                // sessionChatLog state is correct
-                sessionChatLog.onMessageSent(dispatchedMessageIt.key(), processedMessage);
+                auto chatLogMessage = ChatLogMessage{false, processedMessage};
+                sessionChatLog.insertIncompleteMessageAtIdx(currentIdx, sender, message.dispName,
+                                                            chatLogMessage, dispatchedMessageIt.key());
             }
             break;
         }
--- a/src/model/sessionchatlog.cpp
+++ b/src/model/sessionchatlog.cpp
@@ -284,8 +284,8 @@ std::vector<IChatLog::DateChatLogIdxPair> SessionChatLog::getDateIdxs(const QDat
     return ret;
 }
 
-void SessionChatLog::insertMessageAtIdx(ChatLogIdx idx, ToxPk sender, QString senderName,
-                                        ChatLogMessage message)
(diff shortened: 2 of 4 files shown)


sudden6:
Do you think it would be possible to add a test case so that this won't happen again?

sphaerophoria:
Yeah I was thinking about that. I wouldn't want to mock History I think, since that's where a lot of the bug potential could come from. I think the easiest thing to do would be to alter RawDatabase to somehow let us work with a database in ram that we can seed with whatever we want. I'm not sure if our sql lib allows that though.

Diadlo (changes requested):
I don't understand, why you don't want to mock `History`? Because it can be the root of some bugs? If so, it doesn't matter for `ChatHistory` test. But how to test `History` it's interesting question

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5735#-:-LijHCQj06AyZRXx0djG:b-e6pise)* status: 0 of 1 LGTMs obtained

___
*[src/model/chathistory.cpp, line 385 at r1](https://reviewable.io/reviews/qtox/qtox/5735#-LijGZ1XE5TvStXzugVY:-LijGZ1Y3SiGgmuYBBob:b55mdq9) ([raw file](https://github.com/qtox/qtox/blob/dc25183bffede1469a199f871281d6568da0e5fb/src/model/chathistory.cpp#L385)):*
> ```cpp
> 
>             if (isComplete) {
>                 auto chatLogMessage = ChatLogMessage{true, processedMessage};
> ```

Looks like the first line of `if` and `else` are the ~same
`auto chatLogMessage = ChatLogMessage{isComplete, processedMessage};`

Diadlo (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5735)* status: 0 of 1 LGTMs obtained

___
*[src/model/sessionchatlog.h, line 51 at r1](https://reviewable.io/reviews/qtox/qtox/5735#-LijI1Bg00qVELSTfDEr:-LijI1Bg00qVELSTfDEs:b-f735x9) ([raw file](https://github.com/qtox/qtox/blob/dc25183bffede1469a199f871281d6568da0e5fb/src/model/sessionchatlog.h#L51)):*
> ```objc
>     void insertCompleteMessageAtIdx(ChatLogIdx idx, ToxPk sender, QString senderName,
>                                     ChatLogMessage message);
>     void insertIncompleteMessageAtIdx(ChatLogIdx idx, ToxPk sender, QString senderName,
> ```

const reference for this function and the previous
___
*[src/persistence/history.cpp, line 630 at r1](https://reviewable.io/reviews/qtox/qtox/5735#-LijI7-i5AtKHYb0-8ux:-LijI7-i5AtKHYb0-8uy:bsw3iuk) ([raw file](https://github.com/qtox/qtox/blob/e7e6bff318a995cf4fe80b7660a67a4148053046/src/persistence/history.cpp#L630)):*
> ```cpp
>         auto display_name = QString::fromUtf8(row[4].toByteArray().replace('\0', ""));
>         auto sender_key = row[5].toString();
>         if (row[6].isNull()) {
> ```

Why now it can't be null?

sudden6:
@sphaerophoria can you rebase?

anthonybilinski:
I'm going to close this pending further investigation, accidentally running it on my main profile just dumped a bunch of chat logs to the wrong friends.

jessica181920:
I strongly suggest we branch pre-refactor for 1.17 as this is a pretty severe issue jeopardizing the security of qTox, and the refactor is too buggy to ship in its current state.

And please do yourself a favor and never run this PR on your real tox profile.

anthonybilinski:
The problem is that e607e6ecb48e559ccd86b92a2f6d7dd453a41d11 has a bug in its SQL select. I don't know SQL very well so this terminoogy is going to be bad. Although the new `History::getMessagesForFriend` correctly joins, the new `History::getUnsentMessagesForFriend`, instead of joining all tables on ids and adding WHERE chat.public_key = arg, joins most tables on ids, but also joins peers table on arg public_key. I think the resultant table is every offline message without restriction, stuck together with peers entries matching the arg public_key. Changing back to the WHERE logic like above works, or otherwise joining both on id and also on pubic key should work.

I'll rebase this PR and update it with the fix + some testing tonight, and then hopefully re-open.

anthonybilinski:
This PR has been rebased to master tip and on top of a message selection fix, to avoid the above problem. I'll address remaining review comments and continuing the review here: https://github.com/qTox/qTox/pull/5773.

PR #5738, opened 2019-07-02
fix(search): application crash when starting a search

bodwok:
The application crashes if:
1. no messages in chat (clear)
2. try to do a backward search

I think this is because std::prev requires iterator which can be decremented
Added a check of this requirement and fixed crash

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5738)


Diadlo (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5738#-:-LinQU921eEbqLoxy1r-:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5742, opened 2019-07-06
Disable IPC in FreeBSD jails because the semget system call is intentionally disabled there

yurivict:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5742)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -39,8 +39,13 @@
 #include <QMutexLocker>
 
 #include <ctime>
+#include <memory>
 #include <sodium.h>
 #include <stdio.h>
+#if defined(__FreeBSD__) || defined(__DragonFly__)
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#endif
 
 #if defined(Q_OS_OSX)
 #include "platform/install_osx.h"
@@ -216,14 +221,36 @@ int main(int argc, char* argv[])
                            QObject::tr("Starts new instance and opens the login screen.")));
     parser.process(*a);
 
-    uint32_t profileId = settings.getCurrentProfileId();
-    IPC ipc(profileId);
-    if (!ipc.isAttached()) {
-        qCritical() << "Can't init IPC";
-        return EXIT_FAILURE;
-    }
+    bool doIpc = true;
+    std::unique_ptr<IPC> ipc;
 
-    QObject::connect(&settings, &Settings::currentProfileIdChanged, &ipc, &IPC::setProfileId);
+#if defined(__FreeBSD__) || defined(__DragonFly__)
+    // IPC can't be used in FreeBSD jails because semget(2) is intentionally disabled in jails.
+    // Determine if we run in a FreeBSD jail, and disable IPC in this case.
+    {
+        int injail;
+        size_t size = sizeof(injail);
+        if (::sysctlbyname("security.jail.jailed", &injail, &size, NULL, 0) == 0) {
+          if (injail != 0) {
+            doIpc = false;
+            qInfo() << "Running in a FreeBSD jail: IPC is disabled because the semget system call is not available";
+          }
+        } else {
+          qWarning() << "Unable to determine the sysctl security.jail.jailed value";
+        }
+    }
+#endif
+
+    if (doIpc) {
+        uint32_t profileId = settings.getCurrentProfileId();
+        ipc.reset(new IPC(profileId));
+        if (!ipc->isAttached()) {
+            qCritical() << "Can't init IPC";
+            return EXIT_FAILURE;
+        }
+
+        QObject::connect(&settings, &Settings::currentProfileIdChanged, ipc.get(), &IPC::setProfileId);
+    }
 
     // For the auto-updater
     if (sodium_init() < 0) {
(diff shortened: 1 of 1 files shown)


sudden6 (changes requested):
The build on TravisCI is failing because you don't follow the commit message format

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742#-:-LjNYPJcBGQiYqJVwTUy:b9z1jbq)* status: 0 of 1 LGTMs obtained

___
*[src/main.cpp, line 225 at r1](https://reviewable.io/reviews/qtox/qtox/5742#-LjNYGXP1sp8ZzHbqADl:-LjNYGXP1sp8ZzHbqADm:b-2wasyi) ([raw file](https://github.com/qtox/qtox/blob/2a860057273d80792cf41a092023588477e223f3/src/main.cpp#L225)):*
> ```cpp
> 
>     bool doIpc = true;
>     std::auto_ptr<IPC> ipc;
> ```

`auto_ptr` is deprecated, I think `unique_ptr` is appropriate here

yurivict:
Ok, I will fix this.

yurivict:
I fixed ```std::unique_ptr``` and the commit message.

sudden6:
@yurivict commit message still doesn't fit: https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#scope

sudden6 (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742)* status: 0 of 1 LGTMs obtained

___
*[src/main.cpp, line 225 at r1](https://reviewable.io/reviews/qtox/qtox/5742#-LjNYGXP1sp8ZzHbqADl:-LjNlnJw3IiQ7styzWn_:bbdbug1) ([raw file](https://github.com/qtox/qtox/blob/2a860057273d80792cf41a092023588477e223f3/src/main.cpp#L225)):*

`auto_ptr` is deprecated, I think `unique_ptr` is appropriate here
</blockquote></details>

ok, now

sudden6 (approved):
code :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742#-:-LjNlyFq5FkJz_8L4k9J:b8012cc)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (approved):
Isn't this issue not really specific to BSD? More generally, IPC failing is non-critical, so should be handled as such, whether because of a BSD jail, another OS jail, or another reason entirely.

If so, instead of ifdefing FreeBSD and checking if we're in a jail, can we just not exit failure on IPC failure, demote log to a warning, and continue execution without IPC?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742#-:-LjXfXGV97Qgit1smxto:b-weklc3)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (changes requested):
oops I meant to block on this question ^

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742#-:-Lk60gFo4rHdLD6CRfit:bw1r059)* status: :shipit: complete! 1 of 1 LGTMs obtained

yurivict:
> Isn't this issue not really specific to BSD?

This is a possible next step improvement. IPC through Qt fails in a non-graceful way, with some cryptic messages, so detecting a FreeBSD jail is a clean way out.

If there is a way to cleanly fail in case of other IPC failures - this should also be implemented, in addition to this.

anthonybilinski (changes requested):
During constructor IPC::IPC we call globalMemory.attach() and check the return, per Qt docs: https://doc.qt.io/qt-5/qsharedmemory.html#attach this states whether IPC was successful. In main.cpp we
```
    if (!ipc.isAttached()) {
        qCritical() << "Can't init IPC";
        return EXIT_FAILURE;
    }
```
can we just handle that `isAttached` differently and not exit failure - instead avoid doing IPC later in main? This seems like a robust, platform-agnostic way to detect when IPC won't work.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742#-:-LkBApYM6anQv1kI02uy:b26t86m)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5742)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/main.cpp, line 381 at r2](https://reviewable.io/reviews/qtox/qtox/5742#-LnQlsEM3dLpb3_qcFA8:-LnQlsEM3dLpb3_qcFA9:b3ydy77) ([raw file](https://github.com/qtox/qtox/blob/c3e4e82049b476de5453bc2c654c19d920e908a3/src/main.cpp#L381)):*
> ```cpp
> 
>     // Start to accept Inter-process communication
>     if (doIpc) {
> ```

just a note, that in the case where doIpc is false for reasons other than IPC being disabled, such as the -p option being passed, we will then not register IPC callbacks in this case, which is undesired. Registering IPC events should only depend on IPC being available.

anthonybilinski:
I created https://github.com/qTox/qTox/pull/5808 based off of this review to run IPC based on `ipc.isAttached` instead of checking if we're in a jail. It's a smaller simpler change that handles more generic failures, I think.

anthonybilinski:
Closing since alternate implementation https://github.com/qTox/qTox/pull/5808 is merged.

Issue #5643, opened 2019-05-03
A(n) (Auto) Updater for Linux (AppImages) using delta update


antony-jr:
##### Brief Description

Currently auto updater is only available for windows but not for Linux , specifically qTox AppImages does not support update. But AppImage does support updating officially using zsync's algorithm. Zsync algorithm is used to do the delta update , the entire update only depends on github releases or a simple http server with http range requests.

So I think it would be nice if qTox implements an auto update for AppImage releases.

**I would be happy to work on this if you would like that**.

#### Additional Information

AppImage update can be implemented using [**AppImageUpdaterBridge**](https://github.com/antony-jr/AppImageUpdaterBridge) (*written in C++ using Qt5* which only depends on Qt base) (Made by me) or you could use the [official AppImage update library](https://github.com/AppImage/AppImageUpdate) also written in C++ but does not depend on Qt so it uses curl and other stuff for networking.

I highly recommend using my library since it only depends on Qt and thus does not add extra dependency for your application. The official documentation is [here](https://antony-jr.github.io/AppImageUpdaterBridge) 

For a reference you can take a look at this application which uses AppImageUpdaterBridge to do the delta update for the appimage -> https://gitlab.com/technic93/e2designer/-/jobs/202568868/artifacts/raw/build/e2designer-v0.1-365-g80ca747-x86_64.AppImage (**To start the updater , Help -> Check for update**)

sudden6:
Hi,

I read through your proposal and took a look at the library you linked, both look very interesting.
A few questions:

- You seem to re-implement the zsync algorithm in your library, why?
- Does your library support using proxies for the download?

antony-jr:
> You seem to re-implement the zsync algorithm in your library, why?

Because the zsync algorithm is old and the sources are also old(It was programed in C with hard coded http client). It also did not have support for modern http servers like what github uses. **Mainly I wanted to re-implement it in C++ and Qt to make it more maintainable and portable , this library was specially created for Qt Applications. Unlike the official library and old sources , this library respects the Qt Event Loop. Therefore using this library will not break anything and does not add any other issues.**.

> Does your library support using proxies for the download

[...]

PR #5745, opened 2019-07-11
refactor(MacOSX): switch to Qt 5.13 compliant methods for OSX

Kribylet:
This commit aims to bring Mac OSX specific code into Qt 5.13 compliance.

This PR will be WIP until ready to utilize TravisCI as I am unable to test Mac OSX functionality directly (Apple licenses forbid running OSX VM on non-Apple hardware.)

- [ ] Commits follow our [git commit guidelines](https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#git-commit-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5745)

PR #5747, opened 2019-07-13
feat/appimage delta updater

antony-jr:

> This PR implements the delta updater as mentioned in #5643 for qTox AppImages. The delta updater is only used for AppImage builds and does not break the source code for other platforms. This delta updater is integrated along with UpdateCheck class and uses the existing UI for notifying the update available button. Without the CMake build options , the delta updater library is never touched.

This PR is a re-post of https://github.com/qTox/qTox/pull/5690, This PR was opened because Travis-CI was not tracking the old PR and this also to fixes merge conflicts that was present in the old PR. This is a fresh PR which does the same thing as https://github.com/qTox/qTox/pull/5690

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 option(PLATFORM_EXTENSIONS "Enable platform specific extensions, requires extra dependencies" ON)
 option(USE_FILTERAUDIO "Enable the echo canceling backend" ON)
 option(UPDATE_CHECK "Enable automatic update check" ON)
+option(APPIMAGE_UPDATER_BRIDGE "Use AppImageUpdaterBridge to do the update" OFF)
 option(USE_CCACHE "Use ccache when available" ON)
 option(SPELL_CHECK "Enable spell cheching support" ON)
 option(SVGZ_ICON "Compress the SVG icon of qTox" ON)
@@ -690,6 +691,13 @@ endif()
 
 if(${UPDATE_CHECK})
     add_definitions(-DUPDATE_CHECK_ENABLED=1)
+    if(APPIMAGE_UPDATER_BRIDGE_SRC_DIR AND APPIMAGE_UPDATER_BRIDGE_BUILD_DIR)
+        message(STATUS "using AppImageUpdaterBridge")
+        add_definitions(-DAPPIMAGE_UPDATER_BRIDGE_ENABLED=1)
+        include_directories(${APPIMAGE_UPDATER_BRIDGE_SRC_DIR})
+    else()
+        message(STATUS "NOT using AppImageUpdaterBridge")
+    endif()
     set(${PROJECT_NAME}_SOURCES ${${PROJECT_NAME}_SOURCES}
         src/net/updatecheck.cpp
         src/net/updatecheck.h)
@@ -729,10 +737,18 @@ add_library(${PROJECT_NAME}_static
   ${${PROJECT_NAME}_SOURCES}
   ${${PROJECT_NAME}_QM_FILES}
   ${${PROJECT_NAME}_RESOURCES})
-target_link_libraries(${PROJECT_NAME}_static
-  ${CMAKE_REQUIRED_LIBRARIES}
-  ${ALL_LIBRARIES})
 
+if(APPIMAGE_UPDATER_BRIDGE_BUILD_DIR AND APPIMAGE_UPDATER_BRIDGE_SRC_DIR AND UPDATE_CHECK)
+    target_link_libraries(${PROJECT_NAME}_static
+    ${CMAKE_REQUIRED_LIBRARIES}
+    ${ALL_LIBRARIES}
+    ${APPIMAGE_UPDATER_BRIDGE_BUILD_DIR}/libAppImageUpdaterBridge.a)
+else()
+    target_link_libraries(${PROJECT_NAME}_static
+    ${CMAKE_REQUIRED_LIBRARIES}
+    ${ALL_LIBRARIES})
+endif()
+  
 add_executable(${PROJECT_NAME}
   WIN32
   MACOSX_BUNDLE
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,11 @@ if(APPLE)
     /usr/local/opt/openal-soft/lib/pkgconfig:$ENV{PKG_CONFIG_PATH})
 endif()
 
+if(${APPIMAGE_UPDATER_BRIDGE})
+  set(ENV{PKG_CONFIG_PATH}
+    /usr/local/lib/pkgconfig:$ENV{PKG_CONFIG_PATH})
+endif()
+
 execute_process(
   COMMAND brew --prefix qt5
   OUTPUT_VARIABLE QT_PREFIX_PATH
@@ -691,12 +696,16 @@ endif()
 
 if(${UPDATE_CHECK})
     add_definitions(-DUPDATE_CHECK_ENABLED=1)
-    if(APPIMAGE_UPDATER_BRIDGE_SRC_DIR AND APPIMAGE_UPDATER_BRIDGE_BUILD_DIR)
-        message(STATUS "using AppImageUpdaterBridge")
-        add_definitions(-DAPPIMAGE_UPDATER_BRIDGE_ENABLED=1)
-        include_directories(${APPIMAGE_UPDATER_BRIDGE_SRC_DIR})
+    if(${APPIMAGE_UPDATER_BRIDGE})
+        search_dependency(AIUB  PACKAGE AppImageUpdaterBridge)
+        if(AIUB_FOUND)
+		message(STATUS "using AppImageUpdaterBridge")
+                add_definitions(-DAPPIMAGE_UPDATER_BRIDGE_ENABLED=1)
+	else()
+                message(STATUS "cannot find AppImageUpdaterBridge , ignoring cmake flag")
+        endif()
     else()
-        message(STATUS "NOT using AppImageUpdaterBridge")
+        message(STATUS "not using AppImageUpdaterBridge")
     endif()
     set(${PROJECT_NAME}_SOURCES ${${PROJECT_NAME}_SOURCES}
         src/net/updatecheck.cpp
@@ -737,18 +746,10 @@ add_library(${PROJECT_NAME}_static
   ${${PROJECT_NAME}_SOURCES}
   ${${PROJECT_NAME}_QM_FILES}
   ${${PROJECT_NAME}_RESOURCES})
+target_link_libraries(${PROJECT_NAME}_static
+  ${CMAKE_REQUIRED_LIBRARIES}
+  ${ALL_LIBRARIES})
 
-if(APPIMAGE_UPDATER_BRIDGE_BUILD_DIR AND APPIMAGE_UPDATER_BRIDGE_SRC_DIR AND UPDATE_CHECK)
-    target_link_libraries(${PROJECT_NAME}_static
-    ${CMAKE_REQUIRED_LIBRARIES}
-    ${ALL_LIBRARIES}
-    ${APPIMAGE_UPDATER_BRIDGE_BUILD_DIR}/libAppImageUpdaterBridge.a)
-else()
-    target_link_libraries(${PROJECT_NAME}_static
-    ${CMAKE_REQUIRED_LIBRARIES}
-    ${ALL_LIBRARIES})
-endif()
-  
 add_executable(${PROJECT_NAME}
   WIN32
   MACOSX_BUNDLE
--- a/appimage/build-appimage.sh
+++ b/appimage/build-appimage.sh
@@ -31,7 +31,7 @@ readonly DEBUG="$1"
 
 # Set this to True to upload the PR version of the 
 # AppImage to transfer.sh for testing.
-readonly UPLOAD_PR_APPIMAGE="True"
+readonly UPLOAD_PR_APPIMAGE="False"
 
 # Fail out on error
 set -exo pipefail
--- a/appimage/build-appimage.sh
+++ b/appimage/build-appimage.sh
@@ -2,7 +2,7 @@
 
 # MIT License
 #
-# Copyright © 2018 by The qTox Project Contributors
+# Copyright © 2019 by The qTox Project Contributors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
--- a/appimage/build-appimage.sh
+++ b/appimage/build-appimage.sh
@@ -2,7 +2,7 @@
 
 # MIT License
 #
-# Copyright © 2019 by The qTox Project Contributors
+# Copyright © 2018 by The qTox Project Contributors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -28,6 +28,10 @@
 # stay open to poke around in the filesystem.
 
 readonly DEBUG="$1"
+
+# Set this to True to upload the PR version of the 
+# AppImage to transfer.sh for testing.
+readonly UPLOAD_PR_APPIMAGE="True"
 
 # Fail out on error
 set -exo pipefail
@@ -57,16 +61,41 @@ then
         /bin/bash
 else
     docker run --rm \
+        -e CIRP_GITHUB_REPO_SLUG \
+        -e TRAVIS_EVENT_TYPE \
+        -e TRAVIS_COMMIT \
+        -e TRAVIS_TAG \
         -v $PWD:/qtox \
         -v $PWD/output:/output \
         debian:stretch-slim \
         /bin/bash -c "/qtox/appimage/build.sh"
 fi
 
+
 # use the version number in the name when building a tag on Travis CI
 if [ -n "$TRAVIS_TAG" ]
 then
+    # the aitool should have written the appimage in the same name
+    # as below so no need to move things , it should have also written
+    # the .zsync meta file as the given name below with .zsync
+    # extension.
     readonly OUTFILE=./output/qTox-"$TRAVIS_TAG".x86_64.AppImage
-    mv ./output/*.AppImage "$OUTFILE"
+ 
+    # just check if the files are in the right place
+    eval "ls $OUTFILE"
+    eval "ls $OUTFILE.zsync"
+
     sha256sum "$OUTFILE" > "$OUTFILE".sha256
+else
+    if [ "$UPLOAD_PR_APPIMAGE" == "True" ]
+    then
+        # upload PR builds to test them.
+        echo "Uploading AppImage to transfer.sh"
+        curl --upload-file "./output/qTox-$TRAVIS_COMMIT-x86_64.AppImage" \
+            "https://transfer.sh/qTox-$TRAVIS_COMMIT-x86_64.AppImage" > ./upload
+        echo "$(cat ./upload)"
+        echo -n "$(cat ./upload)\\n" >> ./uploaded-to
+        rm -rf ./upload ./uploaded-to
+        sha256sum "./output/qTox-$TRAVIS_COMMIT-x86_64.AppImage"
+    fi
 fi
--- a/appimage/build.sh
+++ b/appimage/build.sh
@@ -2,7 +2,7 @@
 
 # MIT License
 #
-# Copyright © 2019 by The qTox Project Contributors
+# Copyright © 2018 by The qTox Project Contributors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -22,6 +22,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+
 # Fail out on error
 set -exuo pipefail
 
@@ -39,11 +40,31 @@ readonly AITOOL_BUILD_DIR="$BUILD_DIR"/aitool
 readonly SQLCIPHER_BUILD_DIR="$BUILD_DIR"/sqlcipher
 # ldqt binary
 readonly LDQT_BIN="/usr/lib/x86_64-linux-gnu/qt5/bin/linuxdeployqt"
+# aitool binary
+readonly AITOOL_BIN="/usr/local/bin/appimagetool"
+readonly APPRUN_BIN="/usr/local/bin/AppRun"
 readonly APT_FLAGS="-y --no-install-recommends"
 # snorenotify source
 readonly SNORE_GIT="https://github.com/KDE/snorenotify"
 # snorenotify build directory
 readonly SNORE_BUILD_DIR="$BUILD_DIR"/snorenotify
+# "appimage updater bridge" becomes aub
+readonly AUB_SRC_DIR="$BUILD_DIR"/aub
+# aub source
+readonly AUB_GIT="https://github.com/antony-jr/AppImageUpdaterBridge"
+# aub build dir
+readonly AUB_BUILD_DIR="$BUILD_DIR"/aub/build
+
+# update information to be embeded in AppImage
+if [ "cron" == "${TRAVIS_EVENT_TYPE:-}" ]
+then
+    # update information for nightly version
+    readonly NIGHTLY_REPO_SLUG=$(echo "$CIRP_GITHUB_REPO_SLUG" | tr "/" "|")
+    readonly UPDATE_INFO="gh-releases-zsync|$NIGHTLY_REPO_SLUG|ci-master-latest|qTox-*-x86_64.AppImage.zsync"
+else
+    # update information for stable version
+    readonly UPDATE_INFO="gh-releases-zsync|qTox|qTox|latest|qTox-*.x86_64.AppImage.zsync"
+fi
 
 # use multiple cores when building
 export MAKEFLAGS="-j$(nproc)"
@@ -89,6 +110,17 @@ LDFLAGS="-lcrypto"
 make
 make install
 
+# build aub into a static library and later use it in 
+# qTox
+git clone "$AUB_GIT" "$AUB_SRC_DIR"
+cd "$AUB_SRC_DIR" # we need to checkout first
+git checkout tags/v1.1.1
+mkdir $AUB_BUILD_DIR
+cd $AUB_BUILD_DIR
+cmake .. -DLOGGING_DISABLED=ON
+make
+
+
 # copy qtox source
 cp -r "$QTOX_SRC_DIR" "$QTOX_BUILD_DIR"
 cd "$QTOX_BUILD_DIR"
@@ -103,7 +135,10 @@ mkdir -p ./_build
 cd _build
 
 # need to build with -DDESKTOP_NOTIFICATIONS=True for snorenotify
-cmake -DDESKTOP_NOTIFICATIONS=True ../
+cmake -DDESKTOP_NOTIFICATIONS=True \
+      -DUPDATE_CHECK=True \
+      -DAPPIMAGE_UPDATER_BRIDGE_SRC_DIR="$AUB_SRC_DIR" \
+      -DAPPIMAGE_UPDATER_BRIDGE_BUILD_DIR="$AUB_BUILD_DIR" ../
 
 make
 
@@ -139,7 +174,7 @@ cd build
 export PKG_CONFIG_PATH=/deps/lib/pkgconfig/
 
 cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=ON \
--DAPPIMAGEKIT_PACKAGE_DEBS=ON -DUPDATE_CHECK=ON
+-DAPPIMAGEKIT_PACKAGE_DEBS=ON
 
 make
 make install
@@ -152,7 +187,18 @@ readonly QTOX_DESKTOP_FILE="$QTOX_APP_DIR"/usr/local/share/applications/*.deskto
 
 eval "$LDQT_BIN $QTOX_DESKTOP_FILE -bundle-non-qt-libs -extra-plugins=libsnore-qt5"
 
-eval "$LDQT_BIN $QTOX_DESKTOP_FILE -appimage"
+# Move the required files to the correct directory
+mv "$QTOX_APP_DIR"/usr/* "$QTOX_APP_DIR/"
+rm -rf "$QTOX_APP_DIR/usr"
+
+# this is important , aitool automatically uses the same filename in .zsync meta file.
+# if this name does not match with the one we upload , the update always fails.
+if [ -n "$TRAVIS_TAG" ]
+then
+    eval "$AITOOL_BIN -u \"$UPDATE_INFO\" $QTOX_APP_DIR qTox-$TRAVIS_TAG.x86_64.AppImage"
+else
+    eval "$AITOOL_BIN -u \"$UPDATE_INFO\" $QTOX_APP_DIR qTox-$TRAVIS_COMMIT-x86_64.AppImage"
+fi
 
 # Chmod since everything is root:root
 chmod 755 -R "$OUTPUT_DIR"
--- a/appimage/build.sh
+++ b/appimage/build.sh
@@ -137,7 +137,8 @@ cd _build
 # need to build with -DDESKTOP_NOTIFICATIONS=True for snorenotify
 cmake -DDESKTOP_NOTIFICATIONS=True \
       -DUPDATE_CHECK=True \
-      -DAPPIMAGE_UPDATER_BRIDGE=True
+      -DAPPIMAGE_UPDATER_BRIDGE=True \
+      ../
 
 make
 
--- a/appimage/build.sh
+++ b/appimage/build.sh
@@ -191,6 +191,12 @@ eval "$LDQT_BIN $QTOX_DESKTOP_FILE -bundle-non-qt-libs -extra-plugins=libsnore-q
 mv "$QTOX_APP_DIR"/usr/* "$QTOX_APP_DIR/"
 rm -rf "$QTOX_APP_DIR/usr"
 
+# copy OpenSSL libs to AppImage
+# Warning: This is hard coded to debain:stretch.
+cp /usr/lib/x86_64-linux-gnu/libssl.so* "$QTOX_APP_DIR/local/lib/"
+cp /usr/lib/x86_64-linux-gnu/libcrypt.so* "$QTOX_APP_DIR/local/lib/"
+cp /usr/lib/x86_64-linux-gnu/libcrypto.so* "$QTOX_APP_DIR/local/lib"
+
 # this is important , aitool automatically uses the same filename in .zsync meta file.
 # if this name does not match with the one we upload , the update always fails.
 if [ -n "$TRAVIS_TAG" ]
--- a/appimage/build.sh
+++ b/appimage/build.sh
@@ -2,7 +2,7 @@
 
 # MIT License
 #
-# Copyright © 2018 by The qTox Project Contributors
+# Copyright © 2019 by The qTox Project Contributors
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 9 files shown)


antony-jr:
@sudden6 I've created an easy way to test the updater, just download this [appimage](https://github.com/antony-jr/qTox-PR5690/releases/download/obsolete/qTox-e8326eefb395540418cb00d01504c1437d39fc58-x86_64.AppImage) and run it. This appimage is an old version so the update available button should show up within 20 seconds.

```
 $ wget "https://github.com/antony-jr/qTox-PR5690/releases/download/obsolete/qTox-e8326eefb395540418cb00d01504c1437d39fc58-x86_64.AppImage"
 $ chmod +x qTox-e8326eefb395540418cb00d01504c1437d39fc58-x86_64.AppImage
 $ ./qTox-e8326eefb395540418cb00d01504c1437d39fc58-x86_64.AppImage
```

**Something like this....**,

When you just open the application nothing shows up.
Screenshot_20190713_213234


You should see a log message something like this

Screenshot_20190713_213453


After few seconds, a green glow in the settings will indicate that an update is available(**this is not my design**)

Screenshot_20190713_213246

Now the about form should have an **update available** button

Screenshot_20190713_213257

Screenshot_20190713_213305


When you click the **update available** button a confirmation dialog box will appear

Screenshot_20190713_213327

When you hit **Install Update**, the delta update will start

Screenshot_20190713_213350


**Again my internet speed is not 20 MB per second but that's my disk read speed. For now between the changes, about 77 MiB was usable of total 88 MiB, so I only needed to download 11 MiB to get the latest version which is a big improvement**

antony-jr:
Also the latest build crashes and its not my fault. Just letting you know.

Screenshot_20190713_213610

**EDIT**: I think this is the commit which caused the crash -> https://github.com/qTox/qTox/commit/e607e6ecb48e559ccd86b92a2f6d7dd453a41d11#commitcomment-34292419

sudden6:
The cleanup looks good!

I found out that I didn't have the "Check for updates" setting enabled, this is why it didn't work the first time.

I tried to test again, but get the following suspicious log output from the AppImage you linked:

```
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve CRYPTO_num_locks
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve CRYPTO_set_id_callback
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve CRYPTO_set_locking_callback
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve ERR_free_strings
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve EVP_CIPHER_CTX_cleanup
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve EVP_CIPHER_CTX_init
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_new_null
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_push
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_free
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_num
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_pop_free
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve sk_value
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve SSL_library_init
[16:42:12.543 UTC] :0 : Warning: QSslSocket: cannot resolve SSL_load_error_strings
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve SSL_get_ex_new_index
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve SSLv23_client_method
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve SSLv23_server_method
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve X509_STORE_CTX_get_chain
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve OPENSSL_add_all_algorithms_noconf
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve OPENSSL_add_all_algorithms_conf
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve SSLeay
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot resolve SSLeay_version
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function SSLeay
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function CRYPTO_num_locks
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function CRYPTO_set_id_callback
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function CRYPTO_set_locking_callback
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function SSL_library_init
[16:42:12.544 UTC] :0 : Warning: QSslSocket: cannot call unresolved function SSLv23_client_method
[16:42:12.549 UTC] :0 : Warning: QSslSocket: cannot call unresolved function sk_num
[16:42:12.627 UTC] :0 : Warning: QSslSocket: cannot call unresolved function SSLv23_client_method
[16:42:12.627 UTC] :0 : Warning: QSslSocket: cannot call unresolved function SSL_library_init
```

No update available notification still.

Seems like something is missing for Qt to support SSL?

For the crash, if this is the latest master branch, please submit an issue report with backtrace.

antony-jr:
> I tried to test again, but get the following suspicious log output from the AppImage you linked:

@sudden6 That's the OpenSSL error.  It happens when the host system has unsupported OpenSSL version. To fix this we need to package the correct OpenSSL libs along with the AppImage. But I'm surprised that you never seen these errors in spite of using **QNetwork module**.

sudden6:
@antony-jr I think qTox doesn't use SSL for anything yet.

antony-jr:
>  I think qTox doesn't use SSL for anything yet.

Then we might need to package the OpenSSL libs. I will add it in the next commit. The easiest way to solve this is to just copy the OpenSSL libs from the build system itself.(works all the time, mostly)

antony-jr:
Or do you need to build it from source?

sudden6:
If it's available from the debian container we can take it, IIRC that's also done with the rest of the libs for the AppImage, right?

antony-jr:
> If it's available from the debian container

Yes it should be available there since Qt libs are installed with SSL support.

> IIRC that's also done with the rest of the libs for the AppImage, right?

Yep.

sudden6:
Ok, then it's fine.

antony-jr:
Finally the build succeeded but we can't test it anyway since all latest builds crashes. Only MacOSX builds are failing for some reason(Maybe using the latest version of Qt is the problem).

sudden6:
@antony-jr macOS builds failing is known and being worked on in https://github.com/qTox/qTox/pull/5745 you can ignore these build failures

nurupo:
>I think qTox doesn't use SSL for anything yet.

Doesn't qTox communicate with toxme's API through https? I think that was the reason why we added OpenSSL build on Windows and made the Qt Windows build aware of the built OpenSSL libs, so that Windows Qt would have ssl/tls capabilities.

nurupo:
___
*[appimage/build-appimage.sh, line 5 at r4](https://reviewable.io/reviews/qtox/qtox/5747#-LjiXTsI4WIibESwhqQU:-LjiXTsI4WIibESwhqQV:b-28y7zg) ([raw file](https://github.com/qtox/qtox/blob/6ca54dc832e15cfd55176c6b4b72a15a9c46904a/appimage/build-appimage.sh#L5)):*
> ```Shell
> # MIT License
> #
> # Copyright © 2018 by The qTox Project Contributors
> ```

I believe this year change should be reverted.

nurupo:
Interesting that linuxdeployqt doesn't package OpenSSL libs for us, I wonder if it does so on purpose?

antony-jr:
> @antony-jr macOS builds failing is known and being worked on in #5745 you can ignore these build failures

Okay.

What I'm excited about is that, the version I updated to was the version with more than **100 commits**, meaning it changed so much(even still it only downloaded at most 11 MiB). If users were to run the **nightly builds** then the data they need to download between new versions will be **2 MiB** at max. This is what the AppImage author calls the **software fluid**.

> Interesting that linuxdeployqt doesn't package OpenSSL libs for us, I wonder if it does so on purpose?

@nurupo I think OpenSSL libs are currently not packaged intentionally(Maybe). It has some complication, example some linux distro's use different location to store the certs and other stuff. OpenSSL is the biggest issue packaging Qt application which uses HTTP(S).

nurupo:
Oh, apparently Qt uses `dlopen()` on OpenSSL, so it's not a link-time dependency but a run-time one, so linuxdeployqt has no way to see QtNetwork depending on OpenSSL, thus it doesn't bundle it. This was discussed in https://github.com/probonopd/linuxdeployqt/issues/209.

antony-jr:
> Oh, apparently Qt uses dlopen() on OpenSSL, so it's not a link-time dependency but a run-time one, so linuxdeployqt has no way to see QtNetwork depending on OpenSSL, thus it doesn't bundle it.

Don't know about that but my local version qt is linked against **OpenSSL**.

```
        linux-vdso.so.1 (0x00007ffe16b6e000)
        libQt5Core.so.5 => /usr/lib/libQt5Core.so.5 (0x00007f8f3d402000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f8f3d3e1000)
        libz.so.1 => /usr/lib/libz.so.1 (0x00007f8f3d1ca000)
        libssl.so.1.1 => /usr/lib/libssl.so.1.1 (0x00007f8f3d13a000)
        libcrypto.so.1.1 => /usr/lib/libcrypto.so.1.1 (0x00007f8f3ce66000)
```

Also **qTox** AppImages have **libcrypt.so** but not libssl and libcrypto.so

antony-jr (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: 0 of 1 LGTMs obtained

___
*[appimage/build-appimage.sh, line 5 at r4](https://reviewable.io/reviews/qtox/qtox/5747#-LjiXTsI4WIibESwhqQU:-Ljiar7J65pgon6b4iQ-:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/6ca54dc832e15cfd55176c6b4b72a15a9c46904a/appimage/build-appimage.sh#L5)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

I believe this year change should be reverted.
</blockquote></details>

Done.

nurupo:
probonopd brings up a good point against bundling OpenSSL in that discussion I have linked: different distributions store CA certificates in different places, so if you bundle OpenSSL from Debian and run the AppImage on Fedora - it won't be able to find the CA certificates as Debian's OpenSSL we have bundled would look for them at `/etc/ssl/certs/ca-certificates.crt`, while Fedora has them at 
`/etc/pki/tls/certs/ca-bundle.crt`.

[Here is a list of CA certificate paths some distros have](https://gitlab.com/probono/platformissues/blob/master/README.md#certificates).

There are also [`SSL_CERT_DIR` and `SSL_CERT_FILE`](https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_default_verify_paths.html) env variables that OpenSSL uses in some cases to find CA certificates, depending if the application that uses OpenSSL allows that (not sure if Qt allows that).

The OpenSSL situation is rather complicated, huh.

nurupo:
>Don't know about that but my local version qt is linked against OpenSSL.

I'm getting a completely different output on Debian:

```
$ ldd /usr/lib/x86_64-linux-gnu/libQt5Network.so
        linux-vdso.so.1 (0x00007ffde93fd000)
        libQt5Core.so.5 => /usr/lib/x86_64-linux-gnu/libQt5Core.so.5 (0x00007f5f0fb7e000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5f0fb5d000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f5f0f93f000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f5f0f7bb000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5f0f5fa000)
        libicui18n.so.63 => /usr/lib/x86_64-linux-gnu/libicui18n.so.63 (0x00007f5f0f31f000)
        libicuuc.so.63 => /usr/lib/x86_64-linux-gnu/libicuuc.so.63 (0x00007f5f0f14e000)
        libpcre2-16.so.0 => /usr/lib/x86_64-linux-gnu/libpcre2-16.so.0 (0x00007f5f0f0d3000)
        libdouble-conversion.so.1 => /usr/lib/x86_64-linux-gnu/libdouble-conversion.so.1 (0x00007f5f0f0bc000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5f0f0b7000)
        libglib-2.0.so.0 => /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f5f0ef98000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f5f0ee15000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5f0edf9000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f5f1024d000)
        libicudata.so.63 => /usr/lib/x86_64-linux-gnu/libicudata.so.63 (0x00007f5f0d409000)
        libpcre.so.3 => /lib/x86_64-linux-gnu/libpcre.so.3 (0x00007f5f0d395000)
```

antony-jr:
> There are also SSL_CERT_DIR and SSL_CERT_FILE env variables that OpenSSL uses in some cases to find CA certificates, depending if the application that uses OpenSSL allows that (not sure if Qt allows that).

IIRC , **This is a issue only if you use libgnutls**. This has never been a issue for me.

> so if you bundle OpenSSL from Debian and run the AppImage on Fedora - it won't be able to find the CA certificates as Debian's OpenSSL we have bundled would look for them at /etc/ssl/certs/ca-certificates.crt, while Fedora has them at
/etc/pki/tls/certs/ca-bundle.crt.

No, I've tested my other AppImages which uses OpenSSL but it never had this issue. **So I think if you are using libgnutls then you have to patch it somehow to look into different locations one by one(I think I've seen someone do that before)**

nurupo:
It's not just gnutls. probonopd has opened this issue against OpenSSL https://github.com/openssl/openssl/issues/7481

I have also faced similar issue with Debian's OpenSSL when packaging [static toxic](https://build.tox.chat/job/toxic-no-x11-musl_build_linux_x86-64_release/lastSuccessfulBuild/artifact/toxic-no-x11-musl_build_linux_x86-64_release.tar.xz), I had to make a startup script that checks for different CA certificate store filepaths since the OpenSSL wouldn't work otherwise on, for example, Fedora.

antony-jr:
> I have also faced similar issue with Debian's OpenSSL when packaging static toxic, I had to make a startup script that checks for different CA certificate store filepaths since the OpenSSL wouldn't work otherwise on, for example, Fedora.

@nurupo So do we need to write a similar script for qTox.AppImage startup too?

nurupo:
Depends on how Qt uses OpenSSL. Qt could already try the different paths on its own, in which case - no. I'm also not sure if a `dlopen()`'ed OpenSSL would be able to see the env variable being set, never tried that. Qt could also unset those env variables for some reason... The best way to find out would be to run qTox AppImage (built on Debian/Ubuntu) on a Fedora system, do something that would make qTox use OpenSSL (e.g. check for an update or use a ToxMe service) and check if that action succeeded.

antony-jr:
> The best way to find out would be to run qTox AppImage (built on Debian/Ubuntu) on a Fedora system, do something that would make qTox use OpenSSL (e.g. check for an update or use a ToxMe service) and check if that action succeeded.

I've already made the required changes assuming it would work, all we need is a build with these changes that works(some commits before this PR causes a crash at startup of the program) to do that. For now we have to ignore these and only focus on the logic and code style. Maybe merge this and hopefully a working **nightly build can be used to test all functionality**. We always have an option to revoke all commits. Or I could create a new repo which builds an AppImage with these changes merged with a stable tree of qTox repo and test it(But this is a tedious work).

sudden6:
@nurupo I agree with @antony-jr here, it's already a huge improvement that OpenSSL is included at all. No one noticed before that toxme.io isn't working in the AppImage anyway.

To merge this PR I just would like to have the update process tested once by someone who is not the PR author, but it looks like that depends a bit on fixing the crash you seem to have.

nurupo:
Shipping it as it is and fixing OpenSSL later (maybe if users complain, so that we have someone to test the fixes?) sounds good to me, as it should at least work on Debian, Ubuntu and some other distros as it is.

@antony-jr do you know of a way to make AppImage run some startup sh script instead of the qTox binary? You seem knowledgeable about AppImage and that's probably the first thing we would try when trying to fix OpenSSL CA certificate path in the future, so I want to ask this while you are still here, for the future reference.

antony-jr:
> @antony-jr do you know of a way to make AppImage run some startup sh script instead of the qTox binary? You seem knowledgeable about AppImage and that's probably the first thing we would try when trying to fix OpenSSL CA certificate path in the future, so I want to ask this while you are still here, for the future reference.

@nurupo its easy you just replace the **AppRun** binary with a bash script called **AppRun** and start the binary from that script. A lot of applications do that, especially python programs packaged with AppImages. 

**EDIT**:
For example, this is what I use for one of my AppImage which packages **[insight](https://github.com/antony-jr/insight)**
```
#!/usr/bin/env bash
export PATH=$PATH:$(pwd)/bin
if [ $# -eq 0 ]
    then
        insight
        exit
fi 
insight $*
exit
```

antony-jr:
> To merge this PR I just would like to have the update process tested once by someone who is not the PR author, but it looks like that depends a bit on fixing the crash you seem to have.

I can do something about that. Just wait a bit.

antony-jr:
Okay @sudden6 I've done the tedious work, here is the version which has **OpenSSL** bundled with it, https://github.com/antony-jr/qTox-AppImage/releases (I copied the changed files to this tree -> ```203720275c6b9926719dabb374be3469e0ea04a8```), I tested this on a live boot of Fedora on a VM but it complained libjack.so is missing **probably because I was using a very very old version of Fedora ISO**. (I can't install it through the package manager because if I did that Fedora would update the entire system which takes a lot of time. I suggest someone with Fedora should test it)

I also tested it on Xubuntu 19.x live boot in a VM and it worked like a charm. To test if it ran without the bundled OpenSSL. I extracted the AppImage and deleted the packaged openssl libs. Now when I ran the AppImage it complained about the incompatible functions from OpenSSL again. So **in conclusion the bundled OpenSSL is working great and without it some distro could not update or use Toxme.**

Also if you want to see it update you have to differ the SHA1 hash of the local or the remote AppImage. To change the SHA1 hash of the local AppImage you can append some garbage data to it to trigger a update. Like this,

```
 $ echo "garbage data" >> qTox-*-x86_64.AppImage
 $ ./qTox-*-x86_64.AppImage # should trigger a update
```

antony-jr:
Okay I've tested the AppImage in a latest version of fedora but still it does not work because it expects **libjack.so.0**,is this lib Debian specific?

But looking through the exclude list, this lib is intentionally excluded from packaging, https://github.com/AppImage/pkg2appimage/blob/5c36cfaf2ce9785cd1f23b7fda99c54a78e2a718/excludelist#L164

This is what I get when I execute the AppImage in Fedora Workstation 30,

broken


**So to fix this should I bundle libjack.so.0 too? Why is this error not reported by any Fedora users?**.

sudden6:
This `libjack.so` is a library for the JACK audio daeamon. qTox shouldn't have any hard dependency on it though. No idea why this happens on Fedora though.

I tried to test `qTox-66474969797f33d98c1698c3d2f083de2787e687-x86_64.AppImage`, but it's still not working for me. I don't get any error messages this time, but no update notification either. I waited ~1min and did have the setting to check for updates enabled this time.

My system is `openSUSE Leap 15.0`, so also RPM based like Fedora, maybe they package something in an unexpected way?

antony-jr:
@sudden6 Sorry just now deleted the continuous tag , so the update should have failed, please wait until current build succeeds.

antony-jr:
This build is currently uploading the next version of the AppImage, https://travis-ci.org/antony-jr/qTox-AppImage/builds/558775779 so when you check for update in middle of it, its gonna fail.

sudden6:
Ok, I'll try again once this build is finished.

antony-jr:
@sudden6 It should definitely work now. Use the same old version.

sudden6 (approved):
worked perfect now! :lgtm_strong: 

Merging is blocked by #5745 though and I'd appreciate if @nurupo can also take a final look before merge.

> Reviewed 7 of 9 files at r1, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747#-:-LjoRgk3846x0E2vFaFT:bqrp19c)* status: :shipit: complete! 1 of 1 LGTMs obtained

antony-jr:
Okay, bundling **libjack.so*** fixed the AppImage and it can now run in any Fedora Workstation, test it with the latest build(continuous build) -> https://github.com/antony-jr/qTox-AppImage/releases

@nurupo I also did a full update and there was no issue with OpenSSL. It worked like a charm.

(The screen is stretched because I used the scale mode without installing the guest VM drivers.)

Screenshot_20190715_125853

antony-jr:
Okay bundling **libjack.so*** fixed the AppImage and it can now run in any Fedora Workstation, test it with the latest build(continuous build) -> https://github.com/antony-jr/qTox-AppImage/releases

@nurupo I also did a full update and there was no issue with OpenSSL. It worked like a charm.

The screen is stretched because I used the scale mode without installing the guest VM drivers.

Screenshot_20190715_125853

nurupo (commented):
> Reviewed 1 of 9 files at r1, 1 of 1 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 29 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjoofmM9PcCEPBkNIEX:-LjoofmM9PcCEPBkNIEY:bn39kqs) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L29)):*
> ```CMake
> option(USE_FILTERAUDIO "Enable the echo canceling backend" ON)
> option(UPDATE_CHECK "Enable automatic update check" ON)
> option(APPIMAGE_UPDATER_BRIDGE "Use AppImageUpdaterBridge to do the update" OFF)
> ```

This doesn't seem to be used anywhere?
___
*[CMakeLists.txt, line 745 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-Ljomql_3zYOmIfY2Uv-:-Ljomql_3zYOmIfY2Uv0:bct3e22) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L745)):*
> ```CMake
>     ${CMAKE_REQUIRED_LIBRARIES}
>     ${ALL_LIBRARIES}
>     ${APPIMAGE_UPDATER_BRIDGE_BUILD_DIR}/libAppImageUpdaterBridge.a)
> ```

Would be nice if AIUB used the same `search_dependency()` that is used for finding all of the dependencies, instead of having to provide `APPIMAGE_UPDATER_BRIDGE_*` the paths manually. However, it doesn't look like AIUB provides pkg-config files... Well, you could still make `search_dependency()` work even without AIUB providing pkg-config files, if you `make install` AIUB into some dir and pass the path to that dir as`-DCMAKE_FIND_ROOT_PATH=/path/to/AIUBs/install/prefix` when calling qTox's cmake.
___
*[CMakeLists.txt, line 751 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-Ljoh8NODKshh9nVSVI4:-Ljoh8NODKshh9nVSVI5:b-88s9jt) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L751)):*
> ```CMake
>     ${ALL_LIBRARIES})
> endif()
>   
> ```

Trailing space.
___
*[appimage/build.sh, line 25 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjojdIk50Azn6Ubeb0O:-LjojdIk50Azn6Ubeb0P:by9i1kb) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/appimage/build.sh#L25)):*
> ```Shell
> # THE SOFTWARE.
> ```

Unnecessary new line.
___
*[appimage/build.sh, line 113 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjohQon6XyKEquSCMQ6:-LjohQon6XyKEquSCMQ7:b-88s9jt) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/appimage/build.sh#L113)):*
> ```Shell
> make install
> 
> # build aub into a static library and later use it in 
> ```

Trailing space.
___
*[appimage/build.sh, line 155 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjoiGyC2BiHiw-8esC3:-LjoiGyC2BiHiw-8esC4:bllzk3n) ([raw file](https://github.com/qtox/qtox/blob/b445031912ae943d7580ea21f6813e9f71bc03ee/appimage/build.sh#L155)):*
> ```Shell
> eval "$LDQT_BIN $QTOX_DESKTOP_FILE -bundle-non-qt-libs -extra-plugins=libsnore-qt5"
> 
> eval "$LDQT_BIN $QTOX_DESKTOP_FILE -appimage"
> ```

So, the call to `linuxdeployqt` gets replaced by a call to `appimagetool`? What's the difference here?

antony-jr (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[appimage/build.sh, line 155 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjoiGyC2BiHiw-8esC3:-LjoqOUC6xG0hJLGmIXi:b-ge20a8) ([raw file](https://github.com/qtox/qtox/blob/b445031912ae943d7580ea21f6813e9f71bc03ee/appimage/build.sh#L155)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

So, the call to `linuxdeployqt` gets replaced by a call to `appimagetool`? What's the difference here?
</blockquote></details>

The key difference is that, ```appimagetool``` writes the update information when packaging the appimage and linuxdeployqt guesses the update information automatically which we can't control. Hence appimagetool is used to package the AppImage along with the required update information.(without the guessing stuff)

sudden6:
@antony-jr please rebase your PR on the lastest master so that TravisCI passes again

antony-jr:
Okay. But I'm little busy right now. I will do all the requested changes and rebase by this weekend.

nurupo:
Btw, while using the suggested `-DCMAKE_FIND_ROOT_PATH=/path/to/AIUBs/install/prefix` would work and would be easier, it would be better (but maybe require a bit more work) if AIUB instead provided a [pkg-config](https://en.wikipedia.org/wiki/Pkg-config) file, like many Linux libraries do. I'm not very familiar with creating `.pc` files, so I can't help much here, but I can link to some hopefully useful stuff:
- https://en.wikipedia.org/wiki/Pkg-config
- https://dyn.manpages.debian.org/buster/pkg-config/pkg-config.1.en.html#METADATA_FILE_SYNTAX
- https://cmake.org/cmake/help/latest/command/configure_file.html
- https://github.com/TokTok/c-toxcore/pull/322/files
- https://github.com/TokTok/c-toxcore/blob/master/other/pkgconfig/toxcore.pc.in
- http://dailycommit.blogspot.com/2016/10/how-to-generate-pkg-config-file-with.html

antony-jr:
I did not rebase but the build succeeded, how ?

antony-jr:
oops, forget to change the build script, please don't merge yet. Also I don't know if I rebased correctly.

antony-jr (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 29 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjoofmM9PcCEPBkNIEX:-LknEyzx3D0VCGQrJ1bI:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L29)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

This doesn't seem to be used anywhere?
</blockquote></details>

Done.
___
*[CMakeLists.txt, line 745 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-Ljomql_3zYOmIfY2Uv-:-LjovjmC0l1izBGI96tI:bhy9ot6) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L745)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Would be nice if AIUB used the same `search_dependency()` that is used for finding all of the dependencies, instead of having to provide `APPIMAGE_UPDATER_BRIDGE_*` the paths manually. However, it doesn't look like AIUB provides pkg-config files... Well, you could still make `search_dependency()` work even without AIUB providing pkg-config files, if you `make install` AIUB into some dir and pass the path to that dir as`-DCMAKE_FIND_ROOT_PATH=/path/to/AIUBs/install/prefix` when calling qTox's cmake.
</blockquote></details>

Will do.
___
*[CMakeLists.txt, line 751 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-Ljoh8NODKshh9nVSVI4:-LknF0Eo3eTU9ol8ESUN:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L751)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Trailing space.
</blockquote></details>

Done.
___
*[appimage/build.sh, line 25 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjojdIk50Azn6Ubeb0O:-LknF1Gc6cNnoMHI7VjT:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/appimage/build.sh#L25)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Unnecessary new line.
</blockquote></details>

Done.
___
*[appimage/build.sh, line 113 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-LjohQon6XyKEquSCMQ6:-LknF2nL3t_8E_pD0za0:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/appimage/build.sh#L113)):*
<details><summary><i>Previously, nurupo wrote…</i></summary><blockquote>

Trailing space.
</blockquote></details>

Done.

antony-jr (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 745 at r6](https://reviewable.io/reviews/qtox/qtox/5747#-Ljomql_3zYOmIfY2Uv-:-LknF7Ae0OesXmT7hYza:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/efefe7c0191e823bec5406e3e5ef12da544b4614/CMakeLists.txt#L745)):*
<details><summary><i>Previously, antony-jr (Antony jr) wrote…</i></summary><blockquote>

Will do.
</blockquote></details>

Done.

antony-jr:
Do I need to squash commits?

antony-jr:
@nurupo The nightly builds still follow the git hash for naming the AppImage but this PR uses the full travis commit hash. If the filename is changed and no new zsync meta file is created then the update will always fail for nightly builds. Is any changes required or nightly builds are uploaded as it is written in the ```output``` directory?

antony-jr:
Everything looks good to me. Any more changes required?

sudden6:
@antony-jr I merged for now, I'll ping you via an GH issue in case we need to make further adjustments^^

nurupo (commented):
> Reviewed 2 of 3 files at r7, 1 of 1 files at r8.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5747)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 61 at r8](https://reviewable.io/reviews/qtox/qtox/5747#-LkwwPYhAjS3I0dUJeB9:-LkwwPYi9sgPK5tni06w:bv6wbu9) ([raw file](https://github.com/qtox/qtox/blob/e9665d517aada920d81e57b9a5370bfb22773fbd/CMakeLists.txt#L61)):*
> ```CMake
> if(${APPIMAGE_UPDATER_BRIDGE})
>   set(ENV{PKG_CONFIG_PATH}
>     /usr/local/lib/pkgconfig:$ENV{PKG_CONFIG_PATH})
> ```

This should be removed, it's something the person building qTox is supposed to set on their own in their shell session, as they might have installed AIUB and its .pc files in some custom path, it doesn't have to be `/usr/local/lib/pkgconfig`.
___
*[CMakeLists.txt, line 702 at r8](https://reviewable.io/reviews/qtox/qtox/5747#-Lkwx96U8heiqMD1E0sr:-Lkwx96VEd1DM6SV2l-s:b-80mwet) ([raw file](https://github.com/qtox/qtox/blob/e9665d517aada920d81e57b9a5370bfb22773fbd/CMakeLists.txt#L702)):*
> ```CMake
>         search_dependency(AIUB  PACKAGE AppImageUpdaterBridge)
>         if(AIUB_FOUND)
> 		message(STATUS "using AppImageUpdaterBridge")
> ```

Tabs.
___
*[CMakeLists.txt, line 704 at r8](https://reviewable.io/reviews/qtox/qtox/5747#-LkwxBU89OU5QmSF7Sd8:-LkwxBU89OU5QmSF7Sd9:b-80mwet) ([raw file](https://github.com/qtox/qtox/blob/e9665d517aada920d81e57b9a5370bfb22773fbd/CMakeLists.txt#L704)):*
> ```CMake
> 		message(STATUS "using AppImageUpdaterBridge")
>                 add_definitions(-DAPPIMAGE_UPDATER_BRIDGE_ENABLED=1)
> 	else()
> ```

Tabs.
___
*[appimage/build.sh, line 136 at r8](https://reviewable.io/reviews/qtox/qtox/5747#-LkwyRVy0gAaBUp38gKQ:-LkwyRVy0gAaBUp38gKR:b-rol110) ([raw file](https://github.com/qtox/qtox/blob/e9665d517aada920d81e57b9a5370bfb22773fbd/appimage/build.sh#L136)):*
> ```Shell
> # build dir of simple_make
> cd _build
> ```

You would set `export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:$PKG_CONFIG_PATH` here, so that cmake would find the deps whenever you have installed them.

For example. if you have instead installed them into `/tmp/deps` instead, you could easily change it to `export PKG_CONFIG_PATH=/tmp/deps/lib/pkgconfig:$PKG_CONFIG_PATH`. CMake has no idea where you have installed it, so hard-coding `/usr/local/lib/pkgconfig` in CMake is a bad idea. Also, the way you have hardcoded it, as the very first path in the list, will break things, e.g. if you have two versions of the same dep built and installed, one in `/usr/local` and one in `/tmp/dep` and you want CMake use the one from `/tmp/dep`, because of your CMake change, it will always pick the `/usr/local` one, so it puts it as the very first path.

nurupo:
Looks good to me, aside from tabs and misplaced `PKG_CONFIG_PATH` update - it should be in the .sh script, not cmake file.

The PR is already merged though, so nothing can be done here.

antony-jr:
I am not very experienced with pkg-config. There should be no issue unless you give the cmake flag.To the best of my knowledge no sane person would include it explicitly.

antony-jr:
@nurupo The nightly builds are working really good! Also tested the update.

Screenshot_20190730_072055

PR #5759, opened 2019-08-02
fix(calls): create dummy source for invalidated callback

anthonybilinski:
Fixes mute friend and group calls when mic is switched to disabled then back to a valid source. Now ToxCall can register to the DummySource, and re-register when the source is changed, like when two normal sources are used.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5759)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,6 +262,8 @@ set(${PROJECT_NAME}_SOURCES
   src/audio/backend/alsink.h
   src/audio/backend/alsource.cpp
   src/audio/backend/alsource.h
+  src/audio/backend/dummysource.cpp
+  src/audio/backend/dummysource.h
   src/audio/backend/openal.cpp
   src/audio/backend/openal.h
   src/audio/iaudiosettings.h
--- a/src/audio/backend/alsource.h
+++ b/src/audio/backend/alsource.h
@@ -36,9 +36,9 @@ public:
     AlSource& operator=(AlSource&& other) = delete;
     ~AlSource();
 
-    operator bool() const;
+    operator bool() const override;
 
-    void kill();
+    void kill() override;
 
 private:
     OpenAL& audio;
--- a/src/audio/backend/dummysource.cpp
+++ b/src/audio/backend/dummysource.cpp
@@ -17,33 +17,33 @@
     along with qTox.  If not, see <http://www.gnu.org/licenses/>.
 */
 
-#ifndef IAUDIOSOURCE_H
-#define IAUDIOSOURCE_H
-
-#include <QObject>
+#include "src/audio/backend/dummysource.h"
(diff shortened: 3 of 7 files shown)


sudden6 (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5759)* status: 0 of 1 LGTMs obtained

___
*[src/audio/backend/openal.h, line 152 at r1](https://reviewable.io/reviews/qtox/qtox/5759#-LlJ40ao5rBbJ_DZY1p4:-LlJ40ao5rBbJ_DZY1p5:bn476z7) ([raw file](https://github.com/qtox/qtox/blob/041e65ff1c7dbba3ba45fc94377368b0fc4052b7/src/audio/backend/openal.h#L152)):*
> ```objc
>     std::unordered_set<AlSink*> sinks;
>     std::unordered_set<AlSink*> soundSinks;
>     std::unordered_set<IAudioSource*> sources;
> ```

Mixing `ALSource` and `DummySource` here is probably a bad idea, because `AlSource` actually represents resources in the audio backend and needs to be destroyed differently
___
*[src/audio/backend/openal.cpp, line 329 at r1](https://reviewable.io/reviews/qtox/qtox/5759#-LlJ3b9F3O5KzinuWZ6d:-LlJ3b9F3O5KzinuWZ6e:b-flfbgd) ([raw file](https://github.com/qtox/qtox/blob/041e65ff1c7dbba3ba45fc94377368b0fc4052b7/src/audio/backend/openal.cpp#L329)):*
> ```cpp
>     std::unique_ptr<IAudioSource> source;
>     if (!Settings::getInstance().getAudioInDevEnabled()) {
>         source = std::unique_ptr<IAudioSource>(new DummySource{*this});
> ```

Bad dependency on the `Settings` singleton
___
*[src/audio/backend/openal.cpp, line 366 at r1](https://reviewable.io/reviews/qtox/qtox/5759#-LlJ3m8NC3jkfk2ER9F9:-LlJ3m8NC3jkfk2ER9FA:bb57i8v) ([raw file](https://github.com/qtox/qtox/blob/041e65ff1c7dbba3ba45fc94377368b0fc4052b7/src/audio/backend/openal.cpp#L366)):*
> ```cpp
>     qDebug() << "Unsubscribed from audio input device [" << sources.size() << "subscriptions left ]";
> 
>     if (sources.empty()) {
> ```

Having dummy sources in `sources` will keep the audio device open even though there are no real sources

sudden6:
> what do you think about IAudioControl handling dummy source/sink logic, and dispatching to a virtual private function that OpenAl implements for creating /real/ audio sources and audio sinks?

So you want to turn the `IAudioControl` interface into a full class, right? ATM I don't see any drawbacks from this approach.

PR #5773, opened 2019-08-09
fix(offlinemsg): Fix offline message dispatching on history load, and move old stuck faux_offline_pending messages to a new table

anthonybilinski:
* Fix callback hookup order in ChatHistory
* Add correct call to SessionChatLog to insert an unfinished message
* Fix incorrect logic in parsing History database response

This is a rebase of https://github.com/qTox/qTox/pull/5735. Review will continue here now that it's rebased.

Note that it's rebased on top of yet unmerged https://github.com/qTox/qTox/pull/5772 - without which all pending messages from all friends will be sent to every friend on qTox start!

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5773)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -290,6 +290,8 @@ set(${PROJECT_NAME}_SOURCES
   src/chatlog/content/text.h
   src/chatlog/content/timestamp.cpp
   src/chatlog/content/timestamp.h
+  src/chatlog/content/broken.cpp
+  src/chatlog/content/broken.h
   src/chatlog/customtextdocument.cpp
   src/chatlog/customtextdocument.h
   src/chatlog/documentcache.cpp
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -109,7 +109,7 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString& sender, const QSt
                    ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
 
     if (!date.isNull())
-        msg->markAsSent(date);
+        msg->markAsDelivered(date);
 
     return msg;
 }
@@ -207,7 +207,7 @@ ChatMessage::Ptr ChatMessage::createBusyNotification()
     return msg;
 }
 
-void ChatMessage::markAsSent(const QDateTime& time)
+void ChatMessage::markAsDelivered(const QDateTime& time)
 {
     QFont baseFont = Settings::getInstance().getChatMessageFont();
 
--- a/src/chatlog/chatmessage.cpp
+++ b/src/chatlog/chatmessage.cpp
@@ -26,6 +26,7 @@
 #include "content/spinner.h"
 #include "content/text.h"
 #include "content/timestamp.h"
+#include "content/broken.h"
 #include "src/widget/style.h"
 
 #include <QDebug>
@@ -33,6 +34,7 @@
 
 #include "src/persistence/settings.h"
 #include "src/persistence/smileypack.h"
+#include "src/persistence/history.h"
 
 #define NAME_COL_WIDTH 90.0
 #define TIME_COL_WIDTH 90.0
@@ -43,7 +45,8 @@ ChatMessage::ChatMessage()
 }
 
 ChatMessage::Ptr ChatMessage::createChatMessage(const QString& sender, const QString& rawMessage,
-                                                MessageType type, bool isMe, const QDateTime& date, bool colorizeName)
+                                                MessageType type, bool isMe, MessageState state,
+                                                const QDateTime& date, bool colorizeName)
 {
     ChatMessage::Ptr msg = ChatMessage::Ptr(new ChatMessage);
 
@@ -105,12 +108,21 @@ ChatMessage::Ptr ChatMessage::createChatMessage(const QString& sender, const QSt
                                                        ? QString("%1 %2").arg(sender, rawMessage)
                                                        : rawMessage),
                    ColumnFormat(1.0, ColumnFormat::VariableSize));
-    msg->addColumn(new Spinner(Style::getImagePath("chatArea/spinner.svg"), QSize(16, 16), 360.0 / 1.6),
-                   ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
 
-    if (!date.isNull())
-        msg->markAsDelivered(date);
-
+    switch (state) {
+        case MessageState::complete:
+            msg->addColumn(new Timestamp(date, Settings::getInstance().getTimestampFormat(), baseFont),
+                        ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
+            break;
+        case MessageState::pending:
+            msg->addColumn(new Spinner(Style::getImagePath("chatArea/spinner.svg"), QSize(16, 16), 360.0 / 1.6),
+                        ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
+            break;
+        case MessageState::broken:
+            msg->addColumn(new Broken(Style::getImagePath("chatArea/error.svg"), QSize(16, 16)),
+                        ColumnFormat(TIME_COL_WIDTH, ColumnFormat::FixedSize, ColumnFormat::Right));
+            break;
+    }
     return msg;
 }
 
--- a/src/chatlog/chatmessage.h
+++ b/src/chatlog/chatmessage.h
@@ -57,7 +57,7 @@ public:
     static ChatMessage::Ptr createTypingNotification();
     static ChatMessage::Ptr createBusyNotification();
 
-    void markAsSent(const QDateTime& time);
+    void markAsDelivered(const QDateTime& time);
     QString toString() const;
     bool isAction() const;
     void setAsAction();
--- a/src/chatlog/chatmessage.h
+++ b/src/chatlog/chatmessage.h
@@ -22,6 +22,8 @@
 
 #include "chatline.h"
 #include "src/core/toxfile.h"
+#include "src/persistence/history.h"
+
 #include <QDateTime>
 
 class QGraphicsScene;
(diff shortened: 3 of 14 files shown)


anthonybilinski:
Addressing https://github.com/qTox/qTox/pull/5735#pullrequestreview-256553969:
Originally from @Diadlo 
>I don't understand, why you don't want to mock History? Because it can be the root of some bugs? If so, it doesn't matter for ChatHistory test. But how to test History it's interesting question

Chatted with @sphaerophoria and he basically agrees with what you've said Diadlo. History is a more likely cause of bugs, but that doesn't mean it can't be mocked to test ChatHistory separately. Personally, I'd rather not spend time on unit testing ChatHistory right now. Is it acceptable to include the fix without?

>Looks like the first line of if and else are the ~same
auto chatLogMessage = ChatLogMessage{isComplete, processedMessage};

Agreed, fixed.

From https://github.com/qTox/qTox/pull/5735#pullrequestreview-256554895
> const reference for this function and the previous

done for sender, senderName, message, and file.

> Why now it can't be null?

I think this change was due to the fact the `message` can never be null, due to the table's definition: `message BLOB NOT NULL`.

https://github.com/qTox/qTox/pull/5735#issuecomment-517859591
From @sudden6 
> @sphaerophoria can you rebase?

Done.

Diadlo (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773#-:-LlqIO7w7gBpFOVF5MI1:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
@anthonybilinski was the source of that "sending unsent messages to all friends" bug the database logic? Can we protect against this in the future?

anthonybilinski:
@sudden6 it was a combination of:
1. Bad SQL select since the initial refactor
2. Offline message sending in load fixed, causing 1. to actually have an effect.
3. me stupidly running `gdb qtox` in my build directory and thinking it would run /use/local/qtox instead of ./qtox.

As soon as the SQL select is complete, we have no way to know that the messages aren't for the friend we're sending them to, so I think the only option would be to have tests for history with some kind of mocked rawdatabase (or just a real rawdatabase like I did for the DB schema version upgrade test), and then to test each History SQL interaction on a known DB state and make sure we get back the expected date.

I think this would be worth doing in general, I some changes in progress WRT unit testing db schema logic already that could be extended.

sudden6:
Ok, thanks for the explanation.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773#-:-LlrvsPF2yIjg0k8I2Yo:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

anthonybilinski:
Fixing this causes some unexpected behaviour due to old messages stuck as pending suddenly being sent. https://github.com/qTox/qTox/issues/5776 created to address this. I'll hold off merging this PR until a choice is made and implemented (if needed).

sphaerophoria:
Just linking this back to PR #5734 so it's obvious it's being worked on

anthonybilinski:
I've added my WIP changes for https://github.com/qTox/qTox/issues/5776, so that the fixing of sending can be merged at the same time as the db "upgrade" to avoid sending old history.

anthonybilinski:
This is feature complete now. I've done some preliminary testing and it looks good, but I'm going to do some more testing with crafted databases. **Careful of testing this on your database!** After some git history cleanup and more testing by me, this should be good to go for review.

anthonybilinski:
This is ready to be reviewed now. I made an effort to split it into logical chunks, so I recommend reviewing per commit.

* The RawDatabase change was cherry-picked from master, originally merged as https://github.com/qTox/qTox/pull/5842, because the corrected sql transaction behaviour was needed for the db upgrade to prepare statements in the transaction based off the creation of the new table earlier in that same transaction.
* the "fix(offlinemsg): fix offline message dispatching on history load" change is what was already reviewed for this PR, unchanged.
* the 8 commits after that are the new changes, to fix #5776

sudden6 (changes requested):
> Reviewed 5 of 5 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 187 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquL5QaEZW_qGOEGvpU:-LquL5QaEZW_qGOEGvpV:b-xnkkk6) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/src/persistence/history.cpp#L187)):*
> ```cpp
>         //   possibility 2) it is a old existing database, before version 1 and before we saved schema version,
>         //       and needs to be updated.
>         bool bSuccess;
> ```

please initialize the variable
___
*[src/persistence/history.cpp, line 670 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquPv0-8bkJSxXL-PE4:-LquPv00AKFXCRot4Ci8:bc1u22l) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/src/persistence/history.cpp#L670)):*
> ```cpp
> 
>         assert(!(isPending && isBroken));
>         MessageState messageState;
> ```

There are multiple copies of the conversion, please extract into a function or make a constructor for MessageState (if C++ supports that)
___
*[test/persistence/dbschema_test.cpp, line 136 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquNUwVCeHKBNo3dwsr:-LquNUwVCeHKBNo3dwss:bkc9rfl) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/test/persistence/dbschema_test.cpp#L136)):*
> ```cpp
> {
>     auto db = std::shared_ptr<RawDatabase>{new RawDatabase{"testIsNewDbTrue.db", {}, {}}};
>     bool bSuccess;
> ```

please initialize
___
*[test/persistence/dbschema_test.cpp, line 155 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquORA_1pRQlS2PWNRL:-LquORA_1pRQlS2PWNRM:b9ziewv) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/test/persistence/dbschema_test.cpp#L155)):*
> ```cpp
> }
> 
> void TestDbSchema::test1to2()
> ```

please add a comment, describing what is tested, especially the reason why we create friends with pending messages here. I know you're doing this to test the logic for the pending messages bug fix, but I'd like this documented here too, possibly with a link to the issue and the discussions.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 187 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquL5QaEZW_qGOEGvpU:-Lqu_itT1J2qhCx0loPC:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/src/persistence/history.cpp#L187)):*

please initialize the variable
</blockquote></details>

Done.
___
*[src/persistence/history.cpp, line 670 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquPv0-8bkJSxXL-PE4:-Lqu_kLuExoDLkYrj5go:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/src/persistence/history.cpp#L670)):*

There are multiple copies of the conversion, please extract into a function or make a constructor for MessageState (if C++ supports that)
</blockquote></details>

Done.
___
*[test/persistence/dbschema_test.cpp, line 136 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquNUwVCeHKBNo3dwsr:-Lqu_ldV8pOzFmsTxLa-:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/test/persistence/dbschema_test.cpp#L136)):*

please initialize
</blockquote></details>

Done.
___
*[test/persistence/dbschema_test.cpp, line 155 at r7](https://reviewable.io/reviews/qtox/qtox/5773#-LquORA_1pRQlS2PWNRL:-Lqu_m8l1Dz41w5AlwfG:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b1dd70a6070a74a7e898e210268305c5c8375f77/test/persistence/dbschema_test.cpp#L155)):*

please add a comment, describing what is tested, especially the reason why we create friends with pending messages here. I know you're doing this to test the logic for the pending messages bug fix, but I'd like this documented here too, possibly with a link to the issue and the discussions.
</blockquote></details>

Done.

sudden6 (approved):
:lgtm_strong:

> Reviewed 13 of 13 files at r9.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773#-:-LqurFQa8wZL012WPq_B:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5773)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 118 at r12](https://reviewable.io/reviews/qtox/qtox/5773#-LrYqeAw7ro4HwzTkGly:-LrYqeAxB23AK0l7r6iV:b-sjgyuh) ([raw file](https://github.com/qtox/qtox/blob/cf63ae1a1b0c4494deca197f474de5c82162c7fd/src/persistence/history.cpp#L118)):*
> ```cpp
>         "LEFT JOIN faux_offline_pending ON history.id = faux_offline_pending.id "
>         "WHERE faux_offline_pending.id IS NULL "
>         "GROUP BY chat_id;");
> ```

I've fixed the SQL in this upgrade logic to fix the issue @sudden6 ran into when firs testing this PR. Previously I sorted the temporary table, then "grouped by" without a defined selection method - which is apparently undefined behaviour. Now the "group by" explicitly selects the max history id, e.g. the latest message. @sudden6 ran it again on the same database and didn't send a spurious message this time. I also noticed on one of my test profiles in one of my chats there was a pending message in the chat log after scrolling up. When using this updated version instead, that message was correctly marked as broken.

PR #5779, opened 2019-08-10
fix: empty username causes mention on ever message

### LINKED ISSUE #2119: Highlighting on empty aliases
Hello,

if you've an empty alias and you're in a groupchat:
Every message somebody writes is highlighted in orange.

I think the better solution would be if no message or only messages with the alias you got, while joining the groupchat, will be highlighted.

sudden6:
This fixes #2119 and additionally introduces the possibility to mention
users by their public key.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5779)

### DIFF
--- a/src/model/message.cpp
+++ b/src/model/message.cpp
@@ -31,6 +31,17 @@ void MessageProcessor::SharedParams::onUserNameSet(const QString& username)
                                               QRegularExpression::CaseInsensitiveOption);
 }
 
+/**
+ * @brief Set the public key on which a message should be highlighted
+ * @param pk ToxPk in its hex string form
+ */
+void MessageProcessor::SharedParams::setPublicKey(const QString& pk)
+{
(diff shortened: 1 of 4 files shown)


Diadlo:
AFAIK we have tests for messages. Can you add test case for regression testing and for a new case?

I.e. check that with empty username there is no mention on some message and that pk mention works right.

sudden6:
Good catch, done.

sudden6:
ping @Diadlo

Diadlo (approved):
Sorry. Forgot about this. Good job!
:lgtm_strong:

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5779#-:-LnBnOES91tcNYBH_K9e:bs8ocmb)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5789, opened 2019-08-16
[WIP] feat(proxy): provide commandline tools for proxy settings

### LINKED ISSUE #2070: proxy settings can not be set before the login

Kribylet:
Adds the ability to set proxy settings through command line before qTox attempts to connect. Attempts some very basic sanity checks on the provided settings, and aborts execution before connecting if any issues with the provided settings are found.

If proxy settings are set but qTox is started with a command line option that is incompatible with proxy usage, the commit will currently disable the proxy and continue running. This is fine under the assumption that the user understands which connection settings are incompatible. It may be desirable to more aggressively ensure the user knows what will disable proxy settings however.

The PR is not polished and tested properly yet, but there is enough here to understand the general approach and provide feedback.

If properly implemented, should close #2070

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)

### DIFF
--- a/doc/user_manual_en.md
+++ b/doc/user_manual_en.md
@@ -10,6 +10,7 @@
 * [Quotes](#quotes)
 * [Multi Window Mode](#multi-window-mode)
 * [Keyboard Shortcuts](#keyboard-shortcuts)
+* [Commandline Options](#commandline-options)
 * [Emoji Packs](#emoji-packs)
 
 
@@ -428,6 +429,25 @@ The following shortcuts are currently supported:
 
 In audio group chat microphone mute state will be changed while `Ctrl` + 
 `p` pressed and reverted on release.
+
+## Commandline Options
+
+| Option                             | Action                                             |
+|------------------------------------|----------------------------------------------------|
+| `-p` `<profile>`                   | Use specified unencrypted profile                  |
+| `-l`                               | Start with loginscreen                             |
+| `-I` `<on/off>`                    | Sets IPv6 toggle [Default: ON]                     |
+| `-U` `<on/off>`                    | Sets UDP toggle [Default: ON]                      |
+| `-L` `<on/off>`                    | Sets LAN toggle [Default: ON]                      |
+| `-P` `<protocol>:<address>:<port>` | Applies [proxy options](#commandline-proxy-options) [Default: NONE]|
+
+### Commandline Proxy Options
+Protocol: NONE, HTTP or SOCKS5 <br>
+Address: Proxy address <br>
+Port: Proxy port number (0-65535) <br>
+Example input: <br>
+`qtox -P SOCKS5:192.168.0.1:2121` <br>
+`qtox -P none`
 
 ## Emoji Packs
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -58,96 +58,6 @@ static QList<QByteArray>* logBuffer =
 QMutex* logBufferMutex = new QMutex();
 #endif
 
-/**
- * Verifies that commandline proxy settings are at least reasonable. Does not verify provided IP
- * or hostname addresses are valid. Code duplication with Settings::applyCommandLineOptions, which
- * also verifies arguments, should be removed in a future refactor.
- * @param parser QCommandLineParser instance
- */
-bool verifyProxySettings(QCommandLineParser& parser)
-{
-    QString IPv6Setting = parser.value("I");
-    QString LANSetting = parser.value("L");
-    QString UDPSetting = parser.value("U");
-    QString proxySettingString = parser.value("proxy");
-    QStringList proxySettings = proxySettingString.split(":");
-    // Check for incompatible settings
-    bool activeProxyType = false;
-
-    if (parser.isSet("P")) {
-        activeProxyType = proxySettings[0].compare(QString("SOCKS5"), Qt::CaseInsensitive) == 0
-                          || proxySettings[0].compare(QString("HTTP"), Qt::CaseInsensitive) == 0;
-    }
-
-
-    if (activeProxyType && (UDPSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)) {
-        qCritical() << "Cannot set UDP on with proxy.";
-        return false;
-    }
-
-    if (activeProxyType && (LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)) {
-        qCritical() << "Cannot set LAN discovery on with proxy.";
-        return false;
-    }
-
-    if ((LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)
-        && (UDPSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
-        qCritical() << "Incompatible UDP/LAN settings.";
-        return false;
-    }
-
-    if (parser.isSet("I")) {
-        if (!(IPv6Setting.compare(QString("on"), Qt::CaseInsensitive) == 0
-              || IPv6Setting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
-            qCritical() << "Unable to parse IPv6 setting.";
-            return false;
-        }
-    }
-
-    if (parser.isSet("U")) {
-        if (!(UDPSetting.compare(QString("on"), Qt::CaseInsensitive) == 0
-              || UDPSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
-            qCritical() << "Unable to parse UDP setting.";
-            return false;
-        }
-    }
-
-    if (parser.isSet("L")) {
-        if (!(LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0
-              || LANSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
-            qCritical() << "Unable to parse LAN setting.";
-            return false;
-        }
-    }
-
-    if (parser.isSet("P")) {
-        if (proxySettings[0].compare(QString("NONE"), Qt::CaseInsensitive) == 0) {
-            return true;
-            // slightly lazy check here, accepting 'NONE[:.*]' is fine since no other
-            // arguments will be investigated when proxy settings are applied.
-        }
-        // Since the first argument isn't 'none', verify format of remaining arguments
-        if (proxySettings.size() != 3) {
-            qCritical() << "Invalid number of proxy arguments.";
-            return false;
-        }
-
-        if (!(proxySettings[0].compare(QString("SOCKS5"), Qt::CaseInsensitive) == 0
-              || proxySettings[0].compare(QString("HTTP"), Qt::CaseInsensitive) == 0)) {
-            qCritical() << "Unable to parse proxy type.";
-            return false;
-        }
-
-        // Kriby: Sanity checking IPv4+IPv6/hostnames sure is a lot of work!
-
-        int portNumber = proxySettings[2].toInt();
-        if (!(portNumber > 0 && portNumber < 65536)) {
-            qCritical() << "Invalid port number range.";
-        }
-    }
-    return true;
-}
-
 void cleanup()
 {
     // force save early even though destruction saves, because Windows OS will
@@ -307,18 +217,21 @@ int main(int argc, char* argv[])
                            QObject::tr("Starts new instance and opens the login screen.")));
     parser.addOption(QCommandLineOption(QStringList() << "I"
                                                       << "IPv6",
-                                        QObject::tr("Sets IPv6 <on>/<off>"), QObject::tr("on/off")));
+                                        QObject::tr("Sets IPv6 <on>/<off>. Default is ON."),
+                                        QObject::tr("on/off")));
     parser.addOption(QCommandLineOption(QStringList() << "U"
                                                       << "UDP",
-                                        QObject::tr("Sets UDP <on>/<off>"), QObject::tr("on/off")));
+                                        QObject::tr("Sets UDP <on>/<off>. Default is ON."),
+                                        QObject::tr("on/off")));
     parser.addOption(
         QCommandLineOption(QStringList() << "L"
                                          << "LAN",
-                           QObject::tr("Sets LAN discovery <on>/<off>. UDP off overrides."),
+                           QObject::tr(
+                               "Sets LAN discovery <on>/<off>. UDP off overrides. Default is ON."),
                            QObject::tr("on/off")));
     parser.addOption(QCommandLineOption(QStringList() << "P"
                                                       << "proxy",
-                                        QObject::tr("Sets proxy settings."),
+                                        QObject::tr("Sets proxy settings. Default is NONE."),
                                         QObject::tr("(SOCKS5/HTTP/NONE):(ADDRESS):(PORT)")));
     parser.process(*a);
 
@@ -436,7 +349,7 @@ int main(int argc, char* argv[])
         }
     }
 
-    if (!verifyProxySettings(parser)) {
+    if (!Settings::verifyProxySettings(parser)) {
         return -1;
     }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -58,6 +58,96 @@ static QList<QByteArray>* logBuffer =
 QMutex* logBufferMutex = new QMutex();
 #endif
 
+/**
+ * Verifies that commandline proxy settings are at least reasonable. Does not verify provided IP
+ * or hostname addresses are valid. Code duplication with Settings::applyCommandLineOptions, which
+ * also verifies arguments, should be removed in a future refactor.
+ * @param parser QCommandLineParser instance
+ */
+bool verifyProxySettings(QCommandLineParser& parser)
+{
+    QString IPv6Setting = parser.value("I");
+    QString LANSetting = parser.value("L");
+    QString UDPSetting = parser.value("U");
+    QString proxySettingString = parser.value("proxy");
+    QStringList proxySettings = proxySettingString.split(":");
+    // Check for incompatible settings
+    bool activeProxyType = false;
+
+    if (parser.isSet("P")) {
+        activeProxyType = proxySettings[0].compare(QString("SOCKS5"), Qt::CaseInsensitive) == 0
+                          || proxySettings[0].compare(QString("HTTP"), Qt::CaseInsensitive) == 0;
+    }
+
+
+    if (activeProxyType && (UDPSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)) {
+        qCritical() << "Cannot set UDP on with proxy.";
+        return false;
+    }
+
+    if (activeProxyType && (LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)) {
+        qCritical() << "Cannot set LAN discovery on with proxy.";
+        return false;
+    }
+
+    if ((LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0)
+        && (UDPSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
+        qCritical() << "Incompatible UDP/LAN settings.";
+        return false;
+    }
+
+    if (parser.isSet("I")) {
+        if (!(IPv6Setting.compare(QString("on"), Qt::CaseInsensitive) == 0
+              || IPv6Setting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
+            qCritical() << "Unable to parse IPv6 setting.";
+            return false;
+        }
+    }
+
+    if (parser.isSet("U")) {
+        if (!(UDPSetting.compare(QString("on"), Qt::CaseInsensitive) == 0
+              || UDPSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
+            qCritical() << "Unable to parse UDP setting.";
+            return false;
+        }
+    }
+
+    if (parser.isSet("L")) {
+        if (!(LANSetting.compare(QString("on"), Qt::CaseInsensitive) == 0
+              || LANSetting.compare(QString("off"), Qt::CaseInsensitive) == 0)) {
+            qCritical() << "Unable to parse LAN setting.";
+            return false;
+        }
+    }
+
+    if (parser.isSet("P")) {
+        if (proxySettings[0].compare(QString("NONE"), Qt::CaseInsensitive) == 0) {
+            return true;
+            // slightly lazy check here, accepting 'NONE[:.*]' is fine since no other
+            // arguments will be investigated when proxy settings are applied.
+        }
+        // Since the first argument isn't 'none', verify format of remaining arguments
+        if (proxySettings.size() != 3) {
+            qCritical() << "Invalid number of proxy arguments.";
+            return false;
+        }
+
+        if (!(proxySettings[0].compare(QString("SOCKS5"), Qt::CaseInsensitive) == 0
+              || proxySettings[0].compare(QString("HTTP"), Qt::CaseInsensitive) == 0)) {
+            qCritical() << "Unable to parse proxy type.";
+            return false;
+        }
+
+        // Kriby: Sanity checking IPv4+IPv6/hostnames sure is a lot of work!
+
+        int portNumber = proxySettings[2].toInt();
+        if (!(portNumber > 0 && portNumber < 65536)) {
+            qCritical() << "Invalid port number range.";
+        }
+    }
+    return true;
+}
+
 void cleanup()
 {
     // force save early even though destruction saves, because Windows OS will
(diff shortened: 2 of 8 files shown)


sudden6:
I didn't yet review this in depth, here are some first comments:

- Why move the parsing logic inside settings? IMO the parsing logic should be implemented in `main.cpp` in a separate function and use the public setters of `Settings`. @anthonybilinski I'd also appreciate your opinion on this matter.
- Can you write down the command line options + examples e.g. in https://github.com/qTox/qTox/blob/master/doc/user_manual_en.md? It's really hard to parse the meaning of each option and their dependencies just from the code.

Kribylet:
> * Why move the parsing logic inside settings? IMO the parsing logic should be implemented in `main.cpp` in a separate function and use the public setters of `Settings`.

It should be possible. I wanted to apply the parser settings as closely to after we loaded the profile as possible to avoid using the wrong proxy settings anywhere, but since those settings only apply during the toxcore creation, which happens before the profile settings are loaded it should be equivalent to keep things in main.cpp.

As an aside, shouldn't the fact that the toxcore is created before reading profile settings lead to failure to apply proper proxy settings according to stored profile settings? Proxy settings are stored in the profile, so my impression is that we do not consider proxy settings global.

> * Can you write down the command line options + examples e.g. in https://github.com/qTox/qTox/blob/master/doc/user_manual_en.md? It's really hard to parse the meaning of each option and their dependencies just from the code.

Absolutely! I wanted feedback on what we actually wanted to call the option flags, though, as well as input formats. I was hoping they were legible during the parser construction statement in main.cpp. As it stands they are:

-I (IPv6) [on/off] - pass exact string or parse fails, case insensitive
-U (UDP) [on/off] - ^^^
-L (LAN) [on/off] ^^^ (UDP Required)
-P (Proxy) [<SOCKS5/HTTP/NONE>:<Address(any format)>:<Portnumber(1-65535)>] (UDP must be off, NONE-proxytype sets address to "", port to 0)

sudden6:
> As an aside, shouldn't the fact that the toxcore is created before reading profile settings lead to failure to apply proper proxy settings according to stored profile settings? Proxy settings are stored in the profile, so my impression is that we do not consider proxy settings global.

I'm not 100% sure what you mean here, but AFAIK `Profile` creates `Core` so the settings should always be loaded before `Core`.

Kribylet:
> I'm not 100% sure what you mean here, but AFAIK `Profile` creates `Core` so the settings should always be loaded before `Core`.

Not the case. It is true that the creation of a Profile creates the Core, too, but the Profile is constructed prior to settings being loaded from a toxsave. There is a settings instance created early in main.cpp with the default constructor. Then that settings instance is passed around until it eventually reaches the Profile constructor. The profile constructor also takes a QByteArray of data read from the .tox file and uses that to create a ToxOptions object. However, the actual connection settings generated during the creation of the ToxOptions are read from the default constructed Settings instance.

Since I've refactored related parts of the code recently I went through blame and compared the previous implementation, and it is the same. The qTox client itself would never report this visually to the user since the proper settings are loaded after Core creation, but before a GUI is ever visible, which may be why we haven't seen any bug reports about it. (?)

sudden6:
> but the Profile is constructed prior to settings being loaded from a toxsave.

It seems to me that the proxy settings are not part of the toxsave, so that should be no problem.

But I think I understand now what you mean:

If this line is reached: https://github.com/qTox/qTox/blob/master/src/main.cpp#L346 the core settings are not yet initialized with the correct proxy.

This is only done later in: https://github.com/qTox/qTox/blob/master/src/main.cpp#L353

This could result in no proxy being applied, correct?

sudden6:
Just confirmed this, when running qTox with the `-p` parameter, proxy settings are not applied.

sudden6:
It might be that  #5797 breaks the logic here because the settings are loaded from the file inside `loadProfile(...)` again.

Kribylet:
> It might be that #5797 breaks the logic here because the settings are loaded from the file inside `loadProfile(...)` again.

Rebased and changed to accommodate. Added user manual segment about commandline options, though it does not cover "tox:<..>" and "<..>.tox" positional argument options

sudden6 (changes requested):
> Reviewed 5 of 7 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[doc/user_manual_en.md, line 438 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnT8m6y6Av6pK6ybJDb:-LnT8m7--SyAhemOrN2x:b-cxat20) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/doc/user_manual_en.md#L438)):*
> ```Markdown
> |------------------------------------|----------------------------------------------------|
> | `-p` `<profile>`                   | Use specified unencrypted profile                  |
> | `-l`                               | Force loginscreen                                  |
> ```

IMO `Start with loginscreen` sounds a bit nicer, decide on your own if you want to change it.
___
*[src/main.cpp, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnT9h3W-dBlzX0lZu-a:-LnT9h3X5g6ZittzSFVG:bpzi4vh) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L69)):*
> ```cpp
> bool verifyProxySettings(QCommandLineParser& parser)
> {
>     QString IPv6Setting = parser.value("I");
> ```

maybe use `toLower()` here so all the comparisons don't need extra flags?
___
*[src/main.cpp, line 78 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTANiy7yBpd7u1KUS6:-LnTANiy7yBpd7u1KUS7:bo17hqv) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L78)):*
> ```cpp
> 
>     if (parser.isSet("P")) {
>         activeProxyType = proxySettings[0].compare(QString("SOCKS5"), Qt::CaseInsensitive) == 0
> ```

Instead of `QString("SOCKS5")` use `QStringLiteral(`SOCKS5`)` according to the coding style and maybe create these only once and then use `==` to compare. Same for all other constant strings
___
*[src/main.cpp, line 81 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTACsx0o0bj2bNlrX7:-LnTACsx0o0bj2bNlrX8:bvxay2k) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L81)):*
> ```cpp
>                           || proxySettings[0].compare(QString("HTTP"), Qt::CaseInsensitive) == 0;
>     }
> ```

extra empty line
___
*[src/main.cpp, line 126 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTAyQH6rOvxe_0gbq6:-LnTAyQH6rOvxe_0gbq7:b7mvywl) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L126)):*
> ```cpp
>         if (proxySettings[0].compare(QString("NONE"), Qt::CaseInsensitive) == 0) {
>             return true;
>             // slightly lazy check here, accepting 'NONE[:.*]' is fine since no other
> ```

IMO return after the comment looks nicer, decide on your own
___
*[src/main.cpp, line 141 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTB511AyoLatkNwudH:-LnTB511AyoLatkNwudI:b-cjuuar) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L141)):*
> ```cpp
>         }
> 
>         // Kriby: Sanity checking IPv4+IPv6/hostnames sure is a lot of work!
> ```

IMO while a bit funny, this comment doesn't really provide any help with understanding the code
___
*[src/persistence/settings.cpp, line 301 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTBmNJ3qZa48c-HcsF:-LnTBmNK-ZpRcKukskFD:bs1nt3a) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/persistence/settings.cpp#L301)):*
> ```cpp
>  * Applies command line options on top of loaded settings. Fails without changes if attempting to
>  * apply contradicting settings.
>  * @param parser
> ```

missing explanation for `parser`, it seems `parser` never `nullptr` here, use a reference intstead?
___
*[src/persistence/settings.cpp, line 304 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTBtQ36aoy7g4PBzNf:-LnTBtQ36aoy7g4PBzNg:b6v25io) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/persistence/settings.cpp#L304)):*
> ```cpp
>  * @return Success indicator (success = true)
>  */
> bool Settings::applyCommandLineOptions(const QCommandLineParser* parser)
> ```

Since this is essentially a copy of `verifyProxySettings(...)` what do you think of making `verifyProxySettings(...)` a static function of `Settings` and calling it here to verify the parameters.

Afterwards we'd only need to print a summary of the settings (for debugging purposes) and then apply the already checked parameters. If you

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[doc/user_manual_en.md, line 438 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnT8m6y6Av6pK6ybJDb:-LnX7m3F5La-YxmQgeJd:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/doc/user_manual_en.md#L438)):*

IMO `Start with loginscreen` sounds a bit nicer, decide on your own if you want to change it.
</blockquote></details>

Done.
___
*[src/main.cpp, line 69 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnT9h3W-dBlzX0lZu-a:-LnX7o2I2xMekW5s2HyC:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L69)):*

maybe use `toLower()` here so all the comparisons don't need extra flags?
</blockquote></details>

Done.
___
*[src/main.cpp, line 78 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTANiy7yBpd7u1KUS6:-LnX7t700FQ1EO7eRYdM:bpwsicc) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L78)):*

Instead of `QString("SOCKS5")` use `QStringLiteral(`SOCKS5`)` according to the coding style and maybe create these only once and then use `==` to compare. Same for all other constant strings
</blockquote></details>

Done, though comment if I misunderstood you.
___
*[src/main.cpp, line 81 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTACsx0o0bj2bNlrX7:-LnX7oFcBJYTD0DQRDO-:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L81)):*

extra empty line
</blockquote></details>

Done.
___
*[src/main.cpp, line 126 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTAyQH6rOvxe_0gbq6:-LnX81dWE7f4W5FaJpe9:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L126)):*

IMO return after the comment looks nicer, decide on your own
</blockquote></details>

Done.
___
*[src/main.cpp, line 141 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTB511AyoLatkNwudH:-LnX82fa7K40BY_dcZ4B:bqyke4t) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/main.cpp#L141)):*

IMO while a bit funny, this comment doesn't really provide any help with understanding the code
</blockquote></details>

Done. Changed to potential future todo.
___
*[src/persistence/settings.cpp, line 301 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTBmNJ3qZa48c-HcsF:-LnX88V3Bmsy2-_ZwASy:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/persistence/settings.cpp#L301)):*

missing explanation for `parser`, it seems `parser` never `nullptr` here, use a reference intstead?
</blockquote></details>

Done.
___
*[src/persistence/settings.cpp, line 304 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTBtQ36aoy7g4PBzNf:-LnX8B66C3i1n2C8qCQB:bvwd4j1) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/persistence/settings.cpp#L304)):*

Since this is essentially a copy of `verifyProxySettings(...)` what do you think of making `verifyProxySettings(...)` a static function of `Settings` and calling it here to verify the parameters.

Afterwards we'd only need to print a summary of the settings (for debugging purposes) and then apply the already checked parameters. If you 
</blockquote></details>

Done, partially. I couldn't think of a way to bring the verified results over neatly from the verifyProxySettings function without using a bunch of argument pointers or similar that increased code  complexity needlessly. Hopefully this solution is agreeable.

sudden6:
___
*[src/persistence/settings.cpp, line 304 at r2](https://reviewable.io/reviews/qtox/qtox/5789#-LnTBtQ36aoy7g4PBzNf:-LnXBSK86NshZxwRp0a4:b-t4blpd) ([raw file](https://github.com/qtox/qtox/blob/e012ec7d65df01c0629c64c3e4f4c94a6b43cdbb/src/persistence/settings.cpp#L304)):*
<details><summary><i>Previously, Kribylet wrote…</i></summary><blockquote>

Done, partially. I couldn't think of a way to bring the verified results over neatly from the verifyProxySettings function without using a bunch of argument pointers or similar that increased code  complexity needlessly. Hopefully this solution is agreeable.
</blockquote></details>

I imagined a solution very similar to his, fits perfectl.

sudden6 (changes requested):
> Reviewed 3 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 419 at r3](https://reviewable.io/reviews/qtox/qtox/5789#-LnX9VRZ2mqynaOXzL-a:-LnX9VR_EL5TCizQsDIf:bostdax) ([raw file](https://github.com/qtox/qtox/blob/fdd4d198cc65cf649475983ef5f5ef80c187ac61/src/persistence/settings.cpp#L419)):*
> ```cpp
> 
>     if (parser.isSet("I")) {
>        enableIPv6 = IPv6Setting == ON;
> ```

intendation error, maybe format everything with `tools/format-code.sh` anyway
___
*[src/persistence/settings.cpp, line 437 at r3](https://reviewable.io/reviews/qtox/qtox/5789#-LnXA69p-JbTg-oW4gcQ:-LnXA69p-JbTg-oW4gcR:bk8vx9y) ([raw file](https://github.com/qtox/qtox/blob/fdd4d198cc65cf649475983ef5f5ef80c187ac61/src/persistence/settings.cpp#L437)):*
> ```cpp
>         enableLanDiscovery = LANSetting == ON;
>         qDebug() << QString("Setting LAN Discovery %1.").arg(LANSetting);
>         if (enableLanDiscovery) {
> ```

IMO it would make sense to give the proxy settings priority here and disable LAN discovery, but we need to check that this doesn't contradict other logic in `Settings`

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 419 at r3](https://reviewable.io/reviews/qtox/qtox/5789#-LnX9VRZ2mqynaOXzL-a:-LnXvdVzAuHRGoPq69Hr:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/fdd4d198cc65cf649475983ef5f5ef80c187ac61/src/persistence/settings.cpp#L419)):*

intendation error, maybe format everything with `tools/format-code.sh` anyway
</blockquote></details>

Done.
___
*[src/persistence/settings.cpp, line 437 at r3](https://reviewable.io/reviews/qtox/qtox/5789#-LnXA69p-JbTg-oW4gcQ:-LnXyIdR4eaALx_gBz99:b6nwj82) ([raw file](https://github.com/qtox/qtox/blob/fdd4d198cc65cf649475983ef5f5ef80c187ac61/src/persistence/settings.cpp#L437)):*

IMO it would make sense to give the proxy settings priority here and disable LAN discovery, but we need to check that this doesn't contradict other logic in `Settings`
</blockquote></details>

I changed the logic here to instead apply proxy type settings first, and then check that proxy type is set to none before applying UDP or LAN settings. If not, it outputs to debug about requiring an explicit -P none parameter to perform the change.

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-LnZPgGzD1amyip695c0:bab7lcq) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*
> ```cpp
>             if (enableLanDiscovery) {
>                 if (forceTCP) {
>                     qDebug() << "Cannot use LAN discovery without UDP; enabling UDP.";
> ```

This message seems to contradict the code, AFAICT UDP is not enabled here and I think it shouldn't be enabled either, because then the proxy is circumvented.

Instead of nesting ifs I think it's cleaner to just use `&&` here

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-LnaUfQp6odidvdirSih:b76e48t) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*

This message seems to contradict the code, AFAICT UDP is not enabled here and I think it shouldn't be enabled either, because then the proxy is circumvented.

Instead of nesting ifs I think it's cleaner to just use `&&` here
</blockquote></details>

I'm afraid I genuinely don't understand what you mean. ForceTCP is true if UDP is disabled. LAN discovery requires UDP be enabled. If a proxy is active, this clause is never entered. Contradicting settings are also caught early by verifyProxySettings. I did replace the nesting with &&. If you still see an issue with the message please elaborate.

Would you like me to make it so that UDP has to be enabled explicitly by the user?

Edit: Yeah okay I see, it didn't enable UDP there, but I think it should, unless we want to force explicitness. Fixed.

Kribylet:
Travis doxygen has failed twice in a row due to external issues.

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-Lnb83ZlCIPhZmuuUcCH:b-4f9308) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*
<details><summary><i>Previously, Kribylet wrote…</i></summary><blockquote>

I'm afraid I genuinely don't understand what you mean. ForceTCP is true if UDP is disabled. LAN discovery requires UDP be enabled. If a proxy is active, this clause is never entered. Contradicting settings are also caught early by verifyProxySettings. I did replace the nesting with &&. If you still see an issue with the message please elaborate.

Would you like me to make it so that UDP has to be enabled explicitly by the user?
</blockquote></details>

Would you like me to make it so that UDP has to be enabled explicitly by the user?

Yes, because now if you set a proxy and enable lan discovery, UDP will be enabled again -> proxy circumvented

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-LnbkPkE7Kn9nxKG5YrE:b-yrdf9) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*

Would you like me to make it so that UDP has to be enabled explicitly by the user?

Yes, because now if you set a proxy and enable lan discovery, UDP will be enabled again -> proxy circumvented
</blockquote></details>

If a proxy is already set when trying to -L on, "shouldEnableLAN && proxyType != ICoreSettings::ProxyType::ptNone" will be true, generating a qDebug prompt and skipping the else clause. If you attempt to run the commandline -P <not NONE:etc:etc> -L on, verifyProxySettings will additionally stop execution before even reaching this point because those are considered contradicting settings, as per:

    if (activeProxyType && LANSettingString == ON) {
        qCritical() << "Cannot set LAN discovery on with proxy.";
        return false;
    }

If it hadn't, proxy type would be set to something not-NONE before reaching this statement and it'd still not enable LAN or UDP, though it'd be sort of unexpected behavior from user perspective, which is why we catch it earlier. I'm starting to feel I may be locked in on my view of the sequence of possible events and missing something obvious.

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-Lnbqlsf66UcctKULXWu:bf3zj9u) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*
<details><summary><i>Previously, Kribylet wrote…</i></summary><blockquote>

If a proxy is already set when trying to -L on, "shouldEnableLAN && proxyType != ICoreSettings::ProxyType::ptNone" will be true, generating a qDebug prompt and skipping the else clause. If you attempt to run the commandline -P <not NONE:etc:etc> -L on, verifyProxySettings will additionally stop execution before even reaching this point because those are considered contradicting settings, as per:

    if (activeProxyType && LANSettingString == ON) {
        qCritical() << "Cannot set LAN discovery on with proxy.";
        return false;
    }

If it hadn't, proxy type would be set to something not-NONE before reaching this statement and it'd still not enable LAN or UDP, though it'd be sort of unexpected behavior from user perspective, which is why we catch it earlier. I'm starting to feel I may be locked in on my view of the sequence of possible events and missing something obvious.
</blockquote></details>

You're right about the proxy setting, I missed that this case is catched earlier.

However if you compare the logic here with the one from the Advanced settings UI it is different. In the UI LAN discovery can only be enabled when UDP is explicitly enabled, but LAN Discovery can never force UDP to "on".

The case with `-U on -L on` seems to be already handled with the check, but `-L on` is confusing compared to the UI. Here `-L on` enables LAN + UDP, but in the UI this is not possible. So I think `-L on` should require `-U on` or we could say (and write down in the docs) that by default IPv6, UDP and LAN discovery are in `on` state.

What do you think?

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 471 at r4](https://reviewable.io/reviews/qtox/qtox/5789#-LnZPgGy4EYJdwvl5Bos:-LnbuQWy7ubhwYegpGio:buoui6d) ([raw file](https://github.com/qtox/qtox/blob/4cb4745d6effd4d69bcbe0819a178e98b7b3a005/src/persistence/settings.cpp#L471)):*

You're right about the proxy setting, I missed that this case is catched earlier.

However if you compare the logic here with the one from the Advanced settings UI it is different. In the UI LAN discovery can only be enabled when UDP is explicitly enabled, but LAN Discovery can never force UDP to "on".

The case with `-U on -L on` seems to be already handled with the check, but `-L on` is confusing compared to the UI. Here `-L on` enables LAN + UDP, but in the UI this is not possible. So I think `-L on` should require `-U on` or we could say (and write down in the docs) that by default IPv6, UDP and LAN discovery are in `on` state.

What do you think?
</blockquote></details>

Thinking again, even if we say the default for all is `on`, when the command line is `-U off` it must also automatically turn off LAN discovery and notify the user about this.

sudden6 (approved):
> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789)* status: 0 of 1 LGTMs obtained

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5789#-:-LnmnV82Byec4X19NdEE:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5795, opened 2019-08-23
[WIP] fix build with Qt 5.14

Chiitoo:

These are basically just lazy sed changes to fix the build.  I'm not sure how far back this would be compatible, as a lot of the 'QLatin1String' things seem to be from Qt 5.10, but I'm also not sure yet if any of those new things are actually being used.

To be safe, I thought about enclosing things within a version check, but since there are a bunch of these, and in rather big blocks, it would at least look kind of... weird... or something.  :]

The 'asprintf' should maybe be safe back as far as Qt 5.5, but these parts should probably use 'QTextStream' or 'arg()' instead (according to Qt documentation ( https://doc.qt.io/qt-5/qstring.html#asprintf )).

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5795)

### DIFF
--- a/src/core/toxlogger.cpp
+++ b/src/core/toxlogger.cpp
@@ -32,7 +32,7 @@ namespace {
 QByteArray cleanPath(const char *file)
 {
     // for privacy, make the path relative to the c-toxcore source directory
-    const QRegularExpression pathCleaner(QLatin1Literal{"[\\s|\\S]*c-toxcore."});
+    const QRegularExpression pathCleaner(QLatin1String{"[\\s|\\S]*c-toxcore."});
     QByteArray cleanedPath = QString{file}.remove(pathCleaner).toUtf8();
     cleanedPath.append('\0');
     return cleanedPath;
--- a/src/net/bootstrapnodeupdater.cpp
+++ b/src/net/bootstrapnodeupdater.cpp
@@ -30,22 +30,22 @@
 
 namespace {
 const QUrl NodeListAddress{"https://nodes.tox.chat/json"};
-const QLatin1Literal jsonNodeArrayName{"nodes"};
-const QLatin1Literal emptyAddress{"-"};
+const QLatin1String jsonNodeArrayName{"nodes"};
+const QLatin1String emptyAddress{"-"};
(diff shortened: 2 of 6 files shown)


sudden6:
For the `QLatin1Literal` to `QLatin1String` change can you also update: https://github.com/qTox/qTox/blob/master/doc/coding_standards.md#strings

Else we will have this same problem again soon.

Chiitoo:
Ah, yeah, definitely a good idea.  ^^

In case it's not backwards supportish enough, any thoughts on big 'ifs' like this for example:
```C++
namespace NodeFields {
#if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
const QLatin1String status_udp{"status_udp"};
const QLatin1String status_tcp{"status_tcp"};
const QLatin1String ipv4{"ipv4"};
const QLatin1String ipv6{"ipv6"};
const QLatin1String public_key{"public_key"};
const QLatin1String port{"port"};
const QLatin1String maintainer{"maintainer"};
// TODO(sudden6): make use of this field once we differentiate between TCP nodes, and bootstrap nodes
const QLatin1String tcp_ports{"tcp_ports"};
#else
const QLatin1Literal status_udp{"status_udp"};
const QLatin1Literal status_tcp{"status_tcp"};
const QLatin1Literal ipv4{"ipv4"};
const QLatin1Literal ipv6{"ipv6"};
const QLatin1Literal public_key{"public_key"};
const QLatin1Literal port{"port"};
const QLatin1Literal maintainer{"maintainer"};
const QLatin1Literal tcp_ports{"tcp_ports"};
#endif
} // namespace NodeFields
```

sudden6:
they should be fine, but AFAICT `QLatin1String` existed before our last supported Qt version of 5.5

Chiitoo:
Most likely yeah.  I was probably being distracted by unrelated, newer parts of QString.

sudden6 (approved):
:lgtm_strong: since it seems complete after some IRC discussion

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5795#-:-LnOpAHY04fyJHudgL5V:b-bttdxq)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5769, opened 2019-08-06
Crash when selecting lines in chat log


anthonybilinski:
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
Commit hash:  
toxcore:  v0.2.10
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Rarely

##### Steps to reproduce
Selecting and deselecting lines and parts of lines in chat log while friend is typing a message. Exact repro unknown.

##### Observed Behavior
Crash:
```
Thread 1 "qtox" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff3df9ed7 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff3ddb535 in __GI_abort () at abort.c:79
#2  0x00007ffff422b9a7 in  () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#3  0x00007ffff422ae83 in  () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#4  0x0000555555880f58 in QVector<std::shared_ptr<ChatLine> >::operator[](int) (this=0x55555b9645f8, i=1506)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qvector.h:437
#5  0x000055555587a9ee in ChatLog::clearSelection() (this=0x55555b9645a0) at /home/abilinski/builds/qTox/src/chatlog/chatlog.cpp:146
#6  0x000055555587adce in ChatLog::mousePressEvent(QMouseEvent*) (this=0x55555b9645a0, ev=0x7fffffffd430)
    at /home/abilinski/builds/qTox/src/chatlog/chatlog.cpp:199
#7  0x00007ffff4e1e60f in QWidget::event(QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#8  0x00007ffff4ec26ce in QFrame::event(QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#9  0x00007ffff510ed23 in QGraphicsView::viewportEvent(QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
[...]

PR #5806, opened 2019-08-28
fix(chatlog): out of bounds access when loading messages

sudden6:
We need to invalidate the selection when adding/removing lines from the
chatlog because the selection stores them as absolute indices.

A nicer solution would be to actually update the selection accordingly,
but this will be a major overhaul of the code.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5806)

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -731,6 +731,8 @@ void ChatLog::removeFirsts(const int num)
         lines.clear();
     }
 
+    clearSelection();
+
     for (int i = 0; i < lines.size(); ++i) {
         lines[i]->setRow(i);
     }
@@ -744,6 +746,8 @@ void ChatLog::removeLasts(const int num)
     } else {
         lines.clear();
     }
+
+    clearSelection();
 }
 
 void ChatLog::setScroll(const bool scroll)


sudden6:
This could fix: #5769

anthonybilinski (changes requested):
I think it would be better to shift the selection to maintain the same lines, if possible, instead of always clearing. removeFirsts gets called while just chatting, so you could have a selection of a few lines near the bottom, receive one message, then all of a sudden have your selection disapear, which I think would be pretty confusing.

Instead of shifting the indexes on removal/addition to top, we could instead use absolute chat line indexes like provided by IChatLog, so that indexes remain valid after the ChatLog is changed.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5806#-:-LnOwhlm2CtJsiW32ubf:b-q6fg3w)* status: 0 of 1 LGTMs obtained

anthonybilinski:
I pushed my implementation of the more complex fix here: https://github.com/qTox/qTox/pull/5811 - it handles out of access from both directions, and an existing bug that doesn't cause crashes of loading history causing the selection to jump to new messages.

sudden6:
Closing in favor of #5811

PR #5808, opened 2019-08-29
fix(ipc): handle ipc failure gracefully

anthonybilinski:
Allows running qTox inside jails that block IPC.

This is a rework of https://github.com/qTox/qTox/pull/5742.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5808)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -219,12 +219,11 @@ int main(int argc, char* argv[])
 
     uint32_t profileId = settings.getCurrentProfileId();
     IPC ipc(profileId);
-    if (!ipc.isAttached()) {
-        qCritical() << "Can't init IPC";
-        return EXIT_FAILURE;
+    if (ipc.isAttached()) {
+        QObject::connect(&settings, &Settings::currentProfileIdChanged, &ipc, &IPC::setProfileId);
+    } else {
+        qWarning() << "Can't init IPC, maybe we're in a jail? Continuing with reduced multi-client functionality.";
     }
-
-    QObject::connect(&settings, &Settings::currentProfileIdChanged, &ipc, &IPC::setProfileId);
 
     // For the auto-updater
     if (sodium_init() < 0) {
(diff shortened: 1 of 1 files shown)


anthonybilinski:
@yurivict could you try testing this in your FreeBSD jail that originally caused the problem?

yurivict:
IMO this patch needs to be added on top of the other one. The original patch deterministically determines the reason of failure as being in jail, and this patch guesses that this might be the case. There can be many possible reasons of this failure. I suggest that you merge both PRs.

anthonybilinski:
I fully agree that your original implementation determined if a jail was the cause of failure, and that this one doesn't. The benefits of determining this seem very small to me, since even if we fail to init IPC for some other reason, continuing to run without IPC enabled seems best. I'd happily remove "maybe we're in a jail?" from the warning message to not pretend like this determines the cause of the failure.

We don't currently run CI for FreeBSD, just Windows, OSX, and Ubuntu, and we currently have no FreeBSD specific code in our entire code base, plus nothing as low level as reading kernel info. I feel like introducing new platform specific code not covered by CI that queries kernel info is a fairly large maintenance cost, and isn't worth the small benefit of a more descriptive warning message.

yurivict:
Ok, merge it then.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5808#-:-LnRDJqQAq_TDzguLVap:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
Looks like travis CI got stuck, and didn't even receive the request on their side. I'll try force pushing a newline addition and removal to see if that re-triggers it.

anthonybilinski:
Nope, I just forgot to enable travis on dev branches, and earlier today enabled CI checks on dev branches, meaning this PR is now blocked on https://github.com/qTox/qTox/pull/5810.

PR #5804, opened 2019-08-27
docs(INSTALL): update Fedora instructions

ferdnyc:
I was recently informed that, while "rpmfusion-free" is a repo name, "RPM Fusion" is the name of the project that maintains it, and is two words. :wink: 

While I was in there, some other tweaks:
* Remove outdated sqlcipher note, as all current Fedora 29+ include it
* Add note about RPM Fusion being required to install ffmpeg
* Update `dnf` commands and package names for Fedora 30

- [x] Commits follow our [git commit guidelines](https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#git-commit-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5804)

PR #5809, opened 2019-08-29
chore(build): update simple_make for Fedora

ferdnyc:
In addition to a package-list refresh, enhance simple_make to:
- also add `/usr/local/lib64/` to `/etc/ld.so.conf.d/`
- export a corresponding `PKG_CONFIG_PATH` before running builds
- run `bootstrap.sh --without-sqlcipher`, which is preinstalled

As requested in #5804

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5809)

### DIFF
--- a/simple_make.sh
+++ b/simple_make.sh
@@ -86,20 +86,25 @@ dnf_install() {
         git
         glib2-devel
         gtk2-devel
+        kf5-sonnet-devel
+        libconfig-devel
+        libexif-devel
         libsodium-devel
-        libtool
         libvpx-devel
         libXScrnSaver-devel
         openal-soft-devel
         openssl-devel
         opus-devel
         qrencode-devel
+        qt5-devel
+        qt5-qtdoc
         qt5-qtsvg
         qt5-qtsvg-devel
         qt5-qttools-devel
-        qt-creator
-        qt-devel
-        qt-doc
+        qtsingleapplication-qt5
+        readline-devel
+        sqlcipher-devel
+        sqlite-devel
     )
     sudo dnf install "${dnf_packages[@]}"
 }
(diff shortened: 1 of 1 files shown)


ferdnyc:
I should probably explain why these changes became so extensive. Originally I just updated the list of package dependencies and ran the build, but:
1. Fedora on x86_64 uses `lib64/` paths for libraries, and CMake will respect that. So if you're installing into `/usr/local/`, both `/usr/local/lib/` and `/usr/local/lib64/` have to be accounted for both in the `ld.so` adjustments, and in packagekit (or what's installed can't be discovered). So, that became some extra code.
2. Originally I'd added the `tcl` package in, and was just letting `bootstrap.sh` build `sqlcipher` locally. But when running that on my development machine with `-l` (because I can't install into `/usr/local/` there), I discovered that the sqlcipher build doesn't respect the `-l` flag and was trying to update Tcl's `/usr/share/tcl8.6/sqlite3` anyway, so that was no good at all. Adding the `--without-sqlcipher` kept the build from trying to escape, and the build for `qtox` discovers and links it with the system `/usr/lib64/libsqlcipher-3.15.2.so.0` just like it's supposed to.

So, these changes let the build process complete on my Fedora 30 systems, and `make test` passes with flying colors.

To get qTox to build on my bleeding-edge Fedora Rawhide VM (Rawhide is a 0-day rolling release that receives every new package build as soon as it's submitted to the repos), I also had to comment out the `-Werror` in `CMAKE_CXX_FLAGS`, because the current system packages are internally inconsistent: gdk-pixbuf2 (which has yet to be updated) is calling functions that are newly-deprecated in the just-built Glib package, and with `-Werror` those deprecation warnings kill the build. 

But that's to be expected on a rawhide system with constant moving-target APIs, and the situation will persist until someone gets around to releasing a gdk-pixbuf2 update which removes those deprecated library calls.

sudden6 (approved):
:lgtm_strong: thank you! I don't see the Rawhide problem as too serious, people who run that probably know what to expect.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5809#-:-LnWhtXV9KhbqanJRz5b:bcpfety)* status: :shipit: complete! 1 of 1 LGTMs obtained

ferdnyc:
> I don't see the Rawhide problem as too serious, people who run that probably know what to expect.

Agreed, I mentioned it only as an FYI.

PR #5833, opened 2019-09-16
fix(core): don't log critical on every group join

### LINKED ISSUE #5118: Critical level error log on every group join
##### Brief Description

OS: Ubuntu 16.04
qTox version: v1.14.1+
Commit hash: 498c04b
toxcore: 0.2.2
Qt: 5.5.1

Reproducible: Almost Always

##### Steps to reproduce

1. Join a toxync group.

##### Observed Behavior
`core/core.cpp:1097 : Critical: Peer not found`

When first joining the group, we get `tox_callback_conference_title` which calls `onGroupTitleChange` with peerId == UINT32_MAX, indicating the peer is not yet known: ` * If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference).` Once we receive this, we call `Core::getGroupPeerName` with peerId == UINT32_MAX. This causes ToxCore to return "invalid peer number" since our peernumber >= g->numpeers:
```
int group_peername(const Group_Chats *g_c, uint32_t groupnumber, int peernumber, uint8_t *name)
{
    Group_c *g = get_group_c(g_c, groupnumber);

    if (!g) {
        return -1;
    }

    if ((uint32_t)peernumber >= g->numpeers) {
        return -2;
    }

    if (g->group[peernumber].nick_len == 0) {
        return 0;
    }
```
which causes us to log the critical error in qTox.

This may be a toxcore issue, in that it should return empty string for UINT32_MAX peer number, since it's known in toxcore that UINT32_MAX == empty, and the toxcore name for unkown peer name is the empty string, but we can trivially avoid calling the name at all by first just checking if peer number if UINT32_MAX in `Core::getGroupPeerName`

##### Expected Behavior
No critical error log during normal operation.

anthonybilinski:
It's expected behaviour that the peer is unknown when first joining a group.

Fix #5118

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5833)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -1160,6 +1160,11 @@ QString Core::getGroupPeerName(int groupId, int peerId) const
 {
     QMutexLocker ml{&coreLoopLock};
 
+    // from tox.h: "If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference)."
+    if (peerId != std::numeric_limits<uint32_t>::max()) {
+        return {};
+    }
+
     Tox_Err_Conference_Peer_Query error;
     size_t length = tox_conference_peer_get_name_size(tox.get(), groupId, peerId, &error);
     if (!parsePeerQueryError(error)) {


sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5833)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 597 at r1](https://reviewable.io/reviews/qtox/qtox/5833#-LoyAOmv8FlLeziFGPDg:-LoyAOmv8FlLeziFGPDh:b-65ytfh) ([raw file](https://github.com/qtox/qtox/blob/e436050b61addf3ee870ffff8924564cba4b49b9/src/core/core.cpp#L597)):*
> ```cpp
>     QString author{};
>     // from tox.h: "If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference)."
>     if (peerId != std::numeric_limits<uint32_t>::max()) {
> ```

I think that this isn't the right place to do this. I think using -1 is a valid use of the toxcore API and we should handle the peerId == UINT32_MAX in our wrapper

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5833)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 597 at r1](https://reviewable.io/reviews/qtox/qtox/5833#-LoyAOmv8FlLeziFGPDg:-LoyBKAB0LVanVq8LI2H:bprpf9w) ([raw file](https://github.com/qtox/qtox/blob/e436050b61addf3ee870ffff8924564cba4b49b9/src/core/core.cpp#L597)):*

I think that this isn't the right place to do this. I think using -1 is a valid use of the toxcore API and we should handle the peerId == UINT32_MAX in our wrapper
</blockquote></details>

Yeah agreed - makes way more sense in getPeerName. Done.

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5833#-:-LoyBU2U1DQUxsmY5TNi:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5848, opened 2019-09-21
fix(chatform): Fix invalid access of empty container

### LINKED ISSUE #5832: Segfault when loading history that has less than one day in it
##### Brief Description

OS:  Linux
qTox version: v1.16.3-603-gbd034e13
Commit hash:  bd034e130d1b0630f272d9b36a33b0e16bc44d9e
toxcore:  0.2.10
Qt:  5.9.8

Reproducible: Always

##### Steps to reproduce

1. Start chat with new friend
2. Clear messages
3. Load messages

##### Observed Behavior
Segfault

##### Expected Behavior
History is loaded

##### Additional Info
Looks like a combo of two bugs here. The History function for getNumMessagesForFriendBeforeDateBoundaries started returning the last item on a date instead of the first. I believe this is because GROUP BY does not define which item it will select.

The segfault is caused from a missing out of bounds check in GenericChatForm::loadHistory

Solving the second part is easy, the first part requires some more investigation.

sphaerophoria:
This partially fixes #5832 in that we no long segfault. There is a
deeper underlying problem that the history still relies on undefined
behavior of GROUP BY causing us to get an empty container in some cases.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5848)

### DIFF
--- a/src/widget/form/genericchatform.cpp
+++ b/src/widget/form/genericchatform.cpp
@@ -632,9 +632,12 @@ void GenericChatForm::loadHistory(const QDateTime &time, const LoadHistoryDialog
 
     if (type == LoadHistoryDialog::from) {
         loadHistoryFrom(time);
-        auto msg = messages.cbegin()->second;
-        chatWidget->setScroll(true);
-        chatWidget->scrollToLine(msg);
+        if (!messages.empty())
+        {
+            auto msg = messages.cbegin()->second;
+            chatWidget->setScroll(true);
+            chatWidget->scrollToLine(msg);
+        }
     } else {
         loadHistoryTo(time);
     }


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5848)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/genericchatform.cpp, line 689 at r1](https://reviewable.io/reviews/qtox/qtox/5848#-LpUtfcT4aY0tAPTqh76:-LpUtfcT4aY0tAPTqh77:bt3qqre) ([raw file](https://github.com/qtox/qtox/blob/aaa4c415612ed5b9d1d0ebfcd67a88e17de30bd9/src/widget/form/genericchatform.cpp#L689)):*
> ```cpp
>     auto end = ChatLogIdx(0);
>     if (time.isNull()) {
>         end = messages.begin()->first;
> ```

Isn't this also a potential crash if `messages` is empty?
___
*[src/widget/form/genericchatform.cpp, line 720 at r1](https://reviewable.io/reviews/qtox/qtox/5848#-LpUtu3R9LAnsERe99OS:-LpUtu3R9LAnsERe99OT:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/aaa4c415612ed5b9d1d0ebfcd67a88e17de30bd9/src/widget/form/genericchatform.cpp#L720)):*
> ```cpp
>     auto begin = ChatLogIdx(0);
>     if (time.isNull()) {
>         begin = messages.rbegin()->first;
> ```

same

anthonybilinski:
rebased to v1.17-dev tip. It does look like there are a couple unguarded usages of `messages`'s first item.

PR #5855, opened 2019-09-29
Add AppArmor 2.13.3 profile

Talkless:
- [x ] Commits follow our [git commit guidelines](https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#git-commit-guidelines)

Ubuntu 19.10 (and current Debian and openSUSE rolling releases) has AppArmor 2.13.3 with updated abstractions. This allows us to create cleaned-up profile, removing rules that are already available in the latest AppArmor release.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5855)

### DIFF
--- a/security/apparmor/2.13.3/install.sh
+++ b/security/apparmor/2.13.3/install.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+#   Copyright © 2019 by The qTox Project Contributors
+#
+#   This file is part of qTox, a Qt-based graphical interface for Tox.
+#   qTox is libre software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   qTox is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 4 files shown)


Talkless:
@setharnold could you review this? You did take a look at original (introducing AppArmor profile) PR. Here, second commit is where you can see actual changes wrt. AppArmor 2.13.3.

sudden6:
@Talkless do you think this should be merged even without some further review?

Talkless:
> @Talkless do you think this should be merged even without some further review?

Ugh, well, I guess so. I believe this is no way critical, as probably I am only user of these AppArmor profiles :-) .

sudden6 (approved):
:lgtm_strong: it's probably easier for downstream packagers to take these as a basis than to come up with their own rules from scratch, so I think we'll ship it.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5855#-:-M2wkzBqC3PyTxu6ua07:bp9e457)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (approved):
I'm not very familiar with apparmor, but read through it and nothing seems obviously wrong. I agree with including it - if it's overly restrictive it will be obvious to users of it, and if it's not restrictive enough it's still better than nothing. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5855#-:-M3-yqBg9iWPuDSCULZm:b3h5e6r)* status: :shipit: complete! 2 of 1 LGTMs obtained

PR #5861, opened 2019-10-05
chore(travis): make travis happy by having some output - but not too much

### LINKED ISSUE #5515: TravisCI times out on OSX compilation
When TravisCI brew installs qt5, qt5 needs to be compiled if it is not cached. TravisCI kills our job because there has been no terminal output for 10 minutes, since brew by default does not show compilation progress, and qt5 takes over 10 minutes to build.

https://github.com/qTox/qTox/pull/5490/files used brew's `--verbose` option to show compilation progress, which stops TravisCI for killing us due to not output, but now we get killed for outputting _too_ much, with non-stop compilation spam for the duration of the Qt build.

The hackiest solution is probably to have no build output, then output `.` once every 10s in parallel, to keep us from being killed by TravisCI. It would be nice if there's some way to get some relevant progress without too much spam, or to let TravisCI know that we're going to be outputting nothing for a while and not to worry about it.

anthonybilinski:
Fix #5515

brew install qt5 might take a long time to build Qt. Travis kills us if we
don't output for 10 minutes. Travis also kills us if we output too much, so
verbose isn't an option. So just output some dots...

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5861)

### DIFF
--- a/osx/qTox-Mac-Deployer-ULTIMATE.sh
+++ b/osx/qTox-Mac-Deployer-ULTIMATE.sh
@@ -170,9 +170,23 @@ install() {
     # needed for kf5-sonnet
     brew tap kde-mac/kde
 
-    # verbose so that build output is shown, otherwise qt5 build has no output for >10 mins
-    # and is killed by Travis CI
-    brew install --verbose ffmpeg libexif qrencode qt5 sqlcipher openal-soft #kf5-sonnet
+    # brew install qt5 might take a long time to build Qt. Travis kills us if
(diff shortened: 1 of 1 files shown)


anthonybilinski:
Since even dots aren't enough to make travis happy (the whole job times out later, even now that it no longer times out because there's no output), the follow up solution is being worked on in https://github.com/qTox/qTox/pull/5866. Closing this.

PR #5872, opened 2019-10-11
fix(history): move stuck action messages to broken_messages table

anthonybilinski:

Based on top of https://github.com/qTox/qTox/pull/5773 since this does another db version upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5872)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -26,7 +26,7 @@
 #include "db/rawdatabase.h"
 
 namespace {
-static constexpr int SCHEMA_VERSION = 2;
+static constexpr int SCHEMA_VERSION = 3;
 
 bool createCurrentSchema(RawDatabase& db)
 {
@@ -150,6 +150,34 @@ bool dbSchema1to2(RawDatabase& db)
     return db.execNow(upgradeQueries);
 }
 
+bool dbSchema2to3(RawDatabase& db)
+{
+    // Any faux_offline_pending message with the content "/me " are action
+    // messages that qTox previously let a user enter, but that will cause an
+    // action type message to be sent to toxcore, with 0 length, which will
+    // always fail. They must be be moved from faux_offline_pending to broken_messages
+    // to avoid qTox from erroring trying to send them on every connect
+
+    const QString emptyActionMessageString = "/me ";
+
+    QVector<RawDatabase::Query> upgradeQueries;
+    upgradeQueries += RawDatabase::Query{QString("INSERT INTO broken_messages "
+            "SELECT faux_offline_pending.id FROM "
+            "history JOIN faux_offline_pending "
+            "ON faux_offline_pending.id = history.id "
+            "WHERE history.message = ?;"),
+            {emptyActionMessageString.toUtf8()}};
+
+    upgradeQueries += QString(
+        "DELETE FROM faux_offline_pending "
+        "WHERE id in ("
+            "SELECT id FROM broken_messages);");
+
+    upgradeQueries += RawDatabase::Query(QStringLiteral("PRAGMA user_version = 3;"));
+
+    return db.execNow(upgradeQueries);
+}
+
 /**
 * @brief Upgrade the db schema
 * @note On future alterations of the database all you have to do is bump the SCHEMA_VERSION
@@ -216,6 +244,13 @@ void dbSchemaUpgrade(std::shared_ptr<RawDatabase>& db)
        }
        qDebug() << "Database upgraded incrementally to schema version 2";
        //fallthrough
+    case 2:
+       if (!dbSchema2to3(*db)) {
+            qCritical() << "Failed to upgrade db to schema version 3, aborting";
+            db.reset();
+            return;
+       }
+       qDebug() << "Database upgraded incrementally to schema version 3";
     // etc.
     default:
         qInfo() << "Database upgrade finished (databaseSchemaVersion" << databaseSchemaVersion
--- a/test/persistence/dbschema_test.cpp
+++ b/test/persistence/dbschema_test.cpp
@@ -36,6 +36,7 @@ private slots:
     void testIsNewDb();
     void test0to1();
     void test1to2();
+    void test2to3();
     void cleanupTestCase();
 private:
     bool initSucess{false};
@@ -48,7 +49,8 @@ const QString testFileList[] = {
     "testIsNewDbTrue.db",
     "testIsNewDbFalse.db",
     "test0to1.db",
-    "test1to2.db"
+    "test1to2.db",
+    "test2to3.db"
 };
 
 const QMap<QString, QString> schema0 {
@@ -76,6 +78,9 @@ const QMap<QString, QString> schema2 {
     {"peers", "CREATE TABLE peers (id INTEGER PRIMARY KEY, public_key TEXT NOT NULL UNIQUE)"},
     {"broken_messages", "CREATE TABLE broken_messages (id INTEGER PRIMARY KEY)"}
 };
+
+// move stuck 0-length action messages to the existing "broken_messages" table. Not a real schema upgrade.
+const auto schema3 = schema2;
 
 void TestDbSchema::initTestCase()
 {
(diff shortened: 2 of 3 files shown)


sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 1 files at r1, 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5872#-:-LqwBoQR5XoG3NseDhN-:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
I quickly run this PR on a test profile and it seems to at least start without issues.

src/persistence/history.cpp
```diff
@@ -26,10 +26,11 @@
 #include "db/rawdatabase.h"
 
 namespace {
-static constexpr int SCHEMA_VERSION = 1;
+static constexpr int SCHEMA_VERSION = 3;
 
-void generateCurrentSchema(QVector<RawDatabase::Query>& queries)
+bool createCurrentSchema(std::shared_ptr<RawDatabase>& db)
```
Diadlo: Why do you pass the pointer by reference? It doesn't look like you change pointer itself. 
Since these functions only uses a pointer, it's should be a raw pointer. See [CppCoreGudlines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r3-a-raw-pointer-a-t-is-non-owning)

test/persistence/dbschema_test.cpp
```diff
@@ -35,19 +35,53 @@ private slots:
...
 };
 
 const QString testFileList[] = {
     "testCreation.db",
     "testIsNewDbTrue.db",
     "testIsNewDbFalse.db",
-    "test0to1.db"
+    "test0to1.db",
+    "test1to2.db",
+    "test2to3.db"
```
Diadlo, line 53: You can add an extra colon in the end, to avoid line change in the future

src/persistence/history.cpp
anthonybilinski: Changed to a raw pointer, rather than a raw reference due to the arguments made here https://google.github.io/styleguide/cppguide.html#Output_Parameters.

Diadlo: 1. Sorry, I meant raw reference.
2. Why `db` is an output parameter?
3. If you use the pointer, but not reference, it will be good to check, that `db` is not `nullptr`. Fill free to ignore this point, because out of the scope.

anthonybilinski: re 2., oops I linked to the wrong section, I meant to link to https://google.github.io/styleguide/cppguide.html#Reference_Arguments.
re 3:, I feel like both the copy option, reference option, and pointer options are bad =/
3.1 copy forces an atomic increment and copy of the smart pointer object
3.2 reference has the pointer implementation with force non-null that I want, but as mentioned in the corrected link "References can be confusing, as they have value syntax but pointer semantics."
3.3 pointer can be null, requiring either null checks on every small function, or weak function documents like `// @param db, pointer to database, can't be null` with no actual language support

stupid C++ =/

of these I prefer the pointer option, but dislike the boilerplate code required on every function to check pointers are non-null when they're basically already known to be non-null... I guess checking pre-conditions is still a good idea. I've added an assert non null in each function.

Diadlo: Stupid not C++ (in this case, at least ;) ), but Google Cpp Guide. If I want to iterate over a container and change it's value I just use `for (auto& x : vector) { x.change(); } `. Nothing more, nothing less. Why I can't do the same with functions?

I suggest using point from [CppCoreGudlines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f17-for-in-out-parameters-pass-by-reference-to-non-const). At a minimum, this is a more generally accepted guide, than GoogleCppGuid.

anthonybilinski: I have no problem with it in a case like `for (auto& x : vector) { x.change(); }`, but for functions, it looks to the caller like they're passing in a copy of the argument, when they're really passing in a modifiable reference to it. Also when the reference is stored in a class long term, it looks again like you hold the object, when you might just have a non-owning reference to someone else's object. Pointers having their own syntax makes it clear every time you use `->` that you're not holding the object directly.

I think having a complete set of coding standards is more valuable than arguing each point, so I'm happy to yield to iso CPP core guidelines. I'll update these to references. I also chatted with @sudden6 and one of us will add a line to our coding standards to refer to cpp core guidelines in that case where something isn't covered by our coding standard (like this).

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5872)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 31 at r6](https://reviewable.io/reviews/qtox/qtox/5872#-LqteRYOK5GdAxzS0gun-r6-31:-Lr6aSAR8rIV0smdX79j:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/7d691ac21b806dd74bab257480b630421994f908/src/persistence/history.cpp#L31)):*

I have no problem with it in a case like `for (auto& x : vector) { x.change(); }`, but for functions, it looks to the caller like they're passing in a copy of the argument, when they're really passing in a modifiable reference to it. Also when the reference is stored in a class long term, it looks again like you hold the object, when you might just have a non-owning reference to someone else's object. Pointers having their own syntax makes it clear every time you use `->` that you're not holding the object directly.

I think having a complete set of coding standards is more valuable than arguing each point, so I'm happy to yield to iso CPP core guidelines. I'll update these to references. I also chatted with @sudden6 and one of us will add a line to our coding standards to refer to cpp core guidelines in that case where something isn't covered by our coding standard (like this).
</blockquote></details>

Done.

Diadlo (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5872#-:-Lr7rrp8CFe_n64pKL8U:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 31 at r6](https://reviewable.io/reviews/qtox/qtox/5872#-LqteRYOK5GdAxzS0gun-r6-31:-Lr7rzNO8mT07_nu2VC1:b-gyzcvt) ([raw file](https://github.com/qtox/qtox/blob/7d691ac21b806dd74bab257480b630421994f908/src/persistence/history.cpp#L31)):*

Done.
</blockquote></details>

I feel like I made you do it :/

It was just an opinion with reference to CCG

anthonybilinski:
Hah no worries, it's good for the project to have a consistent style, and CCG is a good default to go to. I'm happy to have that decided as the fallback when our coding standard doesn't cover a topic, so that we don't have to discuss each item that comes up :)

anthonybilinski (commented):
@sudden6 tested this and sent me exactly one old message when he started qTox after the upgrade. I'll take a look at the upgrade logic and extend the test to try to repro whatever case he hit.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5872#-:-LrOGUcK0vd9o0drmi4T:b-ts7ndp)* status: :shipit: complete! 2 of 1 LGTMs obtained

sudden6:
The problematic message has the following row in the `history` table before the upgrade:

```
_rowid_ , id, timestamp, chat_id, sender_alias, message, file_id
"34697"	"34697"	"1565085926105"	"86"	"2"	"going afk for a few days, see you"
```

And after the upgrade it's still in the `history` table and there are no messages in the `broken_messages` table:
```
"34697"	"34697"	"1565085926105"	"86"	"2"	"going afk for a few days, see you"	
```

anthonybilinski:
Oops, last two comments were actually regarding https://github.com/qTox/qTox/pull/5773, and the issue has been fixed as mentioned in https://github.com/qTox/qTox/pull/5773#pullrequestreview-304209945.

anthonybilinski:
#5773 has been in a while and working well, this one's been pretty thoroughly tested and is next in the db upgrade chain, so I'll go ahead and merge. Note this is another version bump in the db, so another checkpoint in qTox's version. databases that are run on this version will not be loadable by older versions of qTox.

PR #5894, opened 2019-10-16
fix(loginScreen): make loginScreen return values comply with Qt standards

### LINKED ISSUE #5781: qTox hangs when login errors or is cancelled (not wrong password)
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
Commit hash:  2dde934abed9433c41d56a7492e12fa3f0041e35
toxcore:  v0.2.10
Qt:  5.12.2
Hardware: Desktop
  
…

Reproducible: Rarely

##### Steps to reproduce

1.  Attempt to open locked profile.

or

1. As login screen but before the main screen, ctrl+c - I'm not able to consistently repro this but have seen it a couple times, I think.

##### Observed Behavior
qTox windows goes away, but qTox does not return - the process hangs indefinitely.
```
$ qtox -p qTest2
[06:17:34.955 UTC] persistence/settings.cpp:123 : Debug: "Loading settings from /home/abilinski/.config/tox/qtox.ini"
[06:17:35.022 UTC] ipc.cpp:61 : Debug: Our global IPC ID is  9129450030786964297
[06:17:35.023 UTC] main.cpp:274 : Debug: commit:  5b4d9dcd7dc04901e31d99758a712932cf43313d
[06:17:35.025 UTC] persistence/profile.cpp:131 : Warning: Failed to lock profile  "qTest2"
[06:17:35.025 UTC] persistence/settings.cpp:287 : Warning: "Could not load new settings (profile change to nullptr)"


^Z
```
##### Expected Behavior
qTox exits.

##### Additional Info
I think this started happening since 0a30a657fcd78bc8163131ae6cac26f4fc7acce8

Kribylet:
This commit makes LoginScreen return QDialog::Rejected (0) and
QDialog::Accepted (1) instead of C standard return values. This should be more
robust with regards to special cases in Qt implementation.

Fixes #5781 

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5894)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -391,8 +391,8 @@ int main(int argc, char* argv[])
     } else {
         nexus.setParser(&parser);
         int returnval = nexus.showLogin(profileName);
-        if (returnval != 0) {
-            return returnval;
+        if (returnval == QDialog::Rejected) {
+            return -1;
         }
     }
 
--- a/src/nexus.cpp
+++ b/src/nexus.cpp
@@ -163,6 +163,11 @@ int Nexus::showLogin(const QString& profileName)
     // The connection order ensures profile will be ready for bootstrap for now
     connect(this, &Nexus::currentProfileChanged, this, &Nexus::bootstrapWithProfile);
     int returnval = loginScreen.exec();
+    if (returnval == QDialog::Rejected) {
+        // Kriby: This will terminate the main application loop, necessary until we refactor
+        // away the split startup/return to login behavior.
+        qApp->quit();
+    }
     disconnect(this, &Nexus::currentProfileChanged, this, &Nexus::bootstrapWithProfile);
     return returnval;
 }
--- a/src/widget/loginscreen.cpp
+++ b/src/widget/loginscreen.cpp
@@ -72,15 +72,6 @@ LoginScreen::~LoginScreen()
     delete ui;
 }
 
-void LoginScreen::closeEvent(QCloseEvent* event)
-{
-
-    // If we are in the bootstrap, returning -1 will give us something to exit with in main.cpp
-    this->setResult(-1);
-    // If we are in application exec, we can quit by closing it, instead.
-    qApp->quit();
-}
-
 /**
  * @brief Resets the UI, clears all fields.
  */
@@ -110,10 +101,11 @@ void LoginScreen::reset(const QString& initialProfileName)
 
 void LoginScreen::onProfileLoaded()
 {
-    done(0);
+    done(QDialog::Accepted);
 }
 
-void LoginScreen::onProfileLoadFailed() {
+void LoginScreen::onProfileLoadFailed()
+{
     QMessageBox::critical(this, tr("Couldn't load this profile"), tr("Wrong password."));
     ui->loginPassword->setFocus();
     ui->loginPassword->selectAll();
--- a/src/widget/loginscreen.h
+++ b/src/widget/loginscreen.h
@@ -21,9 +21,9 @@
 #ifndef LOGINSCREEN_H
 #define LOGINSCREEN_H
 
+#include <QDialog>
 #include <QShortcut>
 #include <QToolButton>
-#include <QDialog>
 
 class Profile;
 
@@ -46,9 +46,6 @@ signals:
     void autoLoginChanged(bool state);
     void createNewProfile(QString name, const QString& pass);
     void loadProfile(QString name, const QString& pass);
-
-protected:
-    virtual void closeEvent(QCloseEvent* event) final override;
 
 public slots:
     void onProfileLoaded();


sudden6:
Our commit linting script complains about a too long commit title, forgotten line break?

Kribylet:
Fixed missing line break

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5894)* status: 0 of 1 LGTMs obtained

___
*[src/widget/loginscreen.h, line 52 at r1](https://reviewable.io/reviews/qtox/qtox/5894#-LrN35vi0t6BRvfQJ_VC:-LrN35vi0t6BRvfQJ_VD:b-dd7yts) ([raw file](https://github.com/qtox/qtox/blob/1fff7f2b8ef4a2ecebf8d9eeb8e5c1cc17bdce05/src/widget/loginscreen.h#L52)):*
This is a bit out of scope of the bug fix, but, according to https://doc.qt.io/qt-5/qdialog.html#return-value-modal-dialogs:
> In order to modify your dialog's close behavior, you can reimplement the functions accept(), reject() or done(). The closeEvent() function should only be reimplemented to preserve the dialog's position or to override the standard close or reject behavior.

From the general description:
> Typically, to get the dialog to close and return the appropriate value, we connect a default button, e.g. OK, to the accept() slot and a Cancel button to the reject() slot. Alternatively you can call the done() slot with Accepted or Rejected.

What do you think about either calling done(Rejected)/done(Accepted) instead of 0/-1, or using the default `rejected` signal, but checking the return code against Accepted/Rejected instead of against 0/1? Then onProfileLoaded could be connected directly to `accept`, and `rejected` could be left unhandled - I think.

Kribylet:
Changed method of issue resolution. Commit title and description changed accordingly. This commit now changes LoginScreen return values to Qt standards (1 success, 0 failure) rather than traditional C values. This is a bit unsightly considering C standards are used elsewhere, but whenever referred to the appropriate enums should be in use, so it shouldn't be too bad. This has the benefit of catching special cases in the Qt implementation, like the one causing the referenced issue, and should be more robust.

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5894)* status: 0 of 1 LGTMs obtained

___
*[src/widget/loginscreen.h, line 52 at r1](https://reviewable.io/reviews/qtox/qtox/5894#-LrN35vi0t6BRvfQJ_VC:-LrSq39F75wihtE94fiz:bychrka) ([raw file](https://github.com/qtox/qtox/blob/1fff7f2b8ef4a2ecebf8d9eeb8e5c1cc17bdce05/src/widget/loginscreen.h#L52)):*

This is a bit out of scope of the bug fix, but, according to https://doc.qt.io/qt-5/qdialog.html#return-value-modal-dialogs:
> In order to modify your dialog's close behavior, you can reimplement the functions accept(), reject() or done(). The closeEvent() function should only be reimplemented to preserve the dialog's position or to override the standard close or reject behavior.

From the general description:
> Typically, to get the dialog to close and return the appropriate value, we connect a default button, e.g. OK, to the accept() slot and a Cancel button to the reject() slot. Alternatively you can call the done() slot with Accepted or Rejected.

What do you think about either calling done(Rejected)/done(Accepted) instead of 0/-1, or using the default `rejected` signal, but checking the return code against Accepted/Rejected instead of against 0/1? Then onProfileLoaded could be connected directly to `accept`, and `rejected` could be left unhandled - I think.
</blockquote></details>

This seems a more robust solution overall. Modified the PR accordingly.

anthonybilinski (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5894)* status: 0 of 1 LGTMs obtained

___
*[src/nexus.cpp, line 171 at r2](https://reviewable.io/reviews/qtox/qtox/5894#-LrVmNDE1T1qxUv4FoHs:-LrVmNDE1T1qxUv4FoHt:b-5zduzk) ([raw file](https://github.com/qtox/qtox/blob/b5bd3c9361e26f39783ec4260b09e4e114af0b2f/src/nexus.cpp#L171)):*
> ```cpp
>         // Kriby: This will terminate the main application loop, necessary until we refactor
>         // away the split startup/return to login behavior.
>         qApp->quit();
> ```

why is this now required? was it also required before but we were missing it?

Kribylet (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5894)* status: 0 of 1 LGTMs obtained

___
*[src/nexus.cpp, line 171 at r2](https://reviewable.io/reviews/qtox/qtox/5894#-LrVmNDE1T1qxUv4FoHs:-LrW4Ci74pv7xTIe2yYd:bdu3baq) ([raw file](https://github.com/qtox/qtox/blob/b5bd3c9361e26f39783ec4260b09e4e114af0b2f/src/nexus.cpp#L171)):*

why is this now required? was it also required before but we were missing it?
</blockquote></details>

It was required previously, check the closeEvent diff in loginScreen. As the comment implies it is an artifact of how qTox handles the initial login as a special case (started first through loginscreen.exec(), but later loginscreens are spawned from the main application). I have tried to refactor this  previously but at the time I could not solve it easily.

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5894#-:-LrW9O4k0iI733OjS-m4:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
I've cherry picked your change to the v1.17-dev branch and updated you fork, and changed the base for this PR so that the change will be included in the v1.17.x release(s), since it's a nice scoped bug fix.

PR #5898, opened 2019-10-17
revert(toxme): remove toxme from qTox - the service is offline permanently

### LINKED ISSUE #5897: Future of qTox and ToxMe after toxme.io goes offline
toxme.io has been taken offline permanently: https://github.com/LittleVulpix/toxme/commit/52c731993ad9bfb6eacde8ace91f896d718a2811

Given this, what should qTox do about ToxMe support? Theoretically someone with other domains or using the same domain could take over the hosting.

1) Remove ToxMe completely from qTox.
2) Support generic name@domain format addresses, and try to fetch a ToxID from whatever address is entered. Still remove registration on ToxMe our Profile form.
3) Something else?

It seems like an important part of info needed is: Is ToxMe a kind of protocol/spec, or just that specific website's API? If it is standardized, does this include account registration?

I'm personally leaning towards just taking ToxMe out of qTox completely. I never really liked the centralized nature and metadata disclosure of hosted name lookup, and there hasn't been any volunteers to take over hosting of toxme.io on IRC so far, meaning qTox will be left with a registration page and friend lookup that times out on every attempt for the foreseeable future, which is a poor user experience.

@qTox/core?

anthonybilinski:
As of 2019-10-09, toxme.io was taken offline permanently. Remove UI and code
in qTox relating to it. Revert this commit if it comes back online in the
future.

Fix #5897

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5898)


Diadlo (approved):
:lgtm_strong:

> Reviewed 15 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5898#-:-LrNJLyu5k4xBpLF8TtI:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5898#-:-LrNcEiKCB_TXu_rJK7N:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

sudden6:
Since there's still no service for replacing toxme.io I'd suggest to remove this now.

anthonybilinski:
From some of the discussion on #5897:
1) toxme could exist on other smaller servers that we don't know about, and they could still support the protocol
2) It may not be easy to get Tox IDs manually from those toxme servers
3) If we keep toxme support, we should add public key verification of any server added by the user

Given this, do we still want to remove toxme support from qTox?

Personally, I don't like that toxme support in qTox seemingly endorses the validity of the ID returned by whatever server is punched in. Currently a user could be sent some random malicious toxme server, the user would enter it in qTox, and everything would work fine. With a Tox ID I think it's more clear to the user that they're trusting the name they're putting in.

I also think that if we do keep toxme, we should add TOFU of servers, and make it clearer in the UI that the user needs to trust the server.

sudden6:
There may be private servers running the toxme protocol, but I doubt it. I think at the moment it only confuses people.

IRC quote:
```
> on qTox client I see there is an option to regster to ToxMe. What is that?
```

>It may not be easy to get Tox IDs manually from those toxme servers

To get the ToxIDs from the server one could still use an old version of qTox if that's really the issue.

> If we keep toxme support, we should add public key verification of any server added by the user

Agree, right now it just seems like an invitation for a MITM attack.

anthonybilinski:
Agreed the UX is confusing and our implementation in insecure. I'll go ahead and merge.

PR #5900, opened 2019-10-18
edit settings for extra themes

TriKriSta:
Screenshot_20191018_193557
Screenshot_20191018_192540

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5900)

### DIFF
--- a/src/persistence/settings.cpp
+++ b/src/persistence/settings.cpp
@@ -213,6 +213,8 @@ void Settings::loadGlobal()
                 style = "None";
         }
         nameColors = s.value("nameColors", false).toBool();
+        extraColor = s.value("extraColor", "#484848").toString();
+        stateExtraColors = s.value("stateExtraColors", 0).toInt();
     }
     s.endGroup();
 
@@ -678,6 +680,8 @@ void Settings::saveGlobal()
         s.setValue("themeColor", themeColor);
         s.setValue("style", style);
         s.setValue("nameColors", nameColors);
+        s.setValue("extraColor", extraColor);
+        s.setValue("stateExtraColors", stateExtraColors);
         s.setValue("statusChangeNotificationEnabled", statusChangeNotificationEnabled);
         s.setValue("spellCheckingEnabled", spellCheckingEnabled);
     }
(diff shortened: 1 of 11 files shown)


sudden6:
AFAICT this allows the user to choose arbitrary colors as the "Accent color", right?

TriKriSta:
Yes

Chiitoo:
Forgot to comment on this waaays back...

The 'Choice' on the button seems a bit weird to me.  'Choose' would probably be a more appropriate form, but that's somewhat weird as well.

Maybe something like 'Choose Colours' (or 'Choose colours'/'Choose colors'), 'Settings' or 'Configure'?

TriKriSta:
I changed to "Choose colors"

TriKriSta:
Can merge it if it doesn't have problem?

anthonybilinski (commented):
Sorry for the long delay, looking at the feature (code not reviewed yet) I really like this.

Before we mixed this "accent colours" with our real full themes (default/dark) which made it hard to notice that "dark" was a real full theme.

With this change, the defaut/dark theme are a lot more obvious since they are the only two options, and instead of having a dropdown of 8 arbitrary colours listed in text, the user can see the colour in the colour picker and apply anything they like.

This seems like it makes UI more clear for themes, reduces options in settings, all while improving customisability for accent colours.

I'll try to get to the code review in the next couple days, and I'll rebase up to master tip resolving conflicts.

I think maybe a couple small UI changes like changing the theme color to a checkbox of just "dark theme" since there are only two options now, and maybe changing wording and spacing of hte extra colors picker, but overall it seems nice.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5900#-:-MFrHyTf4K0lBeSkQ0__:b-6yatc9)* status: 0 of 1 LGTMs obtained

TriKriSta:
`I think maybe a couple small UI changes like changing the theme color to a checkbox of just "dark theme"`
I think need to stay a combobox. Because I'm going to add functions that allow to load new themes from some directory without recompile program. And after user can choose new themes from this combobox.

Issue #5918, opened 2019-11-08
Goup call UI covers most of chat with useless icons


anthonybilinski:
##### Brief Description

qTox version: v1.16.3+
Commit hash:  1a726b54cd2e1d72c6cb6f0f9a733e904f5e5371
toxcore:  0.2.10
Qt:  5.12.4
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Join group audio call.

##### Observed Behavior
1. A blank avatar is shown, covering half the group's chat.
2. The name on that blank avatar is your own, but is elided after 2 characters if the divider is moved down.
3. Resizing the divider between the blank avatar and the avatars of friends causes inconsistent splits between the avatars of friends and the large blank avatar.
5. The section of group member avatars is scrollable, changing between fully visible and half visible.
6. The large blank friend picture rescales strangely, causing clipping when the divider is moved.

##### Expected Behavior
1. The list of friends is shown, and which is in the call is shown up top, as is done with the bolding of group member names.
2. The rest of the chat area is used for chat.

This would be consistent with 1-1 calls, where the difference between being in a call and not is just the colour of the call button. Covering 3/4 of the chat area with friend avatars and a large blank icon is weird. Showing friend avatars may be nice, but then it should be done in text chats as well, like proposed in https://github.com/qTox/qTox/issues/5702.

Screenshot of weird UI:
![image](https://user-images.githubusercontent.com/10469203/68467106-7ff4e980-01ca-11ea-8b9e-d8d4531c34ae.png)

PR #5920, opened 2019-11-10
Fix group call SIGSEGV on window resize

sphaerophoria:
Avatars for group members currently in a call are resized depending on
the area they are displayed in. Previously a scrollbar would appear and
disapear based on the size of the contents. This resulted in
oscillations that ended in a SIGSEGV.
    
This fix avoids the oscillations by fixing the scrollbar to always be
shown.

After changing the layout I noticed that label text was incorrect so I fixed that too.

This fix is _far_ from ideal but my main goal here was to avoid the crashes while disrupting UI as little as possible. Given #5918 exists I'm not motivated to put a lot of work in to make sure it looks good at the same time.

New UI looks as follows
image

Note the permanently visible scrollbar in the bottom.

Test steps to produce the SIGSEGV are to basically attempt to resize your window to look like this

image

As the avatars start scrolling if you play with the size of the window you can end up in a case where a SIGSEGV occurs

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5920)

### DIFF
--- a/src/video/groupnetcamview.cpp
+++ b/src/video/groupnetcamview.cpp
@@ -37,7 +37,7 @@
 class LabeledVideo : public QFrame
 {
 public:
-    LabeledVideo(const QPixmap& avatar, QWidget* parent = nullptr, bool expanding = true)
+    LabeledVideo(const QPixmap& avatar, QString fontColorString, QWidget* parent = nullptr, bool expanding = true)
         : QFrame(parent)
     {
         qDebug() << "Created expanding? " << expanding;
@@ -48,7 +48,7 @@ public:
         connect(videoSurface, &VideoSurface::ratioChanged, this, &LabeledVideo::updateSize);
         label = new CroppingLabel(this);
         label->setTextFormat(Qt::PlainText);
-        label->setStyleSheet("color: white");
+        label->setStyleSheet(QString("color: %1").arg(fontColorString));
 
         label->setAlignment(Qt::AlignCenter);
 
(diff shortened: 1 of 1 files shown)


sudden6:
Do you know why the oscillations cause a segfault?

sphaerophoria:
I couldnt quite tie it together. I had a like 90000 deep call stack though
which i assumed was related.

sudden6:
ok, still better to have a workaround then instead of the crash.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5920#-:-LtLp8hfAvFY6G0VQrvu:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sphaerophoria:
> ok, still better to have a workaround then instead of the crash.

yeah my feelings too. Not ideal but better than crashing

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5920#-:-LtUCJLU9iPWSd41UHUx:b-jbsf8z)* status: :shipit: complete! 2 of 1 LGTMs obtained

anthonybilinski:
Rebased to branch `v1.17-dev` and PR base branch was updated accordingly.

sudden6:
@anthonybilinski this PR seems merged to v1.17-dev and should be closed, right?

anthonybilinski:
It wasn't, but it is now :P

PR #5926, opened 2019-11-14
refactor(coreav): fix locking and threading between Core and CoreAV

sudden6:
Use a standard mutex instead of trying to build proper locking
ourselfes.

This should make it easier to fix all the concurrency issues in Core and CoreAV. We also need to rework the calls to c-toxcore with regards to toxav threading.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -628,7 +628,7 @@ ToxCorePtr Core::makeToxCore(const QByteArray& savedata, const ICoreSettings* co
 
     // toxcore is successfully created, create toxav
     // TODO(sudden6): don't create CoreAv here, Core should be usable without CoreAV
-    core->av = CoreAV::makeCoreAV(core->tox.get());
+    core->av = CoreAV::makeCoreAV(core->tox.get(), core->coreLoopLock);
     if (!core->av) {
         qCritical() << "Toxav failed to start";
         if (err) {
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -484,12 +484,14 @@ Core::Core(QThread* coreThread)
 
 Core::~Core()
 {
-    // need to reset av first, because it uses tox
-    av.reset();
-
+    /*
+     * First stop the thread to stop the timer and avoid Core emitting callbacks
+     * into an already destructed CoreAV.
+     */
     coreThread->exit(0);
     coreThread->wait();
 
+    av.reset();
     tox.reset();
 }
 
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -68,11 +68,12 @@
  * deadlock.
  */
 
-CoreAV::CoreAV(std::unique_ptr<ToxAV, ToxAVDeleter> toxav)
+CoreAV::CoreAV(std::unique_ptr<ToxAV, ToxAVDeleter> toxav, QMutex& toxCoreLock)
     : audio{nullptr}
     , toxav{std::move(toxav)}
     , coreavThread{new QThread{this}}
     , iterateTimer{new QTimer{this}}
+    , coreLock{toxCoreLock}
 {
     assert(coreavThread);
     assert(iterateTimer);
@@ -104,7 +105,7 @@ void CoreAV::connectCallbacks(ToxAV& toxav)
  * @param core pointer to the Tox instance
  * @return CoreAV instance on success, {} on failure
  */
-CoreAV::CoreAVPtr CoreAV::makeCoreAV(Tox* core)
+CoreAV::CoreAVPtr CoreAV::makeCoreAV(Tox* core, QMutex &toxCoreLock)
 {
     TOXAV_ERR_NEW err;
     std::unique_ptr<ToxAV, ToxAVDeleter> toxav{toxav_new(core, &err)};
@@ -124,7 +125,7 @@ CoreAV::CoreAVPtr CoreAV::makeCoreAV(Tox* core)
 
     assert(toxav != nullptr);
 
-    return CoreAVPtr{new CoreAV{std::move(toxav)}};
+    return CoreAVPtr{new CoreAV{std::move(toxav), toxCoreLock}};
 }
 
 /**
@@ -235,9 +236,7 @@ bool CoreAV::isCallVideoEnabled(const Friend* f) const
 bool CoreAV::answerCall(uint32_t friendNum, bool video)
 {
     QMutexLocker locker{&callsLock};
-
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != coreavThread.get());
+    QMutexLocker coreLocker{&coreLock};
 
     qDebug() << QString("Answering call %1").arg(friendNum);
     auto it = calls.find(friendNum);
@@ -260,9 +259,7 @@ bool CoreAV::answerCall(uint32_t friendNum, bool video)
 bool CoreAV::startCall(uint32_t friendNum, bool video)
 {
     QMutexLocker locker{&callsLock};
-
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != coreavThread.get());
+    QMutexLocker coreLocker{&coreLock};
 
     qDebug() << QString("Starting call with %1").arg(friendNum);
     auto it = calls.find(friendNum);
@@ -280,17 +277,14 @@ bool CoreAV::startCall(uint32_t friendNum, bool video)
     assert(audio != nullptr);
     ToxFriendCallPtr call = ToxFriendCallPtr(new ToxFriendCall(friendNum, video, *this, *audio));
     assert(call != nullptr);
-    auto ret = calls.emplace(friendNum, std::move(call));
-    ret.first->second->startTimeout(friendNum);
+    calls.emplace(friendNum, std::move(call));
     return true;
 }
 
 bool CoreAV::cancelCall(uint32_t friendNum)
 {
     QMutexLocker locker{&callsLock};
-
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != coreavThread.get());
+    QMutexLocker coreLocker{&coreLock};
 
     qDebug() << QString("Cancelling call with %1").arg(friendNum);
     if (!toxav_call_control(toxav.get(), friendNum, TOXAV_CALL_CONTROL_CANCEL, nullptr)) {
@@ -380,6 +374,7 @@ void CoreAV::sendCallVideo(uint32_t callId, std::shared_ptr<VideoFrame> vframe)
 
     if (call.getNullVideoBitrate()) {
         qDebug() << "Restarting video stream to friend" << callId;
+        QMutexLocker coreLocker{&coreLock};
         toxav_video_set_bit_rate(toxav.get(), callId, VIDEO_DEFAULT_BITRATE, nullptr);
         call.setNullVideoBitrate(false);
     }
@@ -756,7 +751,6 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
     } else {
         // If our state was null, we started the call and were still ringing
         if (!call.getState() && state) {
-            call.stopTimeout();
             call.setActive(true);
             emit self->avStart(friendNum, call.getVideoEnabled());
         } else if ((call.getState() & TOXAV_FRIEND_CALL_STATE_SENDING_V)
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -365,7 +365,7 @@ bool CoreAV::sendCallAudio(uint32_t callId, const int16_t* pcm, size_t samples, 
 
 void CoreAV::sendCallVideo(uint32_t callId, std::shared_ptr<VideoFrame> vframe)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     // We might be running in the FFmpeg thread and holding the CameraSource lock
     // So be careful not to deadlock with anything while toxav locks in toxav_video_send_frame
@@ -420,7 +420,7 @@ void CoreAV::sendCallVideo(uint32_t callId, std::shared_ptr<VideoFrame> vframe)
  */
 void CoreAV::toggleMuteCallInput(const Friend* f)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     auto it = calls.find(f->getId());
     if (f && (it != calls.end())) {
@@ -435,7 +435,7 @@ void CoreAV::toggleMuteCallInput(const Friend* f)
  */
 void CoreAV::toggleMuteCallOutput(const Friend* f)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     auto it = calls.find(f->getId());
     if (f && (it != calls.end())) {
@@ -503,7 +503,7 @@ void CoreAV::groupCallCallback(void* tox, uint32_t group, uint32_t peer, const i
  */
 void CoreAV::invalidateGroupCallPeerSource(int group, ToxPk peerPk)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     auto it = groupCalls.find(group);
     if (it == groupCalls.end()) {
@@ -594,7 +594,7 @@ bool CoreAV::sendGroupCallAudio(int groupId, const int16_t* pcm, size_t samples,
  */
 void CoreAV::muteCallInput(const Group* g, bool mute)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     auto it = groupCalls.find(g->getId());
     if (g && (it != groupCalls.end())) {
@@ -609,7 +609,7 @@ void CoreAV::muteCallInput(const Group* g, bool mute)
  */
 void CoreAV::muteCallOutput(const Group* g, bool mute)
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     auto it = groupCalls.find(g->getId());
     if (g && (it != groupCalls.end())) {
@@ -693,7 +693,7 @@ bool CoreAV::isCallOutputMuted(const Friend* f) const
  */
 void CoreAV::sendNoVideo()
 {
-    QReadLocker locker{&callsLock};
+    QWriteLocker locker{&callsLock};
 
     // We don't change the audio bitrate, but we signal that we're not sending video anymore
     qDebug() << "CoreAV: Signaling end of video sending";
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -731,9 +731,10 @@ void CoreAV::callCallback(ToxAV* toxav, uint32_t friendNum, bool audio, bool vid
         state |= TOXAV_FRIEND_CALL_STATE_SENDING_V | TOXAV_FRIEND_CALL_STATE_ACCEPTING_V;
     it.first->second->setState(static_cast<TOXAV_FRIEND_CALL_STATE>(state));
 
+    // Must explicitely unlock, because a deadlock can happen via ChatForm/Audio
     locker.unlock();
 
-    emit reinterpret_cast<CoreAV*>(self)->avInvite(friendNum, video);
+    emit self->avInvite(friendNum, video);
 }
 
 void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, void* vSelf)
@@ -741,6 +742,7 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
     Q_UNUSED(toxav);
     CoreAV* self = static_cast<CoreAV*>(vSelf);
 
+    // we must unlock this lock before emitting any signals
     QWriteLocker locker{&self->callsLock};
 
     auto it = self->calls.find(friendNum);
@@ -765,14 +767,18 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
         // If our state was null, we started the call and were still ringing
         if (!call.getState() && state) {
             call.setActive(true);
+            bool videoEnabled = call.getVideoEnabled();
+            call.setState(static_cast<TOXAV_FRIEND_CALL_STATE>(state));
             locker.unlock();
-            emit self->avStart(friendNum, call.getVideoEnabled());
+            emit self->avStart(friendNum, videoEnabled);
         } else if ((call.getState() & TOXAV_FRIEND_CALL_STATE_SENDING_V)
                    && !(state & TOXAV_FRIEND_CALL_STATE_SENDING_V)) {
             qDebug() << "Friend" << friendNum << "stopped sending video";
             if (call.getVideoSource()) {
                 call.getVideoSource()->stopSource();
             }
+
+            call.setState(static_cast<TOXAV_FRIEND_CALL_STATE>(state));
         } else if (!(call.getState() & TOXAV_FRIEND_CALL_STATE_SENDING_V)
                    && (state & TOXAV_FRIEND_CALL_STATE_SENDING_V)) {
             // Workaround toxav sometimes firing callbacks for "send last frame" -> "stop sending
@@ -783,9 +789,9 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
             if (call.getVideoSource()) {
                 call.getVideoSource()->restartSource();
             }
-        }
 
-        call.setState(static_cast<TOXAV_FRIEND_CALL_STATE>(state));
+            call.setState(static_cast<TOXAV_FRIEND_CALL_STATE>(state));
+        }
     }
 }
 
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -151,9 +151,16 @@ IAudioControl* CoreAV::getAudio()
 
 CoreAV::~CoreAV()
 {
+    /* Gracefully leave calls and group calls to avoid deadlocks in destructor */
     for (const auto& call : calls) {
         cancelCall(call.first);
     }
+    for (const auto& call : groupCalls) {
+        leaveGroupCall(call.first);
+    }
+
+    assert(calls.empty());
+    assert(groupCalls.empty());
 
     coreavThread->exit(0);
     coreavThread->wait();
@@ -452,12 +459,19 @@ void CoreAV::toggleMuteCallOutput(const Friend* f)
 void CoreAV::groupCallCallback(void* tox, uint32_t group, uint32_t peer, const int16_t* data,
                                unsigned samples, uint8_t channels, uint32_t sample_rate, void* core)
 {
+    /*
+     * Currently group call audio decoding is handled in the Tox thread by c-toxcore,
+     * so we can be sure that this function is always called from the Core thread.
+     * To change this, an API change in c-toxcore is needed and this function probably must be
+     * changed.
+     * See https://github.com/TokTok/c-toxcore/issues/1364 for details.
+     */
+
     Q_UNUSED(tox);
     Core* c = static_cast<Core*>(core);
     CoreAV* cav = c->getAv();
 
     QReadLocker locker{&cav->callsLock};
-    assert(QThread::currentThread() == cav->coreavThread.get());
 
     const ToxPk peerPk = c->getGroupPeerPk(group, peer);
     const Settings& s = Settings::getInstance();
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -555,7 +555,6 @@ bool CoreAV::sendGroupCallAudio(int groupId, const int16_t* pcm, size_t samples,
                                 uint32_t rate) const
 {
     QMutexLocker locker{&callsLock};
-    assert(QThread::currentThread() == coreavThread.get());
 
     std::map<int, ToxGroupCallPtr>::const_iterator it = groupCalls.find(groupId);
     if (it == groupCalls.end()) {
@@ -679,8 +678,6 @@ bool CoreAV::isCallOutputMuted(const Friend* f) const
 void CoreAV::sendNoVideo()
 {
     QMutexLocker locker{&callsLock};
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != coreavThread.get());
 
     // We don't change the audio bitrate, but we signal that we're not sending video anymore
     qDebug() << "CoreAV: Signaling end of video sending";
@@ -696,8 +693,6 @@ void CoreAV::callCallback(ToxAV* toxav, uint32_t friendNum, bool audio, bool vid
     CoreAV* self = static_cast<CoreAV*>(vSelf);
 
     QMutexLocker locker{&self->callsLock};
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != self->coreavThread.get());
 
     // Audio backend must be set before receiving a call
     assert(self->audio != nullptr);
@@ -727,8 +722,6 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
 {
     Q_UNUSED(toxav);
     CoreAV* self = static_cast<CoreAV*>(vSelf);
-    // This callback should come from the Core thread
-    assert(QThread::currentThread() != self->coreavThread.get());
 
     QMutexLocker locker{&self->callsLock};
 
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -181,7 +181,7 @@ void CoreAV::process()
  */
 bool CoreAV::isCallStarted(const Friend* f) const
 {
-    QMutexLocker locker{&callsLock};
+    QReadLocker locker{&callsLock};
     return f && (calls.find(f->getId()) != calls.end());
 }
 
@@ -192,7 +192,7 @@ bool CoreAV::isCallStarted(const Friend* f) const
  */
 bool CoreAV::isCallStarted(const Group* g) const
 {
-    QMutexLocker locker{&callsLock};
+    QReadLocker locker{&callsLock};
     return g && (groupCalls.find(g->getId()) != groupCalls.end());
 }
 
@@ -203,7 +203,7 @@ bool CoreAV::isCallStarted(const Group* g) const
  */
 bool CoreAV::isCallActive(const Friend* f) const
 {
-    QMutexLocker locker{&callsLock};
+    QReadLocker locker{&callsLock};
     auto it = calls.find(f->getId());
     if (it == calls.end()) {
         return false;
@@ -218,7 +218,7 @@ bool CoreAV::isCallActive(const Friend* f) const
  */
 bool CoreAV::isCallActive(const Group* g) const
 {
-    QMutexLocker locker{&callsLock};
+    QReadLocker locker{&callsLock};
     auto it = groupCalls.find(g->getId());
     if (it == groupCalls.end()) {
         return false;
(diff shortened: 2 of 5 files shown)


sudden6:
Be careful when testing, there seems to be at least one threading error which for occasionally resulted in a crash: 

```
[14:43:55.644 UTC] :0 : Warning: QObject::killTimer: Timers cannot be stopped from another thread
[14:43:55.644 UTC] :0 : Warning: QObject::~QObject: Timers cannot be stopped from another thread
```

sudden6:
This should now be mostly working. Not sure if we should add the call timeout with a new implementation or permanently remove it. I'd prefer to remove it, because it makes it possible to let it ring for longer.

sudden6:
@anthonybilinski I'd be happy if you could do some real world testing.

I tested:

- Starting/Stopping video calls on this PR, answering on master
- Starting/Stopping video calls on master, answering on this PR

sudden6:
should fix https://github.com/qTox/qTox/issues/5843

anthonybilinski:
I've tested briefly with 8e56dc222128f0c2372d55f47919bee03667f5b5 merged to master tip (22787a7f3c52672f1d0c9ba7c736dff240091d8a) and created a group with 3 profiles. As soon as the first one clicks join call, I hit an assert:
```
#3  0x00007ffff465d006 in __GI___assert_fail
    (assertion=0x555555e8c0c0 "QThread::currentThread() == coreavThread.get()", file=0x555555e8bf90 "/home/abilinski/builds/qTox/src/core/coreav.cpp", line=558, function=0x555555e8c628 "bool CoreAV::sendGroupCallAudio(int, const int16_t*, size_t, uint8_t, uint32_t) const") at assert.c:101
#4  0x00005555558af46b in CoreAV::sendGroupCallAudio(int, short const*, unsigned long, unsigned char, unsigned int) const
    (this=0x555556047fe0, groupId=0, pcm=0x5555565a4e00, samples=960, chans=2 '\002', rate=48000) at /home/abilinski/builds/qTox/src/core/coreav.cpp:558
#5  0x00005555558cd4e4 in ToxGroupCall::<lambda(const int16_t*, size_t, uint8_t, uint32_t)>::operator()(const int16_t *, size_t, uint8_t, uint32_t) const
    (__closure=0x555556a11180, pcm=0x5555565a4e00, samples=960, chans=2 '\002', rate=48000) at /home/abilinski/builds/qTox/src/core/toxcall.cpp:219
#6  0x00005555558d09fb in QtPrivate::FunctorCall<QtPrivate::IndexesList<0, 1, 2, 3>, QtPrivate::List<short int const*, long unsigned int, unsigned char, unsigned int>, void, ToxGroupCall::ToxGroupCall(int, CoreAV&, IAudioControl&)::<lambda(const int16_t*, size_t, uint8_t, uint32_t)> >::call(ToxGroupCall::<lambda(const int16_t*, size_t, uint8_t, uint32_t)> &, void **) (f=..., arg=0x7fffce7fac90)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:146
#7  0x00005555558d0539 in QtPrivate::Functor<ToxGroupCall::ToxGroupCall(int, CoreAV&, IAudioControl&)::<lambda(const int16_t*, size_t, uint8_t, uint32_t)>, 4>::call<QtPrivate::List<short const*, unsigned long, unsigned char, unsigned int>, void>(ToxGroupCall::<lambda(const int16_t*, size_t, uint8_t, uint32_t)> &, void *, void **) (f=..., arg=0x7fffce7fac90)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:256
--Type <RET> for more, q to quit, c to continue without paging--
#8  0x00005555558cff79 in QtPrivate::QFunctorSlotObject<ToxGroupCall::ToxGroupCall(int, CoreAV&, IAudioControl&)::<lambda(const int16_t*, size_t, uint8_t, uint32_t)>, 4, QtPrivate::List<short int const*, long unsigned int, unsigned char, unsigned int>, void>::impl(int, QtPrivate::QSlotObjectBase *, QObject *, void **, bool *) (which=1, this_=0x555556a11170, r=0x55555752e310, a=0x7fffce7fac90, ret=0x0)
    at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:439
#9  0x00007ffff4cb85c8 in QMetaObject::activate(QObject*, int, int, void**) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#10 0x0000555555858801 in IAudioSource::frameAvailable(short const*, unsigned long, unsigned char, unsigned int) (this=0x55555752e310, _t1=0x5555565a4e00, _t2=960, _t3=2 '\002', _t4=48000)
    at /home/abilinski/builds/qTox/_build/qtox_static_autogen/MYYQ2Y7GGF/moc_iaudiosource.cpp:164
#11 0x0000555555881da3 in OpenAL::doInput() (this=0x555556113640) at /home/abilinski/builds/qTox/src/audio/backend/openal.cpp:682
#12 0x0000555555881e87 in OpenAL::doAudio() (this=0x555556113640) at /home/abilinski/builds/qTox/src/audio/backend/openal.cpp:703
```

sudden6:
@anthonybilinski thanks for testing. This was an over-agressive assert and I removed it. I also found some other assertions that don't quite make sense and removed them.

anthonybilinski:
I'm hitting a couple deadlocks with the latest version. The first time:
1) A calls B, it rings, B answers then A hangs.
2) A calls B, B answers, they talk, B ends the call, A hangs.

They happen in most of the calls I've made.

sudden6:
I should have tested with actual audio running in the background :man_facepalming: 

That was problem number 0, the other two problems are listed in the commit message.

sudden6:
This seems to have fixed at least the problem in 5949, so IMO we should merge it soon.

anthonybilinski:
I ran this again just now on the latest version, 1-1 calls looked good for audio or video, but when I created a group with 1 member, both running the PR, and one of the members joined the call, the other member hits an assert failure:
```
[22:41:09.966 UTC] core/core.cpp:923 : Debug: "Group 0 peerlist changed"                                                                                                                                    
[22:41:10.015 UTC] core/core.cpp:932 : Debug: "Group 0, peer 1, name changed to qTest5"               
qtox: /home/abilinski/builds/qTox/src/core/coreav.cpp:460: static void CoreAV::groupCallCallback(void*, uint32_t, uint32_t, const int16_t*, unsigned int, uint8_t, uint32_t, void*): Assertion `QThread::currentThread() == cav->coreavThread.get()' failed.                                                                                                                                                            
                                                                                                      
Thread 19 "qTox Core" received signal SIGABRT, Aborted.                                                                                                                                                     
```

sudden6:
Can you provide a backtrace of that event?

anthonybilinski:
Sure, reprod a second time:
```
(gdb) bt
#0  0x00007ffff46bf3eb in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff469e899 in __GI_abort () at abort.c:79
#2  0x00007ffff469e769 in __assert_fail_base (fmt=0x7ffff4832fb8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", assertion=0x555555e8a1b8 "QThread::currentThread() == cav->coreavThread.get()", file=0x555555e89c90 "/home/abilinski/builds/qTox/src/core/coreav.cpp", line=460, function=<optimized out>) at assert.c:92
#3  0x00007ffff46b0006 in __GI___assert_fail (assertion=0x555555e8a1b8 "QThread::currentThread() == cav->coreavThread.get()", file=0x555555e89c90 "/home/abilinski/builds/qTox/src/core/coreav.cpp", line=460, function=0x555555e8a138 "static void CoreAV::groupCallCallback(void*, uint32_t, uint32_t, const int16_t*, unsigned int, uint8_t, uint32_t, void*)") at assert.c:101
#4  0x00005555558ac767 in CoreAV::groupCallCallback(void*, unsigned int, unsigned int, short const*, unsigned int, unsigned char, unsigned int, void*) (tox=0x5555565bfe00, group=0, peer=1, data=0x7fffb001f660, samples=960, channels=2 '\002', sample_rate=48000, core=0x5555560c0ff0) at /home/abilinski/builds/qTox/src/core/coreav.cpp:460
#5  0x00007ffff623a9b0 in decode_audio_packet () at /usr/local/lib/libtoxcore.so.2
#6  0x00007ffff623aaed in handle_group_audio_packet () at /usr/local/lib/libtoxcore.so.2
#7  0x00007ffff6232a04 in handle_lossy () at /usr/local/lib/libtoxcore.so.2
#8  0x00007ffff6221d55 in handle_lossy_packet () at /usr/local/lib/libtoxcore.so.2
#9  0x00007ffff6213a9c in handle_data_packet_core () at /usr/local/lib/libtoxcore.so.2
#10 0x00007ffff6213ee1 in handle_packet_connection () at /usr/local/lib/libtoxcore.so.2
#11 0x00007ffff6215717 in udp_handle_packet () at /usr/local/lib/libtoxcore.so.2
#12 0x00007ffff61fb9e4 in networking_poll () at /usr/local/lib/libtoxcore.so.2
#13 0x00007ffff62295c6 in do_messenger () at /usr/local/lib/libtoxcore.so.2
#14 0x00007ffff6235b55 in tox_iterate () at /usr/local/lib/libtoxcore.so.2
#15 0x00005555558b9d93 in Core::process() (this=0x5555560c0ff0) at /home/abilinski/builds/qTox/src/core/core.cpp:740
#16 0x00005555558c34c7 in QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (Core::*)()>::call(void (Core::*)(), Core*, void**) (f=(void (Core::*)(class Core * const)) 0x5555558b9cfe <Core::process()>, o=0x5555560c0ff0, arg=0x7fffcd7f8f30) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:152
#17 0x00005555558c2db6 in QtPrivate::FunctionPointer<void (Core::*)()>::call<QtPrivate::List<>, void>(void (Core::*)(), Core*, void**) (f=(void (Core::*)(class Core * const)) 0x5555558b9cfe <Core::process()>, o=0x5555560c0ff0, arg=0x7fffcd7f8f30) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:185
#18 0x00005555558c23a0 in QtPrivate::QSlotObject<void (Core::*)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) (which=1, this_=0x555556259bb0, r=0x5555560c0ff0, a=0x7fffcd7f8f30, ret=0x0) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qobjectdefs_impl.h:414
#19 0x00007ffff4e585c8 in QMetaObject::activate(QObject*, int, int, void**) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#20 0x00007ffff4e6566b in QTimer::timeout(QTimer::QPrivateSignal) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#21 0x00007ffff4e58e55 in QObject::event(QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#22 0x00007ffff583ea86 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#23 0x00007ffff5847e00 in QApplication::notify(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#24 0x00007ffff4e2ca9a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#25 0x00007ffff4e83a00 in QTimerInfoList::activateTimers() () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#26 0x00007ffff4e84324 in  () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#27 0x00007ffff38ca84d in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#28 0x00007ffff38caad0 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#29 0x00007ffff38cab73 in g_main_context_iteration () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#30 0x00007ffff4e846a5 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#31 0x00007ffff4e2b63b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#32 0x00007ffff4c64a75 in QThread::exec() () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#33 0x00007ffff4c65cc2 in  () at /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#34 0x00007ffff465d669 in start_thread (arg=<optimized out>) at pthread_create.c:479
#35 0x00007ffff479b323 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

sudden6:
Seems this one is actually a c-toxcore problem, according to my understanding it shouldn't decode audio on the main thread. I'll look into fixing c-toxcore.

anthonybilinski:
I ran this PR in a group voice call with 3 members including myself for about an hour and had no issues (no stuttering, crashes, toxcore logs). I'll do a full code review soon.

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.cpp, line 703 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-LzZ4vaB9C8X__pxZywA:-LzZ4vaB9C8X__pxZywB:bwj3xzw) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/coreav.cpp#L703)):*
> ```cpp
>     // caller (toxcore) holds
>     // Also run the code to switch to the CoreAV thread in yet another thread, in case CoreAV
>     // has threadSwitchLock and wants a toxcore lock that our call stack is holding...
> ```

I went through the CoreAv API before, and 7 were always run from the CoreAv thread, but 20 were run from other threads while still accessing CoreAv state. Your mutex change (first commit) definitely fixes that. But what about the comment removed stating that this should be run on the CoreAv since it's slow, and to not block the Core thread? Also adding mutex locks to all the GUI facing API (e.g. toggle mute) could make the GUI temporarily stutter when interacting with CoreAv.

What do you think about instead of mutex locking all public API, switching to the CoreAv thread on all public API like was done in a couple places before (e.g. timeoutCall)? I'd think this would help Core not become blocked, and make the GUI more responsive.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.cpp, line 703 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-LzZ4vaB9C8X__pxZywA:-LzZCt0WDaXDwjHfJFam:b-pqg6ib) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/coreav.cpp#L703)):*
> But what about the comment removed stating that this should be run on the CoreAv since it's slow, and to not block the Core thread? 

I can not see how this function could be slow and I think without doing some profiling, switching the thread could be more expensive than doing the work directly.

As for the GUI becoming unresponsive, IMO the way to go with that is to use signals and slots more.

Also note that in one of the following commits the mutex is replaced with a read-write lock, further reducing blocking case.

sudden6:
From chat:
>  isn't calling an invoke method if not on the right thread equivalent to a signal?

AFAICT signals are much safer, they are typed and checked at compile time, invoke is only resolved at runtime and IMO we should avoid it at all cost.

At the moment I much prefer readable and maintainable code over the last few percent of speed. If it turns out that we have performance problems on low end devices we can easily try to optimize, but I think now would be to early.

anthonybilinski (changes requested):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r3, 5 of 5 files at r4, 1 of 1 files at r5, 2 of 2 files at r6, 2 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 150 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_hYek5icFcs06ffR4:-Lz_hYel1Rmo5wbcgCS7:bk1qezv) ([raw file](https://github.com/qtox/qtox/blob/d8236a97713fda0ada449da27b603ce8f472a377/src/core/coreav.h#L150)):*
> ```objc
>     std::map<int, ToxGroupCallPtr> groupCalls;
> 
>     // protect 'calls' and 'groupCalls' from being modified by ToxAV and Tox threads
> ```

the "ToxAv and Tox threads" part is incomplete, Video, Audio, and GUI all also call in. Just removing the second half of the sentence seems fine.
___
*[src/core/coreav.h, line 115 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaRhvK7s2UKKXuZ9fH:-LzaRhvK7s2UKKXuZ9fI:b4wxz) ([raw file](https://github.com/qtox/qtox/blob/8e56dc222128f0c2372d55f47919bee03667f5b5/src/core/coreav.h#L115)):*
> ```objc
>     };
> 
>     explicit CoreAV(std::unique_ptr<ToxAV, ToxAVDeleter> tox, QMutex &toxCoreLock);
> ```

`explicit` could be removed now that the constructor has two arguments
___
*[src/core/coreav.h, line 152 at r6](https://reviewable.io/reviews/qtox/qtox/5926#-Lza_Srj042J6aT2FVub:-Lza_Srj042J6aT2FVuc:bgge8dv) ([raw file](https://github.com/qtox/qtox/blob/e359514f7ea45b2e7899328d7308367a3406bccf/src/core/coreav.h#L152)):*
> ```objc
> 
>     // protect 'calls' and 'groupCalls' from being modified by ToxAV and Tox threads
>     mutable QReadWriteLock callsLock{QReadWriteLock::Recursive};
> ```

1) are ReadWrite locks the right solution here? It doesn't seem like the typical case of "many concurrent reads and writing occurs infrequently". Also in places where the ReadLocker is used, `calls`'s members are modified, which seems unsafe if other ReadLockers are reading those members at the same time.

2) Is emitting a signal while holding a lock unsafe? The signal would be processed on asynchronously, and the slot would just block on the mutex, while the emitter of the signal continues and eventually releases the mutex, right? I guess if the slot locked its own mutex, then tried to lock the mutex on the signaller, while the signaller tries to lock the mutex on the slot's class?

this would be easier with all signals >.<
___
*[src/core/coreav.cpp, line 170 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_t0U01GbsIP7TAFed:-Lz_t0U01GbsIP7TAFee:b-m9agcu) ([raw file](https://github.com/qtox/qtox/blob/d8236a97713fda0ada449da27b603ce8f472a377/src/core/coreav.cpp#L170)):*
> ```cpp
>     if (QThread::currentThread() != coreavThread.get())
>         return (void)QMetaObject::invokeMethod(this, "start", Qt::BlockingQueuedConnection);
>     iterateTimer->start();
> ```

where is the iterateTimer started from now?
___
*[src/core/coreav.cpp, line 703 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-LzZ4vaB9C8X__pxZywA:-Lz_kV59CuJrs8liv3Ne:bngq5k3) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/coreav.cpp#L703)):*
Fair enough. I agree with it probably being faster with a mutex, just possibly being more blocking on the core/GUI threads. From your comment
> AFAICT signals are much safer, they are typed and checked at compile time, invoke is only resolved at runtime and IMO we should avoid it at all cost.

that's a good point. I would be happy to use more signals/slots and fewer mutexes going forward, but doing the runtime invoke doesn't seem worth it, and fixing the locking is definitely higher priority.
___
*[src/core/coreav.cpp, line 727 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_py1QCq7FA23PjMie:-Lz_py1S3BVbLWvfh-ST:bnjf4x) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/coreav.cpp#L727)):*
> ```cpp
>         /// but since currently the toxav callbacks are fired from the toxcore thread,
>         /// we'll always reach this point through a non-blocking queud connection, so not in the
>         /// callback.
> ```

after this change to mutexes, now instead of being guaranteed to be called from a slot, we're guaranteed to be called directly from the callback. Is this really undefined behaviour like the comment says? I checked tox.h and toxav.h at first glance didn't see comments saying to not call API from a callback, but that comment was probably written for a reason..
___
*[src/core/coreav.cpp, line 239 at r3](https://reviewable.io/reviews/qtox/qtox/5926#-LzaP5An-Dp3p3qq8wpH:-LzaP5An-Dp3p3qq8wpI:b-hzzfl0) ([raw file](https://github.com/qtox/qtox/blob/d096ad6093c7235347bfa924c401b3c9d77d3d49/src/core/coreav.cpp#L239)):*
> ```cpp
>     QMutexLocker locker{&callsLock};
> 
>     // This callback should come from the Core thread
> ```

what's up with this comment (and others)? isn't this not a callback? It's called from e.g. chatform
___
*[src/core/coreav.cpp, line 704 at r3](https://reviewable.io/reviews/qtox/qtox/5926#-LzaPrKlE6A-Mq7Rs5C_:-LzaPrKlE6A-Mq7Rs5Ca:b-le3zm5) ([raw file](https://github.com/qtox/qtox/blob/d096ad6093c7235347bfa924c401b3c9d77d3d49/src/core/coreav.cpp#L704)):*
> ```cpp
> 
>     QMutexLocker locker{&self->callsLock};
>     // This callback should come from the Core thread
> ```

this one I agree with, but everyone above here seems wrong
___
*[src/core/coreav.cpp, line 239 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaT7wm2OyCGV211MfK:-LzaT7wm2OyCGV211MfL:b-5tn6aa) ([raw file](https://github.com/qtox/qtox/blob/8e56dc222128f0c2372d55f47919bee03667f5b5/src/core/coreav.cpp#L239)):*
> ```cpp
> {
>     QMutexLocker locker{&callsLock};
>     QMutexLocker coreLocker{&coreLock};
> ```

ah, comments are removed in the next commit. feel free to leave them if you want.
___
*[src/core/toxcall.cpp, line 192 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaVU89C_-sNMeK-swb:-LzaVU89C_-sNMeK-swc:b-xwkv5e) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/toxcall.cpp#L192)):*
> ```cpp
>     sink = std::move(newSink);
> }
> ```

I don't like removing the timeout timer just because we were starting it from the wrong thread. We could fix that and make it start from the right thread, and it seems like a nice-to-have feature that was working fine before.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 150 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_hYek5icFcs06ffR4:-Lzd2rfmEchmuH3fOp0p:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d8236a97713fda0ada449da27b603ce8f472a377/src/core/coreav.h#L150)):*

the "ToxAv and Tox threads" part is incomplete, Video, Audio, and GUI all also call in. Just removing the second half of the sentence seems fine.
</blockquote></details>

Done.
___
*[src/core/coreav.h, line 115 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaRhvK7s2UKKXuZ9fH:-Lzd2r8hAlAHuo2e1TcA:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8e56dc222128f0c2372d55f47919bee03667f5b5/src/core/coreav.h#L115)):*

`explicit` could be removed now that the constructor has two arguments
</blockquote></details>

Done.
___
*[src/core/coreav.h, line 152 at r6](https://reviewable.io/reviews/qtox/qtox/5926#-Lza_Srj042J6aT2FVub:-Lzd6F7nA54Un4hH9hSK:boqz3af) ([raw file](https://github.com/qtox/qtox/blob/e359514f7ea45b2e7899328d7308367a3406bccf/src/core/coreav.h#L152)):*

1) are ReadWrite locks the right solution here? It doesn't seem like the typical case of "many concurrent reads and writing occurs infrequently". Also in places where the ReadLocker is used, `calls`'s members are modified, which seems unsafe if other ReadLockers are reading those members at the same time.

2) Is emitting a signal while holding a lock unsafe? The signal would be processed on asynchronously, and the slot would just block on the mutex, while the emitter of the signal continues and eventually releases the mutex, right? I guess if the slot locked its own mutex, then tried to lock the mutex on the signaller, while the signaller tries to lock the mutex on the slot's class?

this would be easier with all signals >.<
</blockquote></details>

1. In this case it is more used as a priority lock, because UI calls into CoreAV directly to get status info (which ideally it wouldn't).

You're right, on some places it should be a WriteLocker and I fixed that.


2. I'd say generally it's safe, because the most common cases work:
  1. Asynchronous signal to other thread, the slot can lock the mutex just like any other thread
  2. Synchronous signal to same thread, works here, because I used a recursive mutex
___
*[src/core/coreav.cpp, line 170 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_t0U01GbsIP7TAFed:-Lzd74cp2CZhR8-up5dh:b-lt0jtx) ([raw file](https://github.com/qtox/qtox/blob/d8236a97713fda0ada449da27b603ce8f472a377/src/core/coreav.cpp#L170)):*

where is the iterateTimer started from now?
</blockquote></details>

In the constructor the `QThread::started` signal is connected to the `CoreAV::process()` slot which then repeatedly starts the timer.
___
*[src/core/coreav.cpp, line 727 at r1](https://reviewable.io/reviews/qtox/qtox/5926#-Lz_py1QCq7FA23PjMie:-Lzd9aZkE-9eLHT8tWjU:bsqocm1) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/coreav.cpp#L727)):*

after this change to mutexes, now instead of being guaranteed to be called from a slot, we're guaranteed to be called directly from the callback. Is this really undefined behaviour like the comment says? I checked tox.h and toxav.h at first glance didn't see comments saying to not call API from a callback, but that comment was probably written for a reason..
</blockquote></details>

The first thing this function does is to lock a recursive mutex, so I think it should be safe. I'm also not sure if we can maybe remove that `if` altogether.  This error handling looks like a workaround for not really threadsafe behavior, because we should never receive two invites to the same call.
___
*[src/core/coreav.cpp, line 239 at r3](https://reviewable.io/reviews/qtox/qtox/5926#-LzaP5An-Dp3p3qq8wpH:-Lzd7RYOC0sSECvti3I-:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d096ad6093c7235347bfa924c401b3c9d77d3d49/src/core/coreav.cpp#L239)):*

what's up with this comment (and others)? isn't this not a callback? It's called from e.g. chatform
</blockquote></details>

Done.
___
*[src/core/coreav.cpp, line 704 at r3](https://reviewable.io/reviews/qtox/qtox/5926#-LzaPrKlE6A-Mq7Rs5C_:-Lzd7hEw3anU4jwu88Jf:b-uueyqq) ([raw file](https://github.com/qtox/qtox/blob/d096ad6093c7235347bfa924c401b3c9d77d3d49/src/core/coreav.cpp#L704)):*

this one I agree with, but everyone above here seems wrong
</blockquote></details>

I'm not sure what you mean with that comment? Changed in follow up commit?
___
*[src/core/coreav.cpp, line 239 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaT7wm2OyCGV211MfK:-Lzd7RzGC5CnXvdpFQUP:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/8e56dc222128f0c2372d55f47919bee03667f5b5/src/core/coreav.cpp#L239)):*

ah, comments are removed in the next commit. feel free to leave them if you want.
</blockquote></details>

Done.
___
*[src/core/toxcall.cpp, line 192 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaVU89C_-sNMeK-swb:-LzdA1_k5mZxq2jIjukj:bjlktwk) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/toxcall.cpp#L192)):*

I don't like removing the timeout timer just because we were starting it from the wrong thread. We could fix that and make it start from the right thread, and it seems like a nice-to-have feature that was working fine before.
</blockquote></details>

IIRC we agreed on removing that timeout, no?

anthonybilinski (changes requested):
> Reviewed 1 of 1 files at r2, 2 of 2 files at r8.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 152 at r6](https://reviewable.io/reviews/qtox/qtox/5926#-Lza_Srj042J6aT2FVub:-LzfTfuY0wdmNxSNH9Zt:bwxz0pt) ([raw file](https://github.com/qtox/qtox/blob/e359514f7ea45b2e7899328d7308367a3406bccf/src/core/coreav.h#L152)):*

1. In this case it is more used as a priority lock, because UI calls into CoreAV directly to get status info (which ideally it wouldn't).

You're right, on some places it should be a WriteLocker and I fixed that.


2. I'd say generally it's safe, because the most common cases work:
  1. Asynchronous signal to other thread, the slot can lock the mutex just like any other thread
  2. Synchronous signal to same thread, works here, because I used a recursive mutex
</blockquote></details>

I didn't consider the prioritization of writers, that seems reasonable.

The reason I ask if it's safe is because you added unlocking of the mutex before emitting signals to avoid deadlocks. But I guess if we have two classes A and B each with a class mutex, A locks, emits signal, B's slot is hit, B locks, A calls into B, B calls into A.. and since these functions aren't necessarily run on the event loop, they might not complete before the slots are processed. Man mixing signal/slots and mutexes adds complexity o.o
___
*[src/core/coreav.h, line 152 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-LzfbW3r2CEYATTevgYO:-LzfbW3r2CEYATTevgYP:bfaktnv) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.h#L152)):*
I didn't notice any instances of this, but careful that we don't recursively call from an e.g. write-locking function into a read-locking function, because although that would seem safe:

> Note that the lock type cannot be changed when trying to lock recursively, i.e. it is not possible to lock for reading in a thread that already has locked for writing (and vice versa).
___
*[src/core/coreav.cpp, line 704 at r3](https://reviewable.io/reviews/qtox/qtox/5926#-LzaPrKlE6A-Mq7Rs5C_:-LzdGQNN4nuEkW2mMA8S:b8buj8) ([raw file](https://github.com/qtox/qtox/blob/d096ad6093c7235347bfa924c401b3c9d77d3d49/src/core/coreav.cpp#L704)):*

I'm not sure what you mean with that comment? Changed in follow up commit?
</blockquote></details>

yup, was talking about the comments. sorry, reviewed per-commit, so some of my earlier comments were addressed later.
___
*[src/core/coreav.cpp, line 483 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-Lzf_iaR91bHofftD7VE:-Lzf_iaS40EicjVgABZc:blfv8r7) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.cpp#L483)):*
> ```cpp
>     }
> 
>     emit c->groupPeerAudioPlaying(group, peerPk);
> ```

does this emit run into the issue of locked signals being emitted that other places were trying to avoid?
___
*[src/core/coreav.cpp, line 769 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-LzfWrRkELe3SCLo37aT:-LzfWrRkELe3SCLo37aU:bg3zf3y) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.cpp#L769)):*
> ```cpp
>             call.setActive(true);
>             locker.unlock();
>             emit self->avStart(friendNum, call.getVideoEnabled());
> ```

here `call` is read after the mutex is unlocked
___
*[src/core/toxcall.cpp, line 192 at r4](https://reviewable.io/reviews/qtox/qtox/5926#-LzaVU89C_-sNMeK-swb:-LzdGsZ5264LcfMchsNv:b-tjieqc) ([raw file](https://github.com/qtox/qtox/blob/84f1a740c13113ce7566d5914e58ddd398283621/src/core/toxcall.cpp#L192)):*

IIRC we agreed on removing that timeout, no?
</blockquote></details>

lol forgot about that. yeah true. for history, reasoning was basically call ringing ending after some time is a kind of weird artificial constraint of old physical phone systems, so it's not even clearly a feature. If users react poorly to it being removed, it should be re-added to the UI layer, not the toxcall layer.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 152 at r6](https://reviewable.io/reviews/qtox/qtox/5926#-Lza_Srj042J6aT2FVub:-Lzn4r3a3cVNmc82bi19:b-diy25x) ([raw file](https://github.com/qtox/qtox/blob/e359514f7ea45b2e7899328d7308367a3406bccf/src/core/coreav.h#L152)):*

I didn't consider the prioritization of writers, that seems reasonable.

The reason I ask if it's safe is because you added unlocking of the mutex before emitting signals to avoid deadlocks. But I guess if we have two classes A and B each with a class mutex, A locks, emits signal, B's slot is hit, B locks, A calls into B, B calls into A.. and since these functions aren't necessarily run on the event loop, they might not complete before the slots are processed. Man mixing signal/slots and mutexes adds complexity o.o
</blockquote></details>

Unfortunately we can not escape that complexity for now.
___
*[src/core/coreav.h, line 152 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-LzfbW3r2CEYATTevgYO:-LznA9hjD-d1T0n6UKZt:bzd3e06) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.h#L152)):*

I didn't notice any instances of this, but careful that we don't recursively call from an e.g. write-locking function into a read-locking function, because although that would seem safe:

> Note that the lock type cannot be changed when trying to lock recursively, i.e. it is not possible to lock for reading in a thread that already has locked for writing (and vice versa).
</blockquote></details>

Yes, that's the reason why in some places the locks are manually unlocked.
___
*[src/core/coreav.cpp, line 483 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-Lzf_iaR91bHofftD7VE:-Lzn9y9m8gu0and4EHB7:bmmmu8u) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.cpp#L483)):*

does this emit run into the issue of locked signals being emitted that other places were trying to avoid?
</blockquote></details>

In this special case we're save for now, because the slots connected to this signal don't call back into CoreAV
___
*[src/core/coreav.cpp, line 769 at r8](https://reviewable.io/reviews/qtox/qtox/5926#-LzfWrRkELe3SCLo37aT:-Lzn9c0Q0UxkPzWGAZT2:bedajpu) ([raw file](https://github.com/qtox/qtox/blob/4e7419642ce13eb874ab9678a58159f768956b2d/src/core/coreav.cpp#L769)):*

here `call` is read after the mutex is unlocked
</blockquote></details>

Fixed in the most primitive way. This function is quite complicated and I doubt the toxcore API is that complicated.

anthonybilinski (approved):
:lgtm_strong: 

> Reviewed 1 of 5 files at r9, 5 of 5 files at r10.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5926#-:-M-JlmlPEGwU0ASrXuYG:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5592, opened 2019-03-27
Changing theme doesn't update file transfer widgets


anthonybilinski:
##### Brief Description

OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash: 993f6fa
toxcore: 0.2.9
Qt: 5.9.5
Hardware: Destkop

Reproducible: Always

##### Steps to reproduce

1. Select default theme.
2. Send or receive a file.
3. Change to dark theme.
4. Send or receive a file.

##### Observed Behavior
File transfer widgets keep the theme that they were created with, instead of updating when the theme is changed, resulting in different widgets with different themes in your chatlog:
![image](https://user-images.githubusercontent.com/10469203/55100280-be47f900-507e-11e9-9af2-3e89d43ffb36.png)

##### Expected Behavior
Existing file transfer widgets are updated to the appropriate theme when application theme is changed.

TriKriSta:
Updating theme without restart program does code harder. I think to delete this function after added functions for save groups histories.

anthonybilinski:
Sorry I don't understand, what does theme changes not propagating have to do with group history? Which function are you talking about deleting?

TriKriSta:
I'm talking about the function that updates themes without restart program.
now if the program is restarted that deletes group history, so I added functions for update themes.
when functions for saving groups histories will be added, the function for update themes will no longer be needed

Issue #5588, opened 2019-03-26
Multi window mode edit box's text is black in dark theme


anthonybilinski:
##### Brief Description

OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash:  769e239661a0b218e4f14aa9ce8710ff7697fe7b
toxcore:  v0.2.9
Qt:  5.9.5
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Select dark theme.
2. Right click friend -> Open chat in new window.
3. Type.

##### Observed Behavior
![image](https://user-images.githubusercontent.com/10469203/55026635-a5790e00-4fc0-11e9-9589-918dbee7ec05.png)

##### Expected Behavior
Same font colour (grey) as normal chat edit box.

Pigpog:
This is always a problem on Windows. I have modified the file `themes/dark/chatArea.css`, and managed to fix the issue. This fix is probably not the greatest since the text colour isn't a variable. But this is the change that is needed. 

Here is my modified `themes/dark/chatArea.css`:
```
QTextEdit
{
    background: @groundBase;
    color: #FFFFFF;
}

QGraphicsView
{
    border: none;
}

QWidget
{
    background: @groundBase;
}
```

I simply changed the color value of QTextEdit.

PR #5934, opened 2019-11-21
feat: change way for update themes (fix #5924, #5592)

### LINKED ISSUE #5924: Dark theme leaves chatform unchanged until restart
##### Brief Description

OS: Ubuntu 19.10
qTox version: v1.16.3+
Commit hash:  1a726b54cd2e1d72c6cb6f0f9a733e904f5e5371
toxcore:  0.2.10
Qt:  5.12.4
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Start qTox with default theme
2. Change theme to "dark"

##### Observed Behavior
Theme is half changed to dark, half kept with default theme, causing an ugly mix:
image

##### Expected Behavior
The entire application is updated to the new theme.

##### Additional Info
Restarting qTox with dark theme selected results in the entire application being correctly styled:
image

TriKriSta:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)


sudden6 (approved):
:lgtm_strong:

> Reviewed 33 of 33 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934#-:-LuF6gGZEFOLfB6mxgph:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
@TriKriSta just a question and not a request for this PR, but do you think it's possible to set the application style sheet globally with https://doc.qt.io/qt-5/qapplication.html#styleSheet-prop instead of applying to each UI element individually?

TriKriSta:
@sudden6 No, we can't. If we want to change the style without restarting, we need to add style at least two places (Widget and ActivateDialog) and need to use for it one style file or generate this file from some files.

anthonybilinski (changes requested):
Sorry for taking so long to get around to reviewing this.

I like the propagation of `reloadTheme` to more classes, fixing multiple GUI bugs.

Qt's pain in the ass multiple inheritance of `QObject` makes it so that our classes can't inherit from two things that inherit from `QObject`. e.g. can't inherit from both `QWidget` and a theoretical generic `ThemedObject` class. This makes it so that you need to specialise the connecting class for each Qt class we want to inherit from, e.g. `WidgetStyle`, `DialogStyle`, `FrameStyle` (as you noticed). This is kind of nasty and doesn't scale well as we inherit from more themable classes that inherit from `QObject` in the future.

Is there a reason to have classes inherit from a class which connects them to the `reloadTheme` signal, rather than each class connecting to the `reloadTheme` signal themselves? It would solve this double inheritance issue and remove the need for `n` different `QObjectStyle` classes, and not add more code or be a signifiant change from the current implementation. This would also (minorly) push the usage of the singleton usage of `GUI` up one level, making it slightly easier to pass in (in the future, I guess).

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934#-:-LxzBXFb6Ce63gmhBFZb:b-nb5azh)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/widget/gui.cpp, line 103 at r1](https://reviewable.io/reviews/qtox/qtox/5934#-LxzB5bgE1shQzk1MwU2:-LxzB5bgE1shQzk1MwU3:bp1pvkm) ([raw file](https://github.com/qtox/qtox/blob/0555bdfe330edaeba38290ab90bd816413c195f0/src/widget/gui.cpp#L103)):*
> ```cpp
>         GUI::getInstance().themeReload();
>     } else {
>         QMetaObject::invokeMethod(&getInstance(), "_reloadTheme", Qt::BlockingQueuedConnection);
> ```

this string would need to be changed to "reloadTheme" to match the changed function name.

TriKriSta (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/widget/gui.cpp, line 103 at r1](https://reviewable.io/reviews/qtox/qtox/5934#-LxzB5bgE1shQzk1MwU2:-Ly3r-6z-znDeSGb2tFu:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/0555bdfe330edaeba38290ab90bd816413c195f0/src/widget/gui.cpp#L103)):*

this string would need to be changed to "reloadTheme" to match the changed function name.
</blockquote></details>

Done.

anthonybilinski:
@TriKriSta, what about my question regarding interfaces vs connecting the classes directly? It would remove the need for having a wrapper class for each Qt GUI class.

TriKriSta:
@anthonybilinski I thought that use inheritance it is good way. but now some classes (like ContentLayout) use connecting without inheritance, so probably it will be a good decision to use a single method (connecting the classes directly) for all classes.
I will try to fix it this week.

anthonybilinski (commented):
doing some testing, it looks like #5592 is mostly fixed, the background colour of the file transfer widgets is now updated, but the text colour still stays with the old theme for existing file transfer widgets: 


and for #5924 the view of the chatform after this change looks pretty similar, with large parts of it not being updated on theme change: 


the text is more minor, but because of the chatform and settings: 
 still being in a kind of half-updated state from #5924, it's still kind of required to restart qTox on theme change. Is this fixable?

re: commit formatting:
1) can you squash your review comment changes into the original changes so that things aren't being changed in one way then another
2) could the commit message be more clear on what's being changed, and the body of the commit explain why? currently it's not really clear what what's being changed or why from the message.
3) could the "Fix #" be moved out of the title, e.g.
> fix(ui): implement and connect reloadTheme in leaf classes
>
> This allows leaf classes to update independently when the GUI has changed themes, without
> their parent having to call updateTheme() manually.
>
>Fix #5924 
>Fix #5592 

> Reviewed 7 of 33 files at r1, 1 of 1 files at r2, 23 of 27 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934#-:-M2I8dtn2FVPJWUYKiZq:b-ffe5p3)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/chatlog/chatlog.cpp, line 127 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I4eMd5-4P5JEKSIQZ:-M2I4eMe9mY0Iwq79F2I:b2vwtsv) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/chatlog/chatlog.cpp#L127)):*
> ```cpp
>     connect(&GUI::getInstance(), &GUI::themeReload, this, &ChatLog::reloadTheme);
> 
>     reloadTheme();
> ```

why is setting stylesheets now done on construction when it wasn't before?
___
*[src/widget/contentdialog.cpp, line 441 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I2ECL792hs0fhGVrS:-M2I2ECL792hs0fhGVrT:b-5ftkme) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/widget/contentdialog.cpp#L441)):*
> ```cpp
> {
>     setStyleSheet(Style::getStylesheet("contentDialog/contentDialog.css"));
>     splitter->widget(0)->setStyleSheet(Style::getStylesheet("friendList/friendList.css"));
> ```

It would be clearer to save `QScrollArea* friendScroll` as a class member then `friendScroll->setStyleSheet` directly rather than relying on the splitter.
___
*[src/widget/genericchatitemwidget.h, line 24 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I13PD3ZwQyqVJaxs3:-M2I13PE-Q2ADUCBy6W5:b1yqrts) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/widget/genericchatitemwidget.h#L24)):*
> ```objc
> 
> #include <QLabel>
> #include <QFrame>
> ```

unnecessary reorder of includes

TriKriSta:
@anthonybilinski 

>  Is this fixable?

maybe. But I cannot catch this problem on my system(

anthonybilinski:
I tested on my arch linux machine and the changes work well going from either dark to light or light to dark. I wonder if the issue I was seeing is specific to my environment or test steps on Ubuntu 19.10. I'll look into it on my side.

TriKriSta (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/chatlog/chatlog.cpp, line 127 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I4eMd5-4P5JEKSIQZ:-M2cjoAlDZE0KZVLp8xu:b-9i8us9) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/chatlog/chatlog.cpp#L127)):*

why is setting stylesheets now done on construction when it wasn't before?
</blockquote></details>

because message for update style aren't sent how program start, so I initialize style in constructor
___
*[src/widget/contentdialog.cpp, line 441 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I2ECL792hs0fhGVrS:-M2cjeF96DspTBRbgsy7:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/widget/contentdialog.cpp#L441)):*

It would be clearer to save `QScrollArea* friendScroll` as a class member then `friendScroll->setStyleSheet` directly rather than relying on the splitter.
</blockquote></details>

Done.
___
*[src/widget/genericchatitemwidget.h, line 24 at r3](https://reviewable.io/reviews/qtox/qtox/5934#-M2I13PD3ZwQyqVJaxs3:-M2cjgWLCB4IHj8LnXVr:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/b8242099a6c2b70d8df2b1ace31b8982206a3e86/src/widget/genericchatitemwidget.h#L24)):*

unnecessary reorder of includes
</blockquote></details>

Done.

anthonybilinski (changes requested):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/5934#-M3-UnmBAmxq6U-Idtml:-M3-UnmBAmxq6U-Idtmm:b-4ymfii) (no related file):*
I tried testing some more on Ubuntu 19.10, and I see the same thing as before (i.e. the fix doesn't work). I tried clearing build directory and everything. I tried on exactly the same commit on arch again and like before the fix did work.

Qt on my Ubuntu desktop is 5.12.4, and Qt on my arch chromebook is 5.14.1. As well the desktop's much faster than the arch machine. I wonder if this indicated there's a race condition in the changes or that somehow a minor Qt version caused such a major change.

I don't think this can be merged until we figure out what's going on and why this doesn't work at all in come cases.

---

TriKriSta:
@anthonybilinski I took advantage of the [recommendation](https://github.com/qTox/qTox/issues/5924#issuecomment-609419548). Can you check it?

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/5934#-M3-UnmBAmxq6U-Idtml:-M4QFh0k84rNbH5qoUUz:b-2ppkrx) (no related file):*

I tried testing some more on Ubuntu 19.10, and I see the same thing as before (i.e. the fix doesn't work). I tried clearing build directory and everything. I tried on exactly the same commit on arch again and like before the fix did work.

Qt on my Ubuntu desktop is 5.12.4, and Qt on my arch chromebook is 5.14.1. As well the desktop's much faster than the arch machine. I wonder if this indicated there's a race condition in the changes or that somehow a minor Qt version caused such a major change.

I don't think this can be merged until we figure out what's going on and why this doesn't work at all in come cases.
</blockquote></details>

Tested, and this looks much better! The settings pane now looks correct when switching to or from dark theme.

I still noticed a couple smaller issues in other places:
1) when switching from dark mode to default, the friends list text stays black, instead of going back to white:
image
2) when switching from default to dark mode, the chatlog has a bright white border around it for all chats. This looks like #5924.
3) the "Type your message here..." text is illegible:
image

Issue 1 is resolved by restarting qTox, so is issue 2's white border, but the "Type your message here..." text is not. 

---

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/5934#-M3-UnmBAmxq6U-Idtml:-M4Q_7L680XFVqPmA0_8:b8zpith) (no related file):*

Tested, and this looks much better! The settings pane now looks correct when switching to or from dark theme.

I still noticed a couple smaller issues in other places:
1) when switching from dark mode to default, the friends list text stays black, instead of going back to white:
image
2) when switching from default to dark mode, the chatlog has a bright white border around it for all chats. This looks like #5924.
3) the "Type your message here..." text is illegible:
image

Issue 1 is resolved by restarting qTox, so is issue 2's white border, but the "Type your message here..." text is not. 
</blockquote></details>

Text edit box text colour is covered by #5588, so that can be ignored for now.

---

bodwok:
>1. when switching from dark mode to default, the friends list text stays black, instead of going back to white

Can be fixed using QSS instead of widget palette. I will write a separate PR for this.

>2. when switching from default to dark mode, the chatlog has a bright white border around it for all chats.

Can be fixed by adding the following code to the beginning GenericChatForm::reloadTheme:

`setStyleSheet("");`
`QList<QWidget*> wgts = findChildren<QWidget*>();`
`for (auto x : wgts) {`
`    x->setStyleSheet("");`
`}`

bodwok:
Fixes friends list color [6079](https://github.com/qTox/qTox/pull/6079)

anthonybilinski (commented):
Sorry about the long delay..

I can't repro the original issues I was seeing even without merging to tip and without your latest commit with the suggestion from @bodwork. I'm not on Ubuntu 20.04 rather than 19.10 like in my last test, so presumably this was a Qt version change that made the difference.

Still the workaround the @bodwok mentioned might help other users with a different Qt version, and sticking it in Widget seems reasonable to catch more than just GenericChatForm.

:lgtm_strong: 

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934#-:-MFosxyx5fRfDml0xgw1:b-bak9nu)* status: :shipit: complete! 2 of 1 LGTMs obtained

anthonybilinski (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5934)* status: :shipit: complete! 2 of 1 LGTMs obtained

anthonybilinski:
I've rebased to master because CI was failing due to a couple things that were fixed on master.

PR #5939, opened 2019-11-26
WIP fix(chatlog): stick scroll to bottom if at bottom, else leave unchanged

anthonybilinski:
This reverts behaviour to be similar to the other insertChatlineAtBottom.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5939)

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -396,6 +396,7 @@ void ChatLog::insertChatlineAtBottom(const QList<ChatLine::Ptr>& newLines)
 
     const bool stickToBtm = stickToBottom();
 
+    auto firstVisible = visibleLines.first()->getRow();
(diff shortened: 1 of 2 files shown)


anthonybilinski:
Here if the scroll bar is at bottom, it stays there on new message. If not, new messages are added to the chatlog out of view, with the view not changing at all, and the scroll bar shifting up slightly with each message.

anthonybilinski:
There are some additional changes required here that I found in testing. The logic to remove messages when there are more than 100 in the log needs to take the scroll position into account. Currently if scrolled up a long way, receiving a message will cause the visible lines to be removed form the chatlog...
This class sure is a mess of mixed model and view :(

I'll take a look at hopefully minimal changes to fix that message trimming logic in the scrolled up case.

PR #5980, opened 2020-02-09
fix(toxcall): handle Null sink correctly

sudden6:
This was introduced in 82a4f1b41292eae981b9374824ff1499ca1cfcb4

@anthonybilinski This prevents a crash when the audio source is set to `Disabled`

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5980)

### DIFF
--- a/src/core/toxcall.cpp
+++ b/src/core/toxcall.cpp
@@ -133,14 +133,15 @@ ToxFriendCall::ToxFriendCall(uint32_t FriendNum, bool VideoEnabled, CoreAV& av, 
                              this->av->sendCallAudio(this->friendId, pcm, samples, chans, rate);
                          });
 
-    audioSrcInvalid = QObject::connect(audioSource.get(), &IAudioSource::invalidated,
-                                       [this]() { this->onAudioSourceInvalidated(); });
+    connect(audioSource.get(), &IAudioSource::invalidated, this, &ToxFriendCall::onAudioSourceInvalidated);
 
     if (!audioInConn) {
         qDebug() << "Audio input connection not working";
     }
 
-    audioSinkInvalid = sink->connectTo_invalidated(this, [this]() { this->onAudioSinkInvalidated(); });
+    if (sink) {
+        audioSinkInvalid = sink->connectTo_invalidated(this, [this]() { this->onAudioSinkInvalidated(); });
+    }
 
     // register video
     if (videoEnabled) {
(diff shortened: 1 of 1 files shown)


sudden6:
@sphaerophoria since you also touched this area, can you please review this PR?

anthonybilinski (changes requested):
just a note for future reference, if https://github.com/qTox/qTox/pull/5759 gets finished and `IAudioControl` returns dummy sinks/sources instead of nullptr when no audio input/output is selected, the conditional checking in ToxCall will no longer be needed.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5980#-:-M1IP2fADt_AZItFEsBT:bjuciod)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxcall.cpp, line 135 at r1](https://reviewable.io/reviews/qtox/qtox/5980#-M1IMhrr4-nEA7GsS6O0:-M1IMhrtDvRR6YlPANcV:b-wtetzw) ([raw file](https://github.com/qtox/qtox/blob/8634baeb6c82dcf861ff1c6c8ed5392bc1d2faee/src/core/toxcall.cpp#L135)):*
> ```cpp
>                          });
> 
>     connect(audioSource.get(), &IAudioSource::invalidated, this, &ToxFriendCall::onAudioSourceInvalidated);
> ```

it looks like we were saving `audioSrcInvalid` before just to be able to disconnect the connection on destruction. Just adding `this` to the third argument of `connect` seems like a much better way to do it, and looks like it would work in every other use of it. What do you think about just removing `audioSrcInvalid` as a member?
___
*[src/core/toxcall.cpp, line 179 at r1](https://reviewable.io/reviews/qtox/qtox/5980#-M1INcuz60xgSfWEx_Tv:-M1INcuz60xgSfWEx_Tw:b-980abu) ([raw file](https://github.com/qtox/qtox/blob/8634baeb6c82dcf861ff1c6c8ed5392bc1d2faee/src/core/toxcall.cpp#L179)):*
> ```cpp
>     audioSource = std::move(newSrc);
> 
>     connect(audioSource.get(), &IAudioSource::invalidated, this, &ToxFriendCall::onAudioSourceInvalidated);
> ```

same here, passing `this` into `connect` seems better, I don't see the need for keeping `audioSrcInvalid` as a member at all.

sudden6:
@anthonybilinski what you're asking for is actually implemented in #5981.

I split of this PR to get it merged faster, because it causes a 100% crash if audio sink `None` is selected.

anthonybilinski (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5980)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxcall.cpp, line 135 at r1](https://reviewable.io/reviews/qtox/qtox/5980#-M1IMhrr4-nEA7GsS6O0:-M1Sv0A48isevzLUGil4:b-179lsx) ([raw file](https://github.com/qtox/qtox/blob/8634baeb6c82dcf861ff1c6c8ed5392bc1d2faee/src/core/toxcall.cpp#L135)):*

it looks like we were saving `audioSrcInvalid` before just to be able to disconnect the connection on destruction. Just adding `this` to the third argument of `connect` seems like a much better way to do it, and looks like it would work in every other use of it. What do you think about just removing `audioSrcInvalid` as a member?
</blockquote></details>

Handled in future PR, #5981
___
*[src/core/toxcall.cpp, line 179 at r1](https://reviewable.io/reviews/qtox/qtox/5980#-M1INcuz60xgSfWEx_Tv:-M1Sv5VK2moycUsEob1l:b-179lsx) ([raw file](https://github.com/qtox/qtox/blob/8634baeb6c82dcf861ff1c6c8ed5392bc1d2faee/src/core/toxcall.cpp#L179)):*

same here, passing `this` into `connect` seems better, I don't see the need for keeping `audioSrcInvalid` as a member at all.
</blockquote></details>

Handled in future PR, #5981

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5980#-:-M1Sv9AbACYRuIZOTV_K:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5982, opened 2020-02-12
feat(groups): Allow being in group call if only member

sphaerophoria:
In other applications chatrooms allow you to idle in a call and have
people hop in and out as desired. If a user is the only one presently
online in a group but knows someone will be joining shortly they should
be able to join the call ahead of time.

Tested in a group call with another qtox instance.
* Muting before anyone else was online worked as expected
* Deafening before anyone else was online worked as expected
* Joining and leaving calls before anyone was online worked as expected

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5982)

### DIFF
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -535,17 +535,17 @@ VideoSource* CoreAV::getVideoSourceFromCall(int friendNum) const
  * @note Call from the GUI thread.
  * @param groupId Id of group to join
  */
-void CoreAV::joinGroupCall(int groupId)
+void CoreAV::joinGroupCall(const Group& group)
 {
     QWriteLocker locker{&callsLock};
 
-    qDebug() << QString("Joining group call %1").arg(groupId);
+    qDebug() << QString("Joining group call %1").arg(group.getId());
 
     // Audio backend must be set before starting a call
     assert(audio != nullptr);
-    ToxGroupCallPtr groupcall = ToxGroupCallPtr(new ToxGroupCall{groupId, *this, *audio});
+    ToxGroupCallPtr groupcall = ToxGroupCallPtr(new ToxGroupCall{group, *this, *audio});
     assert(groupcall != nullptr);
-    auto ret = groupCalls.emplace(groupId, std::move(groupcall));
+    auto ret = groupCalls.emplace(group.getId(), std::move(groupcall));
     if (ret.second == false) {
         qWarning() << "This group call already exists, not joining!";
         return;
--- a/src/core/coreav.h
+++ b/src/core/coreav.h
@@ -68,7 +68,7 @@ public:
     VideoSource* getVideoSourceFromCall(int callNumber) const;
     void sendNoVideo();
 
-    void joinGroupCall(int groupNum);
+    void joinGroupCall(const Group& group);
     void leaveGroupCall(int groupNum);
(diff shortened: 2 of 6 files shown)


sudden6:
> Deafening before anyone else comes online prevents audio from being sent until undeafened. This might be unexpected behavior.

Do you mean that if I set the speakers to silent (deafening?) before someone joins the call, no audio (from mic) is being sent?

sphaerophoria:
yeah, it isnt great and i was planning on looking into it, just noticed
after i filed the PR :(

sudden6:
Ok, then I'll be waiting for a fix for that problem.

sphaerophoria:
Sorry for the bad initial PR. This update is a little less clean than I would have liked... the joinGroupCall and leaveGroupCall are now asymmetric APIs. This is because Core needs to be able to leave the call based on groupId not Group*, but the call needs Group* to determine the number of peers in the chat.

We can't use the ToxGroupCall "peers" member variable because it only flags peers that have sent audio. Using the internal peers varialbe would result in us not sending audio if all of our peers are muted.

sudden6 (changes requested):
I think passing `Group` is better than dealing with the group numbers in the API and IMO all the API of Core and CoreAV should be moved in that direction, so for me that's perfectly fine.

> Reviewed 2 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5982#-:-M0CZ69o2comFp5k-2sJ:b-klok18)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.cpp, line 538 at r1](https://reviewable.io/reviews/qtox/qtox/5982#-M0CYTqn6AFfGjdafDz2:-M0CYTqn6AFfGjdafDz3:b-mnz3f2) ([raw file](https://github.com/qtox/qtox/blob/be563f0d180bd36db9e9919a07fe86e3616887a9/src/core/coreav.cpp#L538)):*
> ```cpp
>  * @param groupId Id of group to join
>  */
> void CoreAV::joinGroupCall(Group* group)
> ```

IMO this should be `const Group& `, because a group call with a `nullptr` group doesn't really make sense. At minimum `const` should be added, since AFAICT no non-const operations are performed on `group`

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5982)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.cpp, line 538 at r1](https://reviewable.io/reviews/qtox/qtox/5982#-M0CYTqn6AFfGjdafDz2:-M0na5GYCCBfd7CxmfC_:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/be563f0d180bd36db9e9919a07fe86e3616887a9/src/core/coreav.cpp#L538)):*

IMO this should be `const Group& `, because a group call with a `nullptr` group doesn't really make sense. At minimum `const` should be added, since AFAICT no non-const operations are performed on `group`
</blockquote></details>

Done.

sudden6 (approved):
:lgtm_strong: will test and then merge

> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5982#-:-M0nj1wA-LegYCLNR7tG:bqfau6p)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #5987, opened 2020-02-24
Improve CI Release Publisher python selection

nurupo:
For some reason it's very hard to select the right python3 version on Travis, and if that wasn't enough they also keep changing - older versions removed and new added. This should make it a bit more future proof against Travis-CI changes. @anthonybilinski has already faced this issue before in https://github.com/qTox/qTox/commit/a44cce65beb60c5f280b651e0c084fa9c2bdb0dc

Also bumped CIRP to 0.2.0. There are very minimal changes between 0.2.0a3 and 0.2.0, so it should work the same.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5987)

### DIFF
--- a/.travis/cirp/install.sh
+++ b/.travis/cirp/install.sh
@@ -28,7 +28,7 @@ else
   python3 --version || true
   pyenv versions || true
 
-  pyenv global 3.7
+  pyenv global $(pyenv versions | grep -o ' 3\.[5-99]\.[1-99]' | tail -n1)
 fi
 
 pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5987#-:-M1Cn6qI5bz3D-0MrGFR:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5882, opened 2019-10-14
Chatlog moves and flashes when a message is received and you are scrolled up


anthonybilinski:
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
Commit hash:  93c9eef51f8143c3e3e5c1e63e0cc64d1d533f6e
toxcore:  0.2.10
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Always

Related to https://github.com/qTox/qTox/issues/5824.

##### Steps to reproduce

1. Scroll up to read history.
2. Receive a message in the same chat.

##### Observed Behavior
1.The chatlog flashes, showing the bottom of the chatlog for ~1 frame, before going back to where you were previously looking. This makes e.g. inline pictures and file transfers obvious to see.
2. The end position shifts by the size of the received chat.
3. Sometimes the chatlog is snapped to the bottom.
4. The scroll position sometimes jumps to a new location.

##### Expected Behavior
1. The viewed chat doesn't move or flash, the scroll bar doesn't change, the loaded chat doesn't change.

anthonybilinski:
Fixed by https://github.com/qTox/qTox/pull/5991, fix present in `v1.17-dev` and `master`.

PR #5989, opened 2020-03-01
fix(chatlog): fix stick to bottom behavior

sudden6:
This commit fixes the behavior when a message is received while the
chatlog is scrolled to the bottom. With this change, the chatlog will
stick to the bottom when it is scrolled all the way down. If it is
somewhere in the middle (e.g. for search) the chatlog will not change
its position.

Please not that this does not attempt to fix any other problems with our chatlog. The fix might also not be perfect, but it should allow us to release qTox v1.17.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5989)


sudden6:
Should fix https://github.com/qTox/qTox/issues/5922 and #5882

sudden6:
Wrong base branch

Issue #5824, opened 2019-09-12
Searching while receiving messages removes history from chatlog


anthonybilinski:
##### Brief Description

OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash:  bd034e130d1b0630f272d9b36a33b0e16bc44d9e
toxcore:  v0.2.10
Qt:  5.9.5
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Search in history a couple days in chat.
2. Continue to send and receive messages while searching.

##### Observed Behavior
1. New messages are appended onto the messages from a couple days ago.
2. All messages from between the loaded history and the new messages are no longer in the chatlog.
3. Scrolling down does not cause the messages to reload.
4. Closing search does not cause the messages to reload.

##### Expected Behavior
Either newly sent and received messages are not appended to the loaded history, or the loaded history is snapped to the bottom with new message displayed in context. I prefer the first option, with an indicator that there are new messages if the user scrolls down.

##### Additional Info
Workaround: right click -> "Go to current date" loads the current history properly. Though this seems to often cause a segfault due to #5818

anthonybilinski:
We're going to remove this from the [v1.17.0 milestone](https://github.com/qTox/qTox/milestone/23), despite it being a regression. It's fairly uncommon to repro, has a workaround, and doesn't cause any data/communication issues, only UI. It's fix would require a fairly large redesign of our chatlog, which we don't want to hold the other features and fixes from v1.17.0 waiting on.

PR #5991, opened 2020-03-01
fix(chatlog): fix stick to bottom behavior

### LINKED ISSUE #5922: Chat window scrolls up upon new message, requires constant scrolling-down
##### Brief Description

OS: Gentoo GNU/Linux
qTox version: 1.16.3-726-g7349b441
Commit hash:  7349b44154f22f6ef0be7880a97cb210d55ca982
toxcore:  0.2.10
Qt:  5.12.5
Hardware:  Intel Core 2 Duo, integrated graphics
Reproducible: Always
##### Steps to reproduce

1. Scroll chat log all the way to the bottom, either through Page Down keys, scroll bar, or middle-mouse.
2. Receive a message or a file, and the chat log will float up, cutting off the typing notification, then bottom of message/file transfer message, etc., going further up each new message.

##### Observed Behavior

Upon receiving a message the chat window will scroll up, instead of staying at the bottom and showing new messages and info without cutting off

##### Expected Behavior

Chat window should stay at the bottom to show new messages without cutting off

sudden6:
This commit fixes the behavior when a message is received while the
chatlog is scrolled to the bottom. With this change, the chatlog will
stick to the bottom when it is scrolled all the way down. If it is
somewhere in the middle (e.g. for search) the chatlog will not change
its position.

Please note that this is only a quick fix for the stick to bottom behaviour and does not fix any other issues with dynamically loading the chatlog.

Fixes #5922 and #5882

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/5991)

### DIFF
--- a/src/chatlog/chatlog.cpp
+++ b/src/chatlog/chatlog.cpp
@@ -407,10 +407,9 @@ void ChatLog::insertChatlineAtBottom(const QList<ChatLine::Ptr>& newLines)
 
     layout(lines.last()->getRow(), lines.size(), useableWidth());
 
-    if (visibleLines.size() > 1) {
-        startResizeWorker(visibleLines[1]);
-    } else {
-        startResizeWorker();
+    // redo layout only when scrolled down
+    if(stickToBottom()) {
+        startResizeWorker(true);
     }
 }
 
@@ -463,9 +462,9 @@ void ChatLog::insertChatlinesOnTop(const QList<ChatLine::Ptr>& newLines)
 
     // redo layout
     if (visibleLines.size() > 1) {
-        startResizeWorker(visibleLines[1]);
+        startResizeWorker(stickToBottom(), visibleLines[1]);
     } else {
-        startResizeWorker();
+        startResizeWorker(stickToBottom());
     }
 
 }
@@ -481,7 +480,7 @@ void ChatLog::scrollToBottom()
     verticalScrollBar()->setValue(verticalScrollBar()->maximum());
 }
 
-void ChatLog::startResizeWorker(ChatLine::Ptr anchorLine)
+void ChatLog::startResizeWorker(bool stick, ChatLine::Ptr anchorLine)
 {
     if (lines.empty()) {
         isScroll = true;
@@ -491,11 +490,11 @@ void ChatLog::startResizeWorker(ChatLine::Ptr anchorLine)
     // (re)start the worker
     if (!workerTimer->isActive()) {
         // these values must not be reevaluated while the worker is running
-        if (anchorLine) {
-            workerAnchorLine = anchorLine;
-            workerStb = false;
+        workerStb = stick;
+        if (stick) {
+            workerAnchorLine = ChatLine::Ptr();
         } else {
-            workerStb = stickToBottom();
+            workerAnchorLine = anchorLine;
         }
     }
 
@@ -765,7 +764,7 @@ int ChatLog::getNumRemove() const
 
 void ChatLog::forceRelayout()
 {
-    startResizeWorker();
+    startResizeWorker(stickToBottom());
 }
 
 void ChatLog::checkVisibility(bool causedWheelEvent)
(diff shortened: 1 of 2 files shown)


anthonybilinski:
If A scrolls up past the max message limit, causing messages from the bottom of their chatlog to be evicted, then B sends a message to A, A's view is pulled downwards, and the message they were scrolled up to is no longer visible.

It seems like A is rendering [new message, max message limit] instead of [old message, oldest message] like they were before.

anthonybilinski:
On my next test, I'm seeing the issue where when A is scrolled up and B sends a new message, A then has a view of {new message, [old message, oldest message]} creating a gap in their message log that isn't resolved by scrolling down.

rolfverh:
what iis  thesence of talking??????

sudden6:
> If A scrolls up past the max message limit, causing messages from the bottom of their chatlog to be evicted, then B sends a message to A, A's view is pulled downwards, and the message they were scrolled up to is no longer visible.

I have observed the same behaviour, it's because the new message gets added to the scene, which forces a re-layout. The chatlog scrolls then a bit, because the scroll position is only remembered by an anchor message and is not perfectly precise. For adding messages at the bottom, the anchor is the second visible message, which is then scrolled such that it is the top most visible message.

> On my next test, I'm seeing the issue where when A is scrolled up and B sends a new message, A then has a view of {new message, [old message, oldest message]} creating a gap in their message log that isn't resolved by scrolling down.

I also ignored this for now. The problem here (and generally when we are not scrolled to the bottom) is that if we don't add the message to the active view, there's no indication of a new message (scrollbar changing), but if we add it the chatlog becomes corrupted. I think for now the chatlog becoming corrupted is preferable over not showing new messages.

The chatlog can be fixed, by "Clear chatlog" and then reloading history btw.

anthonybilinski:
Ah sorry my original review wasn't looking at the scope of these fixes, specifically
> Please note that this is only a quick fix for the stick to bottom behaviour and does not fix any other issues with dynamically loading the chatlog.

I can confirm #5922 and #5882 are fixed, yay :D. That just leaves #5824 I believe from our existing chatlog scroll issues.

re: this:
> I have observed the same behaviour, it's because the new message gets added to the scene, which forces a re-layout. The chatlog scrolls then a bit, because the scroll position is only remembered by an anchor message and is not perfectly precise. For adding messages at the bottom, the anchor is the second visible message, which is then scrolled such that it is the top most visible message.

What I'm seeing is A's view changes by multiple full pages, so far down that the messages I was viewing are evicted from the top of the chatlog, and when I scroll up to see them I need to load more messages. I think this is more than just due to the anchor rounding issue, and is a new issue that wasn't previously present? e.g. if A scrolls to the beginning of their chatlog, causing multiple loadings and evictions of messages, then receives a message from B, they're jumped all the way to the ~100th latest message, and A can scroll down to the latest message without ever triggering a load.

sudden6:
> What I'm seeing is A's view changes by multiple full pages, so far down that the messages I was viewing are evicted from the top of the chatlog

Ok, this shoudn't happen indeed, I'll look into it.

sudden6:
> > What I'm seeing is A's view changes by multiple full pages, so far down that the messages I was viewing are evicted from the top of the chatlog

> Ok, this shoudn't happen indeed, I'll look into it.

This issue is fixed in the latest commit, but the chatlog corruption issues are not.

Paricularly, if the chatlog is scrolled up and two messages are received, only the last one will be rendered. To fix this one can "Clear displayed messages" and reload the history.

anthonybilinski:
Ok sounds good, the remaining chatlog corruption can fall under https://github.com/qTox/qTox/issues/5824. Getting two of the chatlog issues fixed helps us narrow down the remaining issues and improves the usability for the time being. I'll review your change and test some more when I get a chance.

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/5991#-:-M2IGQAI8wM9pOybhRVw:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
How should we proceed with merging this to master?

anthonybilinski:
`v1.17-dev` is manually merged into master periodically, which will pull any `v1-17-dev` changes into `master`.

Alternatively if it's originally merged to `master` and needed on `v1.17-dev`, it has to be cherry-picked to `v1.17-dev`, and its diff won't be present in the merge.

I'll make another merge to `master` now since this would be a great fix to get into the nightly.

PR #6003, opened 2020-03-16
fix(flatpak): workaround unstable flathub downloads

sudden6:
works around https://github.com/flathub/flathub/issues/845

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6003)

### DIFF
--- a/flatpak/build.sh
+++ b/flatpak/build.sh
@@ -46,6 +46,15 @@ cd "$FP_BUILD_DIR"
 # Add 'https://flathub.org' remote:
 flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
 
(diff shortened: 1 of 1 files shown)


anthonybilinski (approved):
We could cache flatpak deps like mentioned in https://github.com/flathub/flathub/issues/845#issuecomment-599137286 too, but caching just adds one more point of failure and we'll still hit 503s on cache misses. I'm fine leaving it out.

:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6003#-:-M2ZHAtuB5nXRWEYri7y:b-k63v59)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
Caching also has the problem that we need to exit the docker container and invalidate the cache properly (could be already handled by flatpak), so I left it out of this quick fix.

PR #6013, opened 2020-03-17
chore(style): remove old style casts and cross sign comparisons

### LINKED ISSUE #6010: Enable -Wsign-compare and fix warnings
https://afterthoughtsoftware.com/posts/C-sign-compare

Explicitly cast one way or the other (note that signed overflow is UB) to make sure the code is doing what you, and any reader of the code, expect it to.

### LINKED ISSUE #6012: Enable -Wold-style-cast, fix warnings
Some of these are static_cast, some are const_cast, maybe a few are reinterpret_cast.

anthonybilinski:
Enable warnings for both.

Fix #6010
Fix #6012

@iphydf this included a lot more reinterpret_casts than https://github.com/qTox/qTox/issues/6012 thought would be necessary. Do you think this should be done differently for all the pointer casts?

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6013)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -948,7 +948,11 @@ void Core::onGroupTitleChange(Tox*, uint32_t groupId, uint32_t peerId, const uin
                               size_t length, void* vCore)
 {
     Core* core = static_cast<Core*>(vCore);
-    QString author = core->getGroupPeerName(groupId, peerId);
+    QString author;
+    // from tox.h: "If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference)."
+    if (peerId != std::numeric_limits<uint32_t>::max()) {
+        author = core->getGroupPeerName(groupId, peerId);
+    }
     emit core->saveRequest();
     emit core->groupTitleChanged(groupId, author, ToxString(cTitle, length).getQString());
 }
@@ -1032,7 +1036,7 @@ bool Core::sendMessageWithType(uint32_t friendId, const QString& message, Tox_Me
                                ReceiptNum& receipt)
 {
     int size = message.toUtf8().size();
-    auto maxSize = tox_max_message_length();
+    auto maxSize = static_cast<int>(tox_max_message_length());
     if (size > maxSize) {
         qCritical() << "Core::sendMessageWithType called with message of size:" << size
                     << "when max is:" << maxSize << ". Ignoring.";
@@ -1077,7 +1081,7 @@ void Core::sendGroupMessageWithType(int groupId, const QString& message, Tox_Mes
     QMutexLocker ml{&coreLoopLock};
 
     int size = message.toUtf8().size();
-    auto maxSize = tox_max_message_length();
+    auto maxSize = static_cast<int>(tox_max_message_length());
     if (size > maxSize) {
         qCritical() << "Core::sendMessageWithType called with message of size:" << size
                     << "when max is:" << maxSize << ". Ignoring.";
@@ -1318,7 +1322,7 @@ QByteArray Core::getToxSaveData()
     uint32_t fileSize = tox_get_savedata_size(tox.get());
     QByteArray data;
     data.resize(fileSize);
-    tox_get_savedata(tox.get(), (uint8_t*)data.data());
+    tox_get_savedata(tox.get(), reinterpret_cast<uint8_t*>(data.data()));
     return data;
 }
 
@@ -1455,11 +1459,6 @@ QString Core::getGroupPeerName(int groupId, int peerId) const
 {
     QMutexLocker ml{&coreLoopLock};
 
-    // from tox.h: "If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference)."
-    if (peerId == std::numeric_limits<uint32_t>::max()) {
-        return {};
-    }
-
     Tox_Err_Conference_Peer_Query error;
     size_t length = tox_conference_peer_get_name_size(tox.get(), groupId, peerId, &error);
     if (!PARSE_ERR(error) || !length) {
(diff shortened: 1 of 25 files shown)


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6013)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1451 at r3](https://reviewable.io/reviews/qtox/qtox/6013#-M2cAIu34kIa7YjW3em_:-M2cAIu34kIa7YjW3ema:buz1m39) ([raw file](https://github.com/qtox/qtox/blob/69677831aba0763f5c8fdc1847db3e868e25c983/src/core/core.cpp#L1451)):*
> ```cpp
>     QMutexLocker ml{&coreLoopLock};
> 
>     // from tox.h: "If peer_number == UINT32_MAX, then author is unknown (e.g. initial joining the conference)."
> ```

Removing this check looks like a bug to me, no?

src/core/core.cpp
```diff
@@ -1070,7 +1074,7 @@ void Core::sendGroupMessageWithType(int groupId, const QString& message, Tox_Mes
     QMutexLocker ml{&coreLoopLock};
 
     int size = message.toUtf8().size();
```
iphydf, line 1083: Can size() return negative values? It may be better to choose the other direction and use size_t for both.

src/video/cameradevice.cpp
```diff
@@ -294,7 +294,7 @@ QVector<QPair<QString, QString>> CameraDevice::getRawDeviceListGeneric()
             return devices;
         }
         if (s->iformat->priv_class) {
-            *(const AVClass**)s->priv_data = s->iformat->priv_class;
+            *reinterpret_cast<const AVClass**>(s->priv_data) = s->iformat->priv_class;
```
iphydf: https://www.ffmpeg.org/doxygen/3.3/structAVFormatContext.html#ac4c0777e54085af2f3f1b27130e2b21b says this is `void*`, so this could be `static_cast`. Are you sure all the `reinterpret_cast`s are necessary?

src/widget/form/genericchatform.cpp
```diff
@@ -728,7 +728,7 @@ bool GenericChatForm::loadHistoryFrom(const QDateTime &time)
 
 void GenericChatForm::removeFirstsMessages(const int num)
 {
-    if (messages.size() > num) {
+    if (static_cast<int>(messages.size()) > num) {
```
iphydf, line 731: Note that signed overflow has undefined behaviour. Can you mention in the PR description why you chose this direction?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6013)* status: 0 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 1076 at r3](https://reviewable.io/reviews/qtox/qtox/6013#-M2b1wAJ-FIz0w3XnFCG-r3-1076:-M2hWGz32R4FO_ruXto_:bwvtb1) ([raw file](https://github.com/qtox/qtox/blob/a6fc9b0e93ee45853667c2259b9040473900d160/src/core/core.cpp#L1076)):*
<details><summary><i>Previously, iphydf wrote…</i></summary><blockquote>

Can size() return negative values? It may be better to choose the other direction and use size_t for both.
</blockquote></details>

No it can't return negative. Qt favours signed over unsigned even in cases where negative isn't expected, and I think that follows with isocpp core guidelines, which is part of qTox's guidelines. Even though both sides are positive here, I don't think using unsigned gives a real benefit since there's no bit twiddling or modulo arithmetic needing defined overflow behaviour.
___
*[src/core/core.cpp, line 1451 at r3](https://reviewable.io/reviews/qtox/qtox/6013#-M2cAIu34kIa7YjW3em_:-M2hRjH1-FAbWYUSK8A_:bwx2lit) ([raw file](https://github.com/qtox/qtox/blob/69677831aba0763f5c8fdc1847db3e868e25c983/src/core/core.cpp#L1451)):*

Removing this check looks like a bug to me, no?
</blockquote></details>

This check was moved to the top of this file, as a precondition for this function.
___
*[src/video/cameradevice.cpp, line 297 at r3](https://reviewable.io/reviews/qtox/qtox/6013#-M2b1wAJ-FIz0w3XnFCO-r3-297:-M2h_uFP2DfrpoQfDQ0T:b-jxuhid) ([raw file](https://github.com/qtox/qtox/blob/a6fc9b0e93ee45853667c2259b9040473900d160/src/video/cameradevice.cpp#L297)):*
<details><summary><i>Previously, iphydf wrote…</i></summary><blockquote>

https://www.ffmpeg.org/doxygen/3.3/structAVFormatContext.html#ac4c0777e54085af2f3f1b27130e2b21b says this is `void*`, so this could be `static_cast`. Are you sure all the `reinterpret_cast`s are necessary?
</blockquote></details>

:+1: fixed.
___
*[src/widget/form/genericchatform.cpp, line 731 at r3](https://reviewable.io/reviews/qtox/qtox/6013#-M2b1wAJ-FIz0w3XnFCP-r3-731:-M2hVQm78L80QjhGFrbn:b89yws9) ([raw file](https://github.com/qtox/qtox/blob/a6fc9b0e93ee45853667c2259b9040473900d160/src/widget/form/genericchatform.cpp#L731)):*
<details><summary><i>Previously, iphydf wrote…</i></summary><blockquote>

Note that signed overflow has undefined behaviour. Can you mention in the PR description why you chose this direction?
</blockquote></details>

Updated commit message. Rationale being https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-signed. In this case we never expect a negative nor a size exceeding int, but generally I favoured casting to `int` to avoid e.g. `int` function arguments causing strange behaviour if negative. If extra range is needed, which I didn't notice in any cases, I'd prefer using `long` to `unsigned`

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6013#-:-M2m3xbm-O_z1CPM1XjR:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6015, opened 2020-03-17
chore(cleanup): don't use 0 as nullptr

### LINKED ISSUE #6008: Avoid the use of 0 where nullptr should be used.
E.g. test/model/groupmessagedispatcher_test.cpp has `return GroupId(0);`, which should be `return GroupId(nullptr);`. A few more cases like this.

anthonybilinski:
Fix #6008

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6015)

### DIFF
--- a/src/nexus.cpp
+++ b/src/nexus.cpp
@@ -113,7 +113,7 @@ void Nexus::start()
 
 #ifdef Q_OS_MAC
     // TODO: still needed?
-    globalMenuBar = new QMenuBar(0);
+    globalMenuBar = new QMenuBar();
     dockMenu = new QMenu(globalMenuBar);
 
     viewMenu = globalMenuBar->addMenu(QString());
--- a/src/persistence/db/rawdatabase.cpp
+++ b/src/persistence/db/rawdatabase.cpp
@@ -738,7 +738,9 @@ void RawDatabase::process()
                     const QByteArray& blob = query.blobs[curParam + i];
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wold-style-cast"
-                    // SQLITE_STATIC uses old-style cast and comes from system headers, so can't be fixed by us
+#pragma GCC diagnostic ignored "-Wzero-as-null-pointer-constant"
+                    // SQLITE_STATIC uses old-style cast and 0 as null pointer butcomes from system headers, so can't
+                    // be fixed by us
                     auto sqliteDataType = SQLITE_STATIC;
 #pragma GCC diagnostic pop
                     if (sqlite3_bind_blob(stmt, i + 1, blob.data(), blob.size(), sqliteDataType)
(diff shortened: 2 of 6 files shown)


sudden6 (approved):
is there a diagnostic we can enable to check for this? Anyway, :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6015#-:-M2cAoQV5w-Z8RSsEZu_:b-9c9pjk)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
Aha, there is. Thanks for asking, it's added and caught one new case, which I've updated.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6015#-:-M2dh_XqEFWcwhTZfTkp:bqj37t)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
CI build is failing on warnings from generated UI files, as well as warnings on some of our usages of Qt API. I don't see any of these warning when compiling locally with Qt 5.12.4 and gcc 9.2.1. I think CI is seeing these additional warnings due to Qt 5.5.1 that's being used and which we still support causing these warnings. To avoid bumping our minimum supported Qt version, which I don't think we should do, I'll have to disable `Wzero-as-null-pointer-constant`. I guess I'll keep in the local disabling of the check around e.g. C header includes that were causing issues, so that people can still use the diagnostic locally with newer Qt versions.

sudden6:
Disabling only for CI is not ideal but better than nothing. We should bump the Qt version after the next release.

anthonybilinski:
I meant to disable `Wzero-as-null-pointer-constant` generally. I don't think disabling just for CI would be good, since we still support the same Qt as CI is using, users would fail to compile if they had the older Qt version.

sudden6:
What do you think of the follwoing:

- We postpone merging this PR
- We bump our Qt requirements after v1.17.0 is tagged
- We merge this PR with the check enabled

anthonybilinski:
Bumping our minimum required Qt version would break support with distros that are still in LTS, e.g. Ubuntu 16.04. I don't think that's a massive cost for the benefit of having the warning on by default.

sudden6:
Support for 16.04 ends with April 2021, I think it's reasonable if qTox 1.17 is the last release for that platform. We should at least consider it.

anthonybilinski:
Yes I think if in the future no OS we support uses a Qt this old, we should bump the minimum required version. I think adding `Wzero-as-null-pointer-constant` as default at that point would be good. This PR as is still allows running on `Wzero-as-null-pointer-constant` on newer Qt versions, and does all the required work for enabling it in the future when we bump our minimum supported version, so I don't really see the need to delay merging it until we can have the warning on by default.

sudden6:
Ok, am convinced now :)

PR #6065, opened 2020-04-03
[WIP] fix(build): don't build with -Werror by default

Piraty:
There's no reason to turn warnings into errors by default.

Generally, nobody else but you wants this flag to be set. In essence,
nobody building from source will have the same compiler as you do which 
makes it very likely to throw warnings along the build. So it's just an
annoyance for everyone else who then has to remove this line...

Related: https://github.com/qTox/qTox/issues/4707#issue-261965122

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6065)

### DIFF
--- a/.travis/build-osx.sh
+++ b/.travis/build-osx.sh
@@ -34,6 +34,11 @@ install_ccache() {
 
 # Build OSX
 build() {
+	# Add all warning flags we can
+	export CFLAGS="-Wall -Wextra -Wpedantic -pedantic-errors -Wconversion"
+	export CFLAGS+=" -Werror" # make all warnings fatal
+	export CXXFLAGS="$CFLAGS"
+
     bash ./osx/qTox-Mac-Deployer-ULTIMATE.sh -i
     bash ./osx/qTox-Mac-Deployer-ULTIMATE.sh -b
     bash ./osx/qTox-Mac-Deployer-ULTIMATE.sh -d
--- a/.travis/build-ubuntu-16-04.sh
+++ b/.travis/build-ubuntu-16-04.sh
@@ -23,7 +23,7 @@ sudo apt-get update -qq
 
 # install needed Qt, OpenAL, opus, qrencode, GTK tray deps, sqlcipher
 # `--force-yes` since we don't care about GPG failing to work with short IDs
-sudo apt-get install -y --force-yes \
+sudo apt-get install --no-install-recommends -y --force-yes \
     automake \
     autotools-dev \
     build-essential \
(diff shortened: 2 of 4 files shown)


sudden6:
@Piraty You're PR completely removes that option, even on our CI, so I can not merge it.

In general I think it's a good idea to build with it enabled by default, because even if there's a different compiler, it's usually a valid warning. @anthonybilinski what do you think?

anthonybilinski:
I agree with @nurupo's points in https://github.com/qTox/qTox/issues/4707#issue-261965122. I think having it on for CI still is good to make sure we don't ignore them, and even without `Werror` on by default, users will still see the warnings, it just won't break their builds. If their compilers are just giving warnings and not errors, it's presumably safe for them to finish the build and use the product. I agree that "it's usually a valid warning", but they will still see the warning.

I think ideally `Werror` should be:
1) on for CI to avoid warnings from getting ignored by us (maintainers).
2) an option in cmake to be enabled.
3) Off by default in cmake, so that users with different compilers aren't prevented from building due to new warnings. New errors will still correctly stop their build, so I don't think there's a risk of their build being unsafe.
4) I think probably still off by default in debug builds, because users with newer or different compilers may want to build debug.

Users and maintainers will still see new warnings that are added by their compilers even if the CI compilers don't see them due to being older or different, so users can still open issues about the warnings and they can still be fixed like they are now. Maybe they'll be fixed a little slower because users will be less motivated to open issues, but that doesn't seem like a huge problem.

sudden6:
@Piraty Ok, so I think the best solution would be to add cmake option that turns `-Werror` on, but is set to `OFF` by default. The CI and release builds will then set it to `ON` explicitly in their respective build scripts. Can you implement that?

Piraty:
@sudden6 you do realize what -Werror means? the compiler fails on errors, and turning every warning into an error means nobody else but you (and your CI) can build it straight away without [removing that line](https://github.com/void-linux/void-packages/blob/298c7c833e908d4743fe83f4a74adc2d8dbcb7c4/srcpkgs/qtox/patches/Werror.patch).

I can take a look at what https://github.com/toktok/c-toxcore did in that regard (they've had the same discussion), but in the end I'm not familiar with your CI setup so it might be better if you find a nice way on your own instead of iterating over my uninformed attempts to suit your desire ;)

sudden6:
@Piraty I think it's better to implement this switch in cmake, just like take e.g. `-DSPELL_CHECK=OFF` as an example. Then the flags need only be specified once in cmake and all CI just needs `-DSTRICT_FLAGS=ON` (or however you want to call it). It also allows to easily build with stricter warnings locally.

anthonybilinski:
I implemented https://github.com/qTox/qTox/pull/6083 based on the feedback, which obsoletes this PR. Thanks for bringing this back to our attention, @Piraty. It will be out by next release.

Piraty:
thanks for picking it up

PR #6083, opened 2020-04-12
fix(build): disable Werror by default, keep enabled on CI

anthonybilinski:
To allow for easier building with different compilers on user systems. Keep
strict checks on CI to make sure new warnings aren't ignored.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6083)

### DIFF
--- a/.travis/build-ubuntu-16-04.sh
+++ b/.travis/build-ubuntu-16-04.sh
@@ -167,6 +167,7 @@ build_qtox() {
         -DSMILEYS=DISABLED \
         -DENABLE_STATUSNOTIFIER=False \
         -DENABLE_GTK_SYSTRAY=False \
+        -DSTRICT_OPTIONS=ON \
         -DSPELL_CHECK=OFF
 
     bdir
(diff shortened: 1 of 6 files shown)


sudden6 (approved):
:lgtm_strong:

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6083#-:-M4kziUwA9tZaqvm5jwY:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

Piraty:
you might want to add more flags like `-Wall -Wpedantic -pedantic-errors` and similar

anthonybilinski:
It would be nice eventually, but it would require a lot of code changes at the same time to become compliant. When they do get added, I'm not sure if adding them only under the `STRICT_OPTIONS` option makes sense because then developers would compile locally without warning, push to PR and fail completely due to the combination of new warnings + Werror.

Piraty:
I don't see the problem. A failing PR's CI pipeline will give a hint on what's causing the failure, and since PRs only get merged with successful CI only clean code will enter the repo.

PR #6085, opened 2020-04-13
feat(ipc): make IPC user specific

### LINKED ISSUE #6076: Make IPC user specific to allow instances for each user
From https://github.com/qTox/qTox/issues/2449#issuecomment-541399690

qTox IPC should be per user, so that if there is a user running qTox on a machine, a second user can log in to their own account and run a second profile.

This could be achieved by simply adding $LOGNAME to the IPC location.

Thanks @RalphCorderoy

anthonybilinski:
Fix #6076

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6085)

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -21,9 +21,37 @@
 #include <QCoreApplication>
 #include <QDebug>
 #include <QThread>
+
 #include <ctime>
 #include <random>
 #include <unistd.h>
+#include <stdlib.h>
+
+namespace
+{
+#ifdef Q_OS_WIN
+    const char* getCurUsername()
+    {
+        return getenv("USERNAME");
+    }
+#else
+    const char* getCurUsername()
+    {
+        return getenv("USER");
+    }
+#endif
+
+    QString getIpcKey()
+    {
+        auto* user = getCurUsername();
+        if (!user)
+        {
+            qWarning() << "Failed to get current username. Will use a global IPC.";
+            user = "";
+        }
+        return QString("qtox-" IPC_PROTOCOL_VERSION "-") + user;
+    }
+} // namespace
 
 /**
  * @var time_t IPC::lastEvent
@@ -40,7 +68,7 @@
 
 IPC::IPC(uint32_t profileId)
     : profileId{profileId}
-    , globalMemory{"qtox-" IPC_PROTOCOL_VERSION}
+    , globalMemory{getIpcKey()}
 {
     qRegisterMetaType<IPCEventHandler>("IPCEventHandler");
 


sudden6:
I was thinking if this is the right way to solve this problem.

I think other users should not even be able to see if we are running a qTox instance, so why is our shared memory even accessible to them? IMO we should setup our shared memory that this is not possible, because it leaks information to other users on the same system.

anthonybilinski:
I think this is more of an OS and Qt problem. It's always possible for an unprivileged user to `ps` and view what processes other users are running, and Qt' `QShareMemory`'s `key` is part of its name for the socket under `/tmp/`. I don't necessarily disagree, but I don't think it's really fixable by qTox.

The qTox IPC file under `/tmp` is writeable only by our own user and readable by only our own group, which seems reasonable, so other users shouldn't be able to actually communicate with us over ICP or read any program state. If I try to start qTox by `root` when it's already running, I see:
```
[16:09:25.644 UTC] ipc.cpp:73 : Debug: Attaching to the global shared memory
[16:09:25.644 UTC] main.cpp:308 : Debug: commit:  7b7950e7f7c8263fcd34ae10dcd312b7cf51ffe7
[16:09:25.644 UTC] ipc.cpp:136 : Debug: postEvent  "activate" to 0
[16:09:30.513 UTC] ipc.cpp:269 : Debug: Previous owner timed out, taking ownership 13747702253042416928 -> 3072002724682861673
```
so it is trying to communicate, but failing. Maybe this is worth some more testing on why #6076 is reporting not being able to open a second instance at all.

sudden6:
Maybe we can make that shared file not world readable?

anthonybilinski:
It is already not world readable, only group readable. i.e. `-rw-r-----`

sudden6:
but isn't one of the default groups in Linux `users` where every user is part of?

anthonybilinski:
> but isn't one of the default groups in Linux users where every user is part of

hmm not on my system, but even if there was, it wouldn't be an issue. Files have a `user` owner and a `group` owner. They don't keep track of all groups the user who created or last edited them is in, only the one group that owns them. So if I create a file, it's owned by user `root` and group `root`, and there are no other members in group `root`. Similarly when I use my user account like with qTox's /tmp file, the file's created as part of my user's group which has no other members.

sudden6:
I think I understand now, so the problem is not that multiple users can read/write to our shared memory, but the problem is that there can be only one shared memory instance per system, because they all use the same key. Now your solution makes the keys unique for each user, by using their username.

Now multiple users can create our IPC mechanism, but it's still not possible to read/write to the IPC location of other users. Correct?

anthonybilinski:
Right, exactly. The key clash causes IPC to attempt (but fail due to permissions) to talk to other instances by other users.

There's still the question of why in my test qTox just gave up on IPC and started (as it should), but in the bug report the user couldn't open a second qTox instance at all. I think that would be worth investigation.

anthonybilinski:
Ah mystery solved:

With v1.16.3, if we failed to init IPC, we exited:
```
[01:25:19.916 UTC] persistence/settings.cpp:133 : Debug: "Loading settings from /home/qtest/.config/tox/qtox.ini"
[01:25:19.928 UTC] ipc.cpp:61 : Debug: Our global IPC ID is  15719266212533128476
[01:25:19.928 UTC] ipc.cpp:75 : Debug: Failed to attach to the global shared memory, giving up. Error: 1
[01:25:19.928 UTC] main.cpp:205 : Critical: Can't init IPC
[01:25:19.928 UTC] ipc.cpp:91 : Warning: Failed to lock in ~IPC
```

This was changed in https://github.com/qTox/qTox/pull/5808 to just not init IPC if it errors:
```
[01:25:53.963 UTC] persistence/settings.cpp:121 : Debug: "Loading settings from /home/qtest/.config/tox/qtox.ini"
[01:25:53.970 UTC] ipc.cpp:61 : Debug: Our global IPC ID is  12400671727463374094
[01:25:53.970 UTC] ipc.cpp:75 : Debug: Failed to attach to the global shared memory, giving up. Error: 1
[01:25:53.970 UTC] main.cpp:255 : Warning: Can't init IPC, maybe we're in a jail? Continuing with reduced multi-client functionality.
[01:25:53.970 UTC] main.cpp:308 : Debug: commit:  7b7950e7f7c8263fcd34ae10dcd312b7cf51ffe7
```

But with this PR, IPC can still create a second user-specific instance:
```
[01:31:41.109 UTC] persistence/settings.cpp:121 : Debug: "Loading settings from /home/qtest/.config/tox/qtox.ini"
[01:31:41.117 UTC] ipc.cpp:82 : Debug: Our global IPC ID is  3210560770336686392
[01:31:41.117 UTC] main.cpp:312 : Debug: commit:  d028f9e39427d8606c7d276979aea63257107852
```

and then that second user's second instance can work as normal:
```
[01:32:12.223 UTC] persistence/settings.cpp:121 : Debug: "Loading settings from /home/qtest/.config/tox/qtox.ini"
[01:32:12.230 UTC] ipc.cpp:82 : Debug: Our global IPC ID is  5646071807914664745
[01:32:12.230 UTC] ipc.cpp:94 : Debug: Attaching to the global shared memory
[01:32:12.230 UTC] main.cpp:312 : Debug: commit:  d028f9e39427d8606c7d276979aea63257107852
[01:32:12.230 UTC] ipc.cpp:157 : Debug: postEvent  "activate" to 0
[01:32:13.024 UTC] main.cpp:360 : Debug: Another qTox instance is already running. If you want to start a second instance, please open login screen (qtox -l) or start with a profile (qtox -p <profile name>).
```

anthonybilinski:
PR updated to use OS-specific env variables rather than always trying both, and commit message updated to not indicate that this allows qTox to be run by multiple users, which was already the case since #5808.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6085#-:-M4xTDp440slad5SS4ub:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6086, opened 2020-04-13
chore(ci): don't use git --no-merges for git commit message checking

### LINKED ISSUE #6051: Travis git commit message verification fails sporadically
Once in a while `verify-commit-messages.sh` fails incorrectly. This happened here: https://travis-ci.org/github/qTox/qTox/jobs/668295578?utm_medium=notification&utm_source=github_status

I think it's happening because travis is doing a shallow clone: `git clone --depth=50 https://github.com/qTox/qTox.git qTox/qTox` and `verify-commit-messages.sh` uses `git log --no-merges`, but in the case where the oldest commit is a merge in the cloned repo, I suspect git can't tell that it's a merge because both of its parent commits are missing. This leads to the error:
```
$ ./.travis/$JOB.sh

Merge pull request #5954

Above ↑ commits don't conform to commit message format:
```

Due to the short clone, we should probably do the less-reliable check of ignoring merge commit messages generically, instead of ignoring merge commits.

anthonybilinski:
When doing a short clone, like travis does, git can't tell if the
oldest commit is a merge of not. Just allow either commit message
format for all commits, and rely on reviewers and common sense to not
name normal commits "Merge.*"

Fix #6051

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6086)

### DIFF
--- a/verify-commit-messages.sh
+++ b/verify-commit-messages.sh
@@ -36,11 +36,10 @@ ARG="$1"
 echo "" # ← formatting
 
 grep_for_invalid() {
-    # differentiate what is allowed for commit messages and merge messages
-    git log --no-merges --format=format:'%s' "$ARG" \
-        | grep -v -E '^(feat|fix|docs|style|refactor|perf|revert|test|chore)(\(.{,12}\))?:.{1,68}$' \
-        || git log --merges --format=format:'%s' "$ARG" \
-            | grep -v -E '^Merge .{1,70}$'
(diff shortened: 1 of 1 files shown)


sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6086#-:-M4nVpxxBfZVACmopeF4:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6094, opened 2020-04-16
chore: Various code cleanups.

iphydf:
* Reorder class data members and/or constructor initialisers to match,
  reducing confusion about when members will be initialised.
* Remove (most) unused variables. Not removed: some global variables with
  `TODO(sudden6)` on them for using them in the future. I don't know how
  far into the future sudden6 wants to use them, so I left them there for
  now.
* Distinguish different bootstrap nodes in the logs by index in the
  bootstrap node list. Originally, we used to log the address/port of the
  node we're bootstrapping to. This was removed out of privacy concerns
  (even though the bootstrap nodes are public). This made the logs much
  less useful when debugging why the client isn't connecting. Having
  indices makes it easier to see that different nodes are being selected,
  and makes it possible to determine which node was selected.
* Explicitly cast unused results of Tox API functions to `void` when all
  we want is to know whether the function succeeds or not.
* Don't try to `#include <unistd.h>` on Windows. It does not exist on
  MSVC.
* Remove extra `;` after function definitions.
* Remove reference indirection of QJsonValueRef, since a copy of that ref
  (small pointer-like object) has to be made anyway when iterating over
  QJsonArrays.
* Make some file-scope global state `static`.
* Use `nullptr` instead of `NULL`.
* Add `#if DESKTOP_NOTIFICATIONS` around the code that implements desktop
  notifications, so it becomes a bit easier to compile everything with a
  single compiler command - useful for manually running static analysers.
* Fix an error on MSVC where `disconnect` is looked up to be a non-static
  member function and the `this` capture is missing.
* Consistently use `struct` and `class` tags for types.
* Use references in ranged-for where it reduces copies.
* Move private static data members out of the Style class and into
  file-local scope. There is no need for them to be in the class. Also
  marked them `const` where possible.
* Removed unused lambda capture.
* Ensure qTox can compile under NDEBUG with `-Wunused-variable` by
  inlining the unused variable into the `assert` that was its only
  target.
* Minor reformatting in core_test.cpp.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6094)

### DIFF
--- a/src/chatlog/content/broken.h
+++ b/src/chatlog/content/broken.h
@@ -38,8 +38,8 @@ public:
     qreal getAscent() const override;
 
 private:
-    QSize size;
     QPixmap pmap;
+    QSize size;
 };
 
 #endif // BROKEN_H
--- a/src/chatlog/content/filetransferwidget.cpp
+++ b/src/chatlog/content/filetransferwidget.cpp
@@ -86,8 +86,6 @@ FileTransferWidget::FileTransferWidget(QWidget* parent, ToxFile file)
         update();
     });
 
-    CoreFile* coreFile = Core::getInstance()->getCoreFile();
-
     connect(ui->leftButton, &QPushButton::clicked, this, &FileTransferWidget::onLeftButtonClicked);
     connect(ui->rightButton, &QPushButton::clicked, this, &FileTransferWidget::onRightButtonClicked);
     connect(ui->previewButton, &QPushButton::clicked, this,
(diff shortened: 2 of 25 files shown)


sudden6 (approved):
:lgtm_strong:

> Reviewed 25 of 25 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6094#-:-M5Hw4V539XIZxzn38O5:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6118, opened 2020-04-28
fix(ui): don't notify of available update if local version is newer

### LINKED ISSUE #6112: Update check should not notify if local version is greater than latest release
##### Brief Description

OS: Fedora
qTox version: v1.17.1-181-g00d08d98
toxcore:  0.2.11
Qt:  5.13.12

Reproducible: Always

##### Observed Behavior

I noticed that there is a new feature that tries to show when an update is available. I built the newest github master version earlier today and immediately after launching it, it's showing the green colored settings cog and the about section showing a green "Update available" button even though the build is the newest.

[Screenshot](https://i.imgur.com/yjLyxq4.png)

Clicking the "Update available" button takes you to the github release page which lists version 1.16.3 as the newest release which is from 2018.

##### Expected Behavior
For it to not show "Update available" when you are running the newest release/a build newer than the current release.

Edit: I can see why it's happening, and why it probably won't be an issue for very long. It looks like it's [comparing](https://github.com/qTox/qTox/blob/00d08d98874d18bbee007c4da295515e0cb8a7c8/src/net/updatecheck.cpp#L84) the version of the program against the last release (which would both have been 1.16.3.x up until fairly recently) but since something in the program got changed to 1.17 and the 1.17 release hasn't happened yet or isn't set as the "latest release" on github, they aren't equal any more.

Instead of checking if latestVersion and curVer are equal or not, maybe it would be better to see if latestVersion is a higher number or not.

anthonybilinski:
This could happen between the time when the release tag is pushed and the time
when the release binaries are published.

Fix #6112

Lower version:
old

Same version:
current

Newer version:
new

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6118)

### DIFF
--- a/src/net/updatecheck.cpp
+++ b/src/net/updatecheck.cpp
@@ -31,6 +31,60 @@
 
 namespace {
 const QString versionUrl{QStringLiteral("https://api.github.com/repos/qTox/qTox/releases/latest")};
+
+struct Version {
+    int major;
+    int minor;
+    int patch;
+};
+
+Version tagToVersion(QString tagName)
+{
+    // capture tag name to avoid showing update available on dev builds which include hash as part of describe
+    QRegularExpression versionFormat{QStringLiteral("v([0-9]+)\\.([0-9]+)\\.([0-9]+)")};
+    auto matches = versionFormat.match(tagName);
+    assert(matches.lastCapturedIndex() == 3);
+
+    bool ok;
+    auto major = matches.captured(1).toInt(&ok);
+    assert(ok);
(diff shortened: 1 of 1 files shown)


sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6118#-:-M613I4B7AJcOTAXu27V:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6119, opened 2020-04-29
[RFC] Refactor cmake project structure

sudden6:
This is the first of a series of PRs that will split qTox into it's major submodules. This should force us to better modularize the qTox code base.

@Diadlo The Plan™ is finally happening.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6119)


Diadlo (approved):
Congratulations. Sorry, I didn't live up to this)
:lgtm_strong:

> Reviewed 18 of 18 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119#-:-M68qcM9EoIT1ry5kiKy:b7t11a5)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6119#-M6EVx-R1uGMp-32s9cN:-M6EVx-R1uGMp-32s9cO:b-7acaa6) (no related file):*
Regarding the layout, I think for these new modules to be stand-alone components that could theoretically be reused, they should contain everything they need to run. For that reason, I think `audio_library` should contain everything it needs to run, e.g. the pcms under `audio/`. In that case, `src/audio` makes less sense as a library than maybe `audio/src`, so that both `audio/src` and `audio/res` could be in the `audio/` library, with everything needed to make that library in the same directory. As-is, `audio_library` can only be used if linked into qTox so that it has the pcm resources.

---

sudden6:
@anthonybilinski @sphaerophoria changed the folder structure according to your comments

anthonybilinski (changes requested):
> Reviewed 9 of 18 files at r1, 32 of 32 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6119#-M6JWxYgA0XZ4duP-uJu:-M6JWxYgA0XZ4duP-uJv:b-excbig) (no related file):*
It would be nice if the include paths could be e.g. `#include "audio/audio.h"` rather than `audio/src/audio.h`. I'm not sure how to achieve this though. I see a few options,
1) make `audio/` the include dir like you have, causing every include path to be `audio/src`
2) make source flat with resources, which would achieve `audio/header.h` but create a messier and worse file structure
3) make `audio/src` the include path, like you have, which would allow us to include without all of the redundant `src/` in includes, but it would also make all our headers in the project effectively flat, removing their module names, which seems worse. Or we could continue to include with the `audio/src` despite this.
4) change our directory structure to be `audio/src/audio/{files}`, then set `audio/src` to be our include path. This would achieve the goal of being able to include "audio/audio.h" but it seems goofy to have a a directory structure like `audio/src/audio` when there's nothing else in `audio/src`.

Out of these I prefer either keeping the `src` in include paths or going to option 4's goofy directory layout, but if you can find a way to achieve a `module/header.h` format without messing with directory structure, that would be great :D If not, it doesn't seem any worse than it was previously, and can always be changed later. I won't block the PR on this.

---
___
*[src/CMakeLists.txt, line 17 at r2](https://reviewable.io/reviews/qtox/qtox/6119#-M6JRj5_2F0jayLbrmAr:-M6JRj5aBmm1nhD7AmGx:b-ivxjzh) ([raw file](https://github.com/qtox/qtox/blob/bdda3d8db47a9a2c45df93a2984405e9933b837c/src/CMakeLists.txt#L17)):*
> ```CMake
> #   along with qTox.  If not, see <http://www.gnu.org/licenses/>
> 
> add_subdirectory(util)
> ```

why keep `src/CMakeLists.txt` and keep utils under `src/utils`? For the same reasons as with `audio`, I think It would make more sense for the module to be in its own directory, e.g. `util/CMakeLists.txt` and `util/src` rather than `src/util`

sudden6:
There's another option, to make the path for public include files `/<module>/include/<module>/header.h`

So for `audio` it would be `/audio/include/audio/audio.h` and `#include "audio/audio.h".

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6119#-M6JWxYgA0XZ4duP-uJu:-M6MPL7H4AYwimnRGJ32:bvq8dek) (no related file):*

It would be nice if the include paths could be e.g. `#include "audio/audio.h"` rather than `audio/src/audio.h`. I'm not sure how to achieve this though. I see a few options,
1) make `audio/` the include dir like you have, causing every include path to be `audio/src`
2) make source flat with resources, which would achieve `audio/header.h` but create a messier and worse file structure
3) make `audio/src` the include path, like you have, which would allow us to include without all of the redundant `src/` in includes, but it would also make all our headers in the project effectively flat, removing their module names, which seems worse. Or we could continue to include with the `audio/src` despite this.
4) change our directory structure to be `audio/src/audio/{files}`, then set `audio/src` to be our include path. This would achieve the goal of being able to include "audio/audio.h" but it seems goofy to have a a directory structure like `audio/src/audio` when there's nothing else in `audio/src`.

Out of these I prefer either keeping the `src` in include paths or going to option 4's goofy directory layout, but if you can find a way to achieve a `module/header.h` format without messing with directory structure, that would be great :D If not, it doesn't seem any worse than it was previously, and can always be changed later. I won't block the PR on this.
</blockquote></details>

I think your suggestion has the same redundant directory structure of my option 4, but while being able to define public headers of a module. That seems like a nice benefit, and it would be easy to gradually move to since we could move headers to `include/` by default, and then move them back to the `src/` dir as the underlying code is refactored, and public API of modules are better defined.

I still don't think of any these are _great_ options, but I think your suggestion is probably the best. I'd be happy to either go with that or defer this to another issue, up to you.

---

iphydf:
That would mean having lots of `-I` flags on all the dependent libraries. I'd rather avoid that.

sudden6:
@iphydf one public include directory per dependent library doesn't seem like "lots" to me?

util/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+#   the Free Software Foundation, either version 3 of the License, or
+#   (at your option) any later version.
+#
+#   qTox is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with qTox.  If not, see <http://www.gnu.org/licenses/>
```
iphydf, line 15: I'm glad we're making sure that if anyone writes `add_library(util_library STATIC ${HEADER_LIST})` in their cmake files, they will be creating a derived work of qtox and have all their stuff under GPLv3.

sudden6: This was also in the global cmake file before, I see no reason or justification to drop that?

anthonybilinski (approved):
:lgtm_strong: 

> Reviewed 38 of 38 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119#-:-M71ZwRX2zUp-k9rp9R4:b-uwkowd)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/CMakeLists.txt, line 17 at r2](https://reviewable.io/reviews/qtox/qtox/6119#-M6JRj5_2F0jayLbrmAr:-M71VCnz1nm2iClxAZ1K:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/bdda3d8db47a9a2c45df93a2984405e9933b837c/src/CMakeLists.txt#L17)):*

why keep `src/CMakeLists.txt` and keep utils under `src/utils`? For the same reasons as with `audio`, I think It would make more sense for the module to be in its own directory, e.g. `util/CMakeLists.txt` and `util/src` rather than `src/util`
</blockquote></details>

Done.
___
*[util/CMakeLists.txt, line 15 at r3](https://reviewable.io/reviews/qtox/qtox/6119#-M6lQSuyCNXZf4NoUC3d-r3-15:-M71XWan4guoDlDUavfE:bujv3fy) ([raw file](https://github.com/qtox/qtox/blob/66e9107699e2352b29892b3b4ead42ee6d4bb7c4/util/CMakeLists.txt#L15)):*

This was also in the global cmake file before, I see no reason or justification to drop that?
</blockquote></details>

I'm in favour of keeping it just for consistency.

anthonybilinski (approved):
> Reviewed 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6119)* status: :shipit: complete! 2 of 1 LGTMs obtained

audio/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    OPTIONS ${RCC_OPTIONS}
+)
+
+# We need this directory, and users of our library will need it too
+target_include_directories(audio_library PUBLIC include/)
+
+target_link_libraries(audio_library util_library)
+target_link_libraries(audio_library Qt5::Core)
+target_link_libraries(audio_library ${OPENAL_LIBRARIES})
+
```
iphydf, line 45: Extra empty line at eof can safely be removed.

PR #6122, opened 2020-05-01
fix(ipc): allow IPC under windows by not using std::random_device with mingw

### LINKED ISSUE #2917: Clicking on Qtox Icon on desktop tries to open separate instance
Clicking on the icon on desktop shortcut tries to open up a separate instance and it doesn't do it. loading the same profile fails. Which is good. But we would like it to just quickly open a opened program instead. 

Any way of actually opening the program just as if you clicked it in the notification task bar?

anthonybilinski:
https://en.cppreference.com/w/cpp/numeric/random/random_device
"A notable implementation where std::random_device is deterministic is old
versions of MinGW (bug 338, fixed since GCC 9.2)."

Although time is less random off Windows, we don't need a strongly random
number for this, and using time avoids having to use platform specific code.

Fix #2917

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6122)

### DIFF
--- a/src/ipc.cpp
+++ b/src/ipc.cpp
@@ -22,6 +22,7 @@
 #include <QDebug>
 #include <QThread>
 
+#include <chrono>
 #include <ctime>
 #include <random>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


sudden6 (approved):
:lgtm_strong: Interesting fix

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6122#-:-M6E4degDzV0ARz6Uble:b-m1nbwm)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6126, opened 2020-05-02
Local bootstrap list

### LINKED ISSUE #5767: Have user editable bootstrap nodelist
When the bootstrap nodes list was stored in the .ini file, it was easily visible and editable by users. Now that it's stored in the json file as provided by nodes.tox.chat, it's still in an easy to edit json format, but is packaged inside the binary, making it near impossible for users to manually edit.

It would be good for the bootstrap list to stay as an external editable config file, so that users can easily create custom private DHTs, view the list, and quickly adds nodes in case of outages / blocks without having to wait for a release.

anthonybilinski:

Fix #5767

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6126)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,6 +379,7 @@ set(${PROJECT_NAME}_SOURCES
   src/model/chathistory.h
   src/model/chathistory.cpp
   src/model/toxclientstandards.h
+  src/model/ibootstraplistgenerator.h
   src/net/bootstrapnodeupdater.cpp
   src/net/bootstrapnodeupdater.h
   src/net/avatarbroadcaster.cpp
--- a/doc/user_manual_en.md
+++ b/doc/user_manual_en.md
@@ -12,6 +12,7 @@
 * [Keyboard Shortcuts](#keyboard-shortcuts)
 * [Commandline Options](#commandline-options)
 * [Emoji Packs](#emoji-packs)
+* [Bootstrap nodes](#bootstrap-nodes)
 
 
 ## Profile corner
@@ -427,7 +428,7 @@ The following shortcuts are currently supported:
 
 ## Push to talk
 
-In audio group chat microphone mute state will be changed while `Ctrl` + 
+In audio group chat microphone mute state will be changed while `Ctrl` +
 `p` pressed and reverted on release.
 
 ## Commandline Options
@@ -457,6 +458,13 @@ for Linux. If these directories don't exist, you have to create them. The emoji
 files have to be in a subfolder also containing `emoticon.xml`, see the
 structure of https://github.com/qTox/qTox/tree/v1.5.2/smileys for further
 information.
+
+## Bootstrap Nodes
+
+qTox uses bootstrap nodes to find its way in to the DHT. The list of nodes is
+stored in `bootstrapNodes.json` and can be found and modified if wanted at
+`~/.config/tox/` on Linux, `%APPDATA%\Roaming\tox` on Windows, and
+`~/Library/Application Support/Tox` on macOS.
 
 
 [ToxMe service]: #register-on-toxme
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -803,22 +803,29 @@ void Core::bootstrapDht()
     // i think the more we bootstrap, the more we jitter because the more we overwrite nodes
     while (i < 2) {
         const DhtServer& dhtServer = bootstrapNodesList[j % listSize];
-        QString dhtServerAddress = dhtServer.address.toLatin1();
         QString port = QString::number(dhtServer.port);
-        QString name = dhtServer.name;
         qDebug("Connecting to bootstrap node %d", j % listSize);
-        QByteArray address = dhtServer.address.toLatin1();
+
+        QByteArray address;
+        if (dhtServer.ipv4.isEmpty() && !dhtServer.ipv6.isEmpty()) {
+            address = dhtServer.ipv6.toLatin1();
+        } else {
+            address = dhtServer.ipv4.toLatin1();
+        }
+
         // TODO: constucting the pk via ToxId is a workaround
         ToxPk pk = ToxId{dhtServer.userId}.getPublicKey();
-
         const uint8_t* pkPtr = pk.getData();
 
         Tox_Err_Bootstrap error;
-        tox_bootstrap(tox.get(), address.constData(), dhtServer.port, pkPtr, &error);
-        PARSE_ERR(error);
-
-        tox_add_tcp_relay(tox.get(), address.constData(), dhtServer.port, pkPtr, &error);
-        PARSE_ERR(error);
+        if (dhtServer.statusUdp) {
+            tox_bootstrap(tox.get(), address.constData(), dhtServer.port, pkPtr, &error);
+            PARSE_ERR(error);
+        }
+        if (dhtServer.statusTcp) {
+            tox_add_tcp_relay(tox.get(), address.constData(), dhtServer.port, pkPtr, &error);
+            PARSE_ERR(error);
+        }
 
         ++j;
         ++i;
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -28,7 +28,7 @@
 #include "src/core/toxstring.h"
 #include "src/model/groupinvite.h"
 #include "src/model/status.h"
-#include "src/net/bootstrapnodeupdater.h"
+#include "src/model/ibootstraplistgenerator.h"
 #include "src/nexus.h"
 #include "src/persistence/profile.h"
 #include "src/util/strongtype.h"
@@ -473,10 +473,11 @@ bool parseErr(Tox_Err_Conference_Delete error, int line)
 
 } // namespace
 
-Core::Core(QThread* coreThread)
+Core::Core(QThread* coreThread, IBootstrapListGenerator& _bootstrapNodes)
     : tox(nullptr)
     , toxTimer{new QTimer{this}}
     , coreThread(coreThread)
+    , bootstrapNodes(_bootstrapNodes)
 {
     assert(toxTimer);
     toxTimer->setSingleShot(true);
@@ -525,7 +526,7 @@ void Core::registerCallbacks(Tox* tox)
  * @return nullptr or a Core object ready to start
  */
 ToxCorePtr Core::makeToxCore(const QByteArray& savedata, const ICoreSettings* const settings,
-                             ToxCoreErrors* err)
+                             IBootstrapListGenerator& bootstrapNodes, ToxCoreErrors* err)
 {
     QThread* thread = new QThread();
     if (thread == nullptr) {
@@ -543,7 +544,7 @@ ToxCorePtr Core::makeToxCore(const QByteArray& savedata, const ICoreSettings* co
         return {};
     }
 
-    ToxCorePtr core(new Core(thread));
+    ToxCorePtr core(new Core(thread, bootstrapNodes));
     if (core == nullptr) {
         if (err) {
             *err = ToxCoreErrors::ERROR_ALLOC;
@@ -785,9 +786,9 @@ void Core::bootstrapDht()
 {
     ASSERT_CORE_THREAD;
 
-    QList<DhtServer> bootstrapNodes = BootstrapNodeUpdater::loadDefaultBootstrapNodes();
+    QList<DhtServer> bootstrapNodesList = bootstrapNodes.getBootstrapnodes();
 
-    int listSize = bootstrapNodes.size();
+    int listSize = bootstrapNodesList.size();
     if (!listSize) {
         qWarning() << "No bootstrap node list";
         return;
@@ -801,7 +802,7 @@ void Core::bootstrapDht()
 #endif
     // i think the more we bootstrap, the more we jitter because the more we overwrite nodes
     while (i < 2) {
-        const DhtServer& dhtServer = bootstrapNodes[j % listSize];
+        const DhtServer& dhtServer = bootstrapNodesList[j % listSize];
         QString dhtServerAddress = dhtServer.address.toLatin1();
         QString port = QString::number(dhtServer.port);
         QString name = dhtServer.name;
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -49,6 +49,7 @@ class ICoreSettings;
 class GroupInvite;
 class Profile;
 class Core;
+class IBootstrapListGenerator;
 
 using ToxCorePtr = std::unique_ptr<Core>;
 
@@ -69,7 +70,7 @@ public:
     };
 
     static ToxCorePtr makeToxCore(const QByteArray& savedata, const ICoreSettings* const settings,
-                                  ToxCoreErrors* err = nullptr);
+                                  IBootstrapListGenerator& bootstrapNodes, ToxCoreErrors* err = nullptr);
     static Core* getInstance();
     const CoreAV* getAv() const;
     CoreAV* getAv();
@@ -188,7 +189,7 @@ signals:
     void failedToRemoveFriend(uint32_t friendId);
 
 private:
-    Core(QThread* coreThread);
+    Core(QThread* coreThread, IBootstrapListGenerator& _bootstrapNodes);
 
     static void onFriendRequest(Tox* tox, const uint8_t* cUserId, const uint8_t* cMessage,
                                 size_t cMessageSize, void* core);
@@ -251,4 +252,5 @@ private:
     mutable QMutex coreLoopLock{QMutex::Recursive};
 
     std::unique_ptr<QThread> coreThread;
+    IBootstrapListGenerator& bootstrapNodes;
 };
--- a/src/core/dhtserver.cpp
+++ b/src/core/dhtserver.cpp
@@ -26,8 +26,14 @@
  */
 bool DhtServer::operator==(const DhtServer& other) const
 {
-    return this == &other || (port == other.port && address == other.address
-                              && userId == other.userId && name == other.name);
+    return this == &other ||
+        (statusUdp == other.statusUdp
+        && statusTcp == other.statusTcp
+        && ipv4 == other.ipv4
+        && ipv6 == other.ipv6
+        && maintainer == other.maintainer
+        && userId == other.userId
+        && port == other.port);
 }
 
 /**
--- a/src/core/dhtserver.h
+++ b/src/core/dhtserver.h
@@ -23,9 +23,12 @@
 
 struct DhtServer
 {
-    QString name;
+    bool statusUdp;
+    bool statusTcp;
+    QString ipv4;
+    QString ipv6;
+    QString maintainer;
     QString userId;
-    QString address;
     quint16 port;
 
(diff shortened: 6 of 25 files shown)


sudden6 (changes requested):
> Reviewed 15 of 24 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126)* status: 0 of 1 LGTMs obtained

___
*[src/core/dhtserver.cpp, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/6126#-M6LcGt6-OSRiY04kXMd:-M6LcGt6-OSRiY04kXMe:b-2e7ukt) ([raw file](https://github.com/qtox/qtox/blob/045f289ca50f98072404231646789f8f2000e07d/src/core/dhtserver.cpp#L29)):*
> ```cpp
> bool DhtServer::operator==(const DhtServer& other) const
> {
>     bool statusUdp;
> ```

This comparison operator now doesn't really make sense to me. Now you're comparing default initialized variables to the other Object?
___
*[src/net/bootstrapnodeupdater.h, line 10 at r1](https://reviewable.io/reviews/qtox/qtox/6126#-M6LcetD3R6gCliMWj0G:-M6LcetD3R6gCliMWj0H:b-w9tsfh) ([raw file](https://github.com/qtox/qtox/blob/045f289ca50f98072404231646789f8f2000e07d/src/net/bootstrapnodeupdater.h#L10)):*
> ```objc
> 
> #include "src/core/dhtserver.h"
> #include "src/net/ibootstraplistgenerator.h"
> ```

IMO that interface should be in `src/core` since `Core needs this`

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126)* status: 0 of 1 LGTMs obtained

___
*[src/core/dhtserver.cpp, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/6126#-M6LcGt6-OSRiY04kXMd:-M6Lwfj31T8gwr-OePAn:b-nko7tu) ([raw file](https://github.com/qtox/qtox/blob/045f289ca50f98072404231646789f8f2000e07d/src/core/dhtserver.cpp#L29)):*

This comparison operator now doesn't really make sense to me. Now you're comparing default initialized variables to the other Object?
</blockquote></details>

Bug, I just copy pasted the members in to list them all :facepalm: Removed.
___
*[src/net/bootstrapnodeupdater.h, line 10 at r1](https://reviewable.io/reviews/qtox/qtox/6126#-M6LcetD3R6gCliMWj0G:-M6M0uIB0pu7yBbeRo95:b-t31xns) ([raw file](https://github.com/qtox/qtox/blob/045f289ca50f98072404231646789f8f2000e07d/src/net/bootstrapnodeupdater.h#L10)):*

IMO that interface should be in `src/core` since `Core needs this`
</blockquote></details>

I agreed `net` doesn't make sense, I've moved it to `model` rather than `core`

sudden6 (approved):
:lgtm_strong:

> Reviewed 5 of 24 files at r1, 6 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126#-:-M6MBOqIDpP8BhwPARet:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sphaerophoria (commented):
I might be misusing the feature, but something doesn't seem right to me? If I change ~/.config/tox/bootstrapNodes.json to

```
{
    "nodes": [
        {
            "ipv4": "127.0.0.1",
            "ipv6": "",
            "maintainer": "sphaerophoria",
            "port": 33445,
            "public_key": "41076B90F4C0869D98F126C551D411C6176CCEC68A7CEF97D55A8C5C63885244",
            "status_tcp": true,
            "status_udp": true
        }
    ]
}
```

I still see the following logs
```
[20:07:38.770 UTC] core/core.cpp:807 : Debug: Connecting to bootstrap node 13
[20:07:38.772 UTC] core/core.cpp:807 : Debug: Connecting to bootstrap node 14
[20:07:43.270 UTC] core/core.cpp:807 : Debug: Connecting to bootstrap node 15
[20:07:43.273 UTC] core/core.cpp:807 : Debug: Connecting to bootstrap node 16
```

but shouldn't there only be one bootstrap node in the list?

Also if I remove some of the required fields from the list I see no indication that anything has failed. Am I using it wrong?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126#-:-M6R8SnsADcxFTD4fup6:b-jxyyp1)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 810 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R4wsQD0FJn_6NSNWS:-M6R4wsQD0FJn_6NSNWT:bd18p6i) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/core/core.cpp#L810)):*
> ```cpp
> 
>         QByteArray address;
>         if (dhtServer.ipv4.isEmpty() && !dhtServer.ipv6.isEmpty()) {
> ```

Should we ensure that only one of these is set in the bootstrap loader? That way we could just use ipv6 if it's valid, otherwise use ipv4.

Also that would ensure we don't try to pass in an empty address to tox_bootstrap
___
*[src/net/bootstrapnodeupdater.cpp, line 67 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R4Qp_2JtcErlAy3W1:-M6R4Qp_2JtcErlAy3W2:b7wnjq9) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L67)):*
> ```cpp
> 
>     // only use nodes that provide at least UDP connection
>     if (!node[NodeFields::status_udp].toBool(false)) {
> ```

What is the point in all of these checks if we already check the needed fields at the top?
___
*[src/net/bootstrapnodeupdater.cpp, line 115 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R604b5zqj3TcEg8Eq:-M6R604b5zqj3TcEg8Er:bg1hnp8) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L115)):*
> ```cpp
> 
>     if (!nodeList.isObject()) {
>         return result;
> ```

We should add warn logs for users since the nodes list is user editable now.
___
*[src/net/bootstrapnodeupdater.cpp, line 120 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R5u1qBgwTFuRZ8-aP:-M6R5u1qBgwTFuRZ8-aQ:bg1hnp8) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L120)):*
> ```cpp
>     QJsonObject rootObj = nodeList.object();
>     if (!(rootObj.contains(jsonNodeArrayName) && rootObj[jsonNodeArrayName].isArray())) {
>         return result;
> ```

We should add warn logs for users since the nodes list is user editable now.
___
*[src/net/bootstrapnodeupdater.cpp, line 142 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R6Ef37c1c2vRhp1le:-M6R6Ef48tvulClvCSHZ:b-fhlf2y) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L142)):*
> ```cpp
>     QString nodesJson = nodesFile.readAll();
>     nodesFile.close();
>     QJsonDocument d = QJsonDocument::fromJson(nodesJson.toUtf8());
> ```

I don't see any reason to use single letter variable names instead of something a little more descriptive
___
*[src/net/bootstrapnodeupdater.cpp, line 261 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R6Sqj3kvZ9kfKl9cJ:-M6R6Sqj3kvZ9kfKl9cK:b4lybs3) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L261)):*
> ```cpp
> }
> 
> QString BootstrapNodeUpdater::getUserNodesFilePath()
> ```

If we're moving to getting paths from the Paths class doesn't it make more sense for this to go there?
___
*[src/persistence/paths.h, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0UdZCDaJt2N8d04A:-M6R0Ud_3otA1DXySzHz:b-acgll4) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.h#L46)):*
> ```objc
>     QString getScreenshotsDir() const;
>     // to be removed when paths migration is complete.
>     QString getLegacySettingsDirPath() const;
> ```

I'm not sure how I feel about these names. If they are the only paths available why are they legacy? If I were trying to retrieve the settings path I would look at this and say "where is the non-legacy settings dir" which seems to not exist.

I'd maybe just note in the doxygen comments that these paths are intended to be moved?

@anthonybilinski indicated that the other paths in here are currently unused as well, which makes things extra confusing since someone would look for the screenshots dir and be pointed to the wrong location. If these paths are a work in progress maybe it's better to feature guard them with an ifdef or something so people know not to use them? 
___
*[src/persistence/paths.cpp, line 263 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0ts6A5XhFkHgL45d:-M6R0ts6A5XhFkHgL45e:byoc8gj) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.cpp#L263)):*
> ```cpp
>  * @note To be removed when path migration is complete.
>  */
> QString Paths::getLegacySettingsDirPath() const
> ```

I'm assuming these are straight copy pastes from the old settings functions. I try to note that type of change in the commit message to ease reviewer load, but that's just me being selfish :P

anthonybilinski (commented):
> I might be misusing the feature, but something doesn't seem right to me

Fixed, `loadNodesFile` was still always using the built-in file rather than the passed in file. It'll now log warnings with invalid config.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126#-:-M6Rrwlm0z_vvTC5mLxf:b-wvjbll)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/core.cpp, line 810 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R4wsQD0FJn_6NSNWS:-M6Rsv5aEZF5Y9BOJsZj:b-d50vqc) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/core/core.cpp#L810)):*
Yeah this part seems iffy. The source, https://nodes.tox.chat/ has ipv4 and ipv6 addresses whenever a node supports both, so setting both in `DhtServer` mirrors that info. I favoured using ipv4 if both are present because the user of qTox might not have an ipv6 connection. I think ideally we would just try to bootstrap from each address, but since our bootstrapping is (questionably) throttled to just trying a couple nodes per bootstrap, and (questionably) says that adding more nodes slows down connecting, I skipped doing that since I figured that would just lead to a slower connection.

If/when our Core bootstrapping is updated, I think it could handle dual-protocol nodes better.

> Also that would ensure we don't try to pass in an empty address to tox_bootstrap

I added a check to `jsonNodeToDhtServer` that will reject nodes that have neither an ipv4 nor ipv6 address.
___
*[src/net/bootstrapnodeupdater.cpp, line 67 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R4Qp_2JtcErlAy3W1:-M6RuMxH330ScQPaPRnJ:bkwb21) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L67)):*

What is the point in all of these checks if we already check the needed fields at the top?
</blockquote></details>

The check at the top is just checking that the keys are present, all the subsequent ones are checking the values of each config.
___
*[src/net/bootstrapnodeupdater.cpp, line 115 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R604b5zqj3TcEg8Eq:-M6RumkP7_d_Fj8tRx0w:b-886gsx) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L115)):*

We should add warn logs for users since the nodes list is user editable now.
</blockquote></details>

Fixed
___
*[src/net/bootstrapnodeupdater.cpp, line 120 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R5u1qBgwTFuRZ8-aP:-M6Runhs1szQbzVh8_UE:b-886gsx) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L120)):*

We should add warn logs for users since the nodes list is user editable now.
</blockquote></details>

Fixed
___
*[src/net/bootstrapnodeupdater.cpp, line 142 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R6Ef37c1c2vRhp1le:-M6RvJOD5YjXtvVPOP_M:bt2vr1b) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L142)):*

I don't see any reason to use single letter variable names instead of something a little more descriptive
</blockquote></details>

Fixed.
___
*[src/net/bootstrapnodeupdater.cpp, line 261 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R6Sqj3kvZ9kfKl9cJ:-M6RwDhc60N5eOuWBHNA:b-rhcrhu) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L261)):*

If we're moving to getting paths from the Paths class doesn't it make more sense for this to go there?
</blockquote></details>

Seems dependent on where Paths goes. The "old"/current way was that we have 3 dirs and every class knows its own path within whichever dir, the "new"/unused way is that `Paths` provides each file path. If we do end up removing all the unused/inaccurate API from `Paths` then I'd be happy to move this in, but if `Paths` is going to stay split between the 3 "old" directories and the many "new"/unused/inaccurate paths, I'd rather not add `getUserNodesFilePath` to it so that it doesn't get mixed in with the other API :S
___
*[src/persistence/paths.h, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0UdZCDaJt2N8d04A:-M6RwmQQ0O0FWd5YoOBR:b-1qzyst) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.h#L46)):*

I'm not sure how I feel about these names. If they are the only paths available why are they legacy? If I were trying to retrieve the settings path I would look at this and say "where is the non-legacy settings dir" which seems to not exist.

I'd maybe just note in the doxygen comments that these paths are intended to be moved?

@anthonybilinski indicated that the other paths in here are currently unused as well, which makes things extra confusing since someone would look for the screenshots dir and be pointed to the wrong location. If these paths are a work in progress maybe it's better to feature guard them with an ifdef or something so people know not to use them? 
</blockquote></details>

Yeah the more I think about Paths, the more this API is extremely confusing. @sudden6 would you be ok with me just deleting all of the current unused API from `Paths` in a single commit? When the paths reword is finished, that commit could be backed out, but in the meantime it would leave `Paths` with an accurate and reasonable API, where each function gives the correct path.
___
*[src/persistence/paths.cpp, line 263 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0ts6A5XhFkHgL45d:-M6Rx99Q7T0nVUCf5G-3:b-jcqilk) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.cpp#L263)):*

I'm assuming these are straight copy pastes from the old settings functions. I try to note that type of change in the commit message to ease reviewer load, but that's just me being selfish :P 
</blockquote></details>

That's fair, I could have split it into a commit along with the code moving in `BootstrapNodeUpdate`r. I'll keep that in mind going forward.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/paths.h, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0UdZCDaJt2N8d04A:-M6TuEX7AwKDRwy3am3o:bhh41dl) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.h#L46)):*

Yeah the more I think about Paths, the more this API is extremely confusing. @sudden6 would you be ok with me just deleting all of the current unused API from `Paths` in a single commit? When the paths reword is finished, that commit could be backed out, but in the meantime it would leave `Paths` with an accurate and reasonable API, where each function gives the correct path.
</blockquote></details>

I'm a bit against completely removing the API, because then we'll forget about it, but ifdef out the problematic parts seems fine.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6126)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/net/bootstrapnodeupdater.cpp, line 261 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R6Sqj3kvZ9kfKl9cJ:-M6WDK5KDi94R6Iz1D9j:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/net/bootstrapnodeupdater.cpp#L261)):*

Seems dependent on where Paths goes. The "old"/current way was that we have 3 dirs and every class knows its own path within whichever dir, the "new"/unused way is that `Paths` provides each file path. If we do end up removing all the unused/inaccurate API from `Paths` then I'd be happy to move this in, but if `Paths` is going to stay split between the 3 "old" directories and the many "new"/unused/inaccurate paths, I'd rather not add `getUserNodesFilePath` to it so that it doesn't get mixed in with the other API :S
</blockquote></details>

Done.
___
*[src/persistence/paths.h, line 46 at r3](https://reviewable.io/reviews/qtox/qtox/6126#-M6R0UdZCDaJt2N8d04A:-M6WDBoh3c8Fze6DzEDx:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/d1553f45482e78b795da32f010bbdd91a08d7d8b/src/persistence/paths.h#L46)):*

I'm a bit against completely removing the API, because then we'll forget about it, but ifdef out the problematic parts seems fine.
</blockquote></details>

Done.

PR #6129, opened 2020-05-03
refactor(db): map foreign key dependency in db schema

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6129)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -27,7 +27,7 @@
 #include "src/core/toxpk.h"
 
 namespace {
-static constexpr int SCHEMA_VERSION = 4;
+static constexpr int SCHEMA_VERSION = 5;
 
 bool createCurrentSchema(RawDatabase& db)
 {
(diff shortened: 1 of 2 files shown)


sphaerophoria (commented):
Just noting for my own understanding.

https://sqlite.org/foreignkeys.html looks like foreign keys will enforce everything we want, cannot delete a referenced key, cannot insert with a reference key that doesn't exist.

So just to be clear, the foreign key and db upgrade doesn't solve anything here right? The actual fix is the one line change reordering the delete calls and the foreign key stuff is just for safety? If so :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6129#-:-M6RBUEwDRu-CRfWcO-V:bacu4vf)* status: :shipit: complete! 1 of 1 LGTMs obtained

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6129#-:-M6RCKJL4WM4_uqo5k8K:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
I'd say even the one line change reordering deletes doesn't fix anything either. We would just be in an invalid state for a couple calls before mid-delete. Now our db will always be consistent, but still if we crashed half way between the mass delete, we might have deleted e.g. `broken_messages` but not `history`, causing us to mass send out messages we were previously holding back. To fix that we should really use transactions more.

I view this change as just additional safety checks going forwards.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6129#-:-M6Raurm1_YBqIEUAoKo:b-k7i7bw)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6130, opened 2020-05-03
refactor(history): enforce using PK instead of ToxId in history

anthonybilinski:
Need to be leniant when reading from db because of ToxIds being saved in the db
from a bug introduced in e07d8d358f6fc890a77e029aa230b69bdecd325e, which used
self ID rather than self Pk. Was subsequently fixed in
033f28e67ecd61cd0fdb6cc0e8de8a2bc920e1d7.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6130)

### DIFF
--- a/src/core/toxid.cpp
+++ b/src/core/toxid.cpp
@@ -114,7 +114,7 @@ ToxId::ToxId(const QByteArray& rawId)
  * If the given rawId isn't a valid Public Key or Tox ID a ToxId with all zero bytes is created.
  *
  * @param rawId Pointer to bytes to convert to ToxId object
- * @param len Number of bytes to read. Must be TOX_SECRET_KEY_SIZE for a Public Key or
+ * @param len Number of bytes to read. Must be TOX_PUBLIC_KEY_SIZE for a Public Key or
  *            TOX_ADDRESS_SIZE for a Tox ID.
  */
 ToxId::ToxId(const uint8_t* rawId, int len)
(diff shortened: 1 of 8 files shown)


sphaerophoria (commented):
Nice, I didn't run this myself, but it looks pretty reasonable. My only concern is switching the type of peers from QHash<QString, int64_t> to QHash<ToxPk, int64_t>, since before we conceivably could have had a peer stored as anything. This seems low risk to me though

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6130#-:-M6RDB-J4tI6utAKmJaK:b-7b7bfw)* status: 0 of 1 LGTMs obtained

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6130#-:-M6RDvFk3TFD2y9l5BjC:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
Removed some needless whitespace changes. I think the map change should be fairly safe. ToxPk asserts on size during construction, I've run it on my old profile full of  content, and every construction of ToxPk looks sane (no one just shoving random 32 byte strings into it).

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6130#-:-M6RmxWt0rfQX1sxRmtv:b-dqvq7a)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6154, opened 2020-05-24
fix(chathistory): file transfers sometimes show wrong author name

### LINKED ISSUE #6147: File transfers from loaded history sometimes show the wrong author
##### Brief Description

OS: Ubuntu 20.04
qTox version: v1.17.2+
Commit hash:  14571abaa30c80f8ebaae33c22c9f500cad7242a
toxcore:  v0.2.12
Qt:  5.12.8

Reproducible: Always

##### Steps to reproduce

1. A sends B text message.
1. A sends B file.
1. B sends A text message.
1. A sends B file.
1. Re-open profile A

##### Observed Behavior
1. A sees the first file as being sent by A.
2. A sees the second file as being sent by B, but with B's name bolded, like is typically done when it's A's own name.

##### Expected Behavior
The same author is shown before and after client restarts.

##### Additional Info
B sees the continues to see the correct sender of the files after restarting.

bodwok:
Sets the name of the file sender, by analogy with: ChatHistory::onMessageReceived and ChatHistory::onMessageSent

Fixes: [6147](https://github.com/qTox/qTox/issues/6147)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6154)

### DIFF
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -215,13 +215,16 @@ void ChatHistory::onFileUpdated(const ToxPk& sender, const ToxFile& file)
     if (canUseHistory()) {
         switch (file.status) {
         case ToxFile::INITIALIZING: {
+            auto selfPk = coreIdHandler.getSelfPublicKey();
+            QString username(selfPk == sender ? coreIdHandler.getUsername() : f.getDisplayedName());
+
             // Note: There is some implcit coupling between history and the current
             // chat log. Both rely on generating a new id based on the state of
             // initializing. If this is changed in the session chat log we'll end up
             // with a different order when loading from history
             history->addNewFileMessage(f.getPublicKey(), file.resumeFileId, file.fileName,
                                        file.filePath, file.filesize, sender,
-                                       QDateTime::currentDateTime(), f.getDisplayedName());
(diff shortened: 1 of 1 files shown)


sudden6:
@anthonybilinski assigned you, because you opened the issue.

anthonybilinski (approved):
Ah the reason the issue was only apparent after the friend sent a text message is because we hide the username when the author public key matches the previous recent message, and we're still saving the author public key correctly, just not the display name, that makes sense.

It's unfortunate that the bug is on saving rather than loading, meaning we'll have incorrect file transfer authors in history even after this fix. We could theoretically heal this since we have the right author public key, maybe by grabbing the author's last display name used from a text message from history or something, or at least showing our own public key instead of the friend's name. I'll open a lower priority issue to do that which I might take a stab at later. I don't think that's any reason to hold up this fix on that though.

Tested with 8e7ae808a38a5d7cfc024b31f1285db9ef8f8959 and could still reproduce the issue, tested with this change and could not. Change looks good.

:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6154#-:-M89pYltD7Il1EeeCWI0:b-655xea)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6166, opened 2020-06-04
fix: Usage of random

nurupo:
- Seed random in Core Thread. Core Thread didn't seed random, resulting
  in always using the same bootstrap nodes, even when you restart qTox
  or change profiles.

- Use QDateTime::currentMSecsSinceEpoch() for seeding random. It
  provides a bigger range of numbers than QTime::currentTime().msec()
  does, and the latter somehow managed to result in approximately the
  same first random number being generated, within a certain range.

- Use something a it more sensible than a mod operation to bound random
  numbers within a range. It's not perfect either, but a lot better.
  Using mod on random skews its distribution too much.

- Use QRandomGenerator's bounded() function to generate random values
  within a range.

- Enable QRandomGenerator's usage starting with Qt 5.10.0.
  QRandomGenerator is present since Qt 5.10.0, not 5.15.0.

- Bootstrap off every 5th node instead of two consecutive nodes. It's
  likely that two consecutive nodes will have the same owner, which
  makes some attacks more likely. The node selection algorithm should be
  scraped and redone from scratch to be honest though.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6166)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -34,8 +34,10 @@
 #include "util/strongtype.h"
 
 #include <QCoreApplication>
-#if (QT_VERSION >= QT_VERSION_CHECK(5, 15, 0))
+#if (QT_VERSION >= QT_VERSION_CHECK(5, 10, 0))
 #include <QRandomGenerator>
+#else
+#include <QDateTime>
 #endif
 #include <QRegularExpression>
 #include <QString>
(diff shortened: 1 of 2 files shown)


sudden6:
> Bootstrap off every 5th node instead of two consecutive nodes. It's
likely that two consecutive nodes will have the same owner, which
makes some attacks more likely. The node selection algorithm should be
scraped and redone from scratch to be honest though.

~IMO it would be simpler to just not boostrap from two consecutive nodes, but from two random nodes instead, what do you think?~

Clarified by IRC discussion. I overlooked that `j` is static so the completely random approach is significantly more work. Complete rewrite of this is necessary at some point.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6166#-:-M900E1t82eAEfDOnw7D:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6176, opened 2020-06-15
refactor: remove unused OSX platform code

sudden6:

- `moveToAppFolder` was commented out and thus not used
- `migrateProfile` says to remove at a future point in time, this was 3 years ago, I think it's a good time now

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6176)


sphaerophoria (commented):
So if I understand correctly

* This removes the unused code for moving the installation to the app folder
* This removes the _currently active_ code to migrateProfiles, but we don't mind because it is intended to be removed at some point "down the line" anyways
* This removes the autoupdater

So the first two points seem okay, but doesn't this potentially break a feature users are using? If that's a conscious decision I guess that's fine, but it seems weird that it wasn't addressed in the PR comments.

> Reviewed 2 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6176#-:-MAIGmAYDGAuMcxDf-Id:by99ucq)* status: 0 of 1 LGTMs obtained

sphaerophoria (commented):
I guess this is working under the assumption that no user calls the updater directly, but it's _only_ called through install_osx.cpp. That seems like a somewhat reasonable expectation... I think I'm okay with removing it (however I have quite low context so I may not be the right person to make this call)

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6176#-:-MAIHYRi2zTzkmf0iS6m:b-xnjn75)* status: 0 of 1 LGTMs obtained

sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6176#-:-MAIHjhe6KiEZcx4JwDN:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
there is no updater anymore, users are supposed to just download and install the newest version

nurupo:
Fixes https://github.com/qTox/qTox/issues/6179

(didn't notice this PR when opening the issue)

PR #6185, opened 2020-07-07
Push-to-talk using Libuiohook

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6185)

This is currently in a working state. TODOs are
- [X] Add multi-key support on scanning end. Settings/saving already work. This would involve tracking which of the keys are pressed.
- [x] Improve key-setting UI to display key names rather than their numbers.
- [x] Make passing of key list to hook callback threadsafe with some non-blocking locking, failing to lock is better than blocking to avoid hanging whole system.
- [X] Improve logging
- [x] Some code cleanup with inline TODOs
- [x] Dynamically destruct when PTT is not selected in AV settings, construct when it is
- [x] Add build for OSX and Windows, update build instructions for Linux
- [x] Remove singleton Settings from HotkeyInput. Either delay construction until AvForm and pass it in, or move functionality that requires Settings outside of class.
- [x] Clean up HotkeyInput (remove globals etc.)
- [x] Dynamically load Libuiohook since it depends on screen, causing unit tests to fail
- [x] Clean up git history

Overall this is way less work than the previous implementations. A couple limitations using this approach:
1. On X11, shortcut keys are still passed through to the underlying application. All libs I saw either use XGrabKey which causes the application to briefly grab focus, eating all other key strokes that happen while the shortcut is held, or have this behaviour. The keys _are_ blockable on OSX and Windows.
1. We get tons of info, way beyond what we need, e.g. all mouse and keyboard input of any kind. This is a bit of a privacy concern, but is the nature of the OS APIs that are provided. De-registering when PTT is not selected in settings and having it off by default should help with that.
1. Libuiohook has no void* in the callback, so we have no way to get back to our class, meaning our impl needs some gross global state. This isn't that bad since this is conceptually a singleton concept anyway.

### DIFF
--- a/.travis/build-ubuntu-16-04.sh
+++ b/.travis/build-ubuntu-16-04.sh
@@ -46,6 +46,7 @@ sudo apt-get install -y --force-yes \
     qttools5-dev \
     libqt5opengl5-dev \
     libqt5svg5-dev \
+    libxtst-dev \
     pkg-config || yes
 
 # ffmpeg
@@ -124,6 +125,16 @@ git checkout tags/1.0.18
 CC="ccache $CC" CXX="ccache $CXX" ./configure
 CC="ccache $CC" CXX="ccache $CXX" make -j$(nproc)
 sudo checkinstall --install --pkgname libsodium --pkgversion 1.0.8 --nodoc -y
+sudo ldconfig
+cd ..
+# libuiohook
+git clone https://github.com/kwhat/libuiohook.git
+cd libuiohook
+git checkout tags/1.0.3
+./bootstrap.sh
+CC="ccache $CC" CXX="ccache $CXX" ./configure
+CC="ccache $CC" CXX="ccache $CXX" make -j$(nproc)
+sudo make install
 sudo ldconfig
 cd ..
 # toxcore
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -406,6 +406,8 @@ set(${PROJECT_NAME}_SOURCES
   src/widget/form/settings/generalform.h
   src/widget/form/settings/genericsettings.cpp
   src/widget/form/settings/genericsettings.h
+  src/widget/form/settings/hotkeyinput.cpp
+  src/widget/form/settings/hotkeyinput.h
   src/widget/form/settings/privacyform.cpp
   src/widget/form/settings/privacyform.h
   src/widget/form/settings/userinterfaceform.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,6 +261,10 @@ set(${PROJECT_NAME}_SOURCES
   src/grouplist.h
   src/ipc.cpp
   src/ipc.h
+  src/globalshortcut.cpp
+  src/globalshortcut.h
+  src/hookmanager.cpp
+  src/hookmanager.h
   src/model/about/aboutfriend.cpp
   src/model/about/aboutfriend.h
   src/model/about/iaboutfriend.h
--- a/audio/include/audio/iaudiosettings.h
+++ b/audio/include/audio/iaudiosettings.h
@@ -26,6 +26,12 @@
 
 class IAudioSettings {
 public:
+    enum class AudioCaptureMode {
+        Continuous = 0,
+        VoiceActivation = 1,
+        PushToTalk = 2
+    };
+
     virtual ~IAudioSettings() = default;
 
     virtual QString getInDev() const = 0;
@@ -51,9 +57,9 @@ public:
     virtual int getOutVolumeMax() const = 0;
     virtual void setOutVolume(int volume) = 0;
 
-    virtual int getAudioCaptureMode() const = 0;
-    virtual void setAudioCaptureMode(int mode) = 0;
-    
+    virtual AudioCaptureMode getAudioCaptureMode() const = 0;
+    virtual void setAudioCaptureMode(AudioCaptureMode mode) = 0;
+
     virtual QList<int> getPttShortcutKeys() const = 0;
     virtual void setPttShortcutKeys(QList<int> keys) = 0;
     
@@ -72,7 +78,7 @@ public:
     DECLARE_SIGNAL(audioInGainDecibelChanged, qreal dB);
     DECLARE_SIGNAL(audioThresholdChanged, qreal dB);
     DECLARE_SIGNAL(outVolumeChanged, int volume);
-    DECLARE_SIGNAL(audioCaptureModeChanged, int mode);
+    DECLARE_SIGNAL(audioCaptureModeChanged, AudioCaptureMode mode);
     DECLARE_SIGNAL(pttShortcutKeysChanged, QList<int> keys);
     DECLARE_SIGNAL(audioBitrateChanged, int bitrate);
     DECLARE_SIGNAL(enableTestSoundChanged, bool newValue);
--- a/audio/include/audio/iaudiosettings.h
+++ b/audio/include/audio/iaudiosettings.h
@@ -22,6 +22,7 @@
 #include "util/interface.h"
 
 #include <QString>
+#include <QList>
 
 class IAudioSettings {
 public:
@@ -50,6 +51,12 @@ public:
     virtual int getOutVolumeMax() const = 0;
     virtual void setOutVolume(int volume) = 0;
 
+    virtual int getAudioCaptureMode() const = 0;
+    virtual void setAudioCaptureMode(int mode) = 0;
+    
+    virtual QList<int> getPttShortcutKeys() const = 0;
+    virtual void setPttShortcutKeys(QList<int> keys) = 0;
+    
     virtual int getAudioBitrate() const = 0;
     virtual void setAudioBitrate(int bitrate) = 0;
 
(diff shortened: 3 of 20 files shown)


anthonybilinski:
I think the code is ready to go, there's just the small issue of libuiohook causing every test to segfault in CI.

This is because it's doing registration with the system during library load, and that's failing because there's no display present. To work around this I'll have to avoid linking libuiohook until qTox can verify that a display is present, then load it in code at that time.

anthonybilinski:
Updated to load libuiohook at runtime instead of linking to it, and made the AV form's option for PTT dynamically dependent on the ability to load the lib, and I've cleaned up git history to a point where I think this is ready to be reviewed and tested.

It looks like CI is failing due to this issue:
```
Resolving download.libsodium.org (download.libsodium.org)... 37.59.238.213
Connecting to download.libsodium.org (download.libsodium.org)|37.59.238.213|:443... connected.
GnuTLS: A TLS fatal alert has been received.
GnuTLS: received alert [70]: Error in protocol version
Unable to establish SSL connection.
```
which I've been seeing for about a week.

anthonybilinski:
Tested on Windows, a couple issues:
1) The library is named `libuiohook-0.dll` so we fail to find it since we're looking for `libuiohook.dll`. On Linux it's called `libuiohook.so` so our name in code is just `libuiohook`. I don't know why the naming is different on the different systems, but at worst we can just rename the library during build.
2) The key blocking works correctly (which isn't possible on X11), but due to the way we set shortcuts, they continue being blocked until a new shortcut is set. So e.g. it's impossible to directly change from "A+S+D" -> "A" since the "A" input will be blocked. To fix this we should clear the setting as soon as the box is selected, then re-set the old combo if no new combo is set.

Other than that it works well. OSX isn't tested yet, but I've got an environment set up to test, just don't see where I can grab the build from from the CI. Will build locally a try it, but at least issue 2) should still be present.

anthonybilinski:
Updated to fix the key-blocking when resetting and tested on macOS as well. It _generally_ works on macOS but there are a couple problems:

* On open, as soon as we check if the lib is there by attempting to load it which is before the GUI is even shown, libuiohook causes this prompt (which then quickly gets hidden behind the main UI =/). If qTox isn't started from the terminal, then it properly says that qTox is asking for the permission.
image
Before the prompt has been answered, libuiohook logs an error:
image
and then doesn't work even once the permission is granted. When switching away from PTT as the input setting or just closing qTox, we call `hook_stop` which segfaults:
image
which looks like a bug inside libuiohook.
* On next start once this permission has been granted, it works great and key blocking works as expected.
* Another minor issue is that on library load, it spits out a bunch of text to stdout rather than going through our registered log callback, which can clobber our logs and generally doesn't respect our log levels:
image


Overall this is a really shit user experience. It seems the best option would be to ourselves ask for this permission before even trying to load the library, or alternatively fix the segfault bug in libuiohook and try to re-load it once the permission is granted. :disappointed:

anthonybilinski:
I've added a single commit backing out the (otherwise working) macOS implementation. This commit can be backed out when #6214 is fixed. I'd still like to merge this PR as-is since setting up permission management is a fairly unrelated task, this works fully on Windows, and this would help testing on macOS in the future by just backing out the last commit.

If the reviewer doesn't want a change that includes then backs out macOS support that's reasonable, but I think it would be nice to have a finished working version of adding hook to the macOS for future use.

anthonybilinski:
On second thought adding macOS support then backing it out in the same PR is dumb. I've just added support on top of this change and pushed that to https://github.com/anthonybilinski/qTox/tree/libuiohook_ptt_macos, which I've mentioned in https://github.com/qTox/qTox/issues/6214. This PR now only adds Linux and Windows support, and doesn't touch macOS at all.

Once permissions are properly handled on macOS in https://github.com/qTox/qTox/issues/6214, it should be very easy to just include the one commit from https://github.com/anthonybilinski/qTox/tree/libuiohook_ptt_macos.

sphaerophoria:
Trying it out and just a couple comments

- It's pretty funny that this is printed in qtox logs on every boot
```
libUIOHook: Cross-platfrom userland keyboard and mouse hooking.
Copyright (C) 2006-2015 Alexander Barker.  All Rights Received.
https://github.com/kwhat/libuiohook/

libUIOHook is free software: you can redistribute it and/or modify
it under the terms of the GNU Lesser General Public License as published
by the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.

libUIOHook is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU Lesser General Public License
along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
This may be related to my libuiohook version. I have version 1.0.3 from https://aur.archlinux.org/packages/libuiohook/

- I think there should be a visual indicator in the audio settings page of when the PTT key is active. This helps first time users understand how the feature works and test out different keys.
- Right ctrl/left ctrl display the same in the preview but which ctrl key you have bound is important
- If I press right alt I see
image
- This empty space is weird
image
- The lack of a way to clear the button is confusing. I agree it's unneeded because of the capture mode dropdown, but maybe writing that in the "Press any key combo" box would make it more clear
- I couldn't bind to a mouse button
- I'd prefer if there was a way to block the input keys from being passed through to other applications, but IIRC from conversations with @anthonybilinski  this is not possible on linux due to X11 limitations. @anthonybilinski it would be nice to link the libuiohook issue here for future reference
- My machine is behaving as push to mute instead of push to talk. I'm not sure if that's a mislabeling or I've done something wrong
- The feature itself seems to work though :+1: 

I'm not sure if any of these should block the PR, just making them known just in case. @anthonybilinski  let me know if you want me to proceed with review now or hold off until some of the above are addressed.

sudden6:
Installed this but seems `libuiohook` is not found by qTox.

Install log:

```
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/lib64/libuiohook.so.1.2.0
-- Installing: /usr/local/lib64/libuiohook.so.1
-- Installing: /usr/local/lib64/libuiohook.so
-- Up-to-date: /usr/local/include/uiohook.h
-- Old export file "/usr/local/lib64/cmake/uiohook/uiohook-config.cmake" will be replaced.  Removing files [/usr/local/lib64/cmake/uiohook/uiohook-config-noconfig.cmake].
-- Installing: /usr/local/lib64/cmake/uiohook/uiohook-config.cmake
-- Installing: /usr/local/lib64/cmake/uiohook/uiohook-config-noconfig.cmake
-- Up-to-date: /usr/local/lib64/pkgconfig/uiohook.pc
```

No relevant output in qTox log. I think it could make sense to log if `libuiohook` couldn't be found, makes debugging easier.

I don't need any special build flags for qTox, do I?

UI for audio activation looks very nice now.

sudden6 (changes requested):
> Reviewed 6 of 19 files at r6.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6185)* status: 0 of 1 LGTMs obtained

___
*[src/hookmanager.cpp, line 43 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGous6n0d_ESmnzA4RW:-MGous6o9od74tRap8ct:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/hookmanager.cpp#L43)):*
> ```cpp
>         case UIOHOOK_SUCCESS:
>             return;
>         
> ```

whitespace error
___
*[src/persistence/settings.h, line 367 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGovwml5uccoIbDoGU3:-MGovwmm2_SGN6VRKL6K:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/persistence/settings.h#L367)):*
> ```objc
>     }
>     void setOutVolume(int volume) override;
>     
> ```

whitespace error
___
*[src/persistence/settings.h, line 370 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGovz6i7Doh6h1aum9W:-MGovz6i7Doh6h1aum9X:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/persistence/settings.h#L370)):*
> ```objc
>     AudioCaptureMode getAudioCaptureMode() const override;
>     void setAudioCaptureMode(AudioCaptureMode mode) override;
>     
> ```

whitespace error
___
*[src/persistence/settings.h, line 376 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGovzoC2WMMa9QVXhBx:-MGovzoD231XgGysfc7g:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/persistence/settings.h#L376)):*
> ```objc
>     QList<int> getPttShortcutNames() const override;
>     void setPttShortcutNames(QList<int> keys) override;   
>     
> ```

whitespace error
___
*[src/widget/form/settings/avform.cpp, line 581 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGowTqT1REPFJ-fhN2e:-MGowTqT1REPFJ-fhN2f:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/widget/form/settings/avform.cpp#L581)):*
> ```cpp
>     pushToTalkShortcutLabel->hide();
>     pushToTalkShortcutInput->hide();
>     
> ```

whitespace error
___
*[src/widget/form/settings/avform.cpp, line 598 at r6](https://reviewable.io/reviews/qtox/qtox/6185#-MGowUl0AHq7C4ejKaKG:-MGowUl0AHq7C4ejKaKH:b-bdq992) ([raw file](https://github.com/qtox/qtox/blob/5908f21c02537bb4b103962ed9eb7d8773af0899/src/widget/form/settings/avform.cpp#L598)):*
> ```cpp
> void AVForm::on_inModeComboBox_currentIndexChanged(int index)
> {
>     const IAudioSettings::AudioCaptureMode mode = static_cast<IAudioSettings::AudioCaptureMode>(inModeComboBox->currentData().toInt()); 
> ```

whitespace error

PR #6216, opened 2020-08-20
refactor(chatlog): remove getInstance from ChatlogItem and remove Core::getInstance()

sudden6:

Goal is to finally remove all users of `Core::getInstance()`.

The problem is currently where to do the lookup of `ToxPk` -> `displayedName´.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6216)

### DIFF
--- a/src/core/core.cpp
+++ b/src/core/core.cpp
@@ -29,7 +29,6 @@
 #include "src/model/groupinvite.h"
 #include "src/model/status.h"
 #include "src/model/ibootstraplistgenerator.h"
-#include "src/nexus.h"
 #include "src/persistence/profile.h"
 #include "util/strongtype.h"
 
@@ -703,15 +702,6 @@ void Core::onStarted()
 void Core::start()
 {
     coreThread->start();
-}
-
-
-/**
- * @brief Returns the global widget's Core instance
- */
-Core* Core::getInstance()
-{
-    return Nexus::getCore();
 }
 
 const CoreAV* Core::getAv() const
--- a/src/core/core.h
+++ b/src/core/core.h
@@ -71,7 +71,6 @@ public:
 
     static ToxCorePtr makeToxCore(const QByteArray& savedata, const ICoreSettings* const settings,
                                   IBootstrapListGenerator& bootstrapNodes, ToxCoreErrors* err = nullptr);
-    static Core* getInstance();
     const CoreAV* getAv() const;
     CoreAV* getAv();
     CoreFile* getCoreFile() const;
--- a/src/model/chatlogitem.cpp
+++ b/src/model/chatlogitem.cpp
@@ -18,9 +18,6 @@
 */
 
 #include "chatlogitem.h"
-#include "src/core/core.h"
-#include "src/friendlist.h"
-#include "src/grouplist.h"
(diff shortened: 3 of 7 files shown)


sudden6:
This is one of the milestones defined in The Plan™ years ago. With this two commits it's now possible to use the Core class without the rest of qTox.

sudden6:
@Diadlo :tada: :tada: :tada:

bodwok:
It fine works for me. 
But I noticed that the empty nickname is not replaced by the public key? (in SessionChatLog)

anthonybilinski (approved):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6216)* status: 0 of 1 LGTMs obtained

___
*[src/model/sessionchatlog.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6216#-MFWf0_e28YmCB0Khcfe:-MFWf0_fAHszyNEQRBNv:bdv8h03) ([raw file](https://github.com/qtox/qtox/blob/c77fac3787f15a8f3c57785780ce104839cd9451/src/model/sessionchatlog.cpp#L124)):*
> ```cpp
>     }
> 
>     return pk.toString();
> ```

@bodwok the empty nickname is replaced by a public key here. `pk.toString()` gives the hek key, and is only called if the pk isn't found in the `FriendList` or `GroupList` which would contain the nickname if there was one. Testing off this change, unsetting one of my profiles names correctly shows the pk for both itself and its friend running this change.

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6216#-:-MFWfLgDC8GXA3G8bJao:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
Specifically tested the case with empty username again, couldn't find the behaviou @bodwok mentioned.

anthonybilinski (changes requested):
:lgtm_cancel: ah wait there is an issue. When there's no name set, it works on the receiving side, but not on the sending side.
Receiving side:

Sending side:


> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6216#-:-MFXE1Ky9euDeW5V9-IC:b-5e036x)* status: 0 of 1 LGTMs obtained

sudden6:
Could reproduce, but it seems this also happens when the name is not empty O_o

sudden6:
Fixed it, I accidentally changed the logic when sending messages.

anthonybilinski (approved):
Fix looks good, and I've re-tested. Behaviour is good on both sending and receiving side. :lgtm_strong: 

Thanks for pointing out the problem @bodwok :)

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6216#-:-MFXQa4t31tKvumYY6br:bpk7kcp)* status: :shipit: complete! 1 of 1 LGTMs obtained

bodwok:
Tested, looks good

sudden6:
@bodwok Thanks for testing and reporting the problem!

PR #6227, opened 2020-09-02
Refactor/core cmake

sudden6:
Working on putting `Core` in a separate cmake project.

- [x] Commits follow our [git commit guidelines](https://github.com/qTox/qTox/blob/master/CONTRIBUTING.md#git-commit-guidelines)

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6227)

PR #6235, opened 2020-09-17
Refactor/coreav calls

sudden6:

@anthonybilinski as we discussed I decided to split this part from #6227 in order to make it easier to merge.

The changes:

- remove functions implemented by `ToxCall` from `CoreAV`, it was mostly just looking up the calls and then forwarding to `ToxCall`
- starting a call returns a `std::shared_pt<ToxCall>`, it must be ended using `endCall()` or there will be a resource leak
- made some properties in `ToxCall` atomic to prevent threading issues, because `ToxCall` objects are used by the Audio, CoreAV and UI threads

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)

### DIFF
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -489,7 +489,7 @@ bool CoreAV::sendGroupCallAudio(int groupNum, const int16_t* pcm, size_t samples
 {
     QReadLocker locker{&callsLock};
 
-    auto it = groupCalls.find(groupId);
+    auto it = groupCalls.find(groupNum);
     if (it == groupCalls.end()) {
         return false;
     }
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -447,27 +447,11 @@ void CoreAV::groupCallCallback(void* tox, uint32_t group, uint32_t peer, const i
 }
 
 /**
- * @brief Called from core to make sure the source for that peer is invalidated when they leave.
- * @param group Group Index
- * @param peer Peer Index
- */
-void CoreAV::invalidateGroupCallPeerSource(const Group& group, ToxPk peerPk)
-{
-    QWriteLocker locker{&callsLock};
-
-    auto it = groupCalls.find(group.getId());
-    if (it == groupCalls.end()) {
-        return;
-    }
-    it->second->removePeer(peerPk);
-}
-
-/**
  * @brief Starts a call in an existing AV groupchat.
  * @note Call from the GUI thread.
  * @param groupId Id of group to join
  */
-void CoreAV::joinGroupCall(const Group& group)
+CoreAV::ToxGroupCallPtr CoreAV::joinGroupCall(const Group& group)
 {
     QWriteLocker locker{&callsLock};
 
@@ -480,12 +464,13 @@ void CoreAV::joinGroupCall(const Group& group)
     // Call Objects must be owned by CoreAV or there will be locking problems with Audio
     groupcall->moveToThread(thread());
     assert(groupcall != nullptr);
-    auto ret = groupCalls.emplace(group.getId(), std::move(groupcall));
+    auto ret = groupCalls.emplace(group.getId(), groupcall);
     if (ret.second == false) {
         qWarning() << "This group call already exists, not joining!";
-        return;
+        return {};
     }
     ret.first->second->setActive(true);
+    return groupcall;
 }
 
 /**
@@ -520,72 +505,6 @@ bool CoreAV::sendGroupCallAudio(int groupNum, const int16_t* pcm, size_t samples
         qDebug() << "toxav_group_send_audio error";
 
     return true;
-}
-
-/**
- * @brief Mutes or unmutes the group call's input (microphone).
- * @param g The group
- * @param mute True to mute, false to unmute
- */
-void CoreAV::muteCallInput(const Group* g, bool mute)
-{
-    QWriteLocker locker{&callsLock};
-
-    auto it = groupCalls.find(g->getId());
-    if (g && (it != groupCalls.end())) {
-        it->second->setMuteMic(mute);
-    }
-}
-
-/**
- * @brief Mutes or unmutes the group call's output (speaker).
- * @param g The group
- * @param mute True to mute, false to unmute
- */
-void CoreAV::muteCallOutput(const Group* g, bool mute)
-{
-    QWriteLocker locker{&callsLock};
-
-    auto it = groupCalls.find(g->getId());
-    if (g && (it != groupCalls.end())) {
-        it->second->setMuteVol(mute);
-    }
-}
-
-/**
- * @brief Returns the group calls input (microphone) state.
- * @param groupId The group id to check
- * @return true when muted, false otherwise
- */
-bool CoreAV::isGroupCallInputMuted(const Group* g) const
-{
-    QReadLocker locker{&callsLock};
-
-    if (!g) {
-        return false;
-    }
-
-    const uint32_t groupId = g->getId();
-    auto it = groupCalls.find(groupId);
-    return (it != groupCalls.end()) && it->second->getMuteMic();
-}
-
-/**
- * @brief Returns the group calls output (speaker) state.
- * @param groupId The group id to check
- * @return true when muted, false otherwise
- */
-bool CoreAV::isGroupCallOutputMuted(const Group* g) const
-{
-    QReadLocker locker{&callsLock};
-
-    if (!g) {
-        return false;
-    }
-
-    const uint32_t groupId = g->getId();
-    auto it = groupCalls.find(groupId);
-    return (it != groupCalls.end()) && it->second->getMuteVol();
 }
 
 /**
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -196,43 +196,6 @@ void CoreAV::process()
 }
 
 /**
- * @brief Checks the call status for a Tox friend.
- * @param f the friend to check
- * @return true, if call is started for the friend, false otherwise
- */
-bool CoreAV::isCallStarted(const Friend* f) const
-{
-    QReadLocker locker{&callsLock};
-    return f && (calls.find(f->getId()) != calls.end());
-}
-
-/**
- * @brief Checks the call status for a Tox group.
- * @param g the group to check
- * @return true, if call is started for the group, false otherwise
- */
-bool CoreAV::isCallStarted(const Group* g) const
-{
-    QReadLocker locker{&callsLock};
-    return g && (groupCalls.find(g->getId()) != groupCalls.end());
-}
-
-/**
- * @brief Checks the call status for a Tox friend.
- * @param f the friend to check
- * @return true, if call is active for the friend, false otherwise
- */
-bool CoreAV::isCallActive(const Friend* f) const
-{
-    QReadLocker locker{&callsLock};
-    auto it = calls.find(f->getId());
-    if (it == calls.end()) {
-        return false;
-    }
-    return isCallStarted(f) && it->second->isActive();
-}
-
-/**
  * @brief Checks the call status for a Tox group.
  * @param g the group to check
  * @return true, if the call is active for the group, false otherwise
@@ -244,17 +207,10 @@ bool CoreAV::isCallActive(const Group* g) const
     if (it == groupCalls.end()) {
         return false;
     }
-    return isCallStarted(g) && it->second->isActive();
+    return it->second->isActive();
 }
 
-bool CoreAV::isCallVideoEnabled(const Friend* f) const
-{
-    QReadLocker locker{&callsLock};
-    auto it = calls.find(f->getId());
-    return isCallStarted(f) && it->second->getVideoEnabled();
-}
-
-bool CoreAV::answerCall(uint32_t friendNum, bool video)
+CoreAV::ToxFriendCallPtr CoreAV::answerCall(uint32_t friendNum, bool video)
 {
     QWriteLocker locker{&callsLock};
     QMutexLocker coreLocker{&coreLock};
@@ -268,18 +224,18 @@ bool CoreAV::answerCall(uint32_t friendNum, bool video)
     if (toxav_answer(toxav.get(), friendNum, audioSettings.getAudioBitrate(),
                      videoBitrate, &err)) {
         it->second->setActive(true);
-        return true;
+        return it->second;
     } else {
         qWarning() << "Failed to answer call with error" << err;
         Toxav_Err_Call_Control controlErr;
         toxav_call_control(toxav.get(), friendNum, TOXAV_CALL_CONTROL_CANCEL, &controlErr);
         PARSE_ERR(controlErr);
         calls.erase(it);
-        return false;
+        return {};
     }
 }
 
-bool CoreAV::startCall(uint32_t friendNum, bool video)
+CoreAV::ToxFriendCallPtr CoreAV::startCall(uint32_t friendNum, bool video)
 {
     QWriteLocker locker{&callsLock};
     QMutexLocker coreLocker{&coreLock};
@@ -288,7 +244,7 @@ bool CoreAV::startCall(uint32_t friendNum, bool video)
     auto it = calls.find(friendNum);
     if (it != calls.end()) {
         qWarning() << QString("Can't start call with %1, we're already in this call!").arg(friendNum);
-        return false;
+        return {};
     }
 
     uint32_t videoBitrate = video ? VIDEO_DEFAULT_BITRATE : 0;
@@ -296,7 +252,7 @@ bool CoreAV::startCall(uint32_t friendNum, bool video)
     toxav_call(toxav.get(), friendNum, audioSettings.getAudioBitrate(), videoBitrate,
                     &err);
     if (!PARSE_ERR(err)) {
-        return false;
+        return {};
     }
 
     // Audio backend must be set before making a call
@@ -306,14 +262,20 @@ bool CoreAV::startCall(uint32_t friendNum, bool video)
     // Call object must be owned by this thread or there will be locking problems with Audio
     call->moveToThread(thread());
     assert(call != nullptr);
-    calls.emplace(friendNum, std::move(call));
-    return true;
+    calls.emplace(friendNum, call);
+    return call;
 }
 
 bool CoreAV::cancelCall(uint32_t friendNum)
 {
     QWriteLocker locker{&callsLock};
     QMutexLocker coreLocker{&coreLock};
+
+    auto it = calls.find(friendNum);
+    if (it == calls.end()) {
+        qWarning() << QString("Can't cancel call with %1, it doesn't exist!").arg(friendNum);
+        return false;
+    }
 
     qDebug() << QString("Cancelling call with %1").arg(friendNum);
     Toxav_Err_Call_Control err;
@@ -322,22 +284,14 @@ bool CoreAV::cancelCall(uint32_t friendNum)
         return false;
     }
 
+    // Set inactive for others
+    it->second->setActive(false);
+
     calls.erase(friendNum);
     locker.unlock();
 
     emit avEnd(friendNum);
     return true;
-}
-
-void CoreAV::timeoutCall(uint32_t friendNum)
-{
-    QWriteLocker locker{&callsLock};
-
-    if (!cancelCall(friendNum)) {
-        qWarning() << QString("Failed to timeout call with %1").arg(friendNum);
-        return;
-    }
-    qDebug() << "Call with friend" << friendNum << "timed out";
 }
 
 /**
@@ -442,36 +396,6 @@ void CoreAV::sendCallVideo(uint32_t callId, std::shared_ptr<VideoFrame> vframe)
 }
 
 /**
- * @brief Toggles the mute state of the call's input (microphone).
- * @param f The friend assigned to the call
- */
-void CoreAV::toggleMuteCallInput(const Friend* f)
-{
-    QWriteLocker locker{&callsLock};
-
-    auto it = calls.find(f->getId());
-    if (f && (it != calls.end())) {
-        ToxCall& call = *it->second;
-        call.setMuteMic(!call.getMuteMic());
-    }
-}
-
-/**
- * @brief Toggles the mute state of the call's output (speaker).
- * @param f The friend assigned to the call
- */
-void CoreAV::toggleMuteCallOutput(const Friend* f)
-{
-    QWriteLocker locker{&callsLock};
-
-    auto it = calls.find(f->getId());
-    if (f && (it != calls.end())) {
-        ToxCall& call = *it->second;
-        call.setMuteVol(!call.getMuteVol());
-    }
-}
-
-/**
  * @brief Called from Tox API when group call receives audio data.
  *
  * @param[in] tox          the Tox object
@@ -536,24 +460,6 @@ void CoreAV::invalidateGroupCallPeerSource(const Group& group, ToxPk peerPk)
         return;
     }
     it->second->removePeer(peerPk);
-}
-
-/**
- * @brief Get a call's video source.
- * @param friendNum Id of friend in call list.
- * @return Video surface to show
- */
-VideoSource* CoreAV::getVideoSourceFromCall(int friendNum) const
-{
-    QReadLocker locker{&callsLock};
-
-    auto it = calls.find(friendNum);
-    if (it == calls.end()) {
-        qWarning() << "CoreAV::getVideoSourceFromCall: No such call, possibly cancelled";
-        return nullptr;
-    }
-
-    return it->second->getVideoSource();
 }
 
 /**
@@ -683,40 +589,6 @@ bool CoreAV::isGroupCallOutputMuted(const Group* g) const
 }
 
 /**
- * @brief Returns the calls input (microphone) mute state.
- * @param f The friend to check
- * @return true when muted, false otherwise
- */
-bool CoreAV::isCallInputMuted(const Friend* f) const
-{
-    QReadLocker locker{&callsLock};
-
-    if (!f) {
-        return false;
-    }
-    const uint32_t friendId = f->getId();
-    auto it = calls.find(friendId);
-    return (it != calls.end()) && it->second->getMuteMic();
-}
-
-/**
- * @brief Returns the calls output (speaker) mute state.
- * @param friendId The friend to check
- * @return true when muted, false otherwise
- */
-bool CoreAV::isCallOutputMuted(const Friend* f) const
-{
-    QReadLocker locker{&callsLock};
-
-    if (!f) {
-        return false;
-    }
-    const uint32_t friendId = f->getId();
-    auto it = calls.find(friendId);
-    return (it != calls.end()) && it->second->getMuteVol();
-}
-
-/**
  * @brief Signal to all peers that we're not sending video anymore.
  * @note The next frame sent cancels this.
  */
@@ -793,11 +665,15 @@ void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, voi
 
     if (state & TOXAV_FRIEND_CALL_STATE_ERROR) {
         qWarning() << "Call with friend" << friendNum << "died of unnatural causes!";
+        // Set inactive for others
+        it->second->setActive(false);
         self->calls.erase(friendNum);
         locker.unlock();
         emit self->avEnd(friendNum, true);
     } else if (state & TOXAV_FRIEND_CALL_STATE_FINISHED) {
         qDebug() << "Call with friend" << friendNum << "finished quietly";
+        // Set inactive for others
+        it->second->setActive(false);
         self->calls.erase(friendNum);
         locker.unlock();
         emit self->avEnd(friendNum);
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -99,6 +99,7 @@ CoreAV::CoreAV(std::unique_ptr<ToxAV, ToxAVDeleter> toxav_, CompatibleRecursiveM
     connect(iterateTimer, &QTimer::timeout, this, &CoreAV::process);
     connect(coreavThread.get(), &QThread::finished, iterateTimer, &QTimer::stop);
     connect(coreavThread.get(), &QThread::started, this, &CoreAV::process);
+    qRegisterMetaType<ToxFriendCallPtr>("ToxFriendCallPtr");
 }
 
 void CoreAV::connectCallbacks()
@@ -210,7 +211,7 @@ bool CoreAV::isCallActive(const Group* g) const
     return it->second->isActive();
 }
 
-CoreAV::ToxFriendCallPtr CoreAV::answerCall(uint32_t friendNum, bool video)
+bool CoreAV::answerCall(uint32_t friendNum, bool video)
 {
     QWriteLocker locker{&callsLock};
     QMutexLocker coreLocker{&coreLock};
@@ -224,14 +225,14 @@ CoreAV::ToxFriendCallPtr CoreAV::answerCall(uint32_t friendNum, bool video)
     if (toxav_answer(toxav.get(), friendNum, audioSettings.getAudioBitrate(),
                      videoBitrate, &err)) {
         it->second->setActive(true);
-        return it->second;
+        return true;
     } else {
         qWarning() << "Failed to answer call with error" << err;
         Toxav_Err_Call_Control controlErr;
         toxav_call_control(toxav.get(), friendNum, TOXAV_CALL_CONTROL_CANCEL, &controlErr);
         PARSE_ERR(controlErr);
         calls.erase(it);
-        return {};
+        return false;
     }
 }
 
@@ -262,7 +263,7 @@ CoreAV::ToxFriendCallPtr CoreAV::startCall(uint32_t friendNum, bool video)
     // Call object must be owned by this thread or there will be locking problems with Audio
     call->moveToThread(thread());
     assert(call != nullptr);
-    calls.emplace(friendNum, call);
+    calls.emplace(friendNum, call.get());
     return call;
 }
 
@@ -273,7 +274,6 @@ bool CoreAV::cancelCall(uint32_t friendNum)
 
     auto it = calls.find(friendNum);
     if (it == calls.end()) {
-        qWarning() << QString("Can't cancel call with %1, it doesn't exist!").arg(friendNum);
         return false;
     }
 
@@ -539,7 +539,7 @@ void CoreAV::callCallback(ToxAV* toxav, uint32_t friendNum, bool audio, bool vid
     call->moveToThread(self->thread());
     assert(call != nullptr);
 
-    auto it = self->calls.emplace(friendNum, std::move(call));
+    auto it = self->calls.emplace(friendNum, call.get());
     if (it.second == false) {
         qWarning() << QString("Rejecting call invite from %1, we're already in that call!").arg(friendNum);
         Toxav_Err_Call_Control err;
@@ -560,7 +560,7 @@ void CoreAV::callCallback(ToxAV* toxav, uint32_t friendNum, bool audio, bool vid
     // Must explicitely unlock, because a deadlock can happen via ChatForm/Audio
     locker.unlock();
 
-    emit self->avInvite(friendNum, video);
+    emit self->avInvite(friendNum, video, call);
 }
 
 void CoreAV::stateCallback(ToxAV* toxav, uint32_t friendNum, uint32_t state, void* vSelf)
--- a/src/core/coreav.cpp
+++ b/src/core/coreav.cpp
@@ -284,9 +284,6 @@ bool CoreAV::cancelCall(uint32_t friendNum)
         return false;
     }
 
-    // Set inactive for others
-    it->second->setActive(false);
-
     calls.erase(friendNum);
     locker.unlock();
 
@@ -437,7 +434,7 @@ void CoreAV::groupCallCallback(void* tox, uint32_t group, uint32_t peer, const i
         return;
     }
 
-    ToxGroupCall& call = *it->second;
+    auto& call = *it->second;
 
     if (call.getMuteVol() || !call.isActive()) {
         return;
@@ -462,14 +459,14 @@ CoreAV::ToxGroupCallPtr CoreAV::joinGroupCall(const Group& group)
 
     ToxGroupCallPtr groupcall = ToxGroupCallPtr(new ToxGroupCall{group, *this, *audio});
     // Call Objects must be owned by CoreAV or there will be locking problems with Audio
-    groupcall->moveToThread(thread());
     assert(groupcall != nullptr);
-    auto ret = groupCalls.emplace(group.getId(), groupcall);
+    groupcall->moveToThread(this->thread());
+    auto ret = groupCalls.emplace(group.getId(), groupcall.get());
     if (ret.second == false) {
         qWarning() << "This group call already exists, not joining!";
         return {};
     }
-    ret.first->second->setActive(true);
+    groupcall->setActive(true);
     return groupcall;
 }
 
@@ -492,7 +489,7 @@ bool CoreAV::sendGroupCallAudio(int groupNum, const int16_t* pcm, size_t samples
 {
     QReadLocker locker{&callsLock};
 
-    std::map<int, ToxGroupCallPtr>::const_iterator it = groupCalls.find(groupNum);
+    auto it = groupCalls.find(groupId);
     if (it == groupCalls.end()) {
         return false;
     }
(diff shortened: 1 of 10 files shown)


sudden6:
Good catch, fixed.

sudden6:
@bodwok can you re-review and give lgtm when you are satisfied?

bodwok (commented):
Sure, sorry for the delay

> Reviewed 7 of 8 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MNUFE0i3e-R5T8_nE9L:b-xznjp0)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/chatform.cpp, line 357 at r2](https://reviewable.io/reviews/qtox/qtox/6235#-MNUENhuAHHZnBLlVj0n:-MNUENhuAHHZnBLlVj0o:buugerp) ([raw file](https://github.com/qtox/qtox/blob/855103d7420fed86b8f22e9813889ef0eb38d3ef/src/widget/form/chatform.cpp#L357)):*
> ```
> if(call) {
> ```
need a space
___
*[src/widget/form/chatform.cpp, line 378 at r2](https://reviewable.io/reviews/qtox/qtox/6235#-MNUEWt0365QhVJsYu9K:-MNUEWt0365QhVJsYu9L:b-w3daky) ([raw file](https://github.com/qtox/qtox/blob/855103d7420fed86b8f22e9813889ef0eb38d3ef/src/widget/form/chatform.cpp#L378)):*
> ```
> if(call) {
> ```
same

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/chatform.cpp, line 357 at r2](https://reviewable.io/reviews/qtox/qtox/6235#-MNUENhuAHHZnBLlVj0n:-MNZfwr088rSsKoa-ODy:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/855103d7420fed86b8f22e9813889ef0eb38d3ef/src/widget/form/chatform.cpp#L357)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

> ```
> if(call) {
> ```
need a space
</blockquote></details>

Done.
___
*[src/widget/form/chatform.cpp, line 378 at r2](https://reviewable.io/reviews/qtox/qtox/6235#-MNUEWt0365QhVJsYu9K:-MNZfxBU47wwfEfw-I1b:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/855103d7420fed86b8f22e9813889ef0eb38d3ef/src/widget/form/chatform.cpp#L378)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

> ```
> if(call) {
> ```
same
</blockquote></details>

Done.

bodwok (commented):
:lgtm_strong:

> Reviewed 4 of 8 files at r1, 1 of 6 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MNchGRW1VW6kkUsjAUB:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (changes requested):
> Reviewed 4 of 8 files at r1, 4 of 6 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 52 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNf9ztK7C925BujrlSj:-MNf9ztK7C925BujrlSk:bz6xhoc) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L52)):*
> ```objc
> public:
>     using CoreAVPtr = std::unique_ptr<CoreAV>;
>     using ToxFriendCallPtr = std::shared_ptr<ToxFriendCall>;
> ```

since ownership isn't actually shared for `ToxFriendCall`s, having owning `unique_ptr`'s held by `CoreAV` and non-owing raw pointers handed out on `startCall` and `answerCall` would make the relationship more clear, and additionally more performant. See note here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-65
___
*[src/core/coreav.h, line 139 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfBC451haa-9if5MoX:-MNfBC451haa-9if5MoY:b7tfwch) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L139)):*
> ```objc
> 
> 
>     using ToxGroupCallPtr = std::shared_ptr<ToxGroupCall>;
> ```

same re: ownership
___
*[src/core/coreav.cpp, line 272 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfC3Zf3CItNMcWjma3:-MNfC3Zf3CItNMcWjma4:brvg8tt) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L272)):*
> ```cpp
> 
>     // Set inactive for others
>     it->second->setActive(false);
> ```

why should anyone be using the `ToxFriendCall` after cancelling? If it's being used as a start of a call is the creation of the object, then I'd think cancelling the call should invalidate the object?
___
*[src/core/coreav.cpp, line 642 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCOMu2P_4LuPXZSl6:-MNfCOMu2P_4LuPXZSl7:b-3yqfj1) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L642)):*
> ```cpp
>         qWarning() << "Call with friend" << friendNum << "died of unnatural causes!";
>         // Set inactive for others
>         it->second->setActive(false);
> ```

same question, marking for reviewable
___
*[src/core/coreav.cpp, line 649 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCT9D9GEVXT9FZ2wW:-MNfCT9D9GEVXT9FZ2wX:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L649)):*
> ```cpp
>         qDebug() << "Call with friend" << friendNum << "finished quietly";
>         // Set inactive for others
>         it->second->setActive(false);
> ```

same
___
*[src/core/toxcall.h, line 109 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCqr56vt7HJzgFdHf:-MNfCqr56vt7HJzgFdHg:bh57s37) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/toxcall.h#L109)):*
> ```objc
>     // ToxCall interface
> public:
>     void endCall() override;
> ```

Keeping with the RAII view of the lifetime of a call being represented as a `ToxCall`, it seems like ending call should be part of the destructor of `ToxCall`. Additionally every call to `endCall` happens right before destroying the `ToxCall` object. To simplify usage and reduce the risk of using `ToxCall` in an invalid state, what about doing the `endCall` logic within the destructor and removing the explicit API?
___
*[src/widget/form/chatform.cpp, line 381 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfDnJxAGNHIxEURWxm:-MNfDnJy4e9o1X75U4kb:b-wv2v3l) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/widget/form/chatform.cpp#L381)):*
> ```cpp
>         call->endCall();
>         call.reset();
>     }
> ```

if `endCall` is moved within destructor, this could unconditionally call `.reset()`
___
*[src/widget/form/chatform.cpp, line 434 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfECLf0tFv_vzj7Umq:-MNfECLf0tFv_vzj7Umr:bmsccvh) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/widget/form/chatform.cpp#L434)):*
> ```cpp
> void ChatForm::onMicMuteToggle()
> {
>     call->setMuteMic(!call->getMuteMic());
> ```

this loses thread safety compared to the previous implementation. Before the get/set were both done within `CoreAV::callsLock`, but now, even though `ToxCall::muteMic`is atomic, there could be a race between the get and set. This seems not trivial to fix: https://stackoverflow.com/a/9806613.

Maybe a toggle method could be added with a class mutex? Or the underlying type could be changed from bool and the xor method could be used, but that doesn't show intent as clearly IMO.
___
*[src/widget/form/groupchatform.cpp, line 244 at r4](https://reviewable.io/reviews/qtox/qtox/6235#-MNfGCoh6qdR4aD7bie8:-MNfGCoh6qdR4aD7bie9:b-58mbm0) ([raw file](https://github.com/qtox/qtox/blob/78cbd034c493e5fd27d4675701fec112d4506a6c/src/widget/form/groupchatform.cpp#L244)):*
> ```cpp
>     if (call) {
>         call->removePeer(user);
>     }
> ```

any reason to check `call != nullptr` here but not in `onMicMuteToggle` or `onVolMuteToggle`? I'd think that on any of these slots our call should still be valid, assuming our signals come in in the right order :P

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 52 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNf9ztK7C925BujrlSj:-MNfR8J9Ezd2xcm1ia4W:b1szjyr) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L52)):*

since ownership isn't actually shared for `ToxFriendCall`s, having owning `unique_ptr`'s held by `CoreAV` and non-owing raw pointers handed out on `startCall` and `answerCall` would make the relationship more clear, and additionally more performant. See note here: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#note-65
</blockquote></details>

ah I should have taken a closer look, and was giving mixed advice before. Ownership is kind of shared because the starter of the call has a reference and CoreAV keeps a reference to all calls as well to handle toxcore callbacks. I was also saying things that were  amix of "CoreAV should return a unique_ptr" and "CoreAV should hold on to the unique_ptr and return a raw pointer".. which was dumb.

I think startCall could still return the owning unique_ptr, CoreAV could hang on to a non-owning raw pointer of it, then the caller can end the call just by deleting the object, and in ToxCall's destructor it can call `CoreAV::cancelCall`?

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 52 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNf9ztK7C925BujrlSj:-MP6L57K2cttWOd8S3Dy:b1b0e99) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L52)):*

ah I should have taken a closer look, and was giving mixed advice before. Ownership is kind of shared because the starter of the call has a reference and CoreAV keeps a reference to all calls as well to handle toxcore callbacks. I was also saying things that were  amix of "CoreAV should return a unique_ptr" and "CoreAV should hold on to the unique_ptr and return a raw pointer".. which was dumb.

I think startCall could still return the owning unique_ptr, CoreAV could hang on to a non-owning raw pointer of it, then the caller can end the call just by deleting the object, and in ToxCall's destructor it can call `CoreAV::cancelCall`?
</blockquote></details>

I tried to do that, but a problem turned up: Calls can be allocated in `CoreAV::callCallback` and if we use raw pointers for this map, we risk never deallocating them properly. We could solve this, by making sure non of the `ToxCall` objects are alive when `CoreAV` is destroyed, but with our current mess of Widget destruction I don't see this happening. I don't think overhead of `shared_ptr` vs `unique_ptr` can be an argument here.
___
*[src/core/coreav.cpp, line 272 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfC3Zf3CItNMcWjma3:-MP6Hyal08PIM8JWpBA4:b-xdf4fl) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L272)):*

why should anyone be using the `ToxFriendCall` after cancelling? If it's being used as a start of a call is the creation of the object, then I'd think cancelling the call should invalidate the object?
</blockquote></details>

I have not yet figured out the complete flow of `setActive`, it's used here and in `stateCallback`, I just remember things broke when I touched it, so it stays in for now.
___
*[src/core/coreav.cpp, line 642 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCOMu2P_4LuPXZSl6:-MP6IG4ICtu34Om9QWqd:b-34vmyd) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L642)):*

same question, marking for reviewable
</blockquote></details>

See above
___
*[src/core/coreav.cpp, line 649 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCT9D9GEVXT9FZ2wW:-MP6IHMdDFpWU_tTrNFe:b-34vmyd) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L649)):*

same
</blockquote></details>

See above
___
*[src/core/toxcall.h, line 109 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfCqr56vt7HJzgFdHf:-MP6HwQKEDPpUQd9sCLh:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/toxcall.h#L109)):*

Keeping with the RAII view of the lifetime of a call being represented as a `ToxCall`, it seems like ending call should be part of the destructor of `ToxCall`. Additionally every call to `endCall` happens right before destroying the `ToxCall` object. To simplify usage and reduce the risk of using `ToxCall` in an invalid state, what about doing the `endCall` logic within the destructor and removing the explicit API?
</blockquote></details>

Done.
___
*[src/widget/form/chatform.cpp, line 381 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfDnJxAGNHIxEURWxm:-MP6HuVM1Go0sOUo7MhE:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/widget/form/chatform.cpp#L381)):*

if `endCall` is moved within destructor, this could unconditionally call `.reset()`
</blockquote></details>

Done.
___
*[src/widget/form/chatform.cpp, line 434 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfECLf0tFv_vzj7Umq:-MP6HF-R-sC3Igwoct3V:b-f1g96d) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/widget/form/chatform.cpp#L434)):*

this loses thread safety compared to the previous implementation. Before the get/set were both done within `CoreAV::callsLock`, but now, even though `ToxCall::muteMic`is atomic, there could be a race between the get and set. This seems not trivial to fix: https://stackoverflow.com/a/9806613.

Maybe a toggle method could be added with a class mutex? Or the underlying type could be changed from bool and the xor method could be used, but that doesn't show intent as clearly IMO.
</blockquote></details>

I'd just leave it at not atomic, because I don't think in this specific case the probability of hitting the non-atomic case is close non existant.
___
*[src/widget/form/groupchatform.cpp, line 244 at r4](https://reviewable.io/reviews/qtox/qtox/6235#-MNfGCoh6qdR4aD7bie8:-MP6GGCE9Lm9Gbpnd7CA:bua08xh) ([raw file](https://github.com/qtox/qtox/blob/78cbd034c493e5fd27d4675701fec112d4506a6c/src/widget/form/groupchatform.cpp#L244)):*

any reason to check `call != nullptr` here but not in `onMicMuteToggle` or `onVolMuteToggle`? I'd think that on any of these slots our call should still be valid, assuming our signals come in in the right order :P
</blockquote></details>

because for the other two cases the `call` is guarded by `audioInputFlag = false;    audioOutputFlag = false;` when reseting the `call` ptr.

anthonybilinski (approved):
> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 52 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNf9ztK7C925BujrlSj:-MRIiqsB-1DwSKOCCxrV:b-zi74cz) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L52)):*

I tried to do that, but a problem turned up: Calls can be allocated in `CoreAV::callCallback` and if we use raw pointers for this map, we risk never deallocating them properly. We could solve this, by making sure non of the `ToxCall` objects are alive when `CoreAV` is destroyed, but with our current mess of Widget destruction I don't see this happening. I don't think overhead of `shared_ptr` vs `unique_ptr` can be an argument here.
</blockquote></details>

I guess the way to push the RAII further would be to not create a `ToxCall` in `CoreAV::callCallback`, but instead do it in `CoreAV::answerCall` and return it there. Then if ChatForm rejects the call, there would be no `ToxCall` to clean up. It would have the added benefit(?) of not opening the webcam and mic before the call is answered.

My bigger concern with the `shared_ptr` is that it muddies ownership rather than the performance hit. It won't block it on this change now, it can always be done later.
___
*[src/core/coreav.cpp, line 272 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfC3Zf3CItNMcWjma3:-MRIfzO801el66VgKjYg:b-br1u9i) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L272)):*

I have not yet figured out the complete flow of `setActive`, it's used here and in `stateCallback`, I just remember things broke when I touched it, so it stays in for now.
</blockquote></details>

Ok, I guess this is at least existing weirdness
___
*[src/widget/form/chatform.cpp, line 434 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfECLf0tFv_vzj7Umq:-MRMb4yP74sQUm219_I6:bypm9km) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/widget/form/chatform.cpp#L434)):*

I'd just leave it at not atomic, because I don't think in this specific case the probability of hitting the non-atomic case is close non existant.
</blockquote></details>

Alright, I might make a quick PR after then, but don't want to keep holding this up.

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRMcPWMCzUcITpVFRcc:-MRMcPWMCzUcITpVFRcd:bclg7nr) (no related file):*
This segfaults if a you join then leave a group call:
```
Thread 1 "qtox" received signal SIGSEGV, Segmentation fault.                                                                                                                                               
std::__atomic_base<bool>::load (__m=std::memory_order_seq_cst, this=0x28) at /usr/include/c++/10/bits/atomic_base.h:426
426             return __atomic_load_n(&_M_i, int(__m));                                                                                                                                                   
(gdb) bt                                                                                             
#0  std::__atomic_base<bool>::load(std::memory_order) const (__m=std::memory_order_seq_cst, this=0x28) at /usr/include/c++/10/bits/atomic_base.h:426
#1  std::atomic<bool>::operator bool() const (this=0x28) at /usr/include/c++/10/atomic:87
#2  0x00005555558305d9 in ToxCall::getMuteMic() const (this=0x0) at ../src/core/toxcall.cpp:88
#3  0x000055555592f48a in GroupChatForm::onCallClicked() (this=0x555558368f90) at ../src/widget/form/groupchatform.cpp:333
```

---

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRMclNR-AWFUCfNeqtK:-MRMclNSC2yYHD74Fpoe:b-m98z03) (no related file):*
The call button in 1:1 also seems to have the wrong state in a few cases, e.g. while it's ringing if you press it again to cancel, it does nothing and logs:
```
[22:34:38.704 UTC] core/coreav.cpp:232 : Debug: "Starting call with 0"
[22:34:38.704 UTC] core/coreav.cpp:235 : Warning: "Can't start call with 0, we're already in this call!"
[22:34:38.704 UTC] widget/form/chatform.cpp:394 : Debug: Failed to start Audio call
```
Once they pick up it then goes green instead of red and you can't end call. The other end gets a red call button when they answer, but clicking it tries to start a new call instead of ending the call.

---

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRNIidM-y6s0nHdkCQl:-MRNIidN7oyxyW6V0JPE:bdqc7t3) (no related file):*
This also doesn't destroy the `ToxCall` in the case of the friend ending the call. CoreAv emits avEnd which reaches ChatForm::onAvEnd, but that doesn't actually clear the call.

I'm working on changing ownership of the ToxCall and handling these three problems, so I can update the PR soon with the fixes.

---

anthonybilinski (commented):
Ok well I'm happy enough with this now, I've addressed my concerns except for the atomic get/set. I don't really want to LGTM since I made some non-trivial changes on top of your branch, if you could review my last 3 commits I'd be happy for it to go it.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MRQk4pyBhUpU3Q_aPC5:bd5vrhq)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRMcPWMCzUcITpVFRcc:-MRQiqgN66G5z2y07hLa:b-8q18ng) (no related file):*

This segfaults if a you join then leave a group call:
```
Thread 1 "qtox" received signal SIGSEGV, Segmentation fault.                                                                                                                                               
std::__atomic_base<bool>::load (__m=std::memory_order_seq_cst, this=0x28) at /usr/include/c++/10/bits/atomic_base.h:426
426             return __atomic_load_n(&_M_i, int(__m));                                                                                                                                                   
(gdb) bt                                                                                             
#0  std::__atomic_base<bool>::load(std::memory_order) const (__m=std::memory_order_seq_cst, this=0x28) at /usr/include/c++/10/bits/atomic_base.h:426
#1  std::atomic<bool>::operator bool() const (this=0x28) at /usr/include/c++/10/atomic:87
#2  0x00005555558305d9 in ToxCall::getMuteMic() const (this=0x0) at ../src/core/toxcall.cpp:88
#3  0x000055555592f48a in GroupChatForm::onCallClicked() (this=0x555558368f90) at ../src/widget/form/groupchatform.cpp:333
```
</blockquote></details>

Fixed, chatform was accessing the null call in the second argument but was checking it in the first.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRMclNR-AWFUCfNeqtK:-MRQiibm9ho-zKNOH3QM:b-d44vqp) (no related file):*

The call button in 1:1 also seems to have the wrong state in a few cases, e.g. while it's ringing if you press it again to cancel, it does nothing and logs:
```
[22:34:38.704 UTC] core/coreav.cpp:232 : Debug: "Starting call with 0"
[22:34:38.704 UTC] core/coreav.cpp:235 : Warning: "Can't start call with 0, we're already in this call!"
[22:34:38.704 UTC] widget/form/chatform.cpp:394 : Debug: Failed to start Audio call
```
Once they pick up it then goes green instead of red and you can't end call. The other end gets a red call button when they answer, but clicking it tries to start a new call instead of ending the call.
</blockquote></details>

Fixed, think this was related to the next issue where `call` wasn't being reset in chatform.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6235#-MRNIidM-y6s0nHdkCQl:-MRQiQSX6c4eD0yUOVGp:b637ngg) (no related file):*

This also doesn't destroy the `ToxCall` in the case of the friend ending the call. CoreAv emits avEnd which reaches ChatForm::onAvEnd, but that doesn't actually clear the call.

I'm working on changing ownership of the ToxCall and handling these three problems, so I can update the PR soon with the fixes.
</blockquote></details>

Fixed. Little more complicated than just `.reset()`ing in `onAvEnd` because that causes a loop where it calls back into cancel call. I changed coreav to erase the call before emitting avEnd, and to ignore cancels to calls that it doesn't know about. It doesn't seem super elegant..

---
___
*[src/core/coreav.h, line 52 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNf9ztK7C925BujrlSj:-MRQjGh01znmkYAXHm2v:buyne5s) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L52)):*

I guess the way to push the RAII further would be to not create a `ToxCall` in `CoreAV::callCallback`, but instead do it in `CoreAV::answerCall` and return it there. Then if ChatForm rejects the call, there would be no `ToxCall` to clean up. It would have the added benefit(?) of not opening the webcam and mic before the call is answered.

My bigger concern with the `shared_ptr` is that it muddies ownership rather than the performance hit. It won't block it on this change now, it can always be done later.
</blockquote></details>

Amended this myself, doing it in `answerCall` isn't great because then it splits ownership between CoreAv and ChatForm, and when ChatForm wants to cancel a ringing call, it has to ask CoreAv to do that instead of doing it itself. Instead I return it through the invite signal, which requires it staying as a shared_ptr since signals require the argument to be copyable. Groupcalls could be made a unique_ptr since those are only ever returned directly. In both cases CoreAv has a purely non-owning map.
___
*[src/core/coreav.cpp, line 272 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfC3Zf3CItNMcWjma3:-MRQjkkf5NKqLh9DDcGD:b-mxvg21) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L272)):*

Ok, I guess this is at least existing weirdness
</blockquote></details>

I removed it only in the cases where it's set right before being destroyed, since it isn't used in the destructor and doesn't emit any signal, it's just dead code. Have tested with that change. Left in other uses of it.

bodwok (commented):
:lgtm_cancel: Additional testing is required

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MRUGPEQ0D72GwIgnY2F:b-vpktmm)* status: 0 of 1 LGTMs obtained

bodwok (commented):
> Reviewed 1 of 1 files at r7, 1 of 5 files at r8.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.cpp, line 516 at r8](https://reviewable.io/reviews/qtox/qtox/6235#-MRVLtTr8O0DW4qdrlxf:-MRVLtTr8O0DW4qdrlxg:bguvg39) ([raw file](https://github.com/qtox/qtox/blob/21d356baf7a8e9ad1bf94a23f2c61d2422df3a00/src/core/coreav.cpp#L516)):*
> ```cpp
>     assert(call != nullptr);
> 
>     auto it = self->calls.emplace(friendNum, std::move(call));
> ```

```calls``` is of type: ```std::map<uint32_t, ToxFriendCall*>```, need to use ```call.get()``` here?

sphaerophoria (commented):
At the request of @anthonybilinski  I'm looking over this. Just throwing my 2c into the ring hoping that it doesn't rub anyone the wrong way. 

Spent some time getting familiar with the layout before the change in order to understand why the change was good. @sudden6 I like the direction you went here. Creating an RAII abstraction around CoreAV so that upper classes don't have to use core directly is definitely a direction I agree with.

As far as I understand before ChatForm didn't see ToxCall at all. ToxCall was more of a helper class for CoreAv to keep track of it's call states and ChatForm asked CoreAv for call information. Something like this...


After this change it looks a little more like this


The ownership is a little more muddled. It looks like @anthonybilinski  wanted to move the ownership to a single spot which is good, but with the current code layout CoreAv still needs access to the ToxCall object. There isn't anything wrong with this in particular, but @anthonybilinski indicated in a private chat that the design goal moving forward is for Core/CoreAv to mostly be Qt wrappers around the toxcore APIs. If that is the case I believe the following changes could solidify ownership of responsibilities while moving us in the desired direction.



What do you guys think?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MRVoxBa2i-xrsOiruqh:b-o5xvv2)* status: 0 of 1 LGTMs obtained

sudden6:
@sphaerophoria I'm not yet deep into the changed code by @anthonybilinski but it seems that your proposed architecture additionally complicates the dependencies (diagram 3). AFAICT with the dispatcher you're mostly moving the routing of signals from `CoreAV` out to a separate class, correct? Personally my intention wasn't for CoreAV to be a think wrapper around c-toxcore, but rather a nice (maybe fat) abstraction around it so it can be conveniently used with Qt. So I see no problem when CoreAV keeps references to ToxCalls. I see it more as a factory, it creates object and as any good factory keeps track of where they go, after they are not needed the resources are recycled by the factory ;P

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 139 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfBC451haa-9if5MoX:-MRWveuV7NKfkptsLjQc:b-hf7ct4) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L139)):*

same re: ownership
</blockquote></details>

I'm still not very happy about storing raw pointers here, what about `weak_ptr`? it seems designed for exactly this problem.
___
*[src/core/coreav.cpp, line 272 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfC3Zf3CItNMcWjma3:-MRWwBjg3iW3Ttm0KRPG:b-i4yp8x) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.cpp#L272)):*

I removed it only in the cases where it's set right before being destroyed, since it isn't used in the destructor and doesn't emit any signal, it's just dead code. Have tested with that change. Left in other uses of it.
</blockquote></details>

LGTM, my problems were probably caused by one of the other things you fixed.
___
*[src/widget/form/chatform.cpp, line 285 at r8](https://reviewable.io/reviews/qtox/qtox/6235#-MRWy-YF1Yil4J5r9ow6:-MRWy-YF1Yil4J5r9ow7:b-c22feh) ([raw file](https://github.com/qtox/qtox/blob/21d356baf7a8e9ad1bf94a23f2c61d2422df3a00/src/widget/form/chatform.cpp#L285)):*
> ```cpp
>     }
> 
>     if (this->call) {
> ```

Minor thing: do you expect to run into this regularly? Else it should be turned into an assert IMO

sphaerophoria (commented):
I think where youre coming from. Why add extra classes resulting in more complex relationships between classes?

My thoughts were that the core layer should basically just convert callbacks to signals and do some minor conversion between tox types and qtox core types. The surface area of the toxcore api is already very wide, if we start mixing responsibilities of the core classes with higher level abstractions (like an RAII call wrapper) the scope of the class will become huge. If we move the ownership into its own class we now have a class with a single responsibility, in theory this can make testing of the ToxCall area simple as we can test the manager/dispatcher and ToxCalls with a mocked out CoreAv as well.

That being said I have no say in this, so feel free to just ignore me and do what you prefer ;) 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MR_xGn3-zetnMwRE6Gc:bpmrsyv)* status: 0 of 1 LGTMs obtained

sudden6:
@sphaerophoria yes that is exactly my main concern.

Another concern is that now instead of one slightly complicated class (CoreAV) you have two classes (CoreAV and the Dispatcher) that need to manage and track the state.

As for testing, I don't really see how additionally testing ToxCall and Dispatcher is better than just mocking CoreAV.

sphaerophoria:
> As for testing, I don't really see how additionally testing ToxCall and Dispatcher is better than just mocking CoreAV.

My thoughts were that we could now test the logic around creation/removal of ToxCalls with a mocked CoreAV. Without the split you have to instantiate CoreAV which I would imagine is not trivial to do easily in a testing context.

Personally I'd prefer more classes with more complex relationships as it makes reasoning about each smaller class easier (part of my understanding of SOLID and why people like it).

sudden6:
> Without the split you have to instantiate CoreAV which I would imagine is not trivial to do easily in a testing context.

I'd have imagined to use a mock of CoreAV for testing the upper layers, and using the real one for testing CoreAV + Core interactions. Not sure if that's enough, but it's already more than we have.

bodwok:
I tested the new changes with the latest version qTox and found no new bugs (other than the compile-level error mentioned above).

It looks like that at the beginning, when I was reviewing the code, I compared this PR with a similar one. Thanks to @anthonybilinski, we managed to discover this. 

A test for CoreAV would also be useful here and for future changes.

anthonybilinski (approved):
Rebased and addressed review comments just to keep this version up to date, even though we're not necessarily going with it.

RE: design conversation,
> Personally my intention wasn't for CoreAV to be a thin wrapper around c-toxcore, but rather a nice (maybe fat) abstraction around it so it can be conveniently used with Qt.

Whoops my mistake for incorrectly relaying that.

I agree with @sphaerophoria that `Core` is already a very wide class with complex responsibilities. Although like @sudden6 said we want a possibly fat abstraction layer that makes using Tox easy with Qt, that layer including multiple classes instead of one bigger class doesn't seem like a problem to me. To keep them together they could be stuck in a `Core` lib or just a namespace, but it doesn't seem like it needs to stay as one class to form that layer.

It's true the inter-class relationship gets more complex, but the complexity is moved out of intra-class relationships in `Core` where there's already a lot going on. Having a few classes working together each with state can be more easily reasoned about e.g. either looking inside of `Core` or `ToxCallDispatcher` with a few things going on in each, or looking at the external interactions between them with a few more things there. I think this is the idea of the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) in [SOLID](https://en.wikipedia.org/wiki/SOLID) like @sphaerophoria said.

So overall I'm in favour of the new class to reduce work done by `Core` and trying to keep `Core` thin even if we want a fat abstraction layer overall.

For ownership, my idea of moving ownership to be owned solely by the starter of the call only really makes sense if they're also the only one ending the call, but if we ever have a case where multiple view classes could start or end a call, e.g. `ChatForm` could start it, but if theoretically the `FriendListWidget` had a right click -> end call or the tray icon did or something, it wouldn't really handle that shared control well. Better might be for `ToxCallDispatcher` to be the owner, when a call is started they emit a signal with a `shared_ptr` of it which anyone interested in the call state can hang onto, and when one of the users wants to end the call instead of just deleting their `ToxCall`, they could ask the dispatcher to end it which could emit a signal to all reference holders which could all remove their `shared_ptr` on their slots to then delete the `ToxCall` when done. This would mean a change back to `shared_pointer` even though the ownership would be managed by `ToxCallDispatcher`. This might even be worth doing now with only 1 view class touching lifetime, since both `CoreAv` and `ChatForm` can want to end a call based on either user input or a call state callback. My current method `CoreAV::stateCallback` -> `ChatForm::onAvEnd` -> `CoreAV::cancelCall` leads to the weird case of sometimes `CoreAV` already knows the call is ended before `cancelCall` is called.

Overall I think my favourite direction would be for `ToxCallDispatcher` to be changed into `ToxCallManager` which would own the lifetime a bit loosely holding a `shared_ptr`, with both `CoreAV` and `ChatForm` getting signals from it when a call is started containing a `shared_ptr`, and both signaling to `ToxCallManager` when they want to end the call, which signals to all owners which causes them each to delete their `shared_ptr` of `ToxCall`.

Thoughts?

> Reviewed 10 of 10 files at r9, 5 of 5 files at r10, 4 of 4 files at r11, 1 of 1 files at r12, 1 of 1 files at r13, 5 of 5 files at r14.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6235#-:-MSOhTAC9-HJKpWJ5ORP:bi5c3z6)* status: 0 of 1 LGTMs obtained

___
*[src/core/coreav.h, line 139 at r3](https://reviewable.io/reviews/qtox/qtox/6235#-MNfBC451haa-9if5MoX:-MSOe-4XAGXjEFQjCtsS:b1q5nk5) ([raw file](https://github.com/qtox/qtox/blob/008e481e0b83460d903a58b48c6a2effb9187a2b/src/core/coreav.h#L139)):*

I'm still not very happy about storing raw pointers here, what about `weak_ptr`? it seems designed for exactly this problem.
</blockquote></details>

I think raw pointers are the recommended way to hold non-owning pointers according to ISOCPP: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-ptr. The lifetime here is guaranteed by ToxCall, so I don't think `weak_ptr` to check if the pointer is still valid is really necessary.

It seems like the only place they recommend `weak_ptr` is for breaking loops of `shared_ptr`s, but I think either with sole ownership `unique_ptr` is better and shared ownership `shared_ptr` is better.
___
*[src/core/coreav.cpp, line 516 at r8](https://reviewable.io/reviews/qtox/qtox/6235#-MRVLtTr8O0DW4qdrlxf:-MSOdHCd-U3jLsTa9yZh:b-wrj5wt) ([raw file](https://github.com/qtox/qtox/blob/21d356baf7a8e9ad1bf94a23f2c61d2422df3a00/src/core/coreav.cpp#L516)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

```calls``` is of type: ```std::map<uint32_t, ToxFriendCall*>```, need to use ```call.get()``` here?
</blockquote></details>

Oops yeah, missed in a commit split I think. Thanks.
___
*[src/widget/form/chatform.cpp, line 285 at r8](https://reviewable.io/reviews/qtox/qtox/6235#-MRWy-YF1Yil4J5r9ow6:-MSOgRKT22SBTEojdS_H:b1dmpbd) ([raw file](https://github.com/qtox/qtox/blob/21d356baf7a8e9ad1bf94a23f2c61d2422df3a00/src/widget/form/chatform.cpp#L285)):*

Minor thing: do you expect to run into this regularly? Else it should be turned into an assert IMO
</blockquote></details>

Agreed, changed :+1:

sudden6:
> Thoughts?

To me this seems a bit over engineered for what we use it now and I fear we'll never use the functionality of the extra manager class to the extent we really need it. On the other hand, making one class do a single thing makes sense and I'd rather agree to the over engineered solution now and get it in in a reasonable timeframe than to drag this issue to infinity.

bodwok:
`ToxCallManager` reduces `CoreAV`'s responsibility which is good. 
In my opinion, the following behavior can help distribute functionality more evenly. Although more changes to the code are required: 
`CoreAV` - passing events from Tox API to `ToxCallManager`.
`ChatForm` - passing events from user to `ToxCallManager`.
`ToxCallManager` - owns and processes `ToxCall` instances, provides an interface to `CoreAV` and `ChatForm`.

All functions that process `ToxCall` (such as `isCallStarted`, `isCallActive`, `toggleMuteCallInput`, etc.) will be transferred to `ToxCallManager`. 

What do you think?

sudden6:
> All functions that process ToxCall (such as isCallStarted, isCallActive, toggleMuteCallInput, etc.) will be transferred to ToxCallManager.

This is exactly the opposite of what I was trying to do here 😅

bodwok:
:smile: OK, just as one of the possible updates.

PR #6244, opened 2020-10-26
Update chathistory.h

jjc98375:
Add the forward declaration of struct Message in chathistory.h header file

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6244)

### DIFF
--- a/src/model/chathistory.h
+++ b/src/model/chathistory.h
@@ -26,6 +26,7 @@
 #include <QSet>
 
 class Settings;
+struct Message;
 
 class ChatHistory : public IChatLog
 {


powerjungle:
What is being fixed with this commit?

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6244)* status: 0 of 1 LGTMs obtained

___
*[src/model/chathistory.h, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/6244#-MSSdDz76PcBnwtfeaOi:-MSSdDz76PcBnwtfeaOj:b-36e6vg) ([raw file](https://github.com/qtox/qtox/blob/69f3d7ffbe0a2da1397db338c017871b8dcd3cce/src/model/chathistory.h#L29)):*
> ```objc
> 
> class Settings;
> struct Message;
> ```

It's not _wrong_ to forward declare `Message` here, it is referenced in the header, but its full definition is already visible through `#include "ichatlog.h"` -> `#include "message.h"`. It's better that each file includes everything it needs directly rather than relying on chain includes, so this change is an improvement IMO, but using a tool like [include what you use](https://include-what-you-use.org/) to automatically do all this cleanup seems a lot more efficient than making a PR for each single line change. There are probably 100s of changes like this that could be made.

I'd rather not have PRs _this_ trivial, but if there was a mass cleanup done that included this e.g. with `include what you use` across the whole project I would be fine with that.

Going to close it, but if you'd like to make a mass cleanup of the same type feel free to open a new one.

PR #6247, opened 2020-11-06
chore: move OSX to use Brewfile

sudden6:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6247)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -152,6 +152,10 @@ jobs:
         ccache: true
         directories:
           - $HOME/Library/Caches/Homebrew
+      addons:
+        homebrew:
+          brewfile: ./osx/Brewfile
+          update: true
       before_cache:
         - brew cleanup
       script:
--- a/.travis/build-osx.sh
+++ b/.travis/build-osx.sh
@@ -20,18 +20,6 @@ set -eu -o pipefail
 
 readonly BIN_NAME="qTox.dmg"
 
-# accelerate builds with ccache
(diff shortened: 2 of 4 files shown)


sudden6:
@anthonybilinski this seems to fix the problems we have with OSX and Travis CI.

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6247#-:-MM3C1Ex0Lyrqa9-Cb0w:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[osx/Brewfile, line 18 at r1](https://reviewable.io/reviews/qtox/qtox/6247#-MM3AoaT-jMSBp01Fn5o:-MM3AoaU69AkvimJ2Lat:btpe00n) ([raw file](https://github.com/qtox/qtox/blob/44008d11bb9f4b6538025528e947ffa45467bb12/osx/Brewfile#L18)):*
> ```Ruby
> brew "openal-soft"
> # accelerate builds with ccache
> brew "ccache"
> ```

I guess this will no longer help with compilation of packages when their binary form isn't available like we were hitting a while ago, but we were only compiling big packages because our env was out of date, so that doesn't seem like something that we should optimize for. Overall this still seems much nicer, but if our deps start taking longer ccache not longer being installed before installing the rest could be something to look at.

PR #6250, opened 2020-11-14
feat(logging): check if current code is tagged

powerjungle:
This commit adds a new define called "GIT_DESCRIBE_EXACT" through cmake.
It is checked with a regex in "updatecheck.cpp" for a version number after the
check for new updates. If there is no version number,
a warning is output to log.

The reason for the new define is to avoid doing too much regex on
"GET_DESCRIBE", since "GIT_DESCRIBE_EXACT" will not contain a version number if
the code is not tagged.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6250)

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -214,6 +214,24 @@ add_definitions(
   -DGIT_VERSION="${GIT_VERSION}"
 )
 
+if (NOT GIT_DESCRIBE_EXACT)
+  execute_process(
+    COMMAND git describe --exact-match --tags HEAD
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE GIT_DESCRIBE_EXACT
+    ERROR_QUIET
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+  )
+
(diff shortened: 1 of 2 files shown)


sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6250)* status: 0 of 1 LGTMs obtained

___
*[src/net/updatecheck.cpp, line 32 at r1](https://reviewable.io/reviews/qtox/qtox/6250#-MMAy_jA09219rc1MWlL:-MMAy_jA09219rc1MWlM:bo9k2o3) ([raw file](https://github.com/qtox/qtox/blob/90cc962802a5909e0692d5a6f84ac6808803fc28/src/net/updatecheck.cpp#L32)):*
> ```cpp
> #include <cassert>
> 
> #define VERSION_REGEX_STRING   "v([0-9]+)\\.([0-9]+)\\.([0-9]+)"
> ```

please change this to a constant in the anonymous namespace (see `versionUrl`) below

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6250#-:-MMSZPqNEjINzDWA7nsE:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

Issue #5700, opened 2019-06-16
Feature Request: Integration with password managers


za3k:
I use a password manager and I'd love an option to integrate it with qtox.

Typically how this works is that qtox would run some arbitrary command (ex `pass tox` or gpg in my case), and use the output of the command as the password. 

The usual way I've seen to do this is to put the command as an option in the tox config file. Running an arbitrary command IMO is not that bad, if you can change a user's config files you can probably already get them to run arbitrary commands easier ways.

Or, if that still tastes bad, qtox could accept a command-line PASSWORD environment file, and people can just make wrappers or aliases

sudden6:
https://github.com/qTox/qTox/issues/4136 seems to also be relevant for this.

IMO executing arbitrary commands to retrieve the password sounds not like a good idea, but I may be wrong here.

Passing the password via environment variable sounds like a good and not too difficult method though.

antony-jr:
We could use a public-key cryptography such as RSA then use the public key to encrypt the qTox password and hand over the burden of keeping the private key to the user. (i.e) They can keep the private key local and have no password for it.

sudden6:
@antony-jr I don't see how that solves the problem of integrating with password managers or OS provided keychains.

Overall I think integrating with the OS provided key chain is the best and most useful option to most users. I don't know how that would work on Windows though.

antony-jr:
Oh... I thought we could do something like how git uses SSH for authentication instead of asking passwords.

antony-jr:
One possible solution for linux would be https://www.passwordstore.org/ but this all seems too much work with too little time.

sudden6:
@antony-jr there was https://github.com/qTox/qTox/pull/3332 as an attempt to implement this, but it was never finished.

za3k:
On 2019-07-30 19:15, Antony jr wrote:
> One possible solution for linux would be
> https://www.passwordstore.org/ but this all seems too much work with
> too little time.

Antony, can you clarify what are you recommending here and how you think 
this would fix some problem? This is a password manager that works with 
linux (and if fact the one I use). Having something like a password 
manager is already part of the assumption in the feature request--the 
entire feature request is to reduce the work-to-time ratio :).

za3k:
sudden6, that seems probably true, but personally I do not have an 'os keychain' on my os, so I'd have a feature request either way--if you want to add keychain integration would it be rude to ask you to open a second issue instead? If you want to close this one as low-pri that's ok.

[...]

PR #6283, opened 2021-01-17
feat(main): add --password command line option

carlor:
Add a command line option for a profile's password, intended to make
integrating with password managers easy.

For example, a user could automatically run:

    $ qtox -p myprofile --password $(pass qtox/myprofile)

Related to #5700

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6283)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -238,6 +238,10 @@ int main(int argc, char* argv[])
                            QObject::tr("Starts new instance and loads specified profile."),
                            QObject::tr("profile")));
     parser.addOption(
+        QCommandLineOption(QStringList() << "password",
+                           QObject::tr("Password for encrypted profile."),
+                           QObject::tr("password")));
+    parser.addOption(
         QCommandLineOption(QStringList() << "l"
                                          << "login",
                            QObject::tr("Starts new instance and opens the login screen.")));
@@ -321,7 +325,7 @@ int main(int argc, char* argv[])
 
     qDebug() << "commit: " << GIT_VERSION;
 
-    QString profileName;
+    QString profileName, profilePassword;
     bool autoLogin = settings.getAutoLogin();
 
     uint32_t ipcDest = 0;
@@ -337,6 +341,9 @@ int main(int argc, char* argv[])
         } else {
             ipcDest = Settings::makeProfileId(profileName);
             autoLogin = true;
+            if (parser.isSet("password")) {
+                profilePassword = parser.value("password");
+            }
         }
     } else if (parser.isSet("l")) {
         doIpc = false;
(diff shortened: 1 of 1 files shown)


anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6283)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6283#-MRGw0yjATeU49ZA3lj3:-MRGw0yjATeU49ZA3lj4:bcnnkxt) (no related file):*
I don't think having passwords passed in by command line is a good idea. Generally reading passes from stdin rather than as an argument is better since it both won't show up in `history`, and won't be visible with `ps` for the whole lifetime of the process.

Would the `--password` argument to cause qTox to block on stdin with no echo for a password to be entered in that way be as useful with the password manager workflow?

---
___
*[src/main.cpp, line 244 at r1](https://reviewable.io/reviews/qtox/qtox/6283#-MRGxwM0DZ8AmNhO7Vl7:-MRGxwM0DZ8AmNhO7Vl8:b3dtms2) ([raw file](https://github.com/qtox/qtox/blob/f924a3731722cbc978e31e84d182179bc696c37a/src/main.cpp#L244)):*
> ```cpp
>                            QObject::tr("Password for encrypted profile."),
>                            QObject::tr("password")));
>     parser.addOption(
> ```

I didn't consider this at first, but on second thought I think it's good that `password` can be set without a profile being set, since it would work in the case of auto-login being enabled and the last profile loaded being the intended one, which is probably common for single-profile uses.
___
*[src/main.cpp, line 397 at r1](https://reviewable.io/reviews/qtox/qtox/6283#-MRGvEbpC3l1MoGrYcRP:-MRGvEbpC3l1MoGrYcRQ:b-dm2bwi) ([raw file](https://github.com/qtox/qtox/blob/ef47c00c8d2e9ff97f4d63511f6f8af38528d2ea/src/main.cpp#L397)):*
> ```cpp
>     Profile* profile = nullptr;
>     if (autoLogin && Profile::exists(profileName) && !Profile::isEncrypted(profileName)) {
>         profile = Profile::loadProfile(profileName, QString(), settings, &parser);
> ```

In case this is changed to a `stdin` reading method,

If no password is passed in but the profile is encrypted, this `if` block will now be entered and the error will be shown, but the login screen for encrypted profiles doesn't have the check box to enable/disable auto load (it's only present if an unencrypted profile is selected from the login screen). This means that if a user just e.g. double clicks on the icon, they might get hit with the error pop up on every start with no obvious way to stop it, instead of just being prompted with the login screen like before.

I think if no `password` is set, this `if` block should be skipped like before, but can be entered and possibly error like now as long as one is set.

PR #5932, opened 2019-11-17
Multipart messages

sphaerophoria:
# Multi-packet messages support

After months of talking about it I've finally taken the time to sit down and clean up the qTox integration of the toxext library to provide multi-packet messages support. Thanks @anthonybilinski  for finally pushing me to sit down and get it done.

Toxcore has a limitation in that any messages bigger than a single UDP packet are not sendable. This is not fixable without adding more handshaking to the protocol which would break backwards compatibility with older toxcores. Unfortunately there is no versioning in the toxcore protocol so until they ship their first breaking change we cannot add any features.

For this reason I wrote a library ToxExt. ToxExt is a general purpose library that implements negotiation of extra features over the toxcore custom packet concept. This allows us to do a couple things
1. Implement new packet level features on top of the toxcore protocol without breaking compatibility with exsiting clients
2. Implement features a little more aggressively. Since all extensions are known to be optional if we find a bug in the feature we can just disable it with little harm done.

This PR implements the first of what are hopefully many more extensions to come.

For more info on the design of ToxExt please see https://github.com/toxext/toxext/blob/master/DESIGN.md

The message extension works by marking sequential packets with a "start"/"part"/"finish" marker. The extension library automatically splits a long message into a "start" packet, several "part" packets, and a "finish" packet. When packets are received by an extension they are cached in the library until a "finish" packet has been seen. When we see a "finish" packet qTox receives the full message and dispatches it through the normal channels.

This PR covers a lot of different areas. I feel like the feature would be unfinished/confusing to users without the following.

* UI element indicating a friend's extension support. This can be green/yellow/red and on hover will show which extensions are supported
* Friends now have a "Negotiating" state. When a friend goes online we try to send them a toxext packet to ask them what extensions they have. If they have toxext on the other end this should be quite quick. If they don't we wait one second before deciding that they won't answer us. In this state we treat them as offline (which I will explain in the next point).
* Offline messages in combination with extensions gets a little confusing. What should the behavior be if I send an extended message, my friend goes offline, and then comes back online without support for the extension? I've decided if the extension set is reduced we fail all messages currently in the offline message queue. This is because sometimes meaning of messages are affected by the ones around it. I wouldn't want to deliver some messages without their surrounding context
    * Since we can have multiple reasons for a message to not be delivered I've added a "reason" field to the broken message concept
[...]

Issue #5812, opened 2019-08-31
Loading chat history is slow and blocks starting qTox


anthonybilinski:
##### Brief Description

OS: Ubuntu 18.04
qTox version: v1.16.3+
Commit hash:  2bba12175ee34b835c3888c391b3169f56d36bf7
toxcore:  v0.2.10
Qt:  5.12.2
Hardware:  Desktop, i7-7700, SSD

Reproducible: Always

##### Steps to reproduce

1. Unencrypted profile (it's not sqlcipher).
2. Have lots (~a couple years of daily use worth) of messages and lots (~45) of friends.
3. Open qTox

##### Observed Behavior
1. After login screen, there's about 20s before the main qTox window appears.
2. Digging into RawDatabase and timing each step, I see, for a single friend 287ms spent in SQL queries. Primarily:
```
SELECT COUNT(history.id) FROM history JOIN peers chat ON chat_id = chat.id WHERE chat.public_key='<censored>'; took 32 ms
SELECT history.id, faux_offline_pending.id, timestamp, chat.public_key, aliases.display_name, sender.public_key, message FROM history JOIN faux_offline_pending ON history.id = faux_offline_pending.id JOIN peers chat on history.chat_id = chat.id JOIN aliases on sender_alias = aliases.id JOIN peers sender on aliases.owner = sender.id WHERE chat.public_key='<censored>'; took 32 ms
SELECT history.id, faux_offline_pending.id, timestamp, chat.public_key, aliases.display_name, sender.public_key, message, file_transfers.file_restart_id, file_transfers.file_path, file_transfers.file_name, file_transfers.file_size, file_transfers.direction, file_transfers.file_state FROM history LEFT JOIN faux_offline_pending ON history.id = faux_offline_pending.id JOIN peers chat ON history.chat_id = chat.id JOIN aliases ON sender_alias = aliases.id JOIN peers sender ON aliases.owner = sender.id LEFT JOIN file_transfers ON history.file_id = file_transfers.id WHERE chat.public_key='<censored>' LIMIT 100 OFFSET 61267; took 109 ms
[...]

PR #6290, opened 2021-01-26
feat(chatlog): Ugrade db schema to support system messages

### LINKED ISSUE #6221: Update history table in db to handle system messages, in addition to chat and file messages
##### Brief Description

"System messages" here is meant to refer to messages that qTox adds to the chatlog, but that weren't sent by either friend. Things like "Bob is calling...", "Call ended after 1h35m", "Bob is now away." etc.

As part of https://github.com/qTox/qTox/projects/5, to allow:
1) A unique reference to any individual chat message in the chatlog. Currently messages are identified in `SessionChatLog` by the `History` index now, but that can't accommodate system messages, which chatlog still needs to handle.
2) Users scrolling to dynamically remove system messages from the chatlog, and also dynamically load them in if they scroll back down, the system messages need to be persisted.
3) To allow UI changes like "Show contacts' status changes" to take immediate effect for the viewed chatlog, by saving system messages even if they aren't displayed.

The history should store system messages, like it does chat and file messages already.

Current the `history` table has all the fields necessary for a chat message, then an optional foreign reference to an id in the `file_transfers` table. Since system messages have quite different requirements from normal chat messages, e.g. no author, and rather than message content just an enum could be saved for which type of system message it was, I think a more flexible design would be:

1. move chat messages to new table, same schema as now. id is a foreign key to history
1. make new system messages table, similar to history but without author
1. update file transfer table to include needed fields that were previously from `history`.
1. make `history` table just an `id` + `type` + `foreign key` that references a row in one of the other 3 tables.

This would allow each of the three sub-history tables to have all of their fields present, and would allow the master history table to still enforce ordering between messages. It should be verified that this wouldn't majorly impact performance and that this is actually a sane schema design.

sphaerophoria:
feat(chatlog): Ugrade db schema to support system messages

* Resolves #6221
* System message schema based off of ChatMessage::createChatInfoMessage
* New table layout required many updates to the queries executed by
  history
* Bonus reduction of history signals/slots by issuing some file transfer
  insertions directly when possible

NOTE: this is on top of PR #5932 so only review the last commit. This is because of the history changes in #5932. @anthonybilinski mentioned in #6221 that this should come on top of #5932.

I'm going to attempt to summarize the changes and how I tested them, more rationale about why changes were made can be found in #6221 

* DB schema upgrade
    * Tested with db schema upgrade test
    * Tested a real world case updating from v1.17.3 (db version 4) to the new DB. I opened the new DB in sqlitebrowser and verified that all tables had the data I expected
* history changes for inserting/retrieving text/file messages
    * Tested sending/receiving files. Ensured no new error logs. Tested around broken messages and faux offline messages as well. I didn't see any dropped messages
* Search query changes to handle new text_messages table
    * Tested search on chat with > 200 messages to ensure it had to search backwards in history. Search text was found successfully

I think I've done my due diligence here. If I were a better developer I would have added some tests to ensure that we have the correct foreign key constraints... but I didn't. Manual tests of the sqlite changes are shown in #6221 

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6290)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -27,7 +27,7 @@
 #include "src/core/toxpk.h"
 
 namespace {
-static constexpr int SCHEMA_VERSION = 6;
+static constexpr int SCHEMA_VERSION = 7;
 
 bool createCurrentSchema(RawDatabase& db)
 {
@@ -42,8 +42,16 @@ bool createCurrentSchema(RawDatabase& db)
         "FOREIGN KEY (owner) REFERENCES peers(id));"
         "CREATE TABLE history "
         "(id INTEGER PRIMARY KEY, "
+        "message_type CHAR(1) NOT NULL DEFAULT 'T' CHECK (message_type in ('T','F','S')), "
         "timestamp INTEGER NOT NULL, "
         "chat_id INTEGER NOT NULL, "
+        // Message subtypes want to reference the following as a foreign key. Foreign keys must be
+        // guaranteed to be unique. Since an ID is already unique, id + message type is also unique
+        "UNIQUE (id, message_type), "
+        "FOREIGN KEY (chat_id) REFERENCES peers(id)); "
+        "CREATE TABLE text_messages "
+        "(id INTEGER PRIMARY KEY, "
+        "message_type CHAR(1) NOT NULL CHECK (message_type = 'T'), "
         "sender_alias INTEGER NOT NULL, "
         // even though technically a message can be null for file transfer, we've opted
         // to just insert an empty string when there's no content, this moderately simplifies
@@ -52,20 +60,30 @@ bool createCurrentSchema(RawDatabase& db)
         // ensure that our blob vector always has the right number of fields. Better to just
         // leave this as NOT NULL for now.
         "message BLOB NOT NULL, "
-        "file_id INTEGER, "
-        "FOREIGN KEY (file_id) REFERENCES file_transfers(id), "
-        "FOREIGN KEY (chat_id) REFERENCES peers(id), "
-        "FOREIGN KEY (sender_alias) REFERENCES aliases(id));"
+        "FOREIGN KEY (id, message_type) REFERENCES history(id, message_type), "
+        "FOREIGN KEY (sender_alias) REFERENCES aliases(id)); "
         "CREATE TABLE file_transfers "
         "(id INTEGER PRIMARY KEY, "
-        "chat_id INTEGER NOT NULL, "
+        "message_type CHAR(1) NOT NULL CHECK (message_type = 'F'), "
+        "sender_alias INTEGER NOT NULL, "
         "file_restart_id BLOB NOT NULL, "
         "file_name BLOB NOT NULL, "
         "file_path BLOB NOT NULL, "
         "file_hash BLOB NOT NULL, "
         "file_size INTEGER NOT NULL, "
         "direction INTEGER NOT NULL, "
-        "file_state INTEGER NOT NULL);"
+        "file_state INTEGER NOT NULL, "
+        "FOREIGN KEY (id, message_type) REFERENCES history(id, message_type), "
+        "FOREIGN KEY (sender_alias) REFERENCES aliases(id)); "
+        "CREATE TABLE system_messages "
+        "(id INTEGER PRIMARY KEY, "
+        "message_type CHAR(1) NOT NULL CHECK (message_type = 'S'), "
+        "system_message_type INTEGER NOT NULL, "
+        "arg1 BLOB, "
+        "arg2 BLOB, "
+        "arg3 BLOB, "
+        "arg4 BLOB, "
+        "FOREIGN KEY (id, message_type) REFERENCES history(id, message_type)); "
         "CREATE TABLE faux_offline_pending (id INTEGER PRIMARY KEY, "
         "required_extensions INTEGER NOT NULL DEFAULT 0, "
         "FOREIGN KEY (id) REFERENCES history(id));"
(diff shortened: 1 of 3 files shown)


sphaerophoria:
After thinking about this a little more and talking with @anthonybilinski I think the schema implemented for system messages isn't good enough. Initially I was thinking that just dropping the string in there would be sufficient, but there are some cases where that doesn't work very well.

1. What if we want to update the system messages later, we'd have to scour the DB for the old version and update it to the new one to ensure consistency. This means we'd have to keep the old versions of the system messages around forever
2. If a user changes their language we cannot re-translate the strings with the current schema.

This gets a little hairy too with arguments in the system message. We can't even re-translate once the arguments have been applied. This means our table needs to have the arguments stored separately from the main message string itself

This actually forces us to have better practices anyways, why store system messages as strings when we don't have to?

After thinking about it a little I think we can get away with a schema of 

`CREATE TABLE system_messages (id INTEGER PRIMARY KEY,  message_type CHAR(1) NOT NULL CHECK (message_type = 'S'), system_message_type INTEGER NOT NULL, arg1 BLOB, arg2 BLOB, arg3 BLOB, arg4 BLOB)`

In this schema system_message_type will be an integer referring to some lookup table of int -> translation string. Then we just apply the arguments and hope we never have a system message with more than 4 arguments. If we do end up having more than 4 in the future we can always add more nullable fields to the table.

Chiitoo:
Probably meant 'Upgrade' here?  :]

```
feat(chatlog): Ugrade db schema to support system messages
```

sphaerophoria:
Proof of concept hook up with chatform shown in https://github.com/qTox/qTox/issues/6222#issuecomment-768054724 proving that the schema is good enough.

anthonybilinski:
Could you rebase now that #5932  is in? Also it looks like `dbschema_test` is failing.

sphaerophoria:
> Could you rebase now that #5932 is in? Also it looks like `dbschema_test` is failing.

I've rebased, but I don't see the dbschema failure locally.

```
└> ./test_dbschema
********* Start testing of TestDbSchema *********
Config: Using QtTest library 5.15.2, Qt 5.15.2 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 10.2.0), arch unknown
PASS   : TestDbSchema::initTestCase()
PASS   : TestDbSchema::testCreation()
PASS   : TestDbSchema::testIsNewDb()
PASS   : TestDbSchema::test0to1()
PASS   : TestDbSchema::test1to2()
PASS   : TestDbSchema::test2to3()
PASS   : TestDbSchema::test3to4()
PASS   : TestDbSchema::test4to5()
PASS   : TestDbSchema::test5to6()
PASS   : TestDbSchema::test6to7()
PASS   : TestDbSchema::cleanupTestCase()
Totals: 11 passed, 0 failed, 0 skipped, 0 blacklisted, 447ms
********* Finished testing of TestDbSchema *********
```

Will check travis after it runs with the rebase.

sphaerophoria:
Failed again. I'm willing to bet it has something to do with sqlite version... I'll try to repro with an older one

sphaerophoria:
Ah nice found the issue. It looks like the foreign key references are not being updated when we use ALTER TABLE to move a table on older sqlite versions. See section 2 of https://www.sqlite.org/lang_altertable.html for further information. It looks like the solution in section 5 of https://www.sqlite.org/lang_altertable.html may help us... I'm hoping that FOREIGN KEYs are based off one of the items they listed....

sphaerophoria:
Patched up the upgrade path for the old alter table behavior. It isn't pretty, but it's the best I could come up with. If you find a better suggestion feel free to let me know.

EDIT: I was confused yesterday and forgot that the dbschema upgrade test doesn't actually test any real conversion. Testing manually on a real profile shows that this change isn't sufficient. I'll figure it out later today

sphaerophoria:
> Ah nice found the issue. It looks like the foreign key references are not being updated when we use ALTER TABLE to move a table on older sqlite versions. See section 2 of https://www.sqlite.org/lang_altertable.html for further information. It looks like the solution in section 5 of https://www.sqlite.org/lang_altertable.html may help us... I'm hoping that FOREIGN KEYs are based off one of the items they listed....

On second read I realize I misread last time... the reason the test was failing is because pragma foreign_key wasn't enabled in our test case but it was in our real upgrade path. All I had to do was enable it before running the upgrade code I had the first time round. I ran the tests on ubuntu 16.04 with an old sqlite and it seems to do the job now

Re-tested real sqlite upgrade on ubuntu 16.04 and a current archlinux (new sqlite and old sqlite). Seems good enough for me!

sphaerophoria:
Just rebased to master, no changes.

sudden6 (commented):
Sorry for the very long delay, I found time to review, everything :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6290#-:-MYfSAQ-4a-lgujWHPHn:b6yqm7t)* status: :shipit: complete! 1 of 1 LGTMs obtained

sudden6:
@sphaerophoria could you rebase on latest master?

anthonybilinski (changes requested):
Also sorry for the long delay, looks like a really well done change. Better than I imagined in #6221.

For reference the upgrade which does a lot of data shuffling took under 2s (on slow SSD) on my main profile, which has years of history and takes about 40s to load normally (damn #5812), so it doesn't seem concerning even for near worst case real world data sets.

:lgtm_strong: 

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6290#-:-MZgZRH_9bGp4CRTAbL5:b-2150t4)* status: :shipit: complete! 2 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 373 at r1](https://reviewable.io/reviews/qtox/qtox/6290#-MZgWWyY48VdVLKD6A2-:-MZgWWyZ9Gwpqce0R456:b-qap27) ([raw file](https://github.com/qtox/qtox/blob/6d80e9c04e560256481b422dc768df3f0f9c6434/src/persistence/history.cpp#L373)):*
> ```cpp
>         RawDatabase::Query("ALTER TABLE faux_offline_pending_new RENAME TO faux_offline_pending");
> 
>     // broken_messages needs to be re-created to reference the new history tablek
> ```

"tablek" typo, will fix myself

anthonybilinski (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6290)* status: :shipit: complete! 2 of 1 LGTMs obtained

Issue #5147, opened 2018-05-17
Flag emoji are incorrectly displayed


sudden6:
##### Brief Description

OS: openSUSE Leap 42.3
qTox version: v1.15.0-45-g1f58e1c0
Commit hash: 1f58e1c0af1f216431528cce3bc787afc481de38
toxcore: 0.2.2
Qt: 5.6.2
Hardware: 
…

Reproducible: Always

##### Steps to reproduce

1. open chat
2. enter the "german flag emoji" 🇩🇪 via emoji picker
3. press "Send" button

##### Observed Behavior

It's displayed as flag in the input box, but if you send it, it becomes the "D emoji" and "E emoji"

##### Expected Behavior

"german flag emoji" is displayed in the chat log

##### Additional Info
"D emoji" and "E emoji"
![spectacle d12727](https://user-images.githubusercontent.com/5585762/40205788-bc70bf9c-5a2d-11e8-8ecd-5cc519ae1d12.png)

happens with all flag emojis

tox-user:
Confirmed. It's been like that since 1.14.0 or 1.14.1.

sudden6:
maybe has something to do with the recent changes around multi byte emois by @anthonybilinski ?

tox-user:
It most likely does.

anthonybilinski:
Yes it does, introduced in 9643e48ef1d68948d52feec4e1be28c3ad61c0da. Parsing logic needs to be updated to favour multi-Ucs4-length emoji (flags) which are made up of multiple single-length valid emoji, i.e. the [US flag] is made up of the emoji [U] and [S] next to each other.

anthonybilinski:
Fixed by https://github.com/qTox/qTox/pull/6302.

PR #6306, opened 2021-02-21
test(persistence): add smileypack unit test

warisb:
A follow-up of issue #5147, add a unit test which ensures the same bug won't come up again.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6306)

### DIFF
--- a/cmake/Testing.cmake
+++ b/cmake/Testing.cmake
@@ -34,15 +34,16 @@ function(auto_test subsystem module extra_deps)
     COMMAND ${TEST_CROSSCOMPILING_EMULATOR} test_${module})
 endfunction()
 
-auto_test(core core ${${PROJECT_NAME}_RESOURCES})
+auto_test(core core "${${PROJECT_NAME}_RESOURCES}")
 auto_test(core contactid "")
 auto_test(core toxid "")
 auto_test(core toxstring "")
 auto_test(chatlog textformatter "")
-auto_test(net bsu ${${PROJECT_NAME}_RESOURCES}) # needs nodes list
+auto_test(net bsu "${${PROJECT_NAME}_RESOURCES}") # needs nodes list
 auto_test(persistence paths "")
 auto_test(persistence dbschema "")
 auto_test(persistence offlinemsgengine "")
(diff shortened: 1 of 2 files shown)


cmake/Testing.cmake
```diff
@@ -34,15 +34,16 @@ function(auto_test subsystem module extra_deps)
...
 auto_test(core contactid "")
 auto_test(core toxid "")
 auto_test(core toxstring "")
 auto_test(chatlog textformatter "")
-auto_test(net bsu ${${PROJECT_NAME}_RESOURCES}) # needs nodes list
+auto_test(net bsu "${${PROJECT_NAME}_RESOURCES}") # needs nodes list
 auto_test(persistence paths "")
 auto_test(persistence dbschema "")
 auto_test(persistence offlinemsgengine "")
+auto_test(persistence smileypack "${${PROJECT_NAME}_RESOURCES}") # need smilies qrc
```
warisb: Note that I found an existing bug in this file: when you pass a list into a cmake function, if you didn't quotate ("") it, it will only take the first entry. This was not noticeable until now since we only needed the first entry in prior usages.

sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6306#-:-MUdrqlZDnzWu9XLUkSL:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[cmake/Testing.cmake, line 46 at r1](https://reviewable.io/reviews/qtox/qtox/6306#-MU0yi8977r-aF41_suk-r1-46:-MUdrhza8Trn6Zfe7b7a:btiqo5c) ([raw file](https://github.com/qtox/qtox/blob/0af1128f49212a18ea2f7114497d2b57d59d9dc0/cmake/Testing.cmake#L46)):*
<details><summary><i>Previously, warisb (Waris) wrote…</i></summary><blockquote>

Note that I found an existing bug in this file: when you pass a list into a cmake function, if you didn't quotate ("") it, it will only take the first entry. This was not noticeable until now since we only needed the first entry in prior usages.
</blockquote></details>

Seems unintuitive and should probably be fixed.

PR #6310, opened 2021-02-28
feat(chatlog): Add SystemMessages to SessionChatLog

sphaerophoria:
* Rendering of system messages consolidated into single place
* API added to ichatlog to insert a system message at current location
* System messages are now used as type + args which will fit nicely with
  the work in #6221

This is the work shown in https://github.com/qTox/qTox/issues/6222#issuecomment-768054724 but without the changes to history. I've split the work so that this can be reviewed without waiting for #6290 to complete. Integration with #6290 will come in a separate PR

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)

### DIFF
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -212,6 +212,13 @@ std::vector<IChatLog::DateChatLogIdxPair> ChatHistory::getDateIdxs(const QDate& 
     }
 }
 
+void ChatHistory::addSystemMessage(const SystemMessage& message)
+{
+    // FIXME: #6221 Insert into history
+    sessionChatLog.addSystemMessage(message);
+}
+
+
 void ChatHistory::onFileUpdated(const ToxPk& sender, const ToxFile& file)
 {
     if (canUseHistory()) {
--- a/src/model/chathistory.h
+++ b/src/model/chathistory.h
@@ -42,11 +42,13 @@ public:
     ChatLogIdx getFirstIdx() const override;
     ChatLogIdx getNextIdx() const override;
     std::vector<DateChatLogIdxPair> getDateIdxs(const QDate& startDate, size_t maxDates) const override;
+    void addSystemMessage(const SystemMessage& message) override;
 
 public slots:
     void onFileUpdated(const ToxPk& sender, const ToxFile& file);
     void onFileTransferRemotePausedUnpaused(const ToxPk& sender, const ToxFile& file, bool paused);
     void onFileTransferBrokenUnbroken(const ToxPk& sender, const ToxFile& file, bool broken);
+
 
 private slots:
     void onMessageReceived(const ToxPk& sender, const Message& message);
--- a/src/model/chatlogitem.cpp
+++ b/src/model/chatlogitem.cpp
@@ -50,6 +50,11 @@ ChatLogItem::ChatLogItem(ToxPk sender_, const QString& displayName, ChatLogMessa
                              ChatLogItemDeleter<ChatLogMessage>::doDelete))
 {}
 
+ChatLogItem::ChatLogItem(SystemMessage systemMessage)
+    : contentType(ContentType::systemMessage)
+    , content(new SystemMessage(std::move(systemMessage)), ChatLogItemDeleter<SystemMessage>::doDelete)
+{}
+
 ChatLogItem::ChatLogItem(ToxPk sender_, const QString& displayName_, ContentType contentType_, ContentPtr content_)
     : sender(std::move(sender_))
     , displayName(displayName_)
@@ -91,6 +96,19 @@ const ChatLogMessage& ChatLogItem::getContentAsMessage() const
     return *static_cast<ChatLogMessage*>(content.get());
 }
 
+SystemMessage& ChatLogItem::getContentAsSystemMessage()
+{
+    assert(contentType == ContentType::systemMessage);
+    return *static_cast<SystemMessage*>(content.get());
+}
+
+const SystemMessage& ChatLogItem::getContentAsSystemMessage() const
+{
+    assert(contentType == ContentType::systemMessage);
+    return *static_cast<SystemMessage*>(content.get());
+}
+
+
 QDateTime ChatLogItem::getTimestamp() const
 {
     switch (contentType) {
@@ -101,6 +119,10 @@ QDateTime ChatLogItem::getTimestamp() const
     case ChatLogItem::ContentType::fileTransfer: {
         const auto& file = getContentAsFile();
         return file.timestamp;
+    }
+    case ChatLogItem::ContentType::systemMessage: {
+        const auto& systemMessage = getContentAsSystemMessage();
+        return systemMessage.timestamp;
     }
     }
 
--- a/src/model/chatlogitem.h
+++ b/src/model/chatlogitem.h
@@ -22,6 +22,7 @@
 #include "src/core/toxfile.h"
 #include "src/core/toxpk.h"
 #include "src/model/message.h"
+#include "src/model/systemmessage.h"
 #include "src/persistence/history.h"
 
 #include <memory>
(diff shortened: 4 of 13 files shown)


anthonybilinski (changes requested):
> Reviewed 13 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/model/chatlogitem.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyCK126DF-pgi-0s9J:-MkyCK126DF-pgi-0s9K:b-13v1) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/model/chatlogitem.cpp#L124)):*
> ```cpp
>     }
>     case ChatLogItem::ContentType::systemMessage: {
>         return QDateTime::currentDateTime();
> ```

Why use currentDateTime here? I think either session messages should store a datetime of just whenever they were initially added, or make make getTimestamp return an optional? It seems like it would be confusing if you export some 1 year old chat and all the system messages have today's date
___
*[src/widget/form/genericchatform.cpp, line 620 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyEvQvE7ZVl8gTf-u1:-MkyEvQvE7ZVl8gTf-u2:bgsajq0) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/widget/form/genericchatform.cpp#L620)):*
> ```cpp
> }
> 
> void GenericChatForm::addSystemInfoMessage(const QDateTime& datetime, SystemMessageType messageType,
> ```

datetime argument is unused (though maybe should be used rather than removed as mentioned above?)

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/model/chatlogitem.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyCK126DF-pgi-0s9J:-MkyU7v56AKEVjGMVyoe:b70ouza) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/model/chatlogitem.cpp#L124)):*

Why use currentDateTime here? I think either session messages should store a datetime of just whenever they were initially added, or make make getTimestamp return an optional? It seems like it would be confusing if you export some 1 year old chat and all the system messages have today's date
</blockquote></details>

Reviewing #6311, it looks like the timestamp is correctly saved and restored from history, but then not carried into the chatlogitem from `ChatHistory::loadHistoryIntoSessionChatLog`. Seems like might as well include it there for a more accurate chatlog?

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/model/chatlogitem.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyCK126DF-pgi-0s9J:-MkzKXlZ1AvFgr6-aJGs:buad0l5) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/model/chatlogitem.cpp#L124)):*

Reviewing #6311, it looks like the timestamp is correctly saved and restored from history, but then not carried into the chatlogitem from `ChatHistory::loadHistoryIntoSessionChatLog`. Seems like might as well include it there for a more accurate chatlog?
</blockquote></details>

I think it's only because there was no reason to. System messages don't show with a timestamp so I didn't see any reason to go out of our way to carry them around. We can add them if you want, I just didn't see any reason to

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/widget/form/genericchatform.h, line 82 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MlGvjShAxaYXwKk-UAX:-MlGvjSi87fOprlNaYYc:bep0a6m) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/widget/form/genericchatform.h#L82)):*
> ```objc
>                               SystemMessage::Args messageArgs);
>     static QString resolveToxPk(const ToxPk& pk);
>     QDateTime getLatestTime() const;
> ```

This causes I think an undesirable interaction with getLatestTime now counting all system messages. Activity based sorting relies on the latest chat message in the chatlog, which used to exclude system messages. So now the friend just coming online is enough to bump them into the "today" group, which doesn't seem desirable.

Including some of the system messages probably makes sense, like calls, but not things like status changes or online/offline. Maybe a new property should be added to them?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/model/chatlogitem.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyCK126DF-pgi-0s9J:-MlH0W348f3xVhMQIllC:blsbrg3) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/model/chatlogitem.cpp#L124)):*

I think it's only because there was no reason to. System messages don't show with a timestamp so I didn't see any reason to go out of our way to carry them around. We can add them if you want, I just didn't see any reason to
</blockquote></details>

Ah fair, thought this was still used in export but all system messages are skipped there. If that were to change, then including time could make sense even though it's not displayed, but for now I think keeping it as default constructed is fine.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

___
*[src/model/chatlogitem.cpp, line 124 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyCK126DF-pgi-0s9J:-Mm9utQv5uMj8ulT2J0a:bxs1s3r) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/model/chatlogitem.cpp#L124)):*

Ah fair, thought this was still used in export but all system messages are skipped there. If that were to change, then including time could make sense even though it's not displayed, but for now I think keeping it as default constructed is fine.
</blockquote></details>

See other comment, opted to add timestamp to system messages
___
*[src/widget/form/genericchatform.h, line 82 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MlGvjShAxaYXwKk-UAX:-Mm9uSLlDp_ZabxibDty:bee2hfb) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/widget/form/genericchatform.h#L82)):*

This causes I think an undesirable interaction with getLatestTime now counting all system messages. Activity based sorting relies on the latest chat message in the chatlog, which used to exclude system messages. So now the friend just coming online is enough to bump them into the "today" group, which doesn't seem desirable.

Including some of the system messages probably makes sense, like calls, but not things like status changes or online/offline. Maybe a new property should be added to them?
</blockquote></details>

Looks like this is only a problem when loading from history, so this only hurts us in combination with the system messages being loaded from history.

_that_ means that we have to store the correct timestamp in history, which means we should use the correct timestamp.

I've altered the behavior in GenericChatForm to ignore messages for certain system message types. No need to add a property I think, we can just filter based off what we think is relevant.
___
*[src/widget/form/genericchatform.cpp, line 620 at r1](https://reviewable.io/reviews/qtox/qtox/6310#-MkyEvQvE7ZVl8gTf-u1:-Mm9uR5GBgL-rL4Hz69C:bq6g5w6) ([raw file](https://github.com/qtox/qtox/blob/22063172af81ab757bcee881dc23e26e546d9266/src/widget/form/genericchatform.cpp#L620)):*

datetime argument is unused (though maybe should be used rather than removed as mentioned above?)
</blockquote></details>

Opted to use it

anthonybilinski (approved):
> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310)* status: 0 of 1 LGTMs obtained

anthonybilinski (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6310#-:-MmmHwc910mxLNtjcQwW:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6311, opened 2021-02-28
6222 save history

sphaerophoria:
*Requires #6290 and #6310*

Final piece tying the db schema upgrades and chatlog work together. Filing now even though it requires other PRs so I don't lose track of it. If reviewing before #6290 and #6310  please only review the final commit

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6311)

### DIFF
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -212,6 +212,13 @@ std::vector<IChatLog::DateChatLogIdxPair> ChatHistory::getDateIdxs(const QDate& 
     }
 }
 
+void ChatHistory::addSystemMessage(const SystemMessage& message)
+{
+    // FIXME: #6221 Insert into history
+    sessionChatLog.addSystemMessage(message);
+}
+
+
 void ChatHistory::onFileUpdated(const ToxPk& sender, const ToxFile& file)
 {
     if (canUseHistory()) {
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -214,7 +214,10 @@ std::vector<IChatLog::DateChatLogIdxPair> ChatHistory::getDateIdxs(const QDate& 
 
 void ChatHistory::addSystemMessage(const SystemMessage& message)
 {
-    // FIXME: #6221 Insert into history
+    if (canUseHistory()) {
+        history->addNewSystemMessage(f.getPublicKey(), message);
+    }
+
     sessionChatLog.addSystemMessage(message);
 }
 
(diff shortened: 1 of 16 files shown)


anthonybilinski (commented):
> Reviewed 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6311)* status: 0 of 1 LGTMs obtained

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6311)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 987 at r2](https://reviewable.io/reviews/qtox/qtox/6311#-MlGrRIY3Wky3_hTrVKR:-MlGrRIY3Wky3_hTrVKS:b-bzzj2l) ([raw file](https://github.com/qtox/qtox/blob/bde6dc0df0583a2c869612a010533bb6246ad7e2/src/persistence/history.cpp#L987)):*
> ```cpp
>         case 'S':
>             // System messages not yet supported
>             assert(false);
> ```

Not that we're expecting backwards compatibility, but this causes an assert on history load on older qTox versions, instead of a version check and controlled failure like normal. I think it would probably be worth bumping the SCHEMA_VERSION in this change to avoid trying at all to load history with old qTox when system messages have been added.

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6311)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 987 at r2](https://reviewable.io/reviews/qtox/qtox/6311#-MlGrRIY3Wky3_hTrVKR:-Mm9So_SBD0dfKtSqly0:b6p0h45) ([raw file](https://github.com/qtox/qtox/blob/bde6dc0df0583a2c869612a010533bb6246ad7e2/src/persistence/history.cpp#L987)):*

Not that we're expecting backwards compatibility, but this causes an assert on history load on older qTox versions, instead of a version check and controlled failure like normal. I think it would probably be worth bumping the SCHEMA_VERSION in this change to avoid trying at all to load history with old qTox when system messages have been added.
</blockquote></details>

Hm I guess that makes sense. The schema didn't change, but our use of it did. I can stomach that.

Confirmed that the schema upgrade prevents the failure by running the previous revision after upgrade.

anthonybilinski (commented):
> Reviewed 7 of 7 files at r3, 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6311)* status: 0 of 1 LGTMs obtained

anthonybilinski (commented):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6311#-:-MmmH59W9EO4vUcLchu2:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6317, opened 2021-03-06
chore(travis): add code coverage report generation

jwestell:
This change adds a cmake configuration switch to enable code coverage
instrumentation during the compilation of the project. When tests are
executed, the instrumentation outputs coverage data to output files in
the build directory. Programs such as lcov/gcovr can turn that data into
reports.

This change also adds steps to the travis CI configuration to build with
this configuration switch and then use lcov to generate the consolidated
report and publish to codecov.io

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6317)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,18 @@ jobs:
     - stage: Linux
       os: linux
       env: JOB=build-ubuntu-16-04
+      addons:
+        apt:
+          packages: 
+            - lcov
       script: "./.travis/$JOB.sh"
+      after_success:
+        # Create lcov report
+        - lcov --directory _debug --capture --output-file coverage.info
+        # Filter out system headers and test sources
+        - lcov --remove coverage.info '/usr/*' '*/test/*' '*/*_autogen/*' --output-file coverage.info
+        # Upload report to codecov.io 
+        - bash <(curl -s https://codecov.io/bash) -f coverage.info || echo "Codecov did not collect coverage reports"
     - stage: "Windows Stage 1: Dependencies (OpenSSL, Qt)"
       os: linux
       # Makes the cache this job creates avaiable only to jobs with WINDOWS_BUILD_ARCH_CACHE_TRICK_VARIABLE=i686,
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,12 +47,12 @@ jobs:
       env: JOB=build-ubuntu-16-04
       addons:
         apt:
-          packages: 
+          packages:
             - lcov
       script: "./.travis/$JOB.sh"
       after_success:
         # Create lcov report
-        - lcov --directory _debug --capture --output-file coverage.info
+        - lcov --directory _build --capture --output-file coverage.info
         # Filter out system headers and test sources
         - lcov --remove coverage.info '/usr/*' '*/test/*' '*/*_autogen/*' --output-file coverage.info
         # Upload report to codecov.io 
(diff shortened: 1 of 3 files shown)


jwestell:
Throwing this out there. Feedback needed on the CI workflow - probably don't want a third build just for code coverage generation, but didn't think the other builds should be affected by gcov instrumentation. To test, merge to local master and push to a fork then ensure travis is enabled on the fork.

sudden6 (commented):
I generally like the idea of adding code coverage reporting.

To keep the build times of the first stage fast (better feedback if the PR builds at all) I'd recommend you move the coverage build to the stage `macOS, AppImage and Flatpak` where the 10min build time won't matter.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6317#-:-MVwtP5XBXvz2APgsDIC:bmdpmyr)* status: 0 of 1 LGTMs obtained

anthonybilinski:
Testing this locally with the original version and a version with code coverage moved to stage 4 and removed from stage 1, and looking at just the longest job within a stage to get the stages time, since the reported stage time or total time includes waiting in queue for a VM to run each job, I see:

before, 15m43s stage 1: https://travis-ci.org/github/anthonybilinski/qTox/builds/763439622
after, 14m32s stage 1: https://travis-ci.org/github/anthonybilinski/qTox/builds/763442414

So we do see a _slight_ time saving in stage 1 by splitting it out, but we're adding an extra job which could mean extra time waiting in queue, and also we're adding 14m32s of wasted build time in stage 4. This doesn't actually increase stage 4's time at all since it's still bottle necked by macOS, but it still seems a bit wasteful.

Also re: faster feedback on PRs, I think even if code coverage is included in the stage 1 job, because compilation and unit tests both run before it, and code coverage only runs `after_success`, I think we would still get as fast notice of error with it in stage 1.

With all this, is it really worth moving to a new stage? It seems like the absolute benefit is about a 1 minute saving on a combined stage 1 and 4 build time of 48m32s, but in the process we're wasting about 14m32s of machine time + queue time.

sudden6:
I don't think Travis reports before the `after_success` thing is hit.

Compared to your example though, the master build is significantly faster (5min stage1): https://travis-ci.org/github/qTox/qTox/builds/761437516 and also other PRs are significantly faster: https://travis-ci.org/github/qTox/qTox/builds/761437516 while only this PR drops down to 10min.

I'd guess that's because the caches in your local test are not working/warmed up as in the builds in this repo. On the other hand this could also mean the caches are not working as designed, because this PR changes the build flags, which makes this whole comparison pointless.

I guess we should try to have it in stage 4 for a time and if the build times are good we can merge it with stage 1.

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6317#-:-MWPobDX2VrucjKLQIsc:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
Just to capture some discussion, you're right that caching was likely messing up my test, and it is also likely messing up this PR build time. I tried pushing to my fork's master branch to update the cache, and I still got weird benchmarking results where it seemed to get slower the more I ran, maybe due to Travis load or throttling.

We don't have a need to keep the produced "full build" binary in the job, so just configuring with coverage support there and not rebuilding during the "test" step would save a good amount of time without needing to split the job and repeat any work. Here I'd predict it still goes up slightly from current master due to running the coverage, but nowhere close to the 3x I was seeing in my benchmark where it wasn't even done the first minimal build by the time the master build was finished the whole job.

If we see job times shoot up for a few builds after merging, we can always split it up after.

PR #6374, opened 2021-09-12
Chatlog Cleanup

### LINKED ISSUE #5878: Chat history only loads on scroll wheel, not on other scroll up methods
##### Brief Description

OS: All
qTox version: v1.16.3+
Commit hash:  16de5f8549354fa2dc775cefefc773a4cd776590
toxcore:  v0.2.10
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Press page up in a chat to scroll up repeatedly

##### Observed Behavior
Chat scrolls up to the top chat line, then stops. Pressing page up again doesn't load more chat history, but scrolling up with the mouse scroll wheel does.

##### Expected Behavior
Any method that scrolls the chatlog should also load history when the top of the chatlog is reached.

### LINKED ISSUE #5940: ChatLog architecture fixes needed to fix scroll
There's a lot of janky ChatLog behaviour since the functionality to remove messages from the ChatLog was added a few months ago. It's definitely a good idea to remove messages from the ChatLog when it becomes too big for memory and performance reasons, but there are some architectural problems:
1) GenericChatForm holds the multiple GUI elements, e.g. search, chat edit, and ChatLog. But GenericChatForm still handles a bunch of message layout logic. IMO this should all be handled only inside of ChatLog.
2) Both GenericChatForm and ChatLog hold a cache of rendered messages. There's no reason to duplicate this any more, since SessionChatLog stores unrendered messages that can be loaded if the ChatLog is scrolls.
3) ChatLog has different message removal behaviour for friends and groups (if history is available), which is no longer required because of SessionChatLog. Message removal from ChatLog should always be possible.
4) SessionChatLog should expose a list of all messages, and ChatLog should just know about the list of rendered messages, and the subset of those rendered messages that are visible. ChatLog currently renders any message added to the bottom below all other messages, which means if you're scrolled up, receiving a new message will incorrectly append it onto your partial view. If instead, ChatLog correctly didn't render the message that was outside of its view, GenericChatForm would still cache that message and think it was rendered, since it has split responsibility and duplicated state.

We have 3 options going forward for fixing ChatLog:

1) Minimal fix: to unblock v1.17.0 release, change a few lines in ChatLog to restore old scroll behaviour, and no longer remove lines form ChatLog, which would be another couple lines would work. This would unfortunately mean we'd lose chat line eviction behaviour for v1.17.0, but would fix the scroll to bottom issue, fix search with scroll issues, and fix adding messages while scrolled up causing a gap in chat. I have this implemented locally.

2) Fix scroll and message gap: As well as fixing scroll like above, add to ChatLog knowledge of what messages are in render range, change logic in ChatLog that removes messages from top or bottom of ChatLog to instead remove messages based on where the current visible messages are. i.e. don't remove messages from top of ChatLog just because a message was added to the bottom, because of the user is scrolled up, messages at top of ChatLog might be what is in view. ChatLog should remove messages that are furthest away from the visible messages. Change GenericChatForm to not cache messages that ChatLog doesn't want to render because they're out of range.

3) For a complete fix: GeneircChatForm and ChatLog need a refactor. All rendered message handling logic should be in ChatLog only, ChatLog shouldn't care if the chat has history, and ChatLog should be aware of its targetted render range. If a message is updated outside of that render range, it should not add it (e.g. new message received when we're scrolled up). GenericChatForm handling wouldn't need to be added for when ChatLog rejects messages, since GenericChatForm would no longer be caching messages.

I feel like option 3) would almost definitely introduce a new slew of bugs since ChatForm is so fragile. I don't want to do a large refactor in that area when it's holding back release. Doing that on master once release is out would be great.

option 2) allows us to keep the message eviction performance benefit. Adding to ChatLog knowledge of its intended render range is a little complicated, and updating GenericChatForm to not cache messages that ChatLog doesn't want to render is a bit complicated too since currently GenericChatForm caches the message before ChatLog is given it. This wouldn't be _too_ hard, but I'm still worried that new bugs would be introduced.

option 1) is simple, and losing the performance benefit of message eviction from ChatLog isn't a regression from last release. IMO we should only re-include the performance benefits when the multiple architectural problems causing bugs are worked out.

@sudden6 what do you think? I'm leaning towards option 1, but think option 2 would be reasonable.

### LINKED ISSUE #6223: Make ChatLog refer to messages by ChatLogIdx, dynamically remove messages from view, and request new messages from IChatLog
##### Brief Description

This issue depends on https://github.com/qTox/qTox/issues/6222.

After #6222 is complete, every message that ChatLog displays will have a uniquely identifying ID, ChatLogIdx, which IChatLog will be aware of. This gives the ability for only ChatLog to care about what messages are currently in view, rather than GenericChatForm having to track it as well. ChatLog can then request messages to be loaded by just calling an API like "load messages from ChatLogIdx A to ChatLogIdx B" which IChatLog can generically fulfil.

This should be able to remove a lot of responsibility from GenericChatForm, and will heavily change the API of ChatLog. ChatLog will need to keep track of the ChatLogIdx of each message it has to be able to know which message indexes come after or before on scroll.

With this change, ChatLog's responsibility should be purely view, and should be the sole class responsibility of which messages are currently loaded. Because of that, ChatLog should be able to remove messages from view on scroll without needing to update any other classes.

With this overhaul of ChatLog, it could maybe be also be renamed to something like ChatView to reduce confusion with IChatLog which is a model concept but which has a very similar name.

sphaerophoria:
**NOTE: Based on top of #6311** 

Refactor/fixes for chatlog to resolve #6223 #5878 #5940

This PR has commits that fall under two categories
* Re-apply reverts from 1.17
** Note there's a fix in the middle of the reverts to deal with an updated Qt since the initial patch
* Implement a more architecturally sound sliding window chat log

Quick sample of what the new sliding window looks like in practice. Given the scope of this change I suggest reviewers try it themselves to see how it feels.

new_chatlog

The only remaining janky behavior is that selections for messages removed from the chatlog are cleared. This means you have to use the export chat log feature to get long logs. A future improvement would be to not purge messages that are selected.

Implementation commit messages for review context...

refactor(chatlog): Rename ChatLog -> ChatWidget

* Makes the distinction between IChatLog (the model class) and
  ChatWidget (the view class) more clear

refactor(chatlog): Remove unused getRow functions from ChatLine

* The getRow functions would not track correctly since the rows indexes
  cannot be fixed to the view anymore

feat(chatlog): Re-implement sliding window ChatLog view

* Replace lines/messages with helper class to synchronize state between
  IChatLog and ChatLog more easily
* selection indexes have been replaced with ChatLine::Ptrs, this is to
  ensure consistency while the view slides around
	* This has another benefit of removing all the code that has to
	  manually slide the selection boxes around
* Replaced all insertion/removal functions with single "insertAtIdx".
  This helps ensure that mappings between ChatLogIdx and position within
  the view are captured correctly as items in the view slide around
* workerTimeout replaced with more generic name "renderFinished" that is
  used in synchronous and asynchronous paths
* Removed unused function ChatForm::insertChatMessage
* Re-implemented "Go to current date" with new ChatLog APIs
* Removed unused GenericChatForm::addSystemDateMessage. This is handled
  by ChatLog now
* Resolves #6223
* Resolves #5878
* Resolves #5940

refactor(chatlog): Store ChatLine::Ptr in messages instead of ChatMessage

* This simplifies future refactoring since the rest of ChatLog expects
  to be working with the base class

refactor(chatlog): Move rendering of messages from GenericChatForm -> ChatLog

* Simplifies reasoning about who owns what functionality between
  GenericChatForm and ChatLog. GenericChatForm is now just a layout
  class and ChatLog handles all interactions with retrieving and
  displaying messages from the model
* Reasoning for work is described in more detail in #6223

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,8 +203,8 @@ set(${PROJECT_NAME}_SOURCES
   src/chatlog/chatline.h
   src/chatlog/chatlinestorage.cpp
   src/chatlog/chatlinestorage.h
-  src/chatlog/chatlog.cpp
-  src/chatlog/chatlog.h
+  src/chatlog/chatwidget.cpp
+  src/chatlog/chatwidget.h
   src/chatlog/chatmessage.cpp
   src/chatlog/chatmessage.h
   src/chatlog/content/filetransferwidget.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,6 +201,8 @@ set(${PROJECT_NAME}_SOURCES
   src/chatlog/chatlinecontentproxy.h
   src/chatlog/chatline.cpp
   src/chatlog/chatline.h
+  src/chatlog/chatlinestorage.cpp
+  src/chatlog/chatlinestorage.h
   src/chatlog/chatlog.cpp
   src/chatlog/chatlog.h
   src/chatlog/chatmessage.cpp
--- a/cmake/Testing.cmake
+++ b/cmake/Testing.cmake
@@ -45,6 +45,7 @@ auto_test(core toxid "")
 auto_test(core toxstring "")
 auto_test(chatlog textformatter "")
 auto_test(net bsu "${${PROJECT_NAME}_RESOURCES}") # needs nodes list
+auto_test(chatlog chatlinestorage "")
 auto_test(persistence paths "")
 auto_test(persistence dbschema "")
 auto_test(persistence offlinemsgengine "")
--- a/src/chatlog/chatline.cpp
+++ b/src/chatlog/chatline.cpp
@@ -37,14 +37,6 @@ ChatLine::~ChatLine()
     }
 }
 
-void ChatLine::setRow(int idx)
-{
-    row = idx;
-
-    for (int c = 0; c < static_cast<int>(content.size()); ++c)
-        content[c]->setIndex(row, c);
-}
-
 void ChatLine::visibilityChanged(bool visible)
 {
     if (isVisible != visible) {
@@ -53,11 +45,6 @@ void ChatLine::visibilityChanged(bool visible)
     }
 
     isVisible = visible;
-}
-
-int ChatLine::getRow() const
-{
-    return row;
 }
 
 ChatLineContent* ChatLine::getContent(int col) const
@@ -153,6 +140,7 @@ void ChatLine::addColumn(ChatLineContent* item, ColumnFormat fmt)
 
     format.push_back(fmt);
     content.push_back(item);
+    item->setIndex(0, content.size() -1 );
 }
 
 void ChatLine::replaceContent(int col, ChatLineContent* lineContent)
@@ -261,9 +249,4 @@ bool ChatLine::lessThanBSRectTop(const ChatLine::Ptr& lhs, const qreal& rhs)
 bool ChatLine::lessThanBSRectBottom(const ChatLine::Ptr& lhs, const qreal& rhs)
 {
     return lhs->sceneBoundingRect().bottom() < rhs;
-}
-
-bool ChatLine::lessThanRowIndex(const ChatLine::Ptr& lhs, const ChatLine::Ptr& rhs)
-{
-    return lhs->getRow() < rhs->getRow();
 }
--- a/src/chatlog/chatline.h
+++ b/src/chatlog/chatline.h
@@ -24,7 +24,7 @@
 #include <QVector>
 #include <memory>
 
-class ChatLog;
+class ChatWidget;
 class ChatLineContent;
 class QGraphicsScene;
 class QStyleOptionGraphicsItem;
@@ -95,7 +95,7 @@ public:
     static bool lessThanBSRectBottom(const ChatLine::Ptr& lhs, const qreal& rhs);
 
 protected:
-    friend class ChatLog;
+    friend class ChatWidget;
 
     QPointF mapToContent(ChatLineContent* c, QPointF pos);
 
--- a/src/chatlog/chatline.h
+++ b/src/chatlog/chatline.h
@@ -84,7 +84,6 @@ public:
     void reloadTheme();
 
     int getColumnCount();
-    int getRow() const;
 
     ChatLineContent* getContent(int col) const;
     ChatLineContent* getContent(QPointF scenePos) const;
@@ -94,7 +93,6 @@ public:
     // comparators
     static bool lessThanBSRectTop(const ChatLine::Ptr& lhs, const qreal& rhs);
     static bool lessThanBSRectBottom(const ChatLine::Ptr& lhs, const qreal& rhs);
-    static bool lessThanRowIndex(const ChatLine::Ptr& lhs, const ChatLine::Ptr& rhs);
 
 protected:
     friend class ChatLog;
@@ -103,7 +101,6 @@ protected:
 
     void addColumn(ChatLineContent* item, ColumnFormat fmt);
     void updateBBox();
-    void setRow(int idx);
     void visibilityChanged(bool visible);
 
 private:
--- a/src/chatlog/chatlinecontent.cpp
+++ b/src/chatlog/chatlinecontent.cpp
@@ -30,11 +30,6 @@ int ChatLineContent::getColumn() const
     return col;
 }
 
-int ChatLineContent::getRow() const
-{
-    return row;
-}
-
 int ChatLineContent::type() const
 {
     return GraphicsItemType::ChatLineContentType;
--- a/src/chatlog/chatlinecontent.h
+++ b/src/chatlog/chatlinecontent.h
@@ -35,7 +35,6 @@ public:
     };
 
     int getColumn() const;
-    int getRow() const;
 
     virtual void setWidth(qreal width) = 0;
     int type() const final;
--- a/src/chatlog/chatlinestorage.cpp
+++ b/src/chatlog/chatlinestorage.cpp
@@ -1,0 +1,203 @@
+#include "chatlinestorage.h"
+
+#include <QDebug>
+
+ChatLineStorage::iterator ChatLineStorage::insertChatMessage(ChatLogIdx idx, QDateTime timestamp, ChatLine::Ptr line)
+{
+    if (idxInfoMap.find(idx) != idxInfoMap.end()) {
+        qWarning() << "Index is already rendered, not updating";
+        return lines.end();
+    }
+
+    auto linePosIncrementIt = infoIteratorForIdx(idx);
+
+    auto insertionPoint = equivalentLineIterator(linePosIncrementIt);
+    insertionPoint = adjustItForDate(insertionPoint, timestamp);
+
+    insertionPoint = lines.insert(insertionPoint, line);
+
+    // All indexes after the insertion have to be incremented by one
+    incrementLinePosAfter(linePosIncrementIt);
+
+    // Newly inserted index is insertinPoint - start
+    IdxInfo info;
+    info.linePos = std::distance(lines.begin(), insertionPoint);
+    info.timestamp = timestamp;
+    idxInfoMap[idx] = info;
+
+    return insertionPoint;
+}
+
+ChatLineStorage::iterator ChatLineStorage::insertDateLine(QDateTime timestamp, ChatLine::Ptr line)
+{
+    // Assume we only need to render one date line per date. I.e. this does
+    // not handle the case of
+    // * Message inserted Jan 3
+    // * Message inserted Jan 4
+    // * Message inserted Jan 3
+    // In this case the second "Jan 3" message will appear to have been sent
+    // on Jan 4
+    // As of right now this should not be a problem since all items should
+    // be sent/received in order. If we ever implement sender timestamps and
+    // the sender screws us by changing their time we may need to revisit this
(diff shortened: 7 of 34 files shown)


anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNgkkeE80FhTORXhPz:-MlNgkkeE80FhTORXhQ-:b-yyyyjx) (no related file):*
While running this, I saw an issue where when lots of chat is loaded, i.e. the scroll bar is small, sometimes the chatlog stays white for a long time, over 10s. Right click -> go to current date, which evicts most messages seems to make this go away. Could be reproduced fairly infrequently by just scrolling up lots then scrolling back down and sending messages.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNh26yBEmgX08VNgwx:-MlNh26yBEmgX08VNgwy:b7ng0cq) (no related file):*
Another issue I saw, after starting up a profile and just quickly sending maybe 100 messages to another profile that was offline, I think while scrolled up, after scrolling back down to the bottom, it didn't have the latest sent messages, and sending more new messages didn't add them either. Right click -> go to current date also didn't show the new messages. Scrolling up and down also didn't. Right click -> load history was the only thing I tried that did.

I haven't been able to repro this trying the same steps :S

---

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNgkkeE80FhTORXhPz:-MlkrTrd1hltg1tfqLEq:b-jib8gf) (no related file):*

While running this, I saw an issue where when lots of chat is loaded, i.e. the scroll bar is small, sometimes the chatlog stays white for a long time, over 10s. Right click -> go to current date, which evicts most messages seems to make this go away. Could be reproduced fairly infrequently by just scrolling up lots then scrolling back down and sending messages.
</blockquote></details>

I haven't been able to reproduce this, but I'll try some more. The only time I saw this behavior was if I manually commented out the resize worker, but I don't see a way that you would have ended up in that state the way the code is now.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNh26yBEmgX08VNgwx:-MlkrnKyACoarfLzg03S:b-8ehfnw) (no related file):*

Another issue I saw, after starting up a profile and just quickly sending maybe 100 messages to another profile that was offline, I think while scrolled up, after scrolling back down to the bottom, it didn't have the latest sent messages, and sending more new messages didn't add them either. Right click -> go to current date also didn't show the new messages. Scrolling up and down also didn't. Right click -> load history was the only thing I tried that did.

I haven't been able to repro this trying the same steps :S
</blockquote></details>

I managed to reproduce something along these lines. Don't have a root cause yet but it's fairly easily for me to reproduce messages not showing after scrolling back down by just not scrolling _all_ the way. I do not see the behavior where re-scrolling does not fix it, but maybe it has the same root cause

---

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNgkkeE80FhTORXhPz:-Mlmu2DABq_JgZnQQ-8m:brcqhca) (no related file):*

I haven't been able to reproduce this, but I'll try some more. The only time I saw this behavior was if I manually commented out the resize worker, but I don't see a way that you would have ended up in that state the way the code is now.
</blockquote></details>

Hmm have seen this a couple more times just during normal conversation. It would happen only when the chat is large, right on an incoming message, last (seemingly indefinitely) with no resize worker overlay, then stop as soon as I receive another new message, and come back as soon as  I receive another. No date line was in the view. I could scroll across the whole chatlog area, scroll against the very top or the very bottom with no change.

Again right click -> "Go to current date" stopped it and made it stop happening on new message, while also evicting most of the chat from view.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNh26yBEmgX08VNgwx:-MlmueBJ3aDk6YTNW4Tf:b4g4t15) (no related file):*

I managed to reproduce something along these lines. Don't have a root cause yet but it's fairly easily for me to reproduce messages not showing after scrolling back down by just not scrolling _all_ the way. I do not see the behavior where re-scrolling does not fix it, but maybe it has the same root cause
</blockquote></details>

I might have seen this again, or maybe something similar, where I was scrolled I think due to search, received a new message, scrolled down and saw it, searched again, and then could never get it to show back up. Scrolling down all the way didn't show it, and "Go to current date" also wouldn't show it, nor would "Load history". Closing qTox and re-opening it did show the message, but this seems like not just not scrolling to the bottom, I don't think.

---

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNgkkeE80FhTORXhPz:-Mm9Na3ABJQG4SwfQPsV:bq7w7op) (no related file):*

Hmm have seen this a couple more times just during normal conversation. It would happen only when the chat is large, right on an incoming message, last (seemingly indefinitely) with no resize worker overlay, then stop as soon as I receive another new message, and come back as soon as  I receive another. No date line was in the view. I could scroll across the whole chatlog area, scroll against the very top or the very bottom with no change.

Again right click -> "Go to current date" stopped it and made it stop happening on new message, while also evicting most of the chat from view.
</blockquote></details>

I haven't managed to reproduce this. I think I can mitigate it though.

The current behavior is already kind of annoying. Every 100 messages the chat log flickers because we have to re-calculate the chatlog to render all our messages. Most chat apps don't clear messages while the window is open I've noticed, so I've just followed that pattern and moved the message purging to when a chat is hidden.

Note: I think this behavior could be improved if we do chat layout from bottom to top instead of top to bottom, but I don't want to take on that complexity here.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNh26yBEmgX08VNgwx:-Mm9OZQREQBglyzhARIi:b9m165u) (no related file):*

I might have seen this again, or maybe something similar, where I was scrolled I think due to search, received a new message, scrolled down and saw it, searched again, and then could never get it to show back up. Scrolling down all the way didn't show it, and "Go to current date" also wouldn't show it, nor would "Load history". Closing qTox and re-opening it did show the message, but this seems like not just not scrolling to the bottom, I don't think.
</blockquote></details>

Hm, interesting. I can't seem to reproduce this. Maybe we get lucky and the removal of chatlog shuffling while it's open will help. One guess is there's a race between re-rendering on messages received and re-rendering on scroll, so it's not an unreasonable hope for an improvement there.

---

anthonybilinski (changes requested):
> Reviewed 33 of 41 files at r1, 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNgkkeE80FhTORXhPz:-Mn4Mf-i4JvPO3d7MiKF:b-889sf0) (no related file):*

I haven't managed to reproduce this. I think I can mitigate it though.

The current behavior is already kind of annoying. Every 100 messages the chat log flickers because we have to re-calculate the chatlog to render all our messages. Most chat apps don't clear messages while the window is open I've noticed, so I've just followed that pattern and moved the message purging to when a chat is hidden.

Note: I think this behavior could be improved if we do chat layout from bottom to top instead of top to bottom, but I don't want to take on that complexity here.
</blockquote></details>

Running the new version with both manual testing with two local profiles for quite a while, sending many messages back and forth with varying timing and with typing notification on both etc, I wasn't able to reproduce this issue. I also had a normal long conversation, which is where I saw the issue many times before, and didn't see it once. So I think this workaround is working as intended :+1: 

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-MlNh26yBEmgX08VNgwx:-Mn4ZVhp00IXsrEr27-M:b-1afz6f) (no related file):*

Hm, interesting. I can't seem to reproduce this. Maybe we get lucky and the removal of chatlog shuffling while it's open will help. One guess is there's a race between re-rendering on messages received and re-rendering on scroll, so it's not an unreasonable hope for an improvement there.
</blockquote></details>

I haven't seen this one again and haven't been able to reproduce it in testing either, so hopefully it's dealt with. I don't think it's worth holding up the merge.

---
*[a discussion](https://reviewable.io/reviews/qtox/qtox/6374#-Mp3xsF34J6gxjP1REtu:-Mp3xsF34J6gxjP1REtv:b-uwkowd) (no related file):*
:lgtm_strong: 

---

anthonybilinski (commented):
> Reviewed 19 of 19 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6374)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6381, opened 2021-09-23
fix(appimage): Handle no TRAVIS_TAG or TRAVIS_COMMIT in AppImage build

anthonybilinski:
With set u, bash aborts when either is not defined. Also neither is set when
building locally.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6381)

### DIFF
--- a/.github/workflows/commit-format.yaml
+++ b/.github/workflows/commit-format.yaml
@@ -1,0 +1,14 @@
+name: Commit Format
+on: pull_request
+jobs:
+  verify-commit-format:
+    name: Verify Commit Format
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # so that we can see the full commit range
+      - name: Run
+        run: ./.travis/verify-commit-format.sh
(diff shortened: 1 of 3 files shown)


sphaerophoria (commented):
General comments

1. We don't use Travis, it's weird to be editing this and propagating travis naming. I can live with it though
2. It seems weird for us to depend on travis here at all. Is it more robust for us to just check if our revision is tagged and use the tag if it is? If the script has the ability to determine it's own git hash/version number why use travis at all?

That being said this looks like it should fix the problem... so :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6381#-:-MkKlT4n2LCsGobJkA67:bh7gzrc)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
Yeah... it's a bit of a band-aid. Travis references will be removed together when GH actions are at feature parity, they're still kind of in a half-implemented state now. Can look at removing passing in of the git tag altogether some time later. I think as-is this change is a small improvement that unbreaks the current state, so I'll merge unchanged, but agreed on your two points.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6381#-:-MmmLA_XC6MDW6qkdzOE:b-cv3b6)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6384, opened 2021-09-23
chore(CI): Run tests on pushed commits

anthonybilinski:
Split commit format check into its own workflow that only runs on PR.

Can be built off for nightly or per-commit releases.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6384)

### DIFF
--- a/.github/workflows/commit-format.yaml
+++ b/.github/workflows/commit-format.yaml
@@ -1,0 +1,14 @@
+name: Commit Format
+on: pull_request
+jobs:
+  verify-commit-format:
+    name: Verify Commit Format
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0 # so that we can see the full commit range
+      - name: Run
+        run: ./.travis/verify-commit-format.sh
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,17 +1,6 @@
-name: test
-on: pull_request
+name: Test
+on: [pull_request, push]
 jobs:
-  verify-commit-format:
-    name: Verify Commit Format
-    runs-on: ubuntu-18.04
-    env:
-      GITHUB_CONTEXT: ${{ github.event.pull_request.commits_url }}
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0 # so that we can see the full commit range
-      - name: Run
-        run: ./.travis/verify-commit-format.sh
   build-docs:
     name: Docs
     runs-on: ubuntu-18.04


sphaerophoria (commented):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6384#-:-MmnvaOG5HGMIsVTwhYv:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[.github/workflows/test.yaml, line 4 at r1](https://reviewable.io/reviews/qtox/qtox/6384#-Mmnqbpl4y2fbwDf1xul:-Mmnqbpm44cpQmIjElJ2:b-j8n2f3) ([raw file](https://github.com/qtox/qtox/blob/db5081d3dce4c2d84e7f1a0aaa8fc81636f47db0/.github/workflows/test.yaml#L4)):*
> ```YAML
> on: pull_request
> jobs:
>   verify-commit-format:
> ```

I think this change exposes a CI behavior that could be improved.

I think we should be running CI on each commit in base_ref..head_ref, before we merge. AFAICT we only run on the last commit of a PR, but check the commit format of the whole range. I think we should only check the commit format of the current commit, but do that for every commit in the stack. This ensures that when we merge we have a commit history where we can check out any old commit, run, and it will work. This makes workflows like git bisect much easier.

That's not the case now though, and forcing you to do that here is unreasonable. So there's two options here

1. Update the travis workflow to only check new commits instead of the whole history. I think that can be done with github context base_ref/head_ref.
2. Do what you did and just say "we only really care if it's buildable, if the commit format is wrong we can't fix that so we shouldn't even bother checking"

Given that we can't fix bad commit messages once they're in master, I think this is reasonable. What does the failure tell us? What do we want the failure to tell us? We want to know if there's something we need to fix, like a broken build from a merge conflict, so running everything except verifying commit format is reasonable.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6384)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[.github/workflows/test.yaml, line 4 at r1](https://reviewable.io/reviews/qtox/qtox/6384#-Mmnqbpl4y2fbwDf1xul:-MmoFcZqB-kbusextMUn:brckm8i) ([raw file](https://github.com/qtox/qtox/blob/d7347773f8b0aa5101db6b1aa15e8f4b503bc90b/.github/workflows/test.yaml#L4)):*
> AFAICT we only run on the last commit of a PR

We're actually running on a test merge of the PR branch into the target branch for most CI, just the verify-commit-format.sh is passed github.event.pull_request.commits_url which just gives the list of commits. For our other CI that use the default cloned state, they should have it merged.

> I think we should be running CI on each commit in base_ref..head_ref, before we merge. .. This ensures that when we merge we have a commit history where we can check out any old commit, run, and it will work. This makes workflows like git bisect much easier.

Hmm yeah that would be nice to have I think, might cause quite a bit of extra friction with contributors without clean git history. Already each merge commit should be fully tested.. but yeah I get wanting to bisect within multiple commits from a PR. I'd probably be in favour of doing that at some time.

Issue #1532, opened 2015-04-25
No support for resuming file transfers across client restarts


zetok:
![muh accepted files](https://cloud.githubusercontent.com/assets/3148759/7333157/64e2ba9e-eb5b-11e4-9d73-34ba9daf9ddb.png)
- `HNNG` client was turned off and on again
- `:3` accepted 1st file transfer when `HNNG` was offline, and second file transfer when `HNNG` came back online
- 2nd file was paused / unpaused few times after accepting file transfer

EDIT:
Both clients are on new api; 
- `:3` is on 2954ee3c958575f6f88fe00beda73a47f600f066
- `HNNG` is on 9ece486e2247a44e77389a80ca42d93233b07466

zetok:
Note: toxcore currently allows that;
Way to implement: https://github.com/irungentoo/Tox_Client_Guidelines/blob/master/Important/File_Transfers.md

ProMcTagonist:
I can't wait for this. It would make bad connections so much less infuriating. As it stands I have to ask people to send image collections as tarballs and hope it gets through before they go off and on again.

sphaerophoria:
Okay I've started taking a look at this.

There was a PR at https://github.com/qTox/qTox/pull/4674 that never got merged. Unfortunately it looks like a potential good contributor was scared off by review comment spam :(. I've started rebasing that PR but there have been some significant changes since then

File transfers are now stored in history, this adds a bit of complexity above the initial PR. I believe at the time of that PR we would only resume transfers over network drops, we now would also want to resume transfers over client restarts.

The PR did introduce some nice concepts though, it added states for "broken paused" and "broken canceled". These flagged to corefile that it should set the state to pause when our peer came back online, or cancel the file transfer. The broken canceled state may be redundant, I haven't looked close enough yet, but the UX concepts there are good.

Filetransferwidget has also been fairly significantly refactored since that point. I think we can take a good chunk of the initial corefile work, but filetransferwidget should be redone.

So in terms of what work there is to do here

* File transfer state needs to be given to the DB
* On client start we need to check for file transfers that are not complete and resume them
* We should probably be hashing files on the sender side when broken file transfers are resumed
    * Looks like we store file hashes in the DB, but I think only when we finish transfering the file. I think no new DB schema will be needed, but we might need to re-jig our usage of the current schema a little 
* We should shoot for a similar UX to the initial PR, however I'm not 100% sure the new states will make as much sense now though.
[...]

PR #6386, opened 2021-09-26
Add in progress file transfers to files form

sphaerophoria:
Identified in #1532, if we add resumable file transfers, we could accidentally lose track of in progress file transfers and sap forgetful users' bandwidth. This PR is an attempt to update the UI so that there is a single place to monitor all in progress file transfers.

High level goals
* Show in progress file transfers in the file transfer list
* Add controls to pause/resume/stop in progress file transfers from the file transfer list

Demo of UI change: 

[](https://user-images.githubusercontent.com/1069811/140666048-e7b3a47d-cd62-426f-9f6e-f1693e20f1e5.mp4)

Future improvement: 
* Add indicator to files form button to show when there are in progress file transfers at a glance

Bonus fix:
* Sometimes the file transfer dialog would end up in a "transmitting..." state even when paused. This was fairly reproducible during testing and fixed.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)


Monsterovich:
Didn't you accidentally fix the bug when the uploading widget remains on the form when you switch a user?

anthonybilinski:
@Monsterovich I can't find an open issue for that, but I know what you mean. It wouldn't be changed as part of this PR, but possibly could have been fixed as part of https://github.com/qTox/qTox/pull/6374 since rendering was changed, though the cause of it hasn't been root caused so there's no guarantee.

anthonybilinski (changes requested):
> Reviewed 1 of 1 files at r1, 14 of 14 files at r2, 20 of 20 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfileprogress.cpp, line 2 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe8WYe-QOEqAhJVHuZ:-Mpe8WYe-QOEqAhJVHu_:b-2j65lm) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L2)):*
> ```cpp
> /*
>     Copyright © 2018-2019 by The qTox Project Contributors
> ```

`2018-2019` should be `2021`
___
*[src/core/toxfileprogress.cpp, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe7c192jEOvy9AHRtH:-Mpe7c192jEOvy9AHRtI:b-tfl3uv) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L41)):*
> ```cpp
> }
> 
> ToxFileProgress::ToxFileProgress(uint64_t filesize, size_t speedSampleCount, int samplePeriodMs)
> ```

What's the benefit of having more than two samples? It looks like we only measure between the head and tail, sometimes looking at tail-1, which would be fine if it was also just head. Why not just hold only two samples at least samplePeriodMs apart?
___
*[src/core/toxfileprogress.cpp, line 47 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe0Gxf2b3wsExRQVQW:-Mpe0Gxf2b3wsExRQVQX:bxixh4y) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L47)):*
> ```cpp
> {
>     if (samplePeriodMs < 0) {
>         qWarning("Invalid sample rate, healing to 1000ms");
> ```

doesn't look like any healing actually takes place here?
___
*[src/core/toxfileprogress.cpp, line 59 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe120U-AlcvWp1eYbB:-Mpe120U-AlcvWp1eYbC:be3z8fl) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L59)):*
> ```cpp
>     }
> 
>     return samples.back().timestamp;
> ```

Isn't the last sample time at `decrementRollingIdx(nextIdx, speedSampleCount)`, not necessarily at `samples.back()` due to the wrapping indexes over the vector?
___
*[src/widget/form/filesform.h, line 137 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenPyE1LNpVtWON0pP:-MpenPyE1LNpVtWON0pQ:b-rx3eyy) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.h#L137)):*
> ```objc
>     QVBoxLayout headLayout;
>     QTabWidget main;
>     QTableView *sent, *recvd;
> ```

`sent` and `recvd` were replaced by `sentModel` and `recvdModel`, and are now unused
___
*[src/widget/form/filesform.cpp, line 45 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-Mpee9cU46jkYL3DTYMM:-Mpee9cU46jkYL3DTYMN:bei4a40) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L45)):*
> ```cpp
>         // our width from it
>         auto buttonXPos = option.rect.x() + buttonHorizontalArea / 2 - controlSize / 2;
>         auto buttonYPos = option.rect.y() + option.rect.height() * 0.1;
> ```

it looks like you might be getting some unintended rounding here, with integer division for `buttonHorizontalArea` and `buttonHorizontalArea / 2` even though `constrolSize` is float
___
*[src/widget/form/filesform.cpp, line 56 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpeeSb657sJSiIpnrIZ:-MpeeSb70YMHD1oniHDa:b-2guby3) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L56)):*
> ```cpp
>         // To center the button, we find the horizontal center and subtract half
>         // our width from it
>         auto buttonXPos = option.rect.x() + buttonHorizontalArea + buttonHorizontalArea / 2 - controlSize / 2;
> ```

same here
___
*[src/widget/form/filesform.cpp, line 454 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenV06ECaIphdKj_yo:-MpenV06ECaIphdKj_yp:b-60a0nn) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L454)):*
> ```cpp
> 
>     connect(sent, &QTableView::activated, this, &FilesForm::onSentFileActivated);
>     connect(recvd, &QTableView::activated, this, &FilesForm::onReceivedFileActivated);
> ```

`sent` and `recvd` can be removed
___
*[src/widget/form/filesform.cpp, line 464 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenhAx9ZA8Ubmd1PUk:-MpenhAx9ZA8Ubmd1PUl:b-vukq1f) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L464)):*
> ```cpp
>     Translator::unregister(this);
>     delete recvd;
>     delete sent;
> ```

again `sent` and `recvd` can be removed
___
*[src/widget/form/filesform.cpp, line 506 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpeobnKD53D2u2JwDej:-MpeobnL7epkMX_FgrI2:bc5j99i) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L506)):*
> ```cpp
> {
>     const auto& filePath = sentModel->data(index, Qt::UserRole).toString();
>     Widget::confirmExecutableOpen(filePath);
> ```

existing behaviour, doesn't need to be changed, but it seems weird to warn a user about a file they've sent
___
*[test/core/fileprogress_test.cpp, line 1 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe8TVx2vgYLXdZZQEB:-Mpe8TVx2vgYLXdZZQEC:b53d2th) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/test/core/fileprogress_test.cpp#L1)):*
> ```cpp
> #include "src/core/toxfileprogress.h"
> ```

missing copyright header
___
*[test/core/fileprogress_test.cpp, line 254 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-MpeIKoL87ZJ9VgQ3tjQ:-MpeIKoL87ZJ9VgQ3tjR:b9nni2d) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/test/core/fileprogress_test.cpp#L254)):*
> ```cpp
> {
>     auto progress = ToxFileProgress(100, 3, 1000);
>     QVERIFY(progress.lastSampleTime() < QTime::currentTime());
> ```

re: comment in fileprogress::addsample, I think this test is flaky since it's using non-monotonic system time 
___
*[test/widget/filesform_test.cpp, line 2 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpeOX4k4A55GSRk901J:-MpeOX4k4A55GSRk901K:bz5r2i6) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/test/widget/filesform_test.cpp#L2)):*
> ```cpp
> /*
>     Copyright © 2014-2021 by The qTox Project Contributors
> ```

copyright year should just be 2021
___
*[src/core/toxfileprogress.h, line 32 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-MpeB-qL0q1uXlw60RAd:-MpeB-qL0q1uXlw60RAe:bldj6j4) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.h#L32)):*
> ```objc
> 
>     QTime lastSampleTime() const;
>     bool addSample(uint64_t bytesSent, QTime now = QTime::currentTime());
> ```

`QTime::currentTime` comes from system clock, so this could break on NTP sync, daylight savings, midnight, time zone change etc. Shouldn't all the samples use a monotonic time?

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfileprogress.cpp, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe7c192jEOvy9AHRtH:-MqfA7yF7gSj0GKf7CFU:be6x9fd) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L41)):*

What's the benefit of having more than two samples? It looks like we only measure between the head and tail, sometimes looking at tail-1, which would be fine if it was also just head. Why not just hold only two samples at least samplePeriodMs apart?
</blockquote></details>

You're right, this is an artifact of the initial design changing over time and me not re-evaluating it. I could easily do this with 2 samples. Fixed.
___
*[src/core/toxfileprogress.cpp, line 47 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe0Gxf2b3wsExRQVQW:-MqfT88bE3VVFe7TOfpH:b-7ql1td) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L47)):*

doesn't look like any healing actually takes place here?
</blockquote></details>

fixed
___
*[src/core/toxfileprogress.cpp, line 59 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe120U-AlcvWp1eYbB:-MqfPrjEC4LoUDwRN1Xm:bv1jw8d) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.cpp#L59)):*

Isn't the last sample time at `decrementRollingIdx(nextIdx, speedSampleCount)`, not necessarily at `samples.back()` due to the wrapping indexes over the vector?
</blockquote></details>

True, fixed
___
*[src/widget/form/filesform.h, line 137 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenPyE1LNpVtWON0pP:-MqfRQVm4t97-F3OwGjm:b-nivyys) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.h#L137)):*

`sent` and `recvd` were replaced by `sentModel` and `recvdModel`, and are now unused
</blockquote></details>

True, but not intentionally. This was a mistake on the last iteration of the refactor. Fixed
___
*[src/widget/form/filesform.cpp, line 45 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-Mpee9cU46jkYL3DTYMM:-MqfRerE1KI5qQkdERw0:b-47vjvg) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L45)):*

it looks like you might be getting some unintended rounding here, with integer division for `buttonHorizontalArea` and `buttonHorizontalArea / 2` even though `constrolSize` is float
</blockquote></details>

Good call, fixed
___
*[src/widget/form/filesform.cpp, line 56 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpeeSb657sJSiIpnrIZ:-MqfT5eG0puv0GYr1YAw:b-7ql1td) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L56)):*

same here
</blockquote></details>

fixed
___
*[src/widget/form/filesform.cpp, line 454 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenV06ECaIphdKj_yo:-MqfSVicByRjouKntzlB:bs2f11c) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L454)):*

`sent` and `recvd` can be removed
</blockquote></details>

Denied
___
*[src/widget/form/filesform.cpp, line 464 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpenhAx9ZA8Ubmd1PUk:-MqfShaq3DqPJwMOf201:bs2f11c) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/src/widget/form/filesform.cpp#L464)):*

again `sent` and `recvd` can be removed
</blockquote></details>

Denied
___
*[test/core/fileprogress_test.cpp, line 1 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-Mpe8TVx2vgYLXdZZQEB:-MqfUV7TBrvWYFioIpQ9:b-7ql1td) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/test/core/fileprogress_test.cpp#L1)):*

missing copyright header
</blockquote></details>

fixed
___
*[test/core/fileprogress_test.cpp, line 254 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-MpeIKoL87ZJ9VgQ3tjQ:-MqfSugY4DIS-GmAb5hT:bavjl8) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/test/core/fileprogress_test.cpp#L254)):*

re: comment in fileprogress::addsample, I think this test is flaky since it's using non-monotonic system time 
</blockquote></details>

No matter what the current time is, the last sample time should be lower since it's an invalid qtime.
___
*[test/widget/filesform_test.cpp, line 2 at r3](https://reviewable.io/reviews/qtox/qtox/6386#-MpeOX4k4A55GSRk901J:-MqfUXqiCEXhZgbDN2mG:b-7ql1td) ([raw file](https://github.com/qtox/qtox/blob/ee99343e79e9b661ad1b6d60ffe1f8ef02e6966d/test/widget/filesform_test.cpp#L2)):*

copyright year should just be 2021
</blockquote></details>

fixed
___
*[src/core/toxfileprogress.h, line 32 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-MpeB-qL0q1uXlw60RAd:-Mqf7blB2ptNqSKSo7mI:b8ufdzx) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.h#L32)):*

`QTime::currentTime` comes from system clock, so this could break on NTP sync, daylight savings, midnight, time zone change etc. Shouldn't all the samples use a monotonic time?
</blockquote></details>

I only did this because it was the way the file transfer dialog worked before. Other classes (filetransferwidget) expect a QTime, and the only way to get a monotonic time is to switch types. I think this change leaks out of the scope of this PR, even though I think it's a desirable change.

I do protect against backwards time changes by resetting the speed count. This means that if the time changes our speed will either reset or seem way too fast for the duration of the sample period.

anthonybilinski (changes requested):
> Reviewed 1 of 14 files at r6, 4 of 20 files at r7, 28 of 30 files at r8, 9 of 15 files at r9, 17 of 21 files at r10.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfileprogress.cpp, line 54 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJ8NmX7caPf4ZasxG1:-MrJ8NmX7caPf4ZasxG2:bhyucun) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L54)):*
> ```cpp
>     }
> 
>     if (now < active->timestamp || now < inactive->timestamp) {
> ```

Isn't the active sample guaranteed to have equal or more bytes and be equal or more recent than the inactive sample? If so, could this check and the one above just check active and rely on that assumption? I was a bit confused when I saw it checking both and assumed the order wasn't guaranteed.
___
*[src/core/toxfileprogress.cpp, line 60 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJClaY1AQNTs_wPZaz:-MrJClaY1AQNTs_wPZb-:bzh0cfo) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L60)):*
> ```cpp
> 
>     // Ensure both samples are initialized
>     if (inactive->timestamp == QTime()) {
> ```

same argument here, couldn't only `active` be checked? or have a third variable to indicate if the samples are initialized?

It would be fine too if the second sample was checked in an assert as more of a sanity check
___
*[src/core/toxfileprogress.cpp, line 110 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJMH7r9RcC6Tbh7LXT:-MrJMH7r9RcC6Tbh7LXU:bxwjmev) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L110)):*
> ```cpp
>     };
> 
>     if (std::any_of(samples.cbegin(), samples.cend(), sampleTimeInvalid)) {
> ```

same thing here
___
*[src/core/toxfileprogress.h, line 32 at r2](https://reviewable.io/reviews/qtox/qtox/6386#-MpeB-qL0q1uXlw60RAd:-MrJ4DfRAE0cPnOTVLB3:b-rdhs64) ([raw file](https://github.com/qtox/qtox/blob/9b13ce83ddda298f63118893fd452478fc58cff6/src/core/toxfileprogress.h#L32)):*

I only did this because it was the way the file transfer dialog worked before. Other classes (filetransferwidget) expect a QTime, and the only way to get a monotonic time is to switch types. I think this change leaks out of the scope of this PR, even though I think it's a desirable change.

I do protect against backwards time changes by resetting the speed count. This means that if the time changes our speed will either reset or seem way too fast for the duration of the sample period.
</blockquote></details>

Alright agreed it's out of scope, good to see there's some mitigation here.

anthonybilinski (commented):
> Reviewed 4 of 21 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)* status: 0 of 1 LGTMs obtained

sphaerophoria (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386)* status: 0 of 1 LGTMs obtained

___
*[src/core/toxfileprogress.cpp, line 54 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJ8NmX7caPf4ZasxG1:-MrJmvx-12MIDgplzRkB:b-m056rk) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L54)):*

Isn't the active sample guaranteed to have equal or more bytes and be equal or more recent than the inactive sample? If so, could this check and the one above just check active and rely on that assumption? I was a bit confused when I saw it checking both and assumed the order wasn't guaranteed.
</blockquote></details>

True, I was trying to be careful around reset and the case where we push data into the inactive sample instead of the active one. That being said, I think there's nothing harmful about this, and makes it more robust to mistakes. I think I'm hesitant to update this to rely on something a little more implicit 
___
*[src/core/toxfileprogress.cpp, line 60 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJClaY1AQNTs_wPZaz:-MrJnqImDsd2RAYauPoz:b-g98a3b) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L60)):*

same argument here, couldn't only `active` be checked? or have a third variable to indicate if the samples are initialized?

It would be fine too if the second sample was checked in an assert as more of a sanity check
</blockquote></details>

When we reset we clear both active and inactive. In this case this makes it so that when we push our first sample, our speed is sane. If we don't initialize them both to the value of the first sample, after the first sample is pushed we'll have active < inactive and a default QTime() effectively at 0, which results in a really large negative speed.

I don't see the benefit of adding a bool to check the state, we already have the validity of each sample stored within the sample themselves.
___
*[src/core/toxfileprogress.cpp, line 110 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJMH7r9RcC6Tbh7LXT:-MrJoNey9EhtgxYH0KHb:b-t3u4ae) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L110)):*

same thing here
</blockquote></details>

I'd rather not rely on it

anthonybilinski (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6386#-:-MrN6jpJ8825xq4X-Bkz:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/core/toxfileprogress.cpp, line 54 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJ8NmX7caPf4ZasxG1:-MrN67rl51Sn2OXKOhG0:b-s1eqiv) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L54)):*

True, I was trying to be careful around reset and the case where we push data into the inactive sample instead of the active one. That being said, I think there's nothing harmful about this, and makes it more robust to mistakes. I think I'm hesitant to update this to rely on something a little more implicit 
</blockquote></details>

Sure agreed functionally it's irrelevant, I found it mildly distracting while reading, but I don't feel strongly and this does have a benefit of protecting against bugs even on release builds unlike an assert. Approving.
___
*[src/core/toxfileprogress.cpp, line 60 at r10](https://reviewable.io/reviews/qtox/qtox/6386#-MrJClaY1AQNTs_wPZaz:-MrLEBO_AWZguwjCZ3JO:b-6bpkap) ([raw file](https://github.com/qtox/qtox/blob/257a19caaaa3a31b0890e3eeb728ef9e19f28ede/src/core/toxfileprogress.cpp#L60)):*

When we reset we clear both active and inactive. In this case this makes it so that when we push our first sample, our speed is sane. If we don't initialize them both to the value of the first sample, after the first sample is pushed we'll have active < inactive and a default QTime() effectively at 0, which results in a really large negative speed.

I don't see the benefit of adding a bool to check the state, we already have the validity of each sample stored within the sample themselves.
</blockquote></details>

Understood that both active and inactive need to be reset, I was just saying that they could both rely on the single condition of `active->timestamp == QTime()`. Regardless approving based on previous comment.

PR #6388, opened 2021-10-04
fix(ui): contact list optimization

### LINKED ISSUE #6372: Friend list flashes and stalls when sorted by activity mode since 6347
##### Brief Description

OS: Arch 2021-08-31
qTox version: v1.17.3+
Commit hash:  bde6dc0df0583a2c869612a010533bb6246ad7e2
toxcore:  v0.2.12
Qt:  5.15.2

Reproducible: Always

##### Steps to reproduce

1. Sort friend list by activity
2. Send a message to a friend

##### Observed Behavior
1. Friend list flashes
2. The more friends in your friend list, the longer the flash takes
3. The GUI freezes for the duration of the flash, which isn't noticeable with 1 friend, but is very noticeable with ~100

##### Expected Behavior
1. Like when sorted by name, sending a message to a friend doesn't cause the friend list to flash or pause GUI.
2. User experience to match before a24f99df0d488450447e0c9360613dbcd9e30fd8, when in activity mode the friend list didn't flash or pause when sending a message.

##### Additional Info
@bodwok do you have any first guesses for how this was changed in your PR or where to fix it? No problem if not.

bodwok:
New changes:

- When the application starts, the contact widgets are drawn after adding the entire list.
- When writing messages, the view will be redrawn if the sort order is changed. 

This should remove flashing during chatting and speed up the launch.

Fix: #6372

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6388)

### DIFF
--- a/src/model/friendlist/friendlistmanager.cpp
+++ b/src/model/friendlist/friendlistmanager.cpp
@@ -20,9 +20,9 @@
 #include "friendlistmanager.h"
 #include "src/widget/genericchatroomwidget.h"
 
-FriendListManager::FriendListManager(QObject *parent) : QObject(parent)
+FriendListManager::FriendListManager(int countContacts, QObject *parent) : QObject(parent)
 {
-
+    this->countContacts = countContacts;
 }
 
 QVector<FriendListManager::IFriendListItemPtr> FriendListManager::getItems() const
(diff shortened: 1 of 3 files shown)


anthonybilinski (approved):
Nice, helps a lot during chatting.

Still pauses and flashes a bit when the order truly does change, which I don't think it did before, but that's much rarer than on every message like before, so :lgtm_strong: 

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6388#-:-MmpxOYc1qbsDd9dnezl:b-iozbix)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6390, opened 2021-10-10
Update Windows dependencies

nurupo:
- Updated base OS to the new Debian Stable release
- Updated dependencies

You can read commit messages for more info.

Tested that everything builds locally: both i686 and x86_64, both debug and release, both with `TRAVIS_CI_STAGE="stage3"` and without.

Tested that i686 and x86_64 release `qtox.exe` binaries run, go online (connect to DHT) and that the desktop capture works in Settings. Didn't add any contacts (zero contacts) or test anything beyond that. Tested under Wine on Linux.

---

It would be nice if qTox supported building with Qt 6. My understanding is that Qt 5 has only one last non-commercial release left - [Qt 5.12.12 sometime in H2 2021](https://wiki.qt.io/Qt_5.12_Release) and that's it. I got Qt 6.2 to cross-compile, so I'm ready on my side, it should be relatively easy to update the build script once qTox supports Qt 6.

I see that qTox uses some APIs that got removed in Qt 6, at least `QRegExp` and `QStringRef` (replaced by `QRegularExpression` and `QStringView`). When building against Qt 5.15, you can turn on a flag which will warn about usage of API that will get removed in Qt 6. The porting guide has instructions on this https://doc.qt.io/qt-6/portingguide.html. Also, using LibsnoreQt5 and KF5Sonnet without Qt 5 might be problematic...

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6390)


anthonybilinski (approved):
Thanks for the heads up on Qt 6 compatibility, I've made a ticket for that here: https://github.com/qTox/qTox/issues/6400.

We can probably pick it up after this upcoming release.

:lgtm_strong: 

> Reviewed 3 of 3 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6390#-:-MmmOT0R-Fsc_vzqcPkO:b-joixq5)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6394, opened 2021-10-15
refactor: Provide a Virtual Method Anchor for Classes in Headers

### LINKED ISSUE #6036: Provide a Virtual Method Anchor for Classes in Headers
*If a class is defined in a header file and has a vtable (either it has virtual methods or it derives from classes with virtual methods), it must always have at least one out-of-line virtual method in the class. Without this, the compiler will copy the vtable and RTTI into every .o file that #includes the header, bloating .o file sizes and increasing link times.*

Source: http://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers

In most cases, this is just a matter of adding an explicitly defined destructor and defining it in the .cpp file.

brandonmosher:
Define at least one virtual method in cpp files for classes with
vtables to improve link time. This is done by defining a defaulted
destructor in the cpp file, or moving an overridden method from the
header to the cpp file. Where neccessary, create new cpp files and
add them to CMakeLists.

See http://llvm.org/docs/CodingStandards.html#provide-a-virtual-method-anchor-for-classes-in-headers.

Fixes #6036.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6394)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,13 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-rtti")
 
 # Hardening flags (ASLR, warnings, etc)
 set(POSITION_INDEPENDENT_CODE True)
+
+include(CheckCXXCompilerFlag)
+CHECK_CXX_COMPILER_FLAG(-Wweak-vtables COMPILER_SUPPORTS_WARNING_WEAK_VTABLES)
+if(COMPILER_SUPPORTS_WARNING_WEAK_VTABLES)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wweak-vtables")
+endif()
+
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstrict-overflow")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstrict-aliasing")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wold-style-cast")
@@ -232,7 +239,18 @@ set(${PROJECT_NAME}_SOURCES
   src/core/core.h
   src/core/dhtserver.cpp
   src/core/dhtserver.h
+  src/core/icoreextpacket.cpp
+  src/core/icoreextpacket.h
+  src/core/icoresettings.cpp
   src/core/icoresettings.h
+  src/core/icorefriendmessagesender.cpp
+  src/core/icorefriendmessagesender.h
+  src/core/icoregroupmessagesender.cpp
+  src/core/icoregroupmessagesender.h
+  src/core/icoregroupquery.cpp
+  src/core/icoregroupquery.h
+  src/core/icoreidhandler.cpp
+  src/core/icoreidhandler.h
   src/core/toxcall.cpp
   src/core/toxcall.h
   src/core/toxencrypt.cpp
@@ -262,11 +280,13 @@ set(${PROJECT_NAME}_SOURCES
   src/ipc.h
   src/model/about/aboutfriend.cpp
   src/model/about/aboutfriend.h
+  src/model/about/iaboutfriend.cpp
   src/model/about/iaboutfriend.h
   src/model/chathistory.cpp
   src/model/chathistory.h
   src/model/chatlogitem.cpp
   src/model/chatlogitem.h
+  src/model/chatroom/chatroom.cpp
   src/model/chatroom/chatroom.h
   src/model/chatroom/friendchatroom.cpp
   src/model/chatroom/friendchatroom.h
@@ -274,11 +294,15 @@ set(${PROJECT_NAME}_SOURCES
   src/model/chatroom/groupchatroom.h
   src/model/contact.cpp
   src/model/contact.h
+  src/model/dialogs/idialogs.cpp
   src/model/dialogs/idialogs.h
+  src/model/dialogs/idialogsmanager.h
+  src/model/dialogs/idialogsmanager.cpp
   src/model/exiftransform.cpp
   src/model/exiftransform.h
   src/model/friendlist/friendlistmanager.cpp
   src/model/friendlist/friendlistmanager.h
+  src/model/friendlist/ifriendlistitem.cpp
   src/model/friendlist/ifriendlistitem.h
   src/model/friendmessagedispatcher.cpp
   src/model/friendmessagedispatcher.h
@@ -290,6 +314,7 @@ set(${PROJECT_NAME}_SOURCES
   src/model/groupmessagedispatcher.h
   src/model/group.cpp
   src/model/group.h
+  src/model/ibootstraplistgenerator.cpp
   src/model/ibootstraplistgenerator.h
   src/model/ichatlog.h
   src/model/imessagedispatcher.h
@@ -297,6 +322,7 @@ set(${PROJECT_NAME}_SOURCES
   src/model/message.h
   src/model/notificationgenerator.cpp
   src/model/notificationgenerator.h
+  src/model/profile/iprofileinfo.cpp
   src/model/profile/iprofileinfo.h
   src/model/profile/profileinfo.cpp
   src/model/profile/profileinfo.h
@@ -317,7 +343,12 @@ set(${PROJECT_NAME}_SOURCES
   src/persistence/db/rawdatabase.h
   src/persistence/history.cpp
   src/persistence/history.h
+  src/persistence/ifriendsettings.cpp
   src/persistence/ifriendsettings.h
+  src/persistence/igroupsettings.cpp
+  src/persistence/igroupsettings.h
+  src/persistence/inotificationsettings.cpp
+  src/persistence/inotificationsettings.h
   src/persistence/offlinemsgengine.cpp
   src/persistence/offlinemsgengine.h
   src/persistence/paths.cpp
@@ -342,6 +373,7 @@ set(${PROJECT_NAME}_SOURCES
   src/video/camerasource.h
   src/video/corevideosource.cpp
   src/video/corevideosource.h
+  src/video/ivideosettings.cpp
   src/video/ivideosettings.h
   src/video/netcamview.cpp
   src/video/netcamview.h
--- a/audio/CMakeLists.txt
+++ b/audio/CMakeLists.txt
@@ -20,6 +20,8 @@ set(SOURCE_FILES
     "include/audio/iaudiosettings.h"
     "include/audio/iaudiosink.h"
     "include/audio/iaudiosource.h"
+    "src/iaudiosettings.cpp"
+    "src/iaudiosink.cpp"
     "src/audio.cpp"
     "src/backend/alsink.cpp"
     "src/backend/alsink.h"
--- a/audio/include/audio/iaudiosettings.h
+++ b/audio/include/audio/iaudiosettings.h
@@ -25,7 +25,12 @@
 
 class IAudioSettings {
 public:
-    virtual ~IAudioSettings() = default;
+    IAudioSettings() = default;
+    virtual ~IAudioSettings();
+    IAudioSettings(const IAudioSettings&) = default;
+    IAudioSettings& operator=(const IAudioSettings&) = default;
+    IAudioSettings(IAudioSettings&&) = default;
+    IAudioSettings& operator=(IAudioSettings&&) = default;
 
     virtual QString getInDev() const = 0;
     virtual void setInDev(const QString& deviceSpecifier) = 0;
--- a/audio/include/audio/iaudiosink.h
+++ b/audio/include/audio/iaudiosink.h
@@ -96,7 +96,13 @@ public:
         return {};
     }
 
-    virtual ~IAudioSink() = default;
+    IAudioSink() = default;
+    virtual ~IAudioSink();
+    IAudioSink(const IAudioSink&) = default;
+    IAudioSink& operator=(const IAudioSink&) = default;
+    IAudioSink(IAudioSink&&) = default;
+    IAudioSink& operator=(IAudioSink&&) = default;
+
     virtual void playAudioBuffer(const int16_t* data, int samples, unsigned channels,
                                  int sampleRate) const = 0;
     virtual void playMono16Sound(const Sound& sound) = 0;
--- a/audio/src/iaudiosettings.cpp
+++ b/audio/src/iaudiosettings.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright © 2020 by The qTox Project Contributors
+    Copyright © 2021 by The qTox Project Contributors
 
     This file is part of qTox, a Qt-based graphical interface for Tox.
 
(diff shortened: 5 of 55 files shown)


jwestell:
By explicitly adding a destructor declaration (even if defaulted in the .cpp file), the implicitly defined move constructor will no longer be generated, thus removing any possibility of move semantics. 

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4713.pdf
> 8 If the definition of a class X does not explicitly declare a move constructor, a non-explicit one will be implicitly
declared as defaulted if and only if
(8.1) — X does not have a user-declared copy constructor,
(8.2) — X does not have a user-declared copy assignment operator,
(8.3) — X does not have a user-declared move assignment operator, and
(8.4) — X does not have a user-declared destructor.

Is removing automatic move semantics worth the strong vtables? Perhaps consider defining default move constructors and move-assignment operators for applicable concrete classes in this PR to retain this behaviour.

anthonybilinski:
Good point @jwestell. To be honest I don't think we are very concerned about link time savings that this warning is suggesting, but also not very concerned about move constructor or inline savings...

So I think the options are
1) Current method, slower linking due to repeated vtable in each object
2) Current proposed version, loses copy constructor and inlining of the destructor.
3) @jwestell method of explicitly defining the copy constructors as default, still losing inlining of the destructor
4) Adding a dummy defined function to each class defined in the source file.

What a mess of a language :D I think of these, 3 is probably best, adding a `= default` definition for the move constructor and move assignment operator. It addresses the clang warning and conforms to llvm coding guidelines for what that's worth, but it's still low impact on us either way.

anthonybilinski (changes requested):
> Reviewed 5 of 44 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6394)* status: 0 of 1 LGTMs obtained

___
*[src/core/icoregroupquery.cpp, line 3 at r1](https://reviewable.io/reviews/qtox/qtox/6394#-MmWiPKS0uDjViqnXy6u:-MmWiPKS0uDjViqnXy6v:bokzboz) ([raw file](https://github.com/qtox/qtox/blob/cdf4f7440c2386db45da1b8f9ff4b18594bb0e26/src/core/icoregroupquery.cpp#L3)):*
> ```cpp
> /*
>     Copyright © 2013 by Maxim Biro <nurupo.contributions@gmail.com>
>     Copyright © 2014-2019 by The qTox Project Contributors
> ```

For new files in general, the date should just be the current year. Also looks like here you copy pasted in nurupo's copyright from another file

brandonmosher:
Defaulting the constructor in this way does in fact make it user-provided rather than implicitly declared: "A function is user-provided if it is user-declared and not explicitly defaulted or deleted on its first declaration."

In addition to explicitly defaulting the move constructor, it may be wise to explicitly default the copy constructor and copy assignment operator as their implicit declaration is now deprecated for classes with a user-declared destructor.

As far as I can tell, a user provided destructor does not prevent implicit declaration of the move assignment operator. That said, it can't hurt to explicitly default it as well.

I propose adding the following to the affected class headers. This has the added benefit of not having to think about which special member functions the compiler is implicitly generating.

T(const T&) = default;
T& operator=(const T&) = default;
T(T&&) = default;
T& operator=(T&&) = default;

If this is more trouble than it's worth, we can also abandon this PR :)

jwestell:
> As far as I can tell, a user provided destructor does not prevent implicit declaration of the move assignment operator. That said, it can't hurt to explicitly default it as well.

A little further down in that standard document I linked above (S15.8.2.4) it also mentions the same thing for move assignment operators.

> In addition to explicitly defaulting the move constructor, it may be wise to explicitly default the copy constructor and copy assignment operator as their implicit declaration is now deprecated for classes with a user-declared destructor.

Agreed. This is in line with the [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three) guidance.

brandonmosher:
Any thoughts on whether it is also worth following [C.67: A polymorphic class should suppress public copy/move](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c67-a-polymorphic-class-should-suppress-public-copymove)?

This would require changes in addition to simply adding the defaulted special member functions. The public copy and move assignment operators in the ContactId class, for example, seem to violate this rule. In this case, if ContactId is always passed by reference, as it appears to be, the change should be trivial.

anthonybilinski:
I don't think pure virtual classes like `ContactId` are vulnerable to slicing since it can't be instantiated without a derived type since `getSize` would be undefined, and I think that's the case for most (but not all) polymorphic classes we have.

In general though, it seems worth following for classes that are vulnerable to slicing.

Re:
> If this is more trouble than it's worth, we can also abandon this PR :)

I'd be happy to accept the change, so as long as you're willing to complete it no need to abandon it. If it's getting to be too much of a pain on your side though, feel free :P

brandonmosher:
Not a pain at all, just want to make sure the extra lines of code are warranted. I went ahead and pushed the changes.

anthonybilinski (changes requested):
> Reviewed 3 of 44 files at r1, 37 of 37 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6394)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6394#-MmqKbV7BE8W8t-xH2jj:-MmqKbV7BE8W8t-xH2jk:b-47t8ei) (no related file):*
Building with clang and `-Wweak-vtables`, I'm still seeing warnings for `IAudioSink`, `ImagePreviewButton` (newly added since start of PR), and `IAudioSink`, and a couple minor formatting issues.

Otherwise looking good.

---
___
*[src/chatlog/chatmessage.h, line 55 at r2](https://reviewable.io/reviews/qtox/qtox/6394#-MmqLgJW-tZjLyRDcEQ1:-MmqLgJX3yY7JeuF_ZrQ:b8zg5oq) ([raw file](https://github.com/qtox/qtox/blob/7050eaeeb04237864317d3bca51d29856c1fd620/src/chatlog/chatmessage.h#L55)):*
> ```objc
>     ChatMessage& operator=(const ChatMessage&) = default;
>     ChatMessage(ChatMessage&&) = default;
>     
> ```

avoid trailing spaces at end of line
___
*[src/core/icoreextpacket.cpp, line 25 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqMGR2CV2DGqPjOuwT:-MmqMGR3754mki9N-Mpc:b-eomijg) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/core/icoreextpacket.cpp#L25)):*
> ```cpp
> ICoreExtPacket::~ICoreExtPacket() = default;
> 
> ICoreExtPacketAllocator::~ICoreExtPacketAllocator() = default;
> ```

all lines should end in newlines, including the last lines of a file
___
*[src/core/icoregroupmessagesender.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqMvC62E-WTtQPjn8B:-MmqMvC62E-WTtQPjn8C:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/core/icoregroupmessagesender.cpp#L22)):*
> ```cpp
> #include "icoregroupmessagesender.h"
> 
> ICoreGroupMessageSender::~ICoreGroupMessageSender() = default;
> ```

same re: missing newline at end of file
___
*[src/core/icoregroupquery.h, line 40 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqMwOO0g3xYiSVpOKG:-MmqMwOO0g3xYiSVpOKH:b-slzui3) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/core/icoregroupquery.h#L40)):*
> ```objc
>     ICoreGroupQuery(ICoreGroupQuery&&) = default;
>     ICoreGroupQuery& operator=(ICoreGroupQuery&&) = default;
>     
> ```

same: re trailing spaces
___
*[src/core/icoregroupquery.cpp, line 3 at r1](https://reviewable.io/reviews/qtox/qtox/6394#-MmWiPKS0uDjViqnXy6u:-MmqExl47x4B8m5Uokde:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/cdf4f7440c2386db45da1b8f9ff4b18594bb0e26/src/core/icoregroupquery.cpp#L3)):*

For new files in general, the date should just be the current year. Also looks like here you copy pasted in nurupo's copyright from another file
</blockquote></details>

Done.
___
*[src/core/icoregroupquery.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqMgCH8tObcV1ukWNZ:-MmqMgCH8tObcV1ukWN_:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/core/icoregroupquery.cpp#L22)):*
> ```cpp
> #include "icoregroupquery.h"
> 
> ICoreGroupQuery::~ICoreGroupQuery() = default;
> ```

same re: missing newline at end of file
___
*[src/core/icoreidhandler.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNIbo3hw8J42uH72f:-MmqNIbo3hw8J42uH72g:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/core/icoreidhandler.cpp#L22)):*
> ```cpp
> #include "icoreidhandler.h"
> 
> ICoreIdHandler::~ICoreIdHandler() = default;
> ```

same re: missing newline at end of file
___
*[src/model/about/iaboutfriend.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNNTjDCHinRLvLnK0:-MmqNNTjDCHinRLvLnK1:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/model/about/iaboutfriend.cpp#L22)):*
> ```cpp
> #include "iaboutfriend.h"
> 
> IAboutFriend::~IAboutFriend() = default;
> ```

same re: missing newline at end of file
___
*[src/model/dialogs/idialogs.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNQ1FBYPItyDq7x0E:-MmqNQ1FBYPItyDq7x0F:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/model/dialogs/idialogs.cpp#L22)):*
> ```cpp
> #include "idialogs.h"
> 
> IDialogs::~IDialogs() = default;
> ```

same re: missing newline at end of file
___
*[src/model/dialogs/idialogsmanager.h, line 36 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNQrT8GiSsb0Akj0p:-MmqNQrU1ISMBamLxjTW:btpm72b) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/model/dialogs/idialogsmanager.h#L36)):*
> ```objc
>     IDialogsManager(IDialogsManager&&) = default;
>     IDialogsManager& operator=(IDialogsManager&&) = default;
>     
> ```

same re: trailing spaces
___
*[src/model/dialogs/idialogsmanager.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNSdU4AH12T1bunm8:-MmqNSdU4AH12T1bunm9:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/model/dialogs/idialogsmanager.cpp#L22)):*
> ```cpp
> #include "idialogsmanager.h"
> 
> IDialogsManager::~IDialogsManager() = default;
> ```

same re: missing newline at end of file
___
*[src/model/friendlist/ifriendlistitem.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNU9-2yuvBROq0wZL:-MmqNU9-2yuvBROq0wZM:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/model/friendlist/ifriendlistitem.cpp#L22)):*
> ```cpp
> #include "ifriendlistitem.h"
> 
> IFriendListItem::~IFriendListItem() = default;
> ```

same re: missing newline at end of file
___
*[src/persistence/ifriendsettings.cpp, line 22 at r3](https://reviewable.io/reviews/qtox/qtox/6394#-MmqNWui8fNsl7lI9t9Y:-MmqNWui8fNsl7lI9t9Z:b-k922xl) ([raw file](https://github.com/qtox/qtox/blob/63c8a7b92ef8ba7392fd0c36c70049aecf5fa505/src/persistence/ifriendsettings.cpp#L22)):*
> ```cpp
> #include "ifriendsettings.h"
> 
> IFriendSettings::~IFriendSettings() = default;
> ```

same re: missing newline at end of file

brandonmosher:
Changes made. I left the `-Wweak-vtables` flag set in `CMakeLists.txt` to warn when future classes fail to comply with this pattern. `MockGroupQuery` and `MockCoreIdHandler` are still generating warnings as a result. I could apply the same changes to these classes, but I'm not sure how to configure CMake to include additional CPP files when building tests, or if that's even desired.

anthonybilinski (changes requested):
re: testing, in `cmake/Testing.cmake`, we define each test and what deps it requires with the function `auto_test`. Its last argument allows for arbitrary additional source files to be given for any individual test. I think using that would allow for fixing this in tests as well, and passing in mock source files needed for the test.

> Reviewed 20 of 20 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6394#-:-Mmt4e334UV_8U2gKQ2P:bynpr4l)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 96 at r4](https://reviewable.io/reviews/qtox/qtox/6394#-MmszXA-7lMwurhdihoF:-MmszXA-7lMwurhdihoG:b-1xluf) ([raw file](https://github.com/qtox/qtox/blob/7dc57774f3b16a72051349dcdc33990d140022a0/CMakeLists.txt#L96)):*
> ```CMake
> set(POSITION_INDEPENDENT_CODE True)
> 
> set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wweak-vtables")
> ```

`-Wweak-vtables` is only available on clang, so this addition causes compilation to fail with gcc. We'd need to conditionally add the warning only for clang or compilers that support it.

It looks like cmake's `CHECK_CXX_COMPILER_FLAG` would allow for this.

brandonmosher:
OK, I think I've got everything working. Now compiling in gcc and clang with no warnings.

anthonybilinski (approved):
:lgtm_strong: 

GH actions CI seems to not be re-running :thinking: I'll try pushing a dummy commit to your branch and then force-push stripping that dummy commit off, sometimes that makes it re-evaluate.

> Reviewed 6 of 6 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6394#-:-MmyomJg8WHfqCYNNkuB:bz78an8)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
Thanks for the change and follow up, @brandonmosher!

PR #6416, opened 2021-12-07
Update Windows dependencies

nurupo:
Mainly motivated by the new Qt 5.12.12 release. Toxcore happened to release while I was bumping the deps, so that's nice too.

Note that SQLCipher has [disabled `PRAGMA cipher_memory_security` in 4.5.0](https://github.com/sqlcipher/sqlcipher/blob/master/CHANGELOG.md#450---october-2021---450-changes), which has been [enabled by default since 4.0.0](https://github.com/sqlcipher/sqlcipher/commit/c40ffaa964666a860068e27ed15e29b328b467df). [When enabled, it securely overwrites the memory before deallocating it, to avoid the secrets remaining in the unallocated memory. It also prevents from the memory being written to swap, so it doesn't hit the disk.](https://discuss.zetetic.net/t/what-is-the-purpose-of-pragma-cipher-memory-security/3953/2) Sounds like something we would want to keep, so I have added it to the list of PRAGMAs qTox enables for SQLCipher 4.x.

Tested that both i686 and x86_64, both debug and release build locally. Tested that i686 and x86_64 release `qtox.exe` binaries run, go online (connect to DHT) and that the desktop capture works in Settings. Didn't add any contacts (zero contacts) or test anything beyond that. Tested under Wine on Linux.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6416)


nurupo:
Tbh I'm surprised that SQLCipher disabled something as important as `cipher_memory_security` in a minor release version. That's clearly a security degradation. If qTox replies on some other SQLCipher defaults, it might make sense to explicitly set them before SQLCipher changes them under us. Users on rolling distros with SQLCipher 4.5.0 available might already be affected by this...

anthonybilinski (approved):
Looks like Arch at least hasn't updated to 4.5.0 yet, but agreed that it's concerning. Opened https://github.com/qTox/qTox/issues/6421 to take a look and explicitly set all needed sqlcipher params ourselves.

:lgtm_strong: 

> Reviewed 1 of 1 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6416#-:-Mqs9Wio9-8jx2cWhlBq:beceosb)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6447, opened 2022-02-09
fix(docs): Add brew prefix to macOS build instructions

anthonybilinski:
Update brew's qt5 to qt@5, following brew naming

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6447)


INSTALL.md
```diff
@@ -446,7 +446,7 @@ Finally, build qTox.
 ```bash
 mkdir -p _build
 cd _build
-cmake ..
+cmake .. -DCMAKE_PREFIX_PATH=$(brew --prefix qt@5)
```
robinlinden, line 449: Both these places use `qt5` and not `qt@5`. Should they be updated?
Is this really needed when you seem to fish out the QT_PREFIX_PATH in your main CMakeLists.txt even without providing it like this?
https://github.com/qTox/qTox/blob/0cb222c0567698d70d4115321def32794a874136/.ci-scripts/build-osx.sh#L28
https://github.com/qTox/qTox/blob/0cb222c0567698d70d4115321def32794a874136/CMakeLists.txt#L65

anthonybilinski: Ah interesting, I didn't realize we had that in our CMakeLists. We used to have it in our user build instructions (implicitly) before when it was handled in `qTox-Mac-Deployer-ULTIMATE.sh` which was removed in ff2cad3b03841775178aaea004c1abb36568e9d7. I'll look at fixing(?) our cmake detection instead. And will update those other occurrences of `qt5`.

anthonybilinski: I've updated the two other uses of `qt5`, but split out fixing the CMakeLists.txt check to https://github.com/qTox/qTox/issues/6505. I haven't looked into why that's broken, but for sure the current build does require explicitly setting `-DCMAKE_PREFIX_PATH="$(brew --prefix qt5)".

robinlinden (approved):

anthonybilinski (commented):
Approving approval: :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6447#-:-MxNfBo6BXWCTGWXkbGf:b-ux2d4e)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6448, opened 2022-02-09
fix(Windows): Add missed openal dll since CI refactor

anthonybilinski:
Old windows/cross-compile/build.sh copied OPENAL_PREFIX_DIR/bin/*.dll to be
included, but the current Dockerfile.windows_builder copies libs one by one
and misses OpenAL. qTox fails to start on launch with due to missing
OpenAL32.dll on Windows because of this.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6448)

### DIFF
--- a/buildscripts/docker/Dockerfile.windows_builder
+++ b/buildscripts/docker/Dockerfile.windows_builder
@@ -191,6 +191,7 @@ RUN mkdir /export && \
   cp /windows/bin/libtoxcore.dll /export && \
   cp /windows/bin/libopus-*.dll /export && \
   cp /windows/lib/libvpx.dll /export && \
+  cp /windows/bin/OpenAL32.dll /export && \
   cp /windows/bin/libssl-*.dll /export && \
   cp /windows/bin/libsnore-qt5.dll /export && \
   mkdir -p /export/libsnore-qt5/ && \
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -219,3 +219,11 @@ then
   echo "Error: Missing some dlls."
   exit 1
 fi
+
+# Check that OpenAL is bundled. It is availabe from WINE, but not on Windows systems
+if grep -q '/root/.wine/drive_c/windows/system32/openal32.dll' dlls-required
+then
+  cat dlls-required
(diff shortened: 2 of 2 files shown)


nurupo:
How come [the mingw-ldd dll check](https://github.com/qTox/qTox/blob/65c42ebd9e2cbc5addab2f1f2502c2e1d5022a8e/windows/cross-compile/build.sh#L161-L216) missed it?

anthonybilinski:
Because the missing dll check is looking for "not found":
```
# Check that no dll is missing
if grep -q 'not found' dlls-required
then
  cat dlls-required
  echo "Error: Missing some dlls."
  exit 1
fi
```
but OpenAL is found in the WINE system dir:
```
# cat dlls-required | grep OpenAL
├── OpenAL32.dll => /root/.wine/drive_c/windows/system32/openal32.dll
```
Which causes it to both not error and not bundle. This is expected for other dlls from the WINE dir, which are found on the target system OS-wide.

So it doesn't seem like an automated check can't really guard against this, since the special case of a lib being provided by WINE but not by a real Windows install is what's causing the mismatch.

nurupo:
Oh wow, didn't expect Wine to have OpenAL32.dll in system32. Yeah, not much we can do about that. We do need to include system32 in the search path, since that's where Windows system dlls are located, otherwise those won't be found.

sphaerophoria:
> So it doesn't seem like an automated check can't really guard against this, since the special case of a lib being provided by WINE but not by a real Windows install is what's causing the mismatch.

Just a note that I don't think this is true. We can filter out the system32 version since we know it exists.

anthonybilinski:
True, I was thinking only in the generic check way. But having an additional check for libs like this (currently just OpenAL) seems reasonable. Updated to include that.

nurupo (approved):

anthonybilinski (commented):
Approving approval: :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6448#-:-Mw81tzqC72eehO0FVXv:b-ux2d4e)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6449, opened 2022-02-10
Fix portable

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6449)

### DIFF
--- a/.github/workflows/build-test-deploy.yaml
+++ b/.github/workflows/build-test-deploy.yaml
@@ -105,7 +105,7 @@ jobs:
         with:
           docker_image_name: almalinux
       - name: Run build
-        run: docker compose run --rm almalinux ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
+        run: docker-compose run --rm almalinux ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
   build-fedora:
     name: Fedora
     runs-on: ubuntu-latest
@@ -121,7 +121,7 @@ jobs:
         with:
           docker_image_name: fedora
       - name: Run build
-        run: docker compose run --rm fedora ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
+        run: docker-compose run --rm fedora ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
   build-opensuse:
     name: Opensuse
     runs-on: ubuntu-latest
@@ -137,7 +137,7 @@ jobs:
         with:
           docker_image_name: opensuse
       - name: Run build
-        run: docker compose run --rm opensuse ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
+        run: docker-compose run --rm opensuse ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
   build-debian:
     name: Debian
     runs-on: ubuntu-latest
@@ -153,7 +153,7 @@ jobs:
         with:
           docker_image_name: debian
       - name: Run build
-        run: docker compose run --rm debian ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
+        run: docker-compose run --rm debian ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
   build-debian-old:
     name: Debian Old
     runs-on: ubuntu-latest
@@ -169,7 +169,7 @@ jobs:
         with:
           docker_image_name: debian_old
       - name: Run build
-        run: docker compose run --rm debian_old ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
+        run: docker-compose run --rm debian_old ./.travis/build-qtox-linux.sh --build-type ${{ matrix.build_type }} --${{ matrix.features }}
   build-ubuntu:
     name: Ubuntu LTS
     runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


anthonybilinski:
Manual testing completed:
1. Tested running a new profile on a fresh VM with no existing profiles.
2. Created a new profile and connected
    * tox profile, global settings, db, user settings, and bootstrap list were all created under ~/.config/tox as expected.
3. Changed to portable mode under advanced settings.
4. Before closing, added a friend and exchanged some messages.
5. Closed qTox
    * tox profile, global settings, db, user settings were all saved to portable location
    * a warning was logged about the lock file not being found
6. Opened qTox
    * logs show everything is loaded from the local location
    * the bootstrap file is re-created in the local location
    * friend from last chat is still added, chat history is still present
7. Disabled portable mode
    * Things continue to work, are saved at regular location on shutdown, load on next start.

So just two slightly unexpected things
* The old location's files stay there are become stale, but can be replaced when the setting is reverted.
* The lock file being lost is a bit sloppy.

Both of these are existing behaviour prior to the regression.

sudden6 (changes requested):
> Reviewed 1 of 1 files at r1, 2 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6449)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/paths.h, line 63 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcLt4jC8t0HeSvIDS-:-MvcLt4jC8t0HeSvIDS0:b2y5qp2) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.h#L63)):*
> ```objc
> private:
>     QString basePath{};
>     std::atomic_bool portable{false};
> ```

Why atomic? I don't see any use case where this class needs to keep atomic state. If this is modified concurrently it might have other problems.
___
*[src/persistence/paths.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcMF-v5UeY2vuSOiQi:-MvcMF-v5UeY2vuSOiQj:bl13dls) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.cpp#L99)):*
> ```cpp
>         break;
>     case Portable::Auto:
>         if (!QFile(portableSettingsPath).exists()) {
> ```

remove inversion for better readability?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6449)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/paths.h, line 63 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcLt4jC8t0HeSvIDS-:-MvcN-xZ0C1f10yR_wGr:b-bbruzk) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.h#L63)):*

Why atomic? I don't see any use case where this class needs to keep atomic state. If this is modified concurrently it might have other problems.
</blockquote></details>

I think it's required because `portable` can be set from UI thread on settings change, and is read every time any path is fetched, which happens from a variety of threads.

There is still some thread unsafety with the load of settings, change of portable mode, save settings to new path pattern, but since no other classes are or should change `portable` state nor save settings, that seems fairly low risk.
___
*[src/persistence/paths.cpp, line 99 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcMF-v5UeY2vuSOiQi:-MvcNkzq2tr603oGYJ_q:b7h0qsc) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.cpp#L99)):*

remove inversion for better readability?
</blockquote></details>

Agreed, moved from Settings :P

Updated.

sudden6 (changes requested):
:lgtm_strong: 

> Reviewed 5 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6449#-:-MvcSHPS4H3VvQvZ5xN7:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/paths.h, line 63 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcLt4jC8t0HeSvIDS-:-MvcRqyC75RvGvKbpy2R:bubu80t) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.h#L63)):*

I think it's required because `portable` can be set from UI thread on settings change, and is read every time any path is fetched, which happens from a variety of threads.

There is still some thread unsafety with the load of settings, change of portable mode, save settings to new path pattern, but since no other classes are or should change `portable` state nor save settings, that seems fairly low risk.
</blockquote></details>

Hmm, allowing such an important setting to be changed any time is most probably problematic. I guess it's acceptable as a workaround for now though.
___
*[src/persistence/paths.cpp, line 135 at r5](https://reviewable.io/reviews/qtox/qtox/6449#-MvcSCWx8bCfTwJ2r3mD:-MvcSCWx8bCfTwJ2r3mE:bzby2c) ([raw file](https://github.com/qtox/qtox/blob/360f9258bc36aa6464734a0c1b4b5a4ae2d247a5/src/persistence/paths.cpp#L135)):*
> ```cpp
> {
>     auto oldVal = portable.exchange(newPortable);
>     if (oldVal != newPortable)
> ```

nit: missing braces

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6449)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/paths.h, line 63 at r2](https://reviewable.io/reviews/qtox/qtox/6449#-MvcLt4jC8t0HeSvIDS-:-MvcU19_03CyjxkWP-N5:b-9nsa3q) ([raw file](https://github.com/qtox/qtox/blob/4b19b3e374708076dc370f567b62eecac21fcbcc/src/persistence/paths.h#L63)):*

Hmm, allowing such an important setting to be changed any time is most probably problematic. I guess it's acceptable as a workaround for now though.
</blockquote></details>

It does seem a bit scary. To be fair, before these things were moved into `Paths`, the paths were switched on setting change from UI just in `Settings` instead, so it's not really a change of behaviour from then.

It does seem to work alright since the tox profile, history etc all get saved to the new location right away, and other classes aren't reading those directly, but it seems like a bit of a precarious situation.
___
*[src/persistence/paths.cpp, line 135 at r5](https://reviewable.io/reviews/qtox/qtox/6449#-MvcSCWx8bCfTwJ2r3mD:-MvcV39n2ONwbF-LBwNk:bt2vr1b) ([raw file](https://github.com/qtox/qtox/blob/360f9258bc36aa6464734a0c1b4b5a4ae2d247a5/src/persistence/paths.cpp#L135)):*

nit: missing braces
</blockquote></details>

Fixed.

PR #6451, opened 2022-02-11
fix(macOS): Update video API usage for newer libavcodec

anthonybilinski:
Newer version of avformat_open_input, av_find_input_format,
avcodec_find_decoder previously used non-const pointers that are now
const. Support both version for compatibiltiy with other platforms.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6451)

### DIFF
--- a/src/video/cameradevice.cpp
+++ b/src/video/cameradevice.cpp
@@ -31,6 +31,9 @@ extern "C" {
 #include "cameradevice.h"
 #include "src/persistence/settings.h"
 
+// no longer needed when avformat version < 59 is no longer supported
+using AvFindInputFormatRet = decltype(av_find_input_format(""));
+
 #if defined(Q_OS_LINUX) || defined(Q_OS_FREEBSD)
 #define USING_V4L 1
 #else
@@ -71,8 +74,8 @@ extern "C" {
 
 QHash<QString, CameraDevice*> CameraDevice::openDevices;
 QMutex CameraDevice::openDeviceLock, CameraDevice::iformatLock;
-AVInputFormat* CameraDevice::iformat{nullptr};
-AVInputFormat* CameraDevice::idesktopFormat{nullptr};
+static AvFindInputFormatRet idesktopFormat{nullptr};
+static AvFindInputFormatRet iformat{nullptr};
 
 CameraDevice::CameraDevice(const QString& devName, AVFormatContext* context)
     : devName{devName}
@@ -92,7 +95,7 @@ CameraDevice* CameraDevice::open(QString devName, AVDictionary** options)
         goto out;
     }
 
-    AVInputFormat* format;
+    AvFindInputFormatRet format;
     if (devName.startsWith("x11grab#")) {
         devName = devName.mid(8);
         format = idesktopFormat;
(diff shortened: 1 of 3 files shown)


sphaerophoria (commented):
Okay, so if I understand correctly

* We cache the output of av_find_input_format, which may or may not return a const pointer
* We use this cached value as an input to avformat_open_input, which will require the same type as av_find_input_format.
* We cannot just unconditionally set our cached version to be const because avformat_open_input's api may say "hey I need a non-const input", so we've solved this with this avconst_59 typedef

Review comments:
* I think at the very least we should be applying this fix to idesktop format as well right? I think this same problem will pop up on linux and windows later and we just haven't hit it yet
* I think a typedef of the whole type is nicer than an avconst_59 typedef. I won't block a pass on this, but I do think it would read nicer
* Given that these are static values that are private, we could actually move them out of the class into a static in the cpp that typedefs with `using AvFindInputFormatRet = decltype(av_find_input_format(""));`. I'd argue that that's nicer but could see the argument that it's worse.

I only really want the first bullet here addressed too, but if you agree about 2/3 I'd be happy to see those changes too

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6451#-:-MvwX2_TAk1tPob8lk-6:b3cq2p5)* status: 0 of 1 LGTMs obtained

anthonybilinski (commented):
Good point about idesktop, I missed it because it wasn't failing, but it's absolutely in the same spot.

Took your second two recommendations too, quite a bit cleaner this way.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6451#-:-MvwwykqC9EY3nq7OQ24:btrhcch)* status: 0 of 1 LGTMs obtained

Geremia:
@anthonybilinski Thanks. [Your patch](https://patch-diff.githubusercontent.com/raw/qTox/qTox/pull/6451.patch) allowed me to build against FFmpeg 5 (libavdevice.so.59 etc.).

warisb:
Passerby trying to help out, but no approval privileges. I pondered using #ifdef to call out backwards compatibility just for the sake of documenting intention, but `decltype` is a very neat one-liner solution. The changes LGTM.

anthonybilinski (commented):
Approving review from non-writer-access-user to pass review check: :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6451#-:-Mw4QX__D4o1JlggKq9N:bukk2nv)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/video/cameradevice.cpp, line 34 at r3](https://reviewable.io/reviews/qtox/qtox/6451#-Mw4QTDg4ajBPo-5Vmhn:-Mw4QTDg4ajBPo-5Vmho:be745gt) ([raw file](https://github.com/qtox/qtox/blob/15673a52b6b4805d482b69281e21947fb7096e05/src/video/cameradevice.cpp#L34)):*
> ```cpp
> #include "src/persistence/settings.h"
> 
> // no longer needed when avformat version < 59 is no longer supported
> ```

I've added a comment to make that explicit, while still use the decltype method rather than ifdef.

PR #6458, opened 2022-02-15
Docker fixes

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6458)

### DIFF
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -115,6 +115,7 @@ if [[ $RUN_TESTS -ne 0 ]]
 then
   export WINEPATH='/export;/windows/bin'
   export CTEST_OUTPUT_ON_FAILURE=1
+  export PATH="$PATH:/opt/wine-stable/bin"
   make test
 fi
 set -u
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -30,11 +30,13 @@
 
 set -euo pipefail
 
+RUN_TESTS=0
 
 while (( $# > 0 )); do
     case $1 in
         --src-dir) QTOX_SRC_DIR=$2; shift 2 ;;
         --arch) ARCH=$2; shift 2 ;;
+        --run-tests) RUN_TESTS=1; shift ;;
         --build-type) BUILD_TYPE=$2; shift 2;;
         *) "Unexpected argument $1"; exit 1 ;;
     esac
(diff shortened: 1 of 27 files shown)


windows/cross-compile/build.sh
```diff
@@ -105,12 +109,13 @@ mkdir -p "${QTOX_PREFIX_DIR}"
 cp qtox.exe $QTOX_PREFIX_DIR
 cp -r /export/* $QTOX_PREFIX_DIR
 
-# Run tests (only on Travis)
+# Run tests
 set +u
-if [[ -n "$TRAVIS_CI_STAGE" ]]
+if [[ $RUN_TESTS -ne 0 ]]
 then
   export WINEPATH='/export;/windows/bin'
```
nurupo, line 116: Does it really work with such `WINEPATH`?
Based on my findings in https://github.com/qTox/qTox/pull/6413#discussion_r781066906 it shouldn't, but it seems like you were able to run tests successfully?

nurupo:
While we are at it, could we copy the source code for the debug build? https://github.com/qTox/qTox/pull/6413#discussion_r778661734

windows/cross-compile/build.sh
anthonybilinski: All I can say is that the tests do run yes, I have no problem with them running locally or in CI.

anthonybilinski:
I'm not really convinced of the benefit either. I understand the use case of gdb and the helper scripts, and those seem quite useful for new users or just users who don't want to deal with installing tools on Windows but regarding your point of:
> Including qTox the source code is useful for gdb to generate nice printouts of where the crash happens in the code and for the more advanced users (or even developers themselves) debugging why the crash is happening.

I feel like the nice printout doesn't add anything for the issue report, since with debug symbols that it already has we'll already have a fully visible back trace, visible variables, and the correct file and line number. It would be nice for a more advanced user who wants to investigate, but at that point I figure they'd want to browse the code rather than just see the gdb printouts, by either cloning it or even just browsing on github, and they could still follow along with gdb's output. It also just doesn't seem very standard for even debug applications to ship with bundled source code.

I don't think there's a big downside with doing it, package size I don't think is a big concern given it's the debug version, but I'm still leaning towards leaving it out.

nurupo (approved):

anthonybilinski (commented):
Approving review: :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6458#-:-Mw80oYk5a6oNuOZ9_Hx:b-c1zd3t)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6459, opened 2022-02-15
feat(CI): Run AddressSanitizer on Fedora build

anthonybilinski:
Keep disabled for jobs that produce artifacts to not effect user performance.
Platforms that do produce artifacts could be tested as well by doubling the
build jobs and using one to test and the other to produce artifacts.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6459)

### DIFF
--- a/.ci-scripts/build-qtox-linux.sh
+++ b/.ci-scripts/build-qtox-linux.sh
@@ -27,6 +27,7 @@ while (( $# > 0 )); do
     case $1 in
     --minimal) MINIMAL=1 ; shift ;;
     --full) MINIMAL=0; shift ;;
+    --sanitize) SANITIZE=1; shift ;;
     --build-type) BUILD_TYPE=$2; shift 2 ;;
     --help|-h) usage; exit 1 ;;
     *) echo "Unexpected argument $1"; usage; exit 1 ;;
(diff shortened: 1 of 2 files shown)


anthonybilinski:
Note our CI fails with ASAN currently, which (I think) is why these jobs are failing, though it's a bit hard to tell since `docker compose run` is outputting no logs as visible from github actions.

The error is Settings not freeing Paths that it constructs, which happens to be fixed by https://github.com/qTox/qTox/pull/6449 EDIT: fix is merged.

anthonybilinski:
And logging will be fixed by https://github.com/qTox/qTox/pull/6460 EDIT: fix is merged.

anthonybilinski:
Welp it turns out running a Qt app with ASAN and TSAN on a bunch of different distros with different Qt versions just ends up finding you a lot of Qt bugs rather than qTox bugs.

I'll see if I can get reliable qTox-related results on just a single distro since that should still cover almost all of qTox, if not will just close :(

anthonybilinski:
Running just ASAN just on our Fedora docker image gives reliably useful results. I think starting with just that make sense, that should still give us nearly the same level of coverage for qTox.

Current failure is due to https://github.com/qTox/qTox/pull/6472. EDIT: fix is merged.

sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6459#-:-MwcGVLG8uGQG7qB452W:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
Fedora build's not being reported on the PR despite it passing in the GH action. Force-pushing re-runs the action but doesn't update the check. Will try opening and closing :shrug:

anthonybilinski:
Ah the issue was that the job name change didn't match our required checks.

PR #6464, opened 2022-02-15
refactor(ToxPk): Remove ability to construct ToxId from ToxPk

anthonybilinski:
Construct ToxPk directly everywhere that used to construct through ToxId.

Now any ToxId should be a valid ToxId, and not possibly just a ToxPk.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6464)

### DIFF
--- a/src/core/toxid.cpp
+++ b/src/core/toxid.cpp
@@ -18,19 +18,21 @@
 */
 
 
+#include "contactid.h"
 #include "toxid.h"
 #include "toxpk.h"
 
 #include <tox/tox.h>
 
 #include <QRegularExpression>
+
+#include <cassert>
 #include <cstdint>
 
 // Tox doesn't publicly define these
 #define NOSPAM_BYTES 4
 #define CHECKSUM_BYTES 2
 
-#define PUBLIC_KEY_HEX_CHARS (2 * TOX_PUBLIC_KEY_SIZE)
 #define NOSPAM_HEX_CHARS (2 * NOSPAM_BYTES)
 #define CHECKSUM_HEX_CHARS (2 * CHECKSUM_BYTES)
 #define TOXID_HEX_CHARS (2 * TOX_ADDRESS_SIZE)
@@ -59,8 +61,7 @@ const QRegularExpression ToxId::ToxIdRegEx(QString("(^|\\s)[A-Fa-f0-9]{%1}($|\\s
  */
 ToxId::ToxId()
     : toxId()
-{
-}
+{}
 
 /**
  * @brief The copy constructor.
@@ -68,8 +69,7 @@ ToxId::ToxId()
  */
 ToxId::ToxId(const ToxId& other)
     : toxId(other.toxId)
-{
-}
+{}
 
 /**
  * @brief Create a Tox ID from a QString.
@@ -82,12 +82,10 @@ ToxId::ToxId(const ToxId& other)
  */
 ToxId::ToxId(const QString& id)
 {
-    // TODO: remove construction from PK only
     if (isToxId(id)) {
         toxId = QByteArray::fromHex(id.toLatin1());
-    } else if (id.length() >= PUBLIC_KEY_HEX_CHARS) {
-        toxId = QByteArray::fromHex(id.left(PUBLIC_KEY_HEX_CHARS).toLatin1());
     } else {
+        assert(!"ToxId constructed with invalid length string");
         toxId = QByteArray(); // invalid id string
     }
 }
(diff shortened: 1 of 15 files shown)


src/core/toxpk.cpp
```diff
@@ -60,6 +61,25 @@ ToxPk::ToxPk(const QByteArray& rawId)
...
 }
 
+/**
+ * @brief Constructs a ToxPk from a QString.
+ *
+  * If the given pk isn't a valid Public Key a ToxPk with all zero bytes is created.
+ *
+ * @param pk Tox Pk string to convert to ToxPk object
+ */
+ToxPk::ToxPk(const QString& pk)
```
robinlinden, line 83: Not sure if it's something you require in qTox, but this should go below the `ToxPk(const uint8_t*)` constructor if the order should match the order in the header.

src/core/toxpk.cpp
```diff
@@ -60,6 +61,25 @@ ToxPk::ToxPk(const QByteArray& rawId)
...
+    : ContactId([pk](){
```
robinlinden: No need to copy this here. A reference is enough.

src/core/contactid.h
```diff
@@ -24,6 +24,9 @@
 #include <cstdint>
 #include <QHash>
 #include <memory>
+#include <tox/tox.h>
+
+#define PUBLIC_KEY_HEX_CHARS (2 * TOX_PUBLIC_KEY_SIZE)
```
robinlinden: Isn't this only used in toxpk.cpp? Might be nicer to put it there?

src/core/toxid.cpp
```diff
@@ -126,12 +123,10 @@ ToxId::ToxId(const uint8_t* rawId, int len)
...
 void ToxId::constructToxId(const QByteArray& rawId)
 {
-    // TODO: remove construction from PK only
-    if (rawId.length() == TOX_PUBLIC_KEY_SIZE) {
-        toxId = QByteArray(rawId); // construct from PK only
-    } else if (rawId.length() == TOX_ADDRESS_SIZE && isToxId(rawId.toHex().toUpper())) {
+    if (rawId.length() == TOX_ADDRESS_SIZE && isToxId(rawId.toHex().toUpper())) {
         toxId = QByteArray(rawId); // construct from full toxid
     } else {
+        assert(!"ToxId constructed with invalid input");
```
robinlinden, line 130: Out of scope for this, but something I like doing is having a static create method that returns a `std::optional<MyType>` when I have a fallible constructor.

robinlinden (approved):

src/core/toxpk.cpp
anthonybilinski: Not something I've thought about generally, but updated because why not.

src/core/toxid.cpp
anthonybilinski: Agreed, ToxId almost has that with the (private) `void ToxId::constructToxId(const QByteArray& rawId)` but it still can't return null.

I'd also like to remove the `isValid` from ToxId and just have it fail construction if not valid, but will leave those for later.

anthonybilinski (commented):
Review comments address, approving Robin's review :lgtm:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6464#-:-Mw1sDqb0MW7-npRD8ag:b-bj5j63)* status: 0 of 1 LGTMs obtained, and 1 stale

___
*[src/core/toxpk.cpp, line 72 at r1](https://reviewable.io/reviews/qtox/qtox/6464#-MvzgiLXsWyCo777PNH5-r1-72:-Mw1s9Ni7GCWGwNR61LV:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/72975adac6fac397bff630bf9f2206f906ddf243/src/core/toxpk.cpp#L72)):*
<details><summary><i>Previously, robinlinden (Robin Lindén) wrote…</i></summary><blockquote>

No need to copy this here. A reference is enough.
</blockquote></details>

Done.
___
*[src/core/contactid.h, line 29 at r1](https://reviewable.io/reviews/qtox/qtox/6464#-MvzgiLXsWyCo777PNH2-r1-29:-Mw1sCGZ7onagiUw8nE_:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/72975adac6fac397bff630bf9f2206f906ddf243/src/core/contactid.h#L29)):*
<details><summary><i>Previously, robinlinden (Robin Lindén) wrote…</i></summary><blockquote>

Isn't this only used in toxpk.cpp? Might be nicer to put it there?
</blockquote></details>

Done.

anthonybilinski (commented):
Re-approving review with lgmt_strong for reviewable check :lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6464#-:-Mw7ajlx5ubKoV4VjrWD:baq6x4k)* status: :shipit: complete! 1 of 1 LGTMs obtained

src/core/toxid.cpp
anthonybilinski: For future reference, this seems like a no-go with Qt. Any class used in signals/slots must have a public default constructor, so a strict factory pattern isn't workable with ToxPk which we want to be able to pass around easily. This also means keeping the `isValid` is basically required, and the recommended "Qt" way of doing things. Oh well =/

PR #6465, opened 2022-02-16
fix(bootstrap): Add one time reset of user bootstrap node list

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)

### DIFF
--- a/src/persistence/globalsettingsupgrader.cpp
+++ b/src/persistence/globalsettingsupgrader.cpp
@@ -1,0 +1,57 @@
+/*
+    Copyright © 2022 by The qTox Project Contributors
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "globalsettingsupgrader.h"
+
+#include <QDebug>
+
+#include <cassert>
+
+bool GlobalSettingsUpgrader::doUpgrade(Settings& settings, int fromVer, int toVer)
+{
+    if (fromVer == toVer) {
+        return true;
+    }
+
+    if (fromVer > toVer) {
+        qWarning().nospace() << "Settings version (" << fromVer
+                             << ") is newer than we currently support (" << toVer
+                             << "). Please upgrade qTox";
+        return false;
+    }
+
+    using SettingsUpgradeFn = bool (*)(Settings&);
+    std::vector<SettingsUpgradeFn> upgradeFns = {};
+
+    assert(fromVer < static_cast<int>(upgradeFns.size()));
+    assert(toVer == static_cast<int>(upgradeFns.size()));
+
+    for (int i = fromVer; i < static_cast<int>(upgradeFns.size()); ++i) {
+        auto const newSettingsVersion = i + 1;
+        if (!upgradeFns[i](settings)) {
+            qCritical() << "Failed to upgrade settings to version " << newSettingsVersion << " aborting";
+            return false;
+        }
+        qDebug() << "Settings upgraded incrementally to schema version " << newSettingsVersion;
+    }
+
+    qInfo() << "Settings upgrade finished (settingsVersion" << fromVer << "->"
+            << toVer << ")";
+    return true;
+}
--- a/src/persistence/globalsettingsupgrader.cpp
+++ b/src/persistence/globalsettingsupgrader.cpp
@@ -18,8 +18,30 @@
 */
 
 #include "globalsettingsupgrader.h"
+#include "src/persistence/settings.h"
+#include "src/persistence/paths.h"
 
 #include <QDebug>
+#include <QFile>
+
+namespace {
+    bool version0to1(Settings& settings)
+    {
+        const auto& paths = settings.getPaths();
+
+        QFile badFile{paths.getUserNodesFilePath()};
+        if (badFile.exists()) {
+            if (!badFile.rename(paths.getBackupUserNodesFilePath())) {
+                qCritical() << "Failed to write to" << paths.getBackupUserNodesFilePath() << \
+                    "aborting bootstrap node upgrade.";
+                return false;
+            }
+        }
+        qWarning() << "Moved" << badFile.fileName() << "to" << paths.getBackupUserNodesFilePath() << \
+            "to mitigate a bug. Resetting bootstrap nodes to default.";
+        return true;
+    }
+} // namespace
 
 #include <cassert>
 
@@ -37,7 +59,7 @@ bool GlobalSettingsUpgrader::doUpgrade(Settings& settings, int fromVer, int toVe
     }
 
     using SettingsUpgradeFn = bool (*)(Settings&);
-    std::vector<SettingsUpgradeFn> upgradeFns = {};
+    std::vector<SettingsUpgradeFn> upgradeFns = {version0to1};
 
     assert(fromVer < static_cast<int>(upgradeFns.size()));
     assert(toVer == static_cast<int>(upgradeFns.size()));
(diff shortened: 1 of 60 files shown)


sudden6:
do you think it makes sense to version the settings too? I have not seen this elsewhere, but I can't come up with a good reason not to do it.

anthonybilinski (commented):
On reconsideration, yes I think that's best and I've changed it to work similar to `History`.

Upgrading is handled by Settings now too, rather than exposing the bug mitigation state to different classes.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465#-:-MxGKLiaEwQTPJZs27C_:b-g37uar)* status: 0 of 1 LGTMs obtained

src/persistence/settings.cpp
```diff
@@ -112,16 +117,34 @@ void Settings::loadGlobal()
...
         filePath = ":/conf/" + globalSettingsFile;
+        defaultSettings = true;
     }
 
     qDebug() << "Loading settings from " + filePath;
 
     QSettings s(filePath, QSettings::IniFormat);
     s.setIniCodec("UTF-8");
 
+    s.beginGroup("Version");
```
sphaerophoria, line 132: Why do you have a group with a single value? Is this just how ini's have to be formatted?

src/persistence/paths.cpp
```diff
@@ -365,11 +365,25 @@ QString Paths::getAppCacheDirPath() const
 #endif
 }
 
+QString Paths::getExampleNodesFilePath() const
+{
+    QDir dir(getSettingsDirPath());
+    constexpr static char nodesFileName[] = "bootstrapNodes.example.json";
```
sphaerophoria, line 371: This is a nice solution. Gives someone a reference without preventing us from updating it later.

src/persistence/globalsettingsupgrader.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <QDebug>
+#include <QFile>
+
+namespace {
+    bool version0to1(Settings& settings)
+    {
+        const auto& paths = settings.getPaths();
+
+        QFile badFile{paths.getUserNodesFilePath()};
+        if (badFile.exists()) {
```
sphaerophoria, line 33: Can we prompt the user for this? No one is going to see the qWarning

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 132 at r4](https://reviewable.io/reviews/qtox/qtox/6465#-MwP2HaK32e0anhnqnvN-r4-132:-MxHwF3A3xeenxck1AtO:b5xk9u4) ([raw file](https://github.com/qtox/qtox/blob/4febe25c27eeec2f9e5b49f0ca6bf70e97b77726/src/persistence/settings.cpp#L132)):*

Why do you have a group with a single value? Is this just how ini's have to be formatted?
</blockquote></details>

It's semi-required by the file format, see bottom of this section: https://doc.qt.io/qt-5/qsettings.html#Format-enum

If it was not explicitly put in a group in code, it would be under "General" in the file, which doesn't seem much different.
___
*[src/persistence/globalsettingsupgrader.cpp, line 33 at r6](https://reviewable.io/reviews/qtox/qtox/6465#-MxGwIuv2LeMZ955T49Y-r6-33:-MxHwa-p9ciWcLznQZJH:b-sjmqje) ([raw file](https://github.com/qtox/qtox/blob/867c21b7fabdaa19dd0bc33e85d9fb4c177d9d12/src/persistence/globalsettingsupgrader.cpp#L33)):*

Can we prompt the user for this? No one is going to see the qWarning
</blockquote></details>

We don't have a GUI at time of global settings load thanks to a mess of initialization order which wouldn't be easy to fix =/

I agree no one's going to see this initially, but they might look into logs or open an issue if qTox refuses to start, which this causes. Same issue in `Settings` where `GlobalSettingsUpgrader::doUpgrade` is called.

sudden6 (changes requested):
> Reviewed 50 of 59 files at r4, 8 of 9 files at r7, 4 of 4 files at r8, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/personalsettingsupgrader.cpp, line 26 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTuV0FBKdBaMY90J67:-MxTuV0FBKdBaMY90J68:btb4xej) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/personalsettingsupgrader.cpp#L26)):*
> ```cpp
> #include <cassert>
> 
> bool PersonalSettingsUpgrader::doUpgrade(SettingsSerializer& settingsSerializer, int fromVer, int toVer)
> ```

same, make these `unsigned`
___
*[src/persistence/settings.h, line 720 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTuZp34Ji8qdiwGvXt:-MxTuZp34Ji8qdiwGvXu:bb7od0r) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/settings.h#L720)):*
> ```objc
>     static QThread* settingsThread;
>     Paths paths;
>     int globalSettingsVersion = 0;
> ```

same
___
*[src/persistence/settings.cpp, line 493 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTupq7-uIiDhiDWpUh:-MxTupq7-uIiDhiDWpUi:b-r8h8u3) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/settings.cpp#L493)):*
> ```cpp
>     bool defaultSettings = false;
>     if (!QFile(filePath).exists()) {
>         defaultSettings = true;
> ```

move to the `else` path below
___
*[src/persistence/globalsettingsupgrader.cpp, line 26 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTuG8g4xn1wm9NeKy0:-MxTuG8g4xn1wm9NeKy1:b-3hznmd) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/globalsettingsupgrader.cpp#L26)):*
> ```cpp
> #include <cassert>
> 
> bool GlobalSettingsUpgrader::doUpgrade(Settings& settings, int fromVer, int toVer)
> ```

if you make these `unsigned int` you can probably remove the casts and I don't expect there will be negative versions.

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/personalsettingsupgrader.cpp, line 26 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTuV0FBKdBaMY90J67:-MxUCciABGxTJHAVHc5x:b7caqk4) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/personalsettingsupgrader.cpp#L26)):*

same, make these `unsigned`
</blockquote></details>

Retracted, based on out-of-band discussions with regards to Qt size types compatibility.

sudden6 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 493 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTupq7-uIiDhiDWpUh:-MxUGzSX9pQJp_3Mh4U0:b-60qyt5) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/settings.cpp#L493)):*

move to the `else` path below
</blockquote></details>

Sorry, I misread that and merging is not possible, but it can be converted to a oneliner: `const  bool defaultSettings = !QFile(filePath).exists()`

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/settings.cpp, line 493 at r7](https://reviewable.io/reviews/qtox/qtox/6465#-MxTupq7-uIiDhiDWpUh:-MxUCeJJ128pCGa9Iw1J:b-1xgh2e) ([raw file](https://github.com/qtox/qtox/blob/77d7d4c96b2bf1011af6643565b2fce2f9ac76fd/src/persistence/settings.cpp#L493)):*

Sorry, I misread that and merging is not possible, but it can be converted to a oneliner: `const  bool defaultSettings = !QFile(filePath).exists()`
</blockquote></details>

arg I was thinking to avoid the case where they have a global settings file but no personal settings file, but this would miss every new profile for a single user. Running it only if it's a new profile settings file could potentially miss the upgrade if a user has a global settings file but no user settings file though.. which looks like a possible case based on this loading behaviour, and digging through history I found that that was in aea9eea8a4e2f95a7175954183959c1070bcc315. So it looks like we have no perfect way to tell if it's a new profile based on Settings files.

That should really be done through an upgrade itself at some point...

For now though, I'll just pass in if it's a new profile from Profile to Settings::loadPersonal, since Profile is aware and is always doing the loading.

sudden6 (approved):
:lgtm_strong: 

> Reviewed 8 of 8 files at r10, 4 of 4 files at r11, 4 of 4 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6465#-:-MxUdsb92tIB_0zF47vf:b-uwkowd)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6468, opened 2022-02-16
fix(appImage): Bundle libpangocairo

### LINKED ISSUE #6454: AppImage doesn't start on Arch Linux
##### Brief Description

OS: Arch Linux
qTox version: v1.17.4+

Reproducible: Always

##### Steps to reproduce

1. Run AppImage

##### Observed Behavior
1. AppImage fails to start with `symbol lookup error: /usr/lib/libpangocairo-1.0.so.0: undefined symbol: pango_font_get_hb_font`

##### Additional Info
https://github.com/qTox/qTox/pull/6413 added:
```
/usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
/usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
/usr/lib/x86_64-linux-gnu/libpango-1.0.so.0
/usr/lib/x86_64-linux-gnu/libpangoft2-1.0.so.0
```
to the AppImage to address some (unsure) OS-specific incompatibility, which likely caused this regression on Arch.

AppImage is a constant whack-a-mole of adding libs that target OSes might not have, not adding too many libs that will bloat the AppImage, and still adding deps of added libs that will be a breakingly different version on some target OS.

I guess in this case we need to add some more deps of the added libs.

Overall the size savings of keeping out libs like this seems pretty minor compared to the constant upkeep and flakyness, but also seems like the general design goal of AppImage, and linuxdeployqt that we use should already bundle the libs it thinks we should have, so I don't think just bundling _everything_ is really the right fix.

It would be nice if we could add test coverage to make sure that the AppImage created starts on a variety of linux distros in CI.

anthonybilinski:
AppImage was failing to start on Arch due to this missing lib.

Fix #6454

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6468)


anthonybilinski:
Manually tested by downloading from action, launching, logging in, going to A/V settings pane which dlopens some more libs, just to ensure no crash due to missing libs on:
* Arch 2022-02-15
* OpenSUSE Tumbleweed
* Fedora 35 - KDE install
* Ubuntu 21.10
* AlmaLinux 8.5 - "Workstation" install

And had no issues on any of them.

sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6468#-:-MwGhlmq8cHRrXUp_DGb:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6474, opened 2022-02-18
fix(history): Replace invalid Tox ID saved in peers table with Tox public key

anthonybilinski:
Due to an old bug that has since be fixed, old history dbs can contain both a
Tox ID version and Tox public key version of the same friend, and always
themselves. They could have n more duplicates if they've updated their nospam.

Tox ID is an invalid length to be stored in strongly typed ToxPk, and in
general having multiple entries belonging to the same user effectively violates
our UNIQUE constraint on public_key.

Introduced in 7168d2b8589a3670f9b5c5d88d5a5d20f1bf5966

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6474)

### DIFF
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -361,9 +361,9 @@ void ChatHistory::loadHistoryIntoSessionChatLog(ChatLogIdx start) const
         // Note that message.id is _not_ a valid conversion here since it is a
         // global id not a per-chat id like the ChatLogIdx
         auto currentIdx = nextIdx++;
-        auto sender = ToxPk(message.sender);
         switch (message.content.getType()) {
         case HistMessageContentType::file: {
+            auto sender = ToxPk(message.sender);
             const auto date = message.timestamp;
             const auto file = message.content.asFile();
             const auto chatLogFile = ChatLogFile{date, file};
@@ -371,6 +371,7 @@ void ChatHistory::loadHistoryIntoSessionChatLog(ChatLogIdx start) const
             break;
         }
         case HistMessageContentType::message: {
+            auto sender = ToxPk(message.sender);
             auto messageContent = message.content.asMessage();
 
             auto isAction = handleActionPrefix(messageContent);
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -27,7 +27,7 @@
 #include "src/core/toxpk.h"
 
 namespace {
-static constexpr int SCHEMA_VERSION = 8;
+static constexpr int SCHEMA_VERSION = 9;
 
 bool createCurrentSchema(RawDatabase& db)
 {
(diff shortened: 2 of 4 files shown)


sudden6 (changes requested):
> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6474)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 421 at r1](https://reviewable.io/reviews/qtox/qtox/6474#-MwGVWUH7hQx1CQ1oxyt:-MwGVWUH7hQx1CQ1oxyu:b-6wq9kz) ([raw file](https://github.com/qtox/qtox/blob/9844e3b98da79aeb1404977e53a171041f06b39a/src/persistence/history.cpp#L421)):*
> ```cpp
> std::vector<BadEntry> getInvalidPeers(RawDatabase& db)
> {
>     std::vector<BadEntry> badPeerIds;
> ```

How many bad Peers do you expect in large dbs? enough to get memory problems? might also consider this for the other queries

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6474)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 421 at r1](https://reviewable.io/reviews/qtox/qtox/6474#-MwGVWUH7hQx1CQ1oxyt:-MwGv-D5Ctp-18zdQpUE:bygg3rw) ([raw file](https://github.com/qtox/qtox/blob/9844e3b98da79aeb1404977e53a171041f06b39a/src/persistence/history.cpp#L421)):*

How many bad Peers do you expect in large dbs? enough to get memory problems? might also consider this for the other queries
</blockquote></details>

For peers, I think 1 is basically a guarantee for old dbs since we used to insert selfId as a peer instead of selfPk. My personal db had two, for some reason, so I guess there was some bug that inserted friends in the past. Changing nospam would increase it by one each time, but I still think that would be in the order of 100 at most.

For aliases, the worst case would be if you set many aliases pre-fix of selfId, then re-set the same ones post-fix, but still probably in the order of 100.

The messages are the only one where I'm worried about it being in the ?? 10s, 100s of thousands or maybe millions, but those are all handled in sqlite which is good for us.

std::vector also heap allocates all of its members, so all together I don't think there's a serious risk here.

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6474#-:-MwO43WK7rI37xDQkTwA:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 421 at r1](https://reviewable.io/reviews/qtox/qtox/6474#-MwGVWUH7hQx1CQ1oxyt:-MwNItVy-_YpUKyoJU-n:bh2cpzk) ([raw file](https://github.com/qtox/qtox/blob/d85db5351f64d55bb80f6f56216ad34b2bca9671/src/persistence/history.cpp#L421)):*

For peers, I think 1 is basically a guarantee for old dbs since we used to insert selfId as a peer instead of selfPk. My personal db had two, for some reason, so I guess there was some bug that inserted friends in the past. Changing nospam would increase it by one each time, but I still think that would be in the order of 100 at most.

For aliases, the worst case would be if you set many aliases pre-fix of selfId, then re-set the same ones post-fix, but still probably in the order of 100.

The messages are the only one where I'm worried about it being in the ?? 10s, 100s of thousands or maybe millions, but those are all handled in sqlite which is good for us.

std::vector also heap allocates all of its members, so all together I don't think there's a serious risk here.
</blockquote></details>

Ok. millions sounds bad, but I guess modern PCs can handle that without running out of RAM
___
*[test/persistence/dbschema_test.cpp, line 57 at r2](https://reviewable.io/reviews/qtox/qtox/6474#-MwO4-L81vCIkoDxBk53:-MwO4-L81vCIkoDxBk54:bq51ow3) ([raw file](https://github.com/qtox/qtox/blob/f3fea27a450b35895cca648a55e1c6a01a883a2c/test/persistence/dbschema_test.cpp#L57)):*
> ```cpp
>     void test6to7();
>     // test7to8 omitted, version only upgrade, versions are not verified in this
>     // test8to9 omitted, data corruption correct upgrade with no schema change
> ```

nit: `correction`?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6474)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 421 at r1](https://reviewable.io/reviews/qtox/qtox/6474#-MwGVWUH7hQx1CQ1oxyt:-MwOEG9Q-n7Slz8JLJ3U:b-4hxhdw) ([raw file](https://github.com/qtox/qtox/blob/d85db5351f64d55bb80f6f56216ad34b2bca9671/src/persistence/history.cpp#L421)):*

Ok. millions sounds bad, but I guess modern PCs can handle that without running out of RAM
</blockquote></details>

The "millions" part just applies to single SQL queries like `UPDATE text_messages SET sender_alias = %1 WHERE sender_alias = %2;`, so wouldn't be loaded into memory at once (or, up to SQL based on caching).
___
*[test/persistence/dbschema_test.cpp, line 57 at r2](https://reviewable.io/reviews/qtox/qtox/6474#-MwO4-L81vCIkoDxBk53:-MwOEjo0ACmh72Ve3rMP:bt2vr1b) ([raw file](https://github.com/qtox/qtox/blob/f3fea27a450b35895cca648a55e1c6a01a883a2c/test/persistence/dbschema_test.cpp#L57)):*

nit: `correction`?
</blockquote></details>

Fixed.

sudden6 (commented):
> Reviewed 2 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6474)* status: :shipit: complete! 1 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 421 at r1](https://reviewable.io/reviews/qtox/qtox/6474#-MwGVWUH7hQx1CQ1oxyt:-MwOIpgiBr1kYm5KRK70:b-tskq06) ([raw file](https://github.com/qtox/qtox/blob/d85db5351f64d55bb80f6f56216ad34b2bca9671/src/persistence/history.cpp#L421)):*

The "millions" part just applies to single SQL queries like `UPDATE text_messages SET sender_alias = %1 WHERE sender_alias = %2;`, so wouldn't be loaded into memory at once (or, up to SQL based on caching).
</blockquote></details>

ah then everything is ok.

PR #6475, opened 2022-02-18
fix(crash): Handle nightly tag name in update check

anthonybilinski:
We didn't use to have a nightly tag, so git describe would include the
verison of the last release even for nightly or dev builds. Now that the
nightly tag exists, which is required for github release craeation, it
breaks out git describe parsing, and there's no way for us to tell if a
user is behind the next release or not.

Instead just don't check update status if the user isn't on a release
tagged version.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6475)


sudden6 (approved):
:lgtm_strong:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6475#-:-MwGUNsP4KGPu_9n21Y7:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6483, opened 2022-02-19
fix(CI): Correct Wine lib path for winehq manual install

### LINKED ISSUE #6481: Windows docker build regressed with new WINE version
Creating this for myself to keep track, will follow up so not providing tons of detail.

79feb19d7d74be5c5b5862298818cc69507fda15 changed WINE form version 5 to version 7 which fixed unit tests under WINE. This causes a new error: `Error: "/root/.wine/drive_c/windows/system32" directory doesn't exist.`.

This probably breaks our DLL checking since it's what references that path.

Likely the change is due to us installing WINE from winehq.org rather than from the debian repos, and the package isn't analogous, so we may be missing up user setup type meta package.

anthonybilinski:
Install path changed in 79feb19d7d74be5c5b5862298818cc69507fda15,
but paths were not updated to match.

Fix #6481

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6483)

### DIFF
--- a/windows/cross-compile/build.sh
+++ b/windows/cross-compile/build.sh
@@ -194,10 +194,10 @@ while IFS= read -r line
 do
   if [[ "$ARCH" == "i686" ]]
   then
-    WINE_DLL_DIR="/root/.wine/drive_c/windows/system32"
+    WINE_DLL_DIR="/opt/wine-stable/lib/wine/i386-windows"
   elif [[ "$ARCH" == "x86_64" ]]
   then
-    WINE_DLL_DIR="/root/.wine/drive_c/windows/system32 /root/.wine/drive_c/windows/syswow64"
+    WINE_DLL_DIR="/opt/wine-stable/lib/wine/i386-windows /opt/wine-stable/lib64/wine/x86_64-windows"
   fi
   python3 /usr/local/bin/mingw-ldd.py $line --dll-lookup-dirs $QTOX_PREFIX_DIR $WINE_DLL_DIR --output-format tree >> dlls-required
 done < <(cat exes runtime-dlls)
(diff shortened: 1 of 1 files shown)


sudden6 (approved):
:lgtm_strong:

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6483#-:-MwNn8wL68pn67EYTXY5:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

windows/cross-compile/build.sh
```diff
@@ -194,10 +194,10 @@ while IFS= read -r line
 do
   if [[ "$ARCH" == "i686" ]]
   then
-    WINE_DLL_DIR="/root/.wine/drive_c/windows/system32"
+    WINE_DLL_DIR="/opt/wine-stable/lib/wine/i386-windows"
   elif [[ "$ARCH" == "x86_64" ]]
   then
-    WINE_DLL_DIR="/root/.wine/drive_c/windows/system32 /root/.wine/drive_c/windows/syswow64"
+    WINE_DLL_DIR="/opt/wine-stable/lib/wine/i386-windows /opt/wine-stable/lib64/wine/x86_64-windows"
```
nurupo, line 200: You have changed the order, it used to check for 64-bit dlls first (`system32`) and only then fallback to 32-bit dlls (`syswow64`), but now it's 32-bit and then 64-bit. It should run slower now, because [it will look for a foo.dll in 32-bit libs directory first, parse PE datastructure of the 32-bit foo.dll, realize that it's 32-bit when qtox is 64-bit -- architecture mismatch, discard whatever was read/parsed and search for foo.dll further, until finally finding it in the 64-bit libs directory.](https://github.com/nurupo/mingw-ldd/blob/67eff2bf8a48b070539f7a6a1855dac0a3174189/mingw_ldd/mingw_ldd.py#L49-L57)

anthonybilinski: Good catch, will open a PR to revert to the previous order.
EDIT: PR: https://github.com/qTox/qTox/pull/6497

PR #6484, opened 2022-02-19
fix(CI): Grant tag update action workflow permissions

anthonybilinski:
Moving a tag between commits with workflows counts as changing the workflow,
so workflow write permission can be required.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6484)

### DIFF
--- a/.github/workflows/build-test-deploy.yaml
+++ b/.github/workflows/build-test-deploy.yaml
@@ -6,7 +6,8 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && github.ref == 'refs/heads/master'
     permissions:
-        contents: write
+      contents: write
+      actions: write # due to possibly updating this file
     steps:
       - uses: actions/checkout@v2
       - name: Move nightly tag to head for nightly release


anthonybilinski:
Marking as draft, need to verify cause in my own fork. Weirdly it doesn't fail on every nightly build, e.g. https://github.com/qTox/qTox/runs/5259106996?check_suite_focus=true failed when the workflow was updated in the commit, but then https://github.com/qTox/qTox/runs/5258928995?check_suite_focus=true also failed despite no workflow updated in the commit. Later https://github.com/qTox/qTox/runs/5262757498?check_suite_focus=true passed with no workflow updated in the commit.

anthonybilinski:
Ok I've convinced myself this is right. Based on workflow APIs requiring the action permission: https://docs.github.com/en/rest/reference/actions#workflows, our failures saying that it must have "workflows" permissions, which isn't available with the same name in actions: https://docs.github.com/en/enterprise-server@3.4/actions/using-workflows/workflow-syntax-for-github-actions#permissions.

I haven't verified this in my fork, and don't understand why on some tag updates it counts as a workflow addition and some it doesn't. It seems like any tag force push, the entire repo is being added at that version so a workflow is always being modified by some logic.

Regardless I'm confident enough to try running this on master, so un-WIP-ing.

sudden6 (commented):
IMO this seems like security issue, is there a way so that we have to explicitly approve updates of our GH workflows?

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6484#-:-MwcFt0l74kw6XBHelbu:b-nh7k6y)* status: 0 of 1 LGTMs obtained

anthonybilinski:
I'm not sure that it's either a security issue (as the workflow already has full write access to the repo), or that it's possible to approve manually - but I'm going to put this back to WIP. We haven't seen any nightly failures recently (for some reason), so clearly I'm missing something about the issue, but also it's become very low priority since it's not happening anymore.

anthonybilinski:
Wellll I haven't seen this again with dozens of master branch builds that updated the nightly tag from an action, so I'm just going to close this. Not sure why it popped up a few times when it was first added, but who cares.

PR #6494, opened 2022-02-23
fix: don't link emoji resources into targets that don't need them

sudden6:
The emoji resouce file is quite big, so link it only when really needed.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6494)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,17 +181,15 @@ if(NOT SMILEYS)
   set(SMILEYS "")
 endif()
 
+set(SMILEY_RESOURCES "")
 if(NOT "${SMILEYS}" STREQUAL "DISABLED")
-  set(${PROJECT_NAME}_RESOURCES
-        ${${PROJECT_NAME}_RESOURCES}
-        smileys/emojione.qrc)
+  set(SMILEY_RESOURCES smileys/emojione.qrc)
 
   if(NOT "${SMILEYS}" STREQUAL "MIN")
-      set(${PROJECT_NAME}_RESOURCES
-            ${${PROJECT_NAME}_RESOURCES}
+      set(SMILEY_RESOURCES
+            ${SMILEY_RESOURCES}
             smileys/smileys.qrc)
   endif()
-
 endif()
 
 set(${PROJECT_NAME}_SOURCES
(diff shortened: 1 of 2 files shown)


anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6494)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6494#-MwgwKWUAwNM8ymdzVr9:-MwgwKWUAwNM8ymdzVrA:b-hg5xi8) (no related file):*
Is this to save on file size? Both tests that don't use smileys and qTox build with -DSMILEYS=DISABLED are exactly the same size before and after this change, so if that was the goal it doesn't seem to be working.

---

sudden6 (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6494)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6494#-MwgwKWUAwNM8ymdzVr9:-Mwi5CQhD0TsYt7man4R:b-d5kmdl) (no related file):*

Is this to save on file size? Both tests that don't use smileys and qTox build with -DSMILEYS=DISABLED are exactly the same size before and after this change, so if that was the goal it doesn't seem to be working.
</blockquote></details>

On my system it seem to cause the autogenerated resource file to be compiled for targets that don't need it. I guess the linker then thows it away because it's unused. Additionally it shows the emojione.qrc file next to the test code in QtCreator, but that isn't really a reason :)

---

anthonybilinski (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6494)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6494#-MwgwKWUAwNM8ymdzVr9:-MwjuBmIC6xk3BTmw-xx:bchqbp6) (no related file):*

On my system it seem to cause the autogenerated resource file to be compiled for targets that don't need it. I guess the linker then thows it away because it's unused. Additionally it shows the emojione.qrc file next to the test code in QtCreator, but that isn't really a reason :)
</blockquote></details>

Ah got it, I see the same thing re: generation step for test_core and test_bsu before the change and not after. Change makes logical sense too. :lgtm_strong: 

---

PR #6496, opened 2022-02-25
add tests for real world Core usage

sudden6:

@anthonybilinski this is the test I told you about, which is completely broken and doesn't even correctly bootstrap. No idea why that is, maybe you can spot something.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6496)


anthonybilinski:
@sudden6 the main issue was that `Core` had static local variables that caused the two `Core` instances to clobber each other's state. I just moved those to be private class variables (`isConnected` and `tolerance`).

There were a few other bugs with the test itself that I've also fixed. The test is now running and passing reliably locally. I've pushed my fix commit to your branch.

anthonybilinski:
Is core_online_test fundamentally different from core_test? core_test was already bootstrapping and checking online status - this seems like an expanded version of that that exercises some more toxcore functionality?

sudden6:
@anthonybilinski no it's not, the only relevant difference is, that there are two `Core` instances which should interact with each other, so I chose to write a different test class for simplicity. ~BTW, this was just the beginning, I wanted to test more actions, but already this doesn't work.~

~EDIT: on CI this seems to work, but not locally, I'm confused.~ Didn't notice your commit.

anthonybilinski:
Pushed one more fixup in change_status_message, passing again now. Then did a bit of minor cleanup based on my own confusion when reading through it re: sleeps. Will review now (commits 1,3) but please review my commits (2,4,5,6) before merging.

anthonybilinski:
And pushed one more to deduplicate MockCoreSettings between core_test and core_online_test.

anthonybilinski (commented):
Reviewed your two commits, :lgtm_strong: on those.

> Reviewed 3 of 3 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6496#-:-Mwr37gw05rPUZ-F2YXu:b-665y67)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6496)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6496#-Mwr39eg3GJ23acRX5cM:-Mwr39eg3GJ23acRX5cN:b-cxnfuo) (no related file):*
Just a note that this significantly increases our unit test run time. This single test takes about 110s on my side. IMO our unit tests were already on the long side. I think ideally a qTox unit test shouldn't rely on toxcore/network at all, but mocking out toxcore obviously isn't trivial so and users can always choose not to run this.

Does that match your view, or is this more of a full qTox/toxcore integration test, rather than a `Core` class unit test?

---

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6496)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6496#-MwrFX4B-IoIZP0NXq09:-MwrFX4B-IoIZP0NXq0A:b-wjbv7o) (no related file):*
https://github.com/qTox/qTox/runs/5345626167?check_suite_focus=true segfaulted in this test, weirdly. I can't reproduce that locally, and ASAN doesn't find any problems. I can't even repro in the same docker image of almalinux, in the debug/minimal configuration... will try running it a bunch of times, but if still no repro I guess we can assume it's a blip in the runner? :S

---

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6496)* status: :shipit: complete! 1 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6496#-MwrFX4B-IoIZP0NXq09:-MwrQQCW9nUCvuoGKxpb:bk6ksln) (no related file):*

https://github.com/qTox/qTox/runs/5345626167?check_suite_focus=true segfaulted in this test, weirdly. I can't reproduce that locally, and ASAN doesn't find any problems. I can't even repro in the same docker image of almalinux, in the debug/minimal configuration... will try running it a bunch of times, but if still no repro I guess we can assume it's a blip in the runner? :S
</blockquote></details>

Could not reproduce this locally with 20+ runs, it's passing in CI on re-run, and ASAN isn't finding anything. I'm ok with assuming that this was a runner error and not a code error.

---

PR #6531, opened 2022-03-07
chore(CI): Set a very long timeout for core_online_test

anthonybilinski:
CI has very flaky network causing everything to take way longer than expected.
None of the tests are actually checking that something happens quickly, so just
set a stupidly long timeout to avoid flakiness.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6531)

### DIFF
--- a/test/core/core_online_test.cpp
+++ b/test/core/core_online_test.cpp
@@ -54,9 +54,8 @@ QList<DhtServer> MockNodeListGenerator::getBootstrapnodes() const {
 }
 
 namespace {
-    const int bootstrap_timeout = 120000; // timeout for bootstrapping
-    const int local_timeout = 500; // timeout for operations happening offline
-    const int connected_message_timeout = 15000; // timeout for alice and bob messaging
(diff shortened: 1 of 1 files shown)


test/core/core_online_test.cpp
```diff
@@ -154,20 +153,20 @@ void TestCoreOnline::change_name()
 
     alice->setUsername(aliceName);
 
-    QTRY_VERIFY_WITH_TIMEOUT(aliceSaveRequest.count() == 1, local_timeout);
```
robinlinden, line 157: I'm guessing the local_timeout ones weren't actually timing out and you just normalized the timeouts a bit as well?

robinlinden (approved):

test/core/core_online_test.cpp
anthonybilinski: Correct. Everything else, from bootstrapping to friend requests, to seeing friend's status change took up to many minutes between all the DHT disconnects and reconnects, so I just gave up on reasonable timeouts anywhere..

PR #6542, opened 2022-03-08
CI script arg parsing and compile options

anthonybilinski:

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6542)

### DIFF
--- a/buildscripts/build_libexif.sh
+++ b/buildscripts/build_libexif.sh
@@ -6,42 +6,20 @@
 
 set -euo pipefail
 
-usage()
-{
-    echo "Download and build libexif for the windows cross compiling environment"
-    echo "Usage: $0 --arch {win64|win32}"
-}
+readonly SCRIPT_DIR="$(dirname "$(realpath "$0")")"
 
-ARCH=""
+source "${SCRIPT_DIR}/build_utils.sh"
 
-while (( $# > 0 )); do
-    case $1 in
-        --arch) ARCH=$2; shift 2 ;;
-        -h|--help) usage; exit 1 ;;
-        *) echo "Unexpected argument $1"; usage; exit 1;;
-    esac
-done
+parse_arch --dep "libexif" --supported "win32 win64" "$@"
 
-if [ "$ARCH" != "win32" ] && [ "$ARCH" != "win64" ]; then
-    echo "Unexpected arch $ARCH"
-    usage
-    exit 1
-fi
+"${SCRIPT_DIR}/download/download_libexif.sh"
 
-"$(dirname "$(realpath "$0")")/download/download_libexif.sh"
-
-if [ "${ARCH}" == "win64" ]; then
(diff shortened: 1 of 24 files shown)


buildscripts/platform_detection.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# SPDX-License-Identifier: GPL-3.0-or-later
+#     Copyright (c) 2022 by The qTox Project Contributors
+
+
+set -euo pipefail
+
+usage()
+{
+    echo "Download and build $DEP_NAME for the Windows cross compiling environment"
```
warisb: maybe nitpick but the filename doesn't seem to match the usage message

warisb: I see it's something the dep downloader sources. Seems fine then.

buildscripts/build_libexif.sh
```diff
@@ -6,42 +6,21 @@
...
-CFLAGS="-O2 -g0" ./configure --host="${HOST}" \
-                         --prefix=/windows/ \
+readonly SCRIPT_DIR="$(dirname "$(realpath "$0")")"
+
+source "${SCRIPT_DIR}/platform_detection.sh"
+
+DEP_NAME="libexif"
+parse_arch "$@"
+
+"${SCRIPT_DIR}/download/download_libexif.sh"
```
warisb, line 15: not sure if this is a level of meta-programming you are comfortable with but you can dedupe even further by using

```
download_${DEP_NAME}.sh
```
then you can move this portion to a common script. but this would mean you have to adhere to this naming convention going forward.

warisb: if you do this a secondary benefit is you can have a common `download.sh` script instead of a mixture of platform detection + download script

warisb: after offline discussion this is fine as is. a common download script would not add any value beyond the build_<dep>.sh script itself.

buildscripts/platform_detection.sh
warisb: after offline discussion I think this file may be due for a rename - since it provides more than one function, but it is only named after one function. will approve once this is addressed.

anthonybilinski: I've updated the DEP_NAME to be passed directly to parse_arch, so that it's not passing anything through global state. File also renamed from `platform_detection.sh` to `build_utils.sh`.

anthonybilinski:
Also updated to pass the supported arches from the build script to parse_arch, since not all will need to support macOS, and removed a couple of the args exported from parse_arch after rebasing the macOS stuff back up (not in PR) since any Windows args that macOS doesn't need need to be nullable from the macOS's section. This meant a little bit more arch handling in a couple of the build scripts.

warisb (approved):
Latest revision LGTM.

warisb (approved):

PR #6543, opened 2022-03-09
Disable online tests in most CI

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6543)

### DIFF
--- a/test/core/core_online_test.cpp
+++ b/test/core/core_online_test.cpp
@@ -28,9 +28,11 @@
 #include <QSignalSpy>
 #include <QtGlobal>
 #include <QtTest/QtTest>
-#include <limits>
-
 #include <QTest>
+#include <QtGlobal>
+
+#include <cassert>
+#include <limits>
 #include <iostream>
 #include <memory>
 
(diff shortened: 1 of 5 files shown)


test/core/core_online_test.cpp
```diff
@@ -83,6 +85,8 @@ TestCoreOnline::TestCoreOnline()
     qRegisterMetaType<ToxPk>("ToxPk");
     qRegisterMetaType<uint32_t>("uint32_t");
     qRegisterMetaType<Status::Status>("Status::Status");
+    auto timeout = QString("%1").arg(very_long_timeout);
+    assert(qputenv("QTEST_FUNCTION_TIMEOUT", timeout.toLatin1()));
```
robinlinden, line 89: This will get removed if you try to run your unit tests with optimizations, so doing things with side-effects inside `assert` is a bit weird. :x

robinlinden (approved):

test/core/core_online_test.cpp
anthonybilinski: Crap missed this comment before merging. That's a good point, will fix.

PR #6548, opened 2022-03-10
feat(history): Add self join/part messages

anthonybilinski:
Makes it so that looking back in chat history, you can see which users you were
connected to for any message. Otherwise self client restarts are unseen.
Follows showGroupJoinLeaveMessages setting which defaults to false, so only
users who opt in will see the messages.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6548)

### DIFF
--- a/src/model/systemmessage.h
+++ b/src/model/systemmessage.h
@@ -42,6 +42,8 @@ enum class SystemMessageType
     incomingCall,
     callEnd,
     messageSendFailed,
+    selfJoinedGroup,
+    selfLeftGroup,
 };
 
 struct SystemMessage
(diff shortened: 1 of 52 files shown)


src/model/systemmessage.h
```diff
@@ -103,6 +105,14 @@ struct SystemMessage
         case SystemMessageType::messageSendFailed:
             translated = QObject::tr("Message failed to send");
             break;
+        case SystemMessageType::selfJoinedGroup:
+            translated = QObject::tr("You have joined the group");
+            numArgs = 0;
```
robinlinden: All other cases with 0 args leave out the numArgs = 0, but I'm fine with leaving this in. It's clearer to always set it with the message imo. :x

robinlinden (approved):

src/model/systemmessage.h
anthonybilinski: Good point, just scrapped the whole numArgs concept and substitute in-place now.

PR #6551, opened 2022-03-10
chore(build): Add LOTS of warnings

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6551)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,12 +99,6 @@ if(COMPILER_SUPPORTS_WARNING_WEAK_VTABLES)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wweak-vtables")
 endif()
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstrict-overflow")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstrict-aliasing")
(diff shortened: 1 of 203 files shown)


CMakeLists.txt
```diff
@@ -104,6 +104,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wstrict-aliasing")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wold-style-cast")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wsign-compare")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wunknown-pragmas")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wswitch")
```
robinlinden: I was about to write that this changes nothing as `-Wswitch` is part of the **basic** `-Wall` warnings, but I guess qTox doesn't have those enabled.. :(

robinlinden (approved):

CMakeLists.txt
anthonybilinski: Good point :grimacing: 

Enabled a ton stolen from you and beat qTox into compliance. Updated with a huge diff.

Each refactor commit should be non-functional (except for trivial things like using float precision instead of double precision in UI). Functional commits are marked as "fix".

robinlinden (approved):

PR #6560, opened 2022-03-12
refactor: Remove Settings singleton

anthonybilinski:
* Make main.cpp's toxSave free functions into a ToxSave class so that it can be
  given Settings on construction.
* Add void* to IPC callbacks so that classes can get back to themselves.

Based on https://github.com/qTox/qTox/pull/6559.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6560)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -55,19 +55,23 @@ QList<QByteArray>* logBuffer =
 QMutex* logBufferMutex = new QMutex();
 #endif
 
+std::unique_ptr<Settings> settings;
+std::unique_ptr<ToxSave> toxSave;
+
 void cleanup()
 {
     // force save early even though destruction saves, because Windows OS will
     // close qTox before cleanup() is finished if logging out or shutting down,
     // once the top level window has exited, which occurs in ~Widget within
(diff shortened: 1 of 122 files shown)


src/main.cpp
```diff
@@ -225,8 +231,12 @@ int main(int argc, char* argv[])
         qWarning() << "Couldn't load font";
     }
 
-    Settings& settings = Settings::getInstance();
-    QString locale = settings.getTranslation();
+    settings = std::unique_ptr<Settings>(new Settings());
+    assert(settings);
```
robinlinden: Afaik this assert will never trigger and the !settings will never evaluate to true since you're not passing `std::nothrow` to new. :P

Even if you're building with `-fno-exceptions`, new will (try to) throw if oom (and qTox will terminate.)

robinlinden (commented):
Not approving yet since I need to look at the singleton removals this one builds on top of. :P

src/main.cpp
anthonybilinski: Ah good point. Removed the check.

robinlinden (approved):

PR #6561, opened 2022-03-12
Group history

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,7 @@ qt5_wrap_ui(${PROJECT_NAME}_FORMS
   src/widget/about/aboutfriendform.ui
   src/widget/form/loadhistorydialog.ui
   src/widget/form/profileform.ui
-  src/widget/form/removefrienddialog.ui
+  src/widget/form/removechatdialog.ui
   src/widget/form/searchsettingsform.ui
   src/widget/form/setpassworddialog.ui
   src/widget/form/settings/aboutsettings.ui
@@ -490,8 +490,8 @@ set(${PROJECT_NAME}_SOURCES
   src/widget/tool/profileimporter.h
   src/widget/tool/recursivesignalblocker.cpp
   src/widget/tool/recursivesignalblocker.h
-  src/widget/tool/removefrienddialog.cpp
-  src/widget/tool/removefrienddialog.h
+  src/widget/tool/removechatdialog.cpp
+  src/widget/tool/removechatdialog.h
   src/widget/tool/screengrabberchooserrectitem.cpp
   src/widget/tool/screengrabberchooserrectitem.h
   src/widget/tool/screengrabberoverlayitem.cpp
--- a/src/core/chatid.h
+++ b/src/core/chatid.h
@@ -41,6 +41,7 @@ public:
     const uint8_t* getData() const;
     bool isEmpty() const;
     virtual int getSize() const = 0;
+    virtual std::unique_ptr<ChatId> clone() const = 0;
 
 protected:
     ChatId();
--- a/src/core/groupid.cpp
+++ b/src/core/groupid.cpp
@@ -67,3 +67,8 @@ int GroupId::getSize() const
 {
     return size;
 }
+
+std::unique_ptr<ChatId> GroupId::clone() const
+{
+    return std::unique_ptr<ChatId>(new GroupId(*this));
+}
--- a/src/core/groupid.h
+++ b/src/core/groupid.h
@@ -31,4 +31,5 @@ public:
     explicit GroupId(const QByteArray& rawId);
     explicit GroupId(const uint8_t* rawId);
     int getSize() const override;
+    std::unique_ptr<ChatId> clone() const override;
 };
--- a/src/core/toxpk.cpp
+++ b/src/core/toxpk.cpp
@@ -87,3 +87,8 @@ int ToxPk::getSize() const
 {
     return size;
 }
+
+std::unique_ptr<ChatId> ToxPk::clone() const
+{
+    return std::unique_ptr<ChatId>(new ToxPk(*this));
+}
--- a/src/core/toxpk.h
+++ b/src/core/toxpk.h
@@ -33,4 +33,5 @@ public:
     explicit ToxPk(const uint8_t* rawId);
     explicit ToxPk(const QString& pk);
     int getSize() const override;
+    std::unique_ptr<ChatId> clone() const override;
 };
--- a/src/model/about/aboutfriend.cpp
+++ b/src/model/about/aboutfriend.cpp
@@ -127,7 +127,7 @@ bool AboutFriend::clearHistory()
     const ToxPk pk = f->getPublicKey();
     History* const history = Nexus::getProfile()->getHistory();
     if (history) {
-        history->removeFriendHistory(pk);
+        history->removeChatHistory(pk);
         return true;
     }
 
--- a/src/model/chat.h
+++ b/src/model/chat.h
@@ -23,6 +23,7 @@
 #include <QObject>
 #include <QString>
 
+class ToxPk;
 class Chat : public QObject
 {
     Q_OBJECT
@@ -31,6 +32,7 @@ public:
 
     virtual void setName(const QString& name) = 0;
     virtual QString getDisplayedName() const = 0;
+    virtual QString getDisplayedName(const ToxPk& contact) const = 0;
     virtual uint32_t getId() const = 0;
     virtual const ChatId& getPersistentId() const = 0;
     virtual void setEventFlag(bool flag) = 0;
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -19,6 +19,7 @@
 
 #include "chathistory.h"
 #include "src/persistence/settings.h"
+#include "src/core/chatid.h"
 #include "src/widget/form/chatform.h"
 
 namespace {
@@ -161,7 +162,7 @@ SearchResult ChatHistory::searchBackward(SearchPos startIdx, const QString& phra
         history->getDateWhereFindPhrase(f.getPublicKey(), earliestMessageDate, phrase,
                                         parameter);
 
-    auto loadIdx = history->getNumMessagesForFriendBeforeDate(f.getPublicKey(), dateWherePhraseFound);
+    auto loadIdx = history->getNumMessagesForChatBeforeDate(f.getPublicKey(), dateWherePhraseFound);
     loadHistoryIntoSessionChatLog(ChatLogIdx(loadIdx));
 
     // Reset search pos to the message we just loaded to avoid a double search
@@ -188,7 +189,7 @@ std::vector<IChatLog::DateChatLogIdxPair> ChatHistory::getDateIdxs(const QDate& 
                                                                    size_t maxDates) const
 {
     if (canUseHistory()) {
-        auto counts = history->getNumMessagesForFriendBeforeDateBoundaries(f.getPublicKey(),
+        auto counts = history->getNumMessagesForChatBeforeDateBoundaries(f.getPublicKey(),
                                                                            startDate, maxDates);
 
         std::vector<IChatLog::DateChatLogIdxPair> ret;
@@ -353,7 +354,7 @@ void ChatHistory::loadHistoryIntoSessionChatLog(ChatLogIdx start) const
     // We know that both history and us have a start index of 0 so the type
     // conversion should be safe
     assert(getFirstIdx() == ChatLogIdx(0));
-    auto messages = history->getMessagesForFriend(f.getPublicKey(), start.get(), end.get());
+    auto messages = history->getMessagesForChat(f.getPublicKey(), start.get(), end.get());
 
     assert(messages.size() == static_cast<int>(end.get() - start.get()));
     ChatLogIdx nextIdx = start;
@@ -424,7 +425,7 @@ void ChatHistory::loadHistoryIntoSessionChatLog(ChatLogIdx start) const
  */
 void ChatHistory::dispatchUnsentMessages(IMessageDispatcher& messageDispatcher)
 {
-    auto unsentMessages = history->getUndeliveredMessagesForFriend(f.getPublicKey());
+    auto unsentMessages = history->getUndeliveredMessagesForChat(f.getPublicKey());
 
     auto requiredExtensions = std::accumulate(
         unsentMessages.begin(), unsentMessages.end(),
@@ -520,7 +521,7 @@ bool ChatHistory::canUseHistory() const
 ChatLogIdx ChatHistory::getInitialChatLogIdx() const
 {
     if (canUseHistory()) {
-        return ChatLogIdx(history->getNumMessagesForFriend(f.getPublicKey()));
+        return ChatLogIdx(history->getNumMessagesForChat(f.getPublicKey()));
     }
     return ChatLogIdx(0);
 }
--- a/src/model/chathistory.cpp
+++ b/src/model/chathistory.cpp
@@ -71,10 +71,10 @@ bool handleActionPrefix(QString& content)
 }
 } // namespace
 
-ChatHistory::ChatHistory(Friend& f_, History* history_, const ICoreIdHandler& coreIdHandler_,
+ChatHistory::ChatHistory(Chat& chat_, History* history_, const ICoreIdHandler& coreIdHandler_,
                          const Settings& settings_, IMessageDispatcher& messageDispatcher,
                          FriendList& friendList)
-    : f(f_)
+    : chat(chat_)
     , history(history_)
     , settings(settings_)
     , coreIdHandler(coreIdHandler_)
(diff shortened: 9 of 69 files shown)


sudden6 (changes requested):
> Reviewed 5 of 5 files at r1, 15 of 15 files at r2, 5 of 5 files at r3, 2 of 2 files at r4, 53 of 53 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[-- commits, line 32 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWtp6mCVLxYjzFvABD):*
`aliases where in which group` -> `aliases were in which group`
`it, and and peer` -> `it, and peer`
___
*[src/model/chat.h, line 35 at r3](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsydq4f4qt2A1ggkN:-MyWsydq4f4qt2A1ggkO:b-c19tzh) ([raw file](https://github.com/qtox/qtox/blob/f1381f4bfc72a3e4e09644a0a8e8e153773bd4a6/src/model/chat.h#L35)):*
> ```objc
>     virtual void setName(const QString& name) = 0;
>     virtual QString getSelfDisplayedName() const = 0;
>     virtual QString getDisplayedName(const ToxPk& contact) const = 0;
> ```

I think that should be an extension for `Group` only, no?
___
*[src/model/friend.h, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsZB5AJRy_5cfSj63:-MyWsZB5AJRy_5cfSj64:b-5ag9rx) ([raw file](https://github.com/qtox/qtox/blob/50558b20dfe7cfb314a48199f4443815df8df3f5/src/model/friend.h#L41)):*
> ```objc
>     void setName(const QString& name) override;
>     void setAlias(const QString& alias);
>     QString getSelfDisplayedName() const override;
> ```

This rename doesn't really make sense to me. What is "self" in this context supposed to mean?
___
*[src/persistence/history.cpp, line 827 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWuFXT39l2KrwW8iKQ:-MyWuFXT39l2KrwW8iKR:bz3m534) ([raw file](https://github.com/qtox/qtox/blob/8086f682a384ae8948cc6aa02f47719647bc541e/src/persistence/history.cpp#L827)):*
> ```cpp
>                                 "DELETE FROM aliases "
>                                 "WHERE id NOT IN ( "
>                                 "    SELECT DISTINCT sender_alias FROM text_messages);"
> ```

Can you write a test for this functionality? It has rather complex logic

sudden6:
Can you quickly outline the design of how Chats/Groups/Friends interact and the relationships between them? Also how they are represented in the db.

anthonybilinski:
Conceptually, Chats represent either a Friend or a Group, so a conversation is always with a Chat, which always has a 32-byte ID (either friend public key, or group persistent ID). The relation between them is simply that both Friends and Groups implement the Chat interface.

In the database, this has some changes to the way we layout data despite having no change to actual schema. Before, `peers` was equal to tox friends + self, each of which except for self had a chat. Every peer had n aliases, each chat had n messages of various types authored by either yourself or the chat owner. 

Now, `peers` contains all of self, tox friends, groups themselves, and group members. Aliases contains n names for any peer other than groups. History messages map to a chat that is any peer other than group members, and message authors maps to any author that is not a group member.

So conceptually "peer" now means "chat or author", "alias" means "name of author", history's chat means "chat", history's author means "author".

Which isn't really _that_ different from before since self was in peer but not a chat, only an author, so  there wasn't a strict relation between peers having chats.

Groups/chats being in the peer table is a bit weird (and was kind of before due to self being there). We could split the "peers" table into two tables, "authors" and "chats", each of which would hold the same 32-byte IDs and would lead to friends being in both tables, but then it would be more clear what aliases and history are depending on (a chat or an author) vs the combined meaning now.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[-- commits, line 32 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWtp6mCVLxYjzFvABD):*

`aliases where in which group` -> `aliases were in which group`
`it, and and peer` -> `it, and peer`
</blockquote></details>

Thanks, fixed.
___
*[src/model/chat.h, line 35 at r3](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsydq4f4qt2A1ggkN:-MyaE7GcAuY8FjMQgur2:b6g5oxa) ([raw file](https://github.com/qtox/qtox/blob/f1381f4bfc72a3e4e09644a0a8e8e153773bd4a6/src/model/chat.h#L35)):*

I think that should be an extension for `Group` only, no?
</blockquote></details>

ChatHistory is dealing with Chats, but needs to be able to resolve an author's name, whether that author is a Friend or a Group member. So to avoid specializing ChatHistory separately for Groups and Friends, we need to have Chat generically implement the getDisplayedName(const ToxPk&), even though it only is really useful for Groups.
___
*[src/model/friend.h, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsZB5AJRy_5cfSj63:-MyaEVt-3XrgiQO2sE_e:b-uew1jq) ([raw file](https://github.com/qtox/qtox/blob/50558b20dfe7cfb314a48199f4443815df8df3f5/src/model/friend.h#L41)):*

This rename doesn't really make sense to me. What is "self" in this context supposed to mean?
</blockquote></details>

From the class' perspective, so for Friend it gets the friend's name, for Group it gets the group's name.

The rename was more to get it out of the way of `getDisplayedName(const ToxPk& contact)`, so that it's clear from the API that one is getting the name of an author within the chat, and the other of the chat itself, even though in the friend's case those are the same thing.
___
*[src/persistence/history.cpp, line 827 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWuFXT39l2KrwW8iKQ:-MyaEoxJCvEsFI1CtnFk:b-kfga52) ([raw file](https://github.com/qtox/qtox/blob/8086f682a384ae8948cc6aa02f47719647bc541e/src/persistence/history.cpp#L827)):*

Can you write a test for this functionality? It has rather complex logic
</blockquote></details>

Yes good point, will do. Will wait until I hear feedback on the option of splitting `peers` into separate `authors` and `chats` tables, since that would change this logic completely and probably simplify it.

sudden6 (changes requested):
> Reviewed 23 of 70 files at r9, 8 of 15 files at r10, 5 of 5 files at r11, 49 of 53 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[src/model/chat.h, line 35 at r3](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsydq4f4qt2A1ggkN:-MycNIY--BZzAexAU_gm:b-i4vzx3) ([raw file](https://github.com/qtox/qtox/blob/f1381f4bfc72a3e4e09644a0a8e8e153773bd4a6/src/model/chat.h#L35)):*

ChatHistory is dealing with Chats, but needs to be able to resolve an author's name, whether that author is a Friend or a Group member. So to avoid specializing ChatHistory separately for Groups and Friends, we need to have Chat generically implement the getDisplayedName(const ToxPk&), even though it only is really useful for Groups.
</blockquote></details>

Why not overload the existing function with a `ToxPk` parameter?
___
*[src/model/friend.h, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsZB5AJRy_5cfSj63:-MycNASoDUvEUS7Rrlcr:b-pl8pxf) ([raw file](https://github.com/qtox/qtox/blob/50558b20dfe7cfb314a48199f4443815df8df3f5/src/model/friend.h#L41)):*

From the class' perspective, so for Friend it gets the friend's name, for Group it gets the group's name.

The rename was more to get it out of the way of `getDisplayedName(const ToxPk& contact)`, so that it's clear from the API that one is getting the name of an author within the chat, and the other of the chat itself, even though in the friend's case those are the same thing.
</blockquote></details>

See other comment
___
*[src/model/friend.cpp, line 124 at r16](https://reviewable.io/reviews/qtox/qtox/6561#-MycNXsF2SidKTORZtbi:-MycNXsF2SidKTORZtbj:b-g47sdj) ([raw file](https://github.com/qtox/qtox/blob/c7f56879f29ee69a0a8e3e0ad26d83cf2b465a91/src/model/friend.cpp#L124)):*
> ```cpp
> {
>     if (contact != friendPk) {
>         return {};
> ```

Is the empty string allowed as displayed name? Could lead to problems

sudden6:
I like the idea of splitting it into chat and authors, mostly because I'm worried that someone will be able to generate a Tox profile and Group with the same 32byte ID, which would completely break our database.

How would messages be stored in this layout? Each message in a separate table referencing a chat and an author?

anthonybilinski (commented):
Sure, I'm not sure that a group ID is any more likely to collide with a Tox key than two group IDs or two Tox keys, but also I don't know that the toxcore API gives us a guarantee of that (haven't checked). I think it's cleaner architecturally though so have no problem with it.

For message history, it wouldn't really be any different. Already each history message references a chat (currently peers) and each text or file message references an alias which references an author (currently peers). System messages reference a chat but not author. Diagram added here: https://github.com/qTox/qTox/pull/6583#issue-1176783978

So all that would change is that history messages reference the new "chat" instead of "peers" as their chat, and aliases references the new "authors" instead of the current "peers", and text/file messages still reference aliases.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561#-:-Mye6_oW7DikuXP4vC-f:b-51nee8)* status: 0 of 1 LGTMs obtained

___
*[src/model/chat.h, line 35 at r3](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsydq4f4qt2A1ggkN:-Mye7AMc4OHol1spWnW3:b-q7zp5d) ([raw file](https://github.com/qtox/qtox/blob/f1381f4bfc72a3e4e09644a0a8e8e153773bd4a6/src/model/chat.h#L35)):*

Why not overload the existing function with a `ToxPk` parameter?
</blockquote></details>

Could do, I didn't think the API was as clear that way, but will make that change.
___
*[src/model/friend.cpp, line 124 at r16](https://reviewable.io/reviews/qtox/qtox/6561#-MycNXsF2SidKTORZtbi:-Mye7QTNEnIXN6YfG5_K:b7382h9) ([raw file](https://github.com/qtox/qtox/blob/c7f56879f29ee69a0a8e3e0ad26d83cf2b465a91/src/model/friend.cpp#L124)):*

Is the empty string allowed as displayed name? Could lead to problems
</blockquote></details>

Will add an assert, this would be an API misuse for Friend.

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[src/model/chat.h, line 35 at r3](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsydq4f4qt2A1ggkN:-Myv8gV_9m2fUXfr3Ipu:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/f1381f4bfc72a3e4e09644a0a8e8e153773bd4a6/src/model/chat.h#L35)):*

Could do, I didn't think the API was as clear that way, but will make that change.
</blockquote></details>

Done.
___
*[src/model/friend.h, line 41 at r2](https://reviewable.io/reviews/qtox/qtox/6561#-MyWsZB5AJRy_5cfSj63:-Myv8B5tC1m-uOW7Mb7t:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/50558b20dfe7cfb314a48199f4443815df8df3f5/src/model/friend.h#L41)):*

See other comment
</blockquote></details>

Done.
___
*[src/model/friend.cpp, line 124 at r16](https://reviewable.io/reviews/qtox/qtox/6561#-MycNXsF2SidKTORZtbi:-Myv8CUv-uPgqeIrwKdn:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/c7f56879f29ee69a0a8e3e0ad26d83cf2b465a91/src/model/friend.cpp#L124)):*

Will add an assert, this would be an API misuse for Friend.
</blockquote></details>

Done.
___
*[src/persistence/history.cpp, line 827 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWuFXT39l2KrwW8iKQ:-Myv8JaEBaHJghzqIWdk:b-sjo1g2) ([raw file](https://github.com/qtox/qtox/blob/8086f682a384ae8948cc6aa02f47719647bc541e/src/persistence/history.cpp#L827)):*

Yes good point, will do. Will wait until I hear feedback on the option of splitting `peers` into separate `authors` and `chats` tables, since that would change this logic completely and probably simplify it.
</blockquote></details>

Split peers PR opened: https://github.com/qTox/qTox/pull/6583.

Will be some rebasing once that's in that will simplify this removal logic.

sudden6 (changes requested):
> Reviewed 1 of 70 files at r17, 5 of 5 files at r18, 2 of 10 files at r19, 51 of 60 files at r20, 1 of 8 files at r21, 1 of 9 files at r22, 8 of 8 files at r23.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[src/model/friend.cpp, line 124 at r16](https://reviewable.io/reviews/qtox/qtox/6561#-MycNXsF2SidKTORZtbi:-Mz9qQ6e102ndPA7ZgPP:bb5mg4r) ([raw file](https://github.com/qtox/qtox/blob/c7f56879f29ee69a0a8e3e0ad26d83cf2b465a91/src/model/friend.cpp#L124)):*

Done.
</blockquote></details>

Is the assert condition inverted or am I reading this wrong? shouldn't it be `assert(contact == friendPk)`?

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[src/model/friend.cpp, line 124 at r16](https://reviewable.io/reviews/qtox/qtox/6561#-MycNXsF2SidKTORZtbi:-Mz9uwXHBxLt9Tz8D3hD:b-u3vnoh) ([raw file](https://github.com/qtox/qtox/blob/c7f56879f29ee69a0a8e3e0ad26d83cf2b465a91/src/model/friend.cpp#L124)):*

Is the assert condition inverted or am I reading this wrong? shouldn't it be `assert(contact == friendPk)`?
</blockquote></details>

Yes :facepalm: , fixed.

sudden6 (approved):
> Reviewed 1 of 63 files at r24, 5 of 5 files at r25, 2 of 2 files at r26, 52 of 53 files at r27, 1 of 1 files at r28, 1 of 2 files at r29, 1 of 1 files at r30, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6561)* status: 0 of 1 LGTMs obtained

___
*[src/persistence/history.cpp, line 827 at r6](https://reviewable.io/reviews/qtox/qtox/6561#-MyWuFXT39l2KrwW8iKQ:-MzARz2e1aAi1aiZTop6:b-h72pga) ([raw file](https://github.com/qtox/qtox/blob/8086f682a384ae8948cc6aa02f47719647bc541e/src/persistence/history.cpp#L827)):*

Split peers PR opened: https://github.com/qTox/qTox/pull/6583.

Will be some rebasing once that's in that will simplify this removal logic.
</blockquote></details>

Ok, waiting for you to rebase this then.

anthonybilinski:
Rebased onto split peers change. Removal logic is simple(r) now - generically removing aliases when they have no associated chat messages, and authors when they have no associated aliases.

sudden6:
> We don't store what users or aliases were in which group, so just remove
any alias that has no associated chat message with it, and peer that
doesn't have an alias (friend with history) and doesn't have a chat
message associated with it as a chat_id (group).

Wouldn't this also remove aliases from friends which don't have any messages in their chat yet? Scenario: Alice accepts Bobs friend request and immediately sets an alias for Bob to `Carol`. No messages are exchanged between Alice and Bob. Now Alice leaves the group `Tox Testers`. (Possibly this is a race condition with the system message that says a friend request was accepted).

Can you confirm this scenario is safe?

Otherwise this looks good to me. Congrats on reaching the group history milestone!

anthonybilinski:
The current friend alias is stored in Settings: `Widget::onFriendAliasChanged` -> `Settings::setFriendAlias` and loaded on construction: `FriendList::addFriend` -> `Settings::getFriendAlias` - so the current set friend alias is unaffected. History's aliases should only be aliases needed to show the author of messages, including back in time.

Before this change, testing the scenario where A add B, neither exchanges a message, A sets B's alias to Bob - A's database has an alias for A for the friend request message, but no alias for B saved.

Database aliases only get inserted in history.cpp from `generateUpdateAlias` which is called in `generateNewTextMessageQueries` and `History::generateNewFileTransferQueries`.

So I'm quite confident that this scenario is safe.

PR #6563, opened 2022-03-13
refactor(Style): Move Style enums to enum class

anthonybilinski:
Based on https://github.com/qTox/qTox/pull/6560.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6563)

### DIFF
--- a/src/widget/style.cpp
+++ b/src/widget/style.cpp
@@ -82,45 +82,46 @@ QString qssifyFont(QFont font)
     return QString("%1 %2px \"%3\"").arg(font.weight() * 8).arg(font.pixelSize()).arg(font.family());
 }
 
-QMap<Style::ColorPalette, QColor> palette;
-
+using ColorPalette = Style::ColorPalette;
+QMap<ColorPalette, QColor> palette;
 QMap<QString, QString> dictColor;
 QMap<QString, QString> dictFont;
 QMap<QString, QString> dictTheme;
 
+using MainTheme = Style::MainTheme;
 const QList<Style::ThemeNameColor> themeNameColors = {
-    {Style::Light, QObject::tr("Default"), QColor()},
-    {Style::Light, QObject::tr("Blue"), QColor("#004aa4")},
-    {Style::Light, QObject::tr("Olive"), QColor("#97ba00")},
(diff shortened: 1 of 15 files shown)


src/widget/style.cpp
```diff
@@ -228,7 +229,7 @@ QFont Style::getFont(Style::Font font)
         appFont(defSize - 1, QFont::Light),  // small light
     };
 
-    return fonts[font];
+    return fonts[static_cast<int>(font)];
```
robinlinden: Unrelated to this PR, but `fonts` should probably be a map, not an array? :x Would make this less spooky. Especially since there's nothing around the Font enum to indicate that it can't be reordered or anything. :P

robinlinden (approved):

anthonybilinski:
Whoops close/open was just trying to get it to re-generate the diff based on the new destination branch, but no luck.

src/widget/style.cpp
anthonybilinski: Yeah agreed, moved it to a std::map.

src/widget/style.cpp
```diff
@@ -219,17 +219,17 @@ QFont Style::getFont(Font font)
...
+        {Font::Big, appFont(defSize + 1, QFont::Normal)},
+        {Font::BigBold, appFont(defSize + 1, QFont::Bold)},
+        {Font::Medium, appFont(defSize, QFont::Normal)},
+        {Font::MediumBold, appFont(defSize, QFont::Bold)},
+        {Font::Small, appFont(defSize - 1, QFont::Normal)},
+        {Font::SmallLight, appFont(defSize - 1, QFont::Light)},
     };
 
-    return fonts[static_cast<int>(font)];
+    return fonts[font];
```
robinlinden: `.at(font)`? `operator[]` will create a new element and return that if the key doesn't exist in the map. :P `.at()` will try to throw and explode instead.

This is why there's no const overload for operator[]. :x If you switch to .at() the map should be able to be const.

anthonybilinski: I like exploding over UB, done :D And moved to const.

PR #6577, opened 2022-03-17
Macos autorun

### LINKED ISSUE #2814: The setting of "Autostart" is disabled on OS X

anthonybilinski:
* [fix(macos): Fix macOS autorun not loading at start](https://github.com/qTox/qTox/pull/6577/commits/fbb75a57fae4e967ef9483a62ceb08904104c79e)
  macOS system.log was reporting "Unknown key for integer: RunAtLoad" and
not loading qTox at start despite
 ~/Library/LaunchAgents/chat.tox.qtox.autorun.plist being present.

  Changing the type from an int to a bool causes it to load successfully
on boot.

  Fix https://github.com/qTox/qTox/issues/2814

* [fix(macos): Represent autorun state in UI persistently](https://github.com/qTox/qTox/pull/6577/commits/170bca87d03adc2a8a55174e75c0514ed2a6b551)
  Check file setting rather than reporting a global value that isn't persisted
across client restart.

This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6577)

### DIFF
--- a/src/platform/autorun_osx.cpp
+++ b/src/platform/autorun_osx.cpp
@@ -25,29 +25,30 @@
 #include <QStandardPaths>
 
 namespace {
-bool state;
+QString getAutorunFile()
+{
+    return QDir::cleanPath(QStandardPaths::writableLocation(QStandardPaths::HomeLocation)
+            + QDir::separator() + "Library" + QDir::separator() + "LaunchAgents"
+            + QDir::separator() + "chat.tox.qtox.autorun.plist");
+}
 } // namespace
 
 bool Platform::setAutorun(const Settings& settings, bool on)
 {
     std::ignore = settings;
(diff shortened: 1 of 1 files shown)


anthonybilinski:
Tested on macOS 10.15, autorun works correct when enabled, correctly doesn't start when disabled even though the file is still present.

This fix has the side-effect of users with an old broken enable file now having autorun show as enabled in settings, but not working. The user toggling the setting off then on will resolve that. I don't think fixing this manually is desirable since users may have turned this on 5 years ago and forgot about it, so qTox opening all of a sudden would be annoying.

robinlinden (approved):

PR #6578, opened 2022-03-18
Core error parsing

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6578)


sudden6:
Can we somehow test/assert/compiler warn that all error cases are covered? Might be handy in case c-toxcore updates it's API and we'll certainly forget about this...

anthonybilinski:
Ah I was about to write that as-is, our CI will fail (only on the clang build) if a new enum value is added in toxcore due to a combination of Werror and Wswitch. This was the benefit of removing the `default` case in  https://github.com/qTox/qTox/pull/6578/commits/350fa9b46292eb7c63795f334e475a47466ab708

i.e.: 
```
qTox/util/src/toxcoreerrorparser.cpp:132:13: error: enumeration value 'TOX_ERR_CONFERENCE_JOIN_FAIL_SEND' not handled in switch [-Werror,-Wswitch]
    switch (error) {
            ^
1 error generated.
ninja: build stopped: subcommand failed.
```

So as long as we stop adding new calls to toxcore APIs without passing in the enum, we should be ok :D Don't think there's something that we can check for there.

_But_ my warnings enabling PR only links the interface to qtox_static, not to util or audio libs, so the util lib actually builds without Werror D:. Will make a separate PR to correct that.

sudden6:
Ok, that sounds like being reasonably well checked and guarded against (famous last words ;P).

sudden6 (approved):

anthonybilinski:
PR to fix the warning on qTox sub-libraries up here: https://github.com/qTox/qTox/pull/6581

PR #6580, opened 2022-03-19
Fix and heal database resume_file_id

### LINKED ISSUE #6553: Export to file: assert failure due to missing file_restart_id in file_history
##### Brief Description

OS: Gentoo Linux
qTox version: git master
Commit hash: 9145dc1cae03f2a316bf88760b9ae0aeec1e29fd
toxcore: 0.2.13
Qt: 5.15.3 (git master with KDE patches)

Reproducible: Always

##### Steps to reproduce

1. Open a chat (with a lot of history?).
2. Right click the chat area and select 'Export to file'.
3. Confirm a file name and start exporting.

##### Observed Behavior

``` gdb
qtox: /build/src/path/persistence/history.cpp:1113: History::getMessagesForFriend(const ToxPk&, size_t, size_t)::<lambda(const QVector<QVariant>&)>: Assertion `!it->isNull()' failed.

Thread 26 "qTox Database" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fff8dffb640 (LWP 31709)]
0x00007ffff4adcebc in ?? () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff4adcebc in  () at /lib64/libc.so.6
#1  0x00007ffff4a8ed22 in raise () at /lib64/libc.so.6
#2  0x00007ffff4a79449 in abort () at /lib64/libc.so.6
#3  0x00007ffff4a79375 in  () at /lib64/libc.so.6
#4  0x00007ffff4a87b02 in  () at /lib64/libc.so.6
#5  0x00005555556ac796 in operator()(QVector<QVariant> const&) const (__closure=0x5555590af258, row=...)
    at /build/src/path/persistence/history.cpp:1113
#6  0x00005555556b513e in std::__invoke_impl<void, History::getMessagesForFriend(const ToxPk&, size_t, size_t)::<lambda(const QVector<QVariant>&)>&, const QVector<QVariant>&>(std::__invoke_other, struct {...} &) (__f=...) at /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/include/g++-v11/bits/invoke.h:61
#7  0x00005555556b3841 in std::__invoke_r<void, History::getMessagesForFriend(const ToxPk&, size_t, size_t)::<lambda(const QVector<QVariant>&)>&, const QVector<QVariant>&>(struct {...} &) (__fn=...) at /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/include/g++-v11/bits/invoke.h:154
#8  0x00005555556b180f in std::_Function_handler<void(const QVector<QVariant>&), History::getMessagesForFriend(const ToxPk&, size_t, size_t)::<lambda(const QVector<QVariant>&)> >::_M_invoke(const std::_Any_data &, const QVector<QVariant> &) (__functor=..., __args#0=...) at /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/include/g++-v11/bits/std_function.h:290
#9  0x000055555569ffcc in std::function<void (QVector<QVariant> const&)>::operator()(QVector<QVariant> const&) const (this=0x5555590af258, __args#0=...)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/11.2.1/include/g++-v11/bits/std_function.h:590
#10 0x000055555569c925 in RawDatabase::process() (this=0x5555564c9460) at /build/src/path/persistence/db/rawdatabase.cpp:827
#11 0x00005555555c3ac3 in RawDatabase::qt_static_metacall(QObject*, QMetaObject::Call, int, void**)
    (_o=0x5555564c9460, _c=QMetaObject::InvokeMetaMethod, _id=6, _a=0x5555590c25c8)
    at /build/path/qtox_static_autogen/E5CNODCRRG/moc_rawdatabase.cpp:106
#12 0x00007ffff5110f2c in QObject::event(QEvent*) () at /usr/lib64/libQt5Core.so.5
#13 0x00007ffff5c07c39 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib64/libQt5Widgets.so.5
#14 0x00007ffff50e352b in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib64/libQt5Core.so.5
#15 0x00007ffff50e6c45 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /usr/lib64/libQt5Core.so.5
#16 0x00007ffff5139863 in  () at /usr/lib64/libQt5Core.so.5
#17 0x00007ffff17a47bd in g_main_context_dispatch () at /usr/lib64/libglib-2.0.so.0
#18 0x00007ffff17a4a78 in  () at /usr/lib64/libglib-2.0.so.0
#19 0x00007ffff17a4b2f in g_main_context_iteration () at /usr/lib64/libglib-2.0.so.0
#20 0x00007ffff51392f4 in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib64/libQt5Core.so.5
#21 0x00007ffff50e1f0b in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib64/libQt5Core.so.5
#22 0x00007ffff4f298fa in QThread::exec() () at /usr/lib64/libQt5Core.so.5
#23 0x00007ffff4f2aaf5 in  () at /usr/lib64/libQt5Core.so.5
#24 0x00007ffff4adb18d in  () at /lib64/libc.so.6
#25 0x00007ffff4b60afc in  () at /lib64/libc.so.6
```

##### Expected Behavior

No crashy crashy.

##### Additional Info

Don't think I can regression test this due to the database incompatibilities.

Not sure if there's something corrupt in the particular history, or if it's just due to its large size (might go back to 2015 with more or less activity almost every day).  I tried exporting some other older logs, which are a lot smaller, and they're fine.

anthonybilinski:
* fix(history): Don't cast file_id to QString before inserting
    
    The implicit cast in addNewFileMessage's argument causes a conversion from the
binary data of file_id to be interpreted as UTF8 characters. On null bytes or
invalid UTF8, the resulting QString can be empty or truncated. Empty IDs causes
null file_restart_id's to be inserted into the database.
    
    Fixes the cause of #6553, but the database damage still needs to be healed.

* fix(history): Heal invalid resume_file_id's saved to history
    
    Replace underlength resume_file_id's with arbitrary data. Loaded file IDs are
not used for anything at this time, but they should not be null nor invalid.
Any 32-byte value is valid, so use all 0's for consistency.
    
    Fix #6553

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6580)

### DIFF
--- a/src/persistence/dbupgrader.cpp
+++ b/src/persistence/dbupgrader.cpp
@@ -26,7 +26,7 @@
 #include <QString>
 
 namespace {
-constexpr int SCHEMA_VERSION = 9;
+constexpr int SCHEMA_VERSION = 10;
 
 struct BadEntry
 {
@@ -262,7 +262,8 @@ bool DbUpgrader::dbSchemaUpgrade(std::shared_ptr<RawDatabase>& db)
     using DbSchemaUpgradeFn = bool (*)(RawDatabase&);
     std::vector<DbSchemaUpgradeFn> upgradeFns = {dbSchema0to1, dbSchema1to2, dbSchema2to3,
                                                  dbSchema3to4, dbSchema4to5, dbSchema5to6,
-                                                 dbSchema6to7, dbSchema7to8, dbSchema8to9};
+                                                 dbSchema6to7, dbSchema7to8, dbSchema8to9,
+                                                 dbSchema9to10};
 
     assert(databaseSchemaVersion < static_cast<int>(upgradeFns.size()));
     assert(upgradeFns.size() == SCHEMA_VERSION);
@@ -608,5 +609,20 @@ bool DbUpgrader::dbSchema8to9(RawDatabase& db)
     const auto badPeers = getInvalidPeers(db);
     mergeDuplicatePeers(upgradeQueries, db, badPeers);
     upgradeQueries += RawDatabase::Query(QStringLiteral("PRAGMA user_version = 9;"));
+    return db.execNow(upgradeQueries);
+}
+
+bool DbUpgrader::dbSchema9to10(RawDatabase& db)
+{
+    // not technically a schema update, but still a database version update based on healing invalid user data.
+    // We inserted some empty resume_file_id's due to #6553. Heal the under-length entries.
+    // The resume file ID isn't actually used for loaded files at this time, so we can heal it to an arbitrary
+    // value of full length.
+    constexpr int resumeFileIdLengthNow = 32;
+    QByteArray dummyResumeId(resumeFileIdLengthNow, 0);
+    QVector<RawDatabase::Query> upgradeQueries;
+    upgradeQueries += RawDatabase::Query(QStringLiteral("UPDATE file_transfers SET file_restart_id = ? WHERE LENGTH(file_restart_id) != 32;"),
+        {dummyResumeId});
+    upgradeQueries += RawDatabase::Query(QStringLiteral("PRAGMA user_version = 10;"));
     return db.execNow(upgradeQueries);
 }
--- a/src/persistence/dbupgrader.h
+++ b/src/persistence/dbupgrader.h
@@ -37,4 +37,5 @@ namespace DbUpgrader
     bool dbSchema6to7(RawDatabase& db);
     bool dbSchema7to8(RawDatabase& db);
     bool dbSchema8to9(RawDatabase& db);
+    bool dbSchema9to10(RawDatabase& db);
 }
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -280,7 +280,7 @@ void History::removeFriendHistory(const ToxPk& friendPk)
     }
 }
 
-void History::onFileInserted(RowId dbId, QString fileId)
+void History::onFileInserted(RowId dbId, QByteArray fileId)
 {
     auto& fileInfo = fileInfos[fileId];
     if (fileInfo.finished) {
@@ -330,7 +330,7 @@ History::generateNewFileTransferQueries(const ToxPk& friendPk, const ToxPk& send
                                .arg(insertionData.size)
                                .arg(insertionData.direction)
                                .arg(ToxFile::CANCELED),
-                           {dispName.toUtf8(), insertionData.fileId.toUtf8(),
+                           {dispName.toUtf8(), insertionData.fileId,
                             insertionData.fileName.toUtf8(), insertionData.filePath.toUtf8(),
                             QByteArray()},
                            [weakThis, fileId](RowId id) {
@@ -362,7 +362,7 @@ RawDatabase::Query History::generateFileFinished(RowId id, bool success, const Q
     }
 }
 
-void History::addNewFileMessage(const ToxPk& friendPk, const QString& fileId,
+void History::addNewFileMessage(const ToxPk& friendPk, const QByteArray& fileId,
                                 const QString& fileName, const QString& filePath, int64_t size,
                                 const ToxPk& sender, const QDateTime& time, QString const& dispName)
 {
@@ -435,7 +435,7 @@ void History::addNewMessage(const ToxPk& friendPk, const QString& message, const
                                                 extensionSet, dispName, insertIdCallback));
 }
 
-void History::setFileFinished(const QString& fileId, bool success, const QString& filePath,
+void History::setFileFinished(const QByteArray& fileId, bool success, const QString& filePath,
                               const QByteArray& fileHash)
 {
     if (historyAccessBlocked()) {
@@ -562,7 +562,7 @@ QList<History::HistMessage> History::getMessagesForFriend(const ToxPk& friendPk,
             it = std::next(row.begin(), fileOffset);
             assert(!it->isNull());
             const auto fileKind = TOX_FILE_KIND_DATA;
-            const auto resumeFileId = (*it++).toString().toUtf8();
+            const auto resumeFileId = (*it++).toByteArray();
             const auto fileName = (*it++).toString();
             const auto filePath = (*it++).toString();
             const auto filesize = (*it++).toLongLong();
--- a/src/persistence/history.h
+++ b/src/persistence/history.h
@@ -117,7 +117,7 @@ struct FileDbInsertionData
     FileDbInsertionData();
 
     RowId historyId;
-    QString fileId;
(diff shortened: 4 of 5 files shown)


robinlinden (approved):
I think this looks fine, but maybe sudden6 should have a look as well since there's db trickery going on and I'm not that familiar with the qTox codebase.

sudden6:
Does this also make the `file_id` have the correct data type in the db? Or do we not have a strict db schema and everything is variable length fields?

Code changes and test lgtm.

anthonybilinski:
The second, SQLite unlike.. pretty much every other SQL database uses duck-typing instead of static typing: https://www.sqlite.org/datatype3.html

> Any column in an SQLite version 3 database, except an [INTEGER PRIMARY KEY](https://www.sqlite.org/lang_createtable.html#rowid) column, may be used to store a value of any storage class.

So even though our file_restart_id was type "BLOB" from the start: https://github.com/qTox/qTox/blob/b067aae207a5439fbe8e2b221a5fe74678a2a1ab/src/persistence/dbupgrader.cpp#L321 it still allowed strings to be inserted.

Testing just with a new empty local db:
```
CREATE TABLE "test_table" ( "test_blob" BLOB );
INSERT INTO "test_table" VALUES('test string data');
INSERT INTO "test_table" VALUES(0x00);
INSERT INTO "test_table" VALUES(x'000102030405060708091011');
INSERT INTO "test_table" VALUES(?);
```
The first two are "Text / Numeric", the third is BLOB, and the fourth is NULL.

@Chiitoo who was experiencing this issue on their db here: https://github.com/qTox/qTox/issues/6553 saw a few of the file_resume_ids still show as strings after the fix - I think that's because their 32 bytes of binary data happen to be interpret-able as valid UTF8 data.

So in short, yes I think our types in the db were/are ok. As long as when we insert and select as binary data, even if sqlite thinks some rows are strings, that shouldn't effect us.

anthonybilinski:
Looking into this a bit more, we always bind arguments as blob using `sqlite3_bind_blob` https://github.com/qTox/qTox/blob/b067aae207a5439fbe8e2b221a5fe74678a2a1ab/src/persistence/db/rawdatabase.cpp#L803 when we could be using any of the binds_* APIs. It doesn't seem like there's much of a difference between them, except that the fixed width type binds don't require a length, and the text binds can calculate their own buffer length.

But then when taking the data back out: https://github.com/qTox/qTox/blob/b067aae207a5439fbe8e2b221a5fe74678a2a1ab/src/persistence/db/rawdatabase.cpp#L894 we construct the QVariant based on sqlite3_column_type, then back in History when handling the callback, we cast the QVariant to the type we expect.

So this seems a bit inconsistent, if we're going to just insert everything as blobs regardless of declared type in the table, we should probably return it all as QByteArray to History, then let it cast back to a UTF8 or whatever it wants. But regardless it doesn't seem like our type isn't `blob` _enough_, if anything it's on the side of over-`blob`-ing data.

sudden6:
It might be an improvement to use the correct APIs to insert the data, not 100% sure if they would have caught this exact issue, since you say SQLite accepts anything in any field, but it's a layer of protection against accidental mistakes.

anthonybilinski:
Reading the APIs, I think literally the only difference of using `sqlite3_bind_text` on text rather than `sqlite3_bind_blob` on the raw representation of UTF8 strings is that:

> ** ^If the fourth parameter to sqlite3_bind_text() or sqlite3_bind_text16()
** is negative, then the length of the string is
** the number of bytes up to the first zero terminator.

But the API documentation isn't very clear. I think it would have literally no benefit to `sqlite3_bind_text()` on text rather than `sqlite3_bind_blob()` either way we'd need to pass in a raw buffer, we'd just have the added option of not supplying length, which we already know.

It would also mean that we'd need to track the bound parameters types when added to a RawDatabase::Query, which are currently all implicitly cast to QByteArray. If we did do that we could use all of:

```
SQLITE_API int sqlite3_bind_blob(sqlite3_stmt*, int, const void*, int n, void(*)(void*));
SQLITE_API int sqlite3_bind_blob64(sqlite3_stmt*, int, const void*, sqlite3_uint64,
                        void(*)(void*));
SQLITE_API int sqlite3_bind_double(sqlite3_stmt*, int, double);
SQLITE_API int sqlite3_bind_int(sqlite3_stmt*, int, int);
SQLITE_API int sqlite3_bind_int64(sqlite3_stmt*, int, sqlite3_int64);
SQLITE_API int sqlite3_bind_null(sqlite3_stmt*, int);
SQLITE_API int sqlite3_bind_text(sqlite3_stmt*,int,const char*,int,void(*)(void*));
SQLITE_API int sqlite3_bind_text16(sqlite3_stmt*, int, const void*, int, void(*)(void*));
SQLITE_API int sqlite3_bind_text64(sqlite3_stmt*, int, const char*, sqlite3_uint64,
                         void(*)(void*), unsigned char encoding);
SQLITE_API int sqlite3_bind_value(sqlite3_stmt*, int, const sqlite3_value*);
SQLITE_API int sqlite3_bind_pointer(sqlite3_stmt*, int, void*, const char*,void(*)(void*));
SQLITE_API int sqlite3_bind_zeroblob(sqlite3_stmt*, int, int n);
SQLITE_API int sqlite3_bind_zeroblob64(sqlite3_stmt*, int, sqlite3_uint64);
```

But we don't support binding numeric values, they're just done in-line since there's no risk of SQL injection there. I don't think we have any use for zerblob, and otherwise if text and blob are the same I think using one API vs thirteen basically has no difference.

From what I've read the difference between e.g. `sqlite3_bind_blob` and `sqlite3_bind_blob64` is that length can be over 2^32, but why on earth would we ever store something that big in a db.

I'm mostly going off the documentation in their header which isn't very complete (many paragraphs about the 13 APIs together, but some parameters or even whole API aren't mentioned at all) and a combination of not very official online sites I can find..

anthonybilinski:
Let me know if not adding support for `sqlite3_bind_text` is acceptable given my last comment - if so I'll merge. A bit ambiguous with the approval but remaining discussion :D

sudden6:
SQLite doesn't seem to provide the kind of typesafe API I had in mind, so I think this is good to merge.

PR #6581, opened 2022-03-19
chore(build): Enable all warnings on qTox libraries as well 

anthonybilinski:
warnings interface was only being linked to qtox_static, so were not being
applied to e.g. util. Link it to each library as well as qtox_static so that
they all inherit all the warnings.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6581)


audio/src/backend/openal.cpp
```diff
@@ -135,7 +135,7 @@ qreal OpenAL::outputVolume() const
         checkAlError();
     }
 
-    return volume;
+    return static_cast<qreal>(volume);
```
robinlinden: You could also change the return type of this function since this is the only implementation. ...or delete it altogether since it looks like it's completely unused?

audio/src/backend/openal.cpp
```diff
@@ -644,7 +644,7 @@ void OpenAL::doInput()
 
     applyGain(inputBuffer, AUDIO_FRAME_SAMPLE_COUNT_TOTAL, gainFactor);
 
-    float volume = getVolume();
+    auto volume = static_cast<qreal>(getVolume());
```
robinlinden: I'm surprised this change doesn't cause a warning on line 657 where `IAudioSource::volumeAvailable(float)` is called. :x Looks like there's a lot of back and forth between double and float in the audio code.

robinlinden (approved):

audio/src/backend/openal.cpp
anthonybilinski: Removed it completely.

audio/src/backend/openal.cpp
anthonybilinski: Removed some flipping back and forth of types, OpenAl::getVolume now returns qreal and IAudioSource::volumeAvailable now sends qreal.

That removes this case.

PR #6583, opened 2022-03-22
Split SQL peers table into new chats table and new authors table

anthonybilinski:
Based on discussion here: https://github.com/qTox/qTox/pull/6561#issuecomment-1073298383

Going from:
image

To:
image

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6583)


anthonybilinski:
Note: testing of this should be done on top of https://github.com/qTox/qTox/pull/6586, although it's independent. The bug there causes corrupt ToxPk's to be used in ChatHistory, which only has the observable effect of self's name not being bolded on chat loaded from history (that I've seen).

Chiitoo:
I'm not sure if I broke my database while trying to find out if there was some regression causing

```
core/core.cpp:839 : Warning: No bootstrap node list
net/bootstrapnodeupdater.cpp:66 : Debug: Node is missing required fields.
```

and no connectivity, which I now again mysteriously have... but testing this (on top of current 'master') and trying to load chat history gives me:

```
[14:39:05.902 UTC] persistence/db/rawdatabase.cpp:780 : Warning: Failed to prepare statement "SELECT (SELECT COUNT(*) - 1 FROM history countHistory JOIN chats ON chat_id = chats.id WHERE chats.uuid = ?AND countHistory.id <= history.id), (timestamp / 1000 / 60 / 60 / 24) AS day FROM history JOIN chats ON chat_id = chats.id WHERE chats.uuid = ? AND timestamp >= 1646085600000 GROUP by day LIMIT 31;" and returned 1
[14:39:05.902 UTC] persistence/db/rawdatabase.cpp:782 : Warning: The full error is 1: no such table: chats
```

Additionally, when I send or receive a message:

```
[14:41:08.990 UTC] persistence/db/rawdatabase.cpp:837 : Warning: Error executing query "BEGIN;"
```

Oh and while going back and forth on commits, at one point I hit this:

```
[14:24:11.205 UTC] persistence/paths.cpp:108 : Debug: Auto portable mode: false
[14:24:11.206 UTC] persistence/settings.cpp:110 : Debug: "Loading settings from /[...]/qtox.ini"
[14:24:11.206 UTC] persistence/globalsettingsupgrader.cpp:35 : Critical: Failed to write to "/[...]/bootstrapNodes.backup.json" aborting bootstrap node upgrade.
[14:24:11.206 UTC] persistence/globalsettingsupgrader.cpp:70 : Critical: Failed to upgrade settings to version 1 aborting
terminate called without an active exception
Aborted
```

I deleted said file and that issue went away.

anthonybilinski:
The "`[14:39:05.902 UTC] persistence/db/rawdatabase.cpp:782 : Warning: The full error is 1: no such table: chats` looks like it missed the upgrade completely, but didn't fail since that would close the database. Are you running on top of https://github.com/qTox/qTox/pull/6580? They're both database version 10 (for now, since neither is merged yet), so if you received that upgrade, this one will skip but assume it's done.

```
[14:24:11.206 UTC] persistence/globalsettingsupgrader.cpp:35 : Critical: Failed to write to "/[...]/bootstrapNodes.backup.json" aborting bootstrap node upgrade.
[14:24:11.206 UTC] persistence/globalsettingsupgrader.cpp:70 : Critical: Failed to upgrade settings to version 1 aborting```
```
Is probably from our GlobalSetttingUpgrade running from version 0 to 1 after it has already run. Its version is stored in the global qtox.ini. So if your qtox.ini was removed or restored from an older version, but your `bootstrapNodes.backup.json` remained - it will break things. This is intentional so that the backed up (possibly) user file isn't overwritten by qTox.. but maybe just numbering the backups would be less disruptive. An abort in that case is expected because we don't load Settings if we can't upgrade it, which is needed for just about everything. That gained a GUI pop-up notification yesterday.

```
core/core.cpp:839 : Warning: No bootstrap node list
net/bootstrapnodeupdater.cpp:66 : Debug: Node is missing required fields.
```

Regarding this, did you have a custom bootstrap node list? If so we added a required field fairly recently that wasn't required before: tcp_ports. It's present in the example nodelist provided, and can be empty, but we check for its presence.

Our lack of backwards compatibility is really ramping up..

Also thanks for testing out the PR :D

Chiitoo:
No, I did not have https://github.com/qTox/qTox/pull/6580 included since this one won't apply with it.

No custom bootstrap node list either.  I did have an ISP-related(?) connection drop around the time, so I'm probably going to blame that, even though other things worked just fine while qTox did not... it was weird.

anthonybilinski:
But even without #6580 applied, had your database been restored after running #6580?

i.e. I'm thinking that #6580 upgraded your db to version 10 which it saved in the db, then this PR saw the db was already version 10, did nothing on upgrade, then the database wasn't in a state it expected (i.e. no chats table). I can't imagine how else it would pass the upgrade but fail to find a chats table.. any error during the upgrade which creates the chats table results in us not accessing history.

Very strange about the boostrap list.. will look into that some more. Neither of the error logs indicate a network problem, just a reading of a local file and populating the bootstrap node list problem (despite net/bootstrapnodeupdater.cpp implying it's network, it only handles the bundled node list currently):

Chiitoo:
Ugh, was writing something here and the computer restarted on itself (been troubleshooting stuff with it just now too and some BIOS setting it needs to be more stable were gone).

I can't be sure about the timeline now.  I tried a backup of the profile and I started hitting the connection issue again... which is... weird.  I need to do some more RAM testing for RMA related things right now though which is not helping.

anthonybilinski:
Gah annoying about the RAM - I had some RAM stability problems myself a few months ago, what a pain.

RE: database version, I think it's fairly safe to assume that the upgrade didn't run. I can't see another way where you wouldn't get a database upgrade failure if the peers table doesn't exist. The error checking is handled by sqlite not us, so should be pretty robust.

RE: connectivity, check if you have a `~/.config/tox/bootstrapNodes.json` - after the global settings upgrade to version 1 fairly recently on master, that file should not exist unless you have a user set file. Instead, your old `~/.config/tox/bootstrapNodes.json` (qTox generated in your case - but could have been user set in the past) is moved to `~/.config/tox/bootstrapNodes.backup.json`, and the new qTox generated list is at `~/.config/tox/bootstrapNodes.example.json`

If you have a `~/.config/tox/bootstrapNodes.json` file running on master tip, try removing it and restarting qTox - then it will use the built-in list instead of whatever's in that file. My current guess is that that file is still around and has some entries that qTox doesn't like i.e. `Node is missing required fields`.

anthonybilinski:
To reviewers: I've done a fair bit of refactoring and extended the unit test on this significantly - so it now depends on https://github.com/qTox/qTox/pull/6589, https://github.com/qTox/qTox/pull/6590, https://github.com/qTox/qTox/pull/6591, https://github.com/qTox/qTox/pull/6592.

test/persistence/dbschema_test.cpp
```diff
@@ -54,192 +48,54 @@ private slots:
...
-    {"faux_offline_pending", "CREATE TABLE faux_offline_pending (id INTEGER PRIMARY KEY, required_extensions INTEGER NOT NULL DEFAULT 0, FOREIGN KEY (id) REFERENCES history(id))"},
-    {"file_transfers", "CREATE TABLE file_transfers (id INTEGER PRIMARY KEY, chat_id INTEGER NOT NULL, file_restart_id BLOB NOT NULL, file_name BLOB NOT NULL, file_path BLOB NOT NULL, file_hash BLOB NOT NULL, file_size INTEGER NOT NULL, direction INTEGER NOT NULL, file_state INTEGER NOT NULL)"},
-    {"history", "CREATE TABLE history (id INTEGER PRIMARY KEY, timestamp INTEGER NOT NULL, chat_id INTEGER NOT NULL, sender_alias INTEGER NOT NULL, message BLOB NOT NULL, file_id INTEGER, FOREIGN KEY (file_id) REFERENCES file_transfers(id), FOREIGN KEY (chat_id) REFERENCES peers(id), FOREIGN KEY (sender_alias) REFERENCES aliases(id))"},
-    {"peers", "CREATE TABLE peers (id INTEGER PRIMARY KEY, public_key TEXT NOT NULL UNIQUE)"},
-    {"broken_messages", "CREATE TABLE broken_messages (id INTEGER PRIMARY KEY, reason INTEGER NOT NULL DEFAULT 0, FOREIGN KEY (id) REFERENCES history(id))"},
-    {"chat_id_idx", "CREATE INDEX chat_id_idx on history (chat_id)"}
+    std::unique_ptr<QTemporaryFile> testDatabaseFile;
```
sphaerophoria, line 97: I thought there would be a way to do this without a unique_ptr, but once again QT's is not what I would guess.

test/persistence/dbschema_test.cpp
```diff
@@ -54,192 +48,54 @@ private slots:
...
-}
-
-void TestDbSchema::createSchemaAtVersion(std::shared_ptr<RawDatabase> db, const std::vector<SqliteMasterEntry>& schema)
-{
-    QVector<RawDatabase::Query> queries;
-    for (auto const& entry : schema) {
-        queries += entry.sql;
-    }
-    QVERIFY(db->execNow(queries));
+    testDatabaseFile.reset();
```
sphaerophoria: Should this resize(0) as well? Docs just say that it seeks to the start, but I could imagine that reading a db with garbage on the end could still produce unexpected results.

src/persistence/db/upgrades/to10.h
```diff
@@ -0,0 +1,55 @@
...
+
+#pragma once
+
+#include "src/persistence/db/rawdatabase.h"
+
+#include <QVector>
+
+class ToxPk;
+
+namespace to10
```
sphaerophoria: Pretty odd global namespace IMO. I think this should be like dbto10 or db::to10 or dbupgrade::to10.

src/persistence/db/upgrades/to10.h
```diff
@@ -0,0 +1,55 @@
+/*
```
sphaerophoria: :O that's a lot of table upgrades

src/persistence/history.cpp
```diff
@@ -270,12 +282,16 @@ void History::removeFriendHistory(const ToxPk& friendPk)
...
                                 "DELETE FROM history WHERE chat_id=%1; "
-                                "DELETE FROM aliases WHERE owner=%1; "
-                                "DELETE FROM peers WHERE id=%1; "
+                                "DELETE FROM chats WHERE id=%1; "
+                                "DELETE FROM aliases WHERE owner=%2; "
+                                "DELETE FROM authors WHERE id=%2; "
                                 "VACUUM;")
-                            .arg(generatePeerIdString(friendPk));
+                            .arg(generateChatIdString().arg(generateAuthorIdString())),
+                            {friendPk.getByteArray(), friendPk.getByteArray(), friendPk.getByteArray(), friendPk.getByteArray(),
```
sphaerophoria: what's happening here?

src/persistence/history.cpp
```diff
@@ -43,23 +43,33 @@ MessageState getMessageState(bool isPending, bool isBroken)
     return messageState;
 }
 
-QString generatePeerIdString(ToxPk const& pk)
+QString generateAuthorIdString()
```
sphaerophoria: I don't have a really good answer here, but I will note that this is pretty confusing. It's not clear when you call this that you have to bind a blob. You have to know what SQL the function is going to write to know what arguments to add.

An expecially bad exmaple is generateNewFileTransferQueries, that's pretttty tough to follow which arguments are binding where.

That being said the alternative is writing this same statement every time you want to get an author :\

src/persistence/db/upgrades/dbupgrader.cpp
```diff
@@ -285,13 +287,15 @@ bool DbUpgrader::createCurrentSchema(RawDatabase& db)
 {
     QVector<RawDatabase::Query> queries;
```
sphaerophoria, line 288: I like the idea of db upgrader, and I know that this isn't introduced in this commit, but I think having createCurrentSchema in history.cpp was a lot easier to follow. Not really requesting a change, just making a note.

src/persistence/history.cpp
```diff
@@ -43,23 +43,33 @@ MessageState getMessageState(bool isPending, bool isBroken)
...
 {
-    return QString("(SELECT id FROM peers WHERE public_key = '%1')").arg(pk.toString());
+    return QString("(SELECT id FROM authors WHERE public_key = ?)");
```
sphaerophoria: Not a new question, but what happens if you accidentally add the same PK twice?

src/persistence/db/upgrades/dbupgrader.cpp
```diff
@@ -285,13 +287,15 @@ bool DbUpgrader::createCurrentSchema(RawDatabase& db)
...
     queries += RawDatabase::Query(QStringLiteral(
-        "CREATE TABLE peers (id INTEGER PRIMARY KEY, "
-        "public_key TEXT NOT NULL UNIQUE);"
+        "CREATE TABLE authors (id INTEGER PRIMARY KEY, "
+        "public_key BLOB NOT NULL UNIQUE);"
+        "CREATE TABLE chats (id INTEGER PRIMARY KEY, "
+        "uuid BLOB NOT NULL UNIQUE);"
```
sphaerophoria, line 293: I guess it makes sense to do this rename now since we won't need to rename again on group history implementation, but I will note that this confused me a little. "Why is this called a uuid? It's always a friend pk".

sphaerophoria (commented):
Haven't reviewed the db9to10 stuff yet, but had some comments I wanted to post before going through the upgrade logic. Generally think the new schema makes more sense, and like the cleanup that came along with it.

test/persistence/dbschema_test.cpp
anthonybilinski: I'm not sure what you mean here? It's resetting the std::unique_ptr, destroying the QTemporaryFile, which deletes the temporary file altogether. Each test case is running with a brand-new QTemporaryFile since init() runs before each test.

test/persistence/dbschema_test.cpp
anthonybilinski: Yeah.. just makes it easy to re-construct each time without having to duplicate the boilerplate open/close.

src/persistence/history.cpp
anthonybilinski: Yeeeeahh..
We used to use Qt's QString::arg to copy the friend pk's hex string representation into the query in all places at once. Since chats and authors are blobs instead of strings, since tox keys and chat ids are inherently binary anyway, we can't just put the public keys into the string anymore. Instead we're binding the arguments, which is done for each t ime the argument is used.

So Qt's QString::arg is replacing all %1 with `(SELECT id FROM chats WHERE uuid = ?)` from `generateChatIdString`, then for each time it's called we're binding the friend pk repeatedly to the query.

It's definitely not nearly as clean as before in this case :S I still think both binary storage of the keys and binding of argument in SQL queries are conceptually better - but out string generation in History as part of queries is taking a hit.

src/persistence/history.cpp
anthonybilinski: Agreed :S

All of the generate.*String functions in History got a lot messier to use in this change. I don't think there's really an easy fix, and don't know that copy pasting the sub-query statement 15 times across the file is much better, but maybe it is.

src/persistence/history.cpp
anthonybilinski: You would get an SQL error on insert that would reject the insert, due to `public_key` being marked as `UNIQUE` in the schema.

In general for sub-queries though, I'm not sure?

src/persistence/db/upgrades/dbupgrader.cpp
anthonybilinski: Yeah fair.. just trying to save on the extra upgrade for group history where group id's are just documented as "IDs".

src/persistence/db/upgrades/to10.h
anthonybilinski: Agreed, changed "to10" to "DbTo10", "peersToAuthors" to "PeersToAuthors", "peersToChats" to "PeersToChats" and renamed the file to dbto10.cpp/.h.

src/persistence/db/upgrades/dbupgrader.cpp
anthonybilinski: Hm, I think it's nice to keep the general version handling in dbupgrader - and dbupgrader depending on history to get the current schema feels a bit weird?

I guess history could handle just `isNewDb` and `createCurrentSchema`, then dispatch to DbUpgrader::dbSchemaUpgrade otherwise.

It would also mean that tests which currently just depend on dbupgrader would depend on history as well to get the current schema made.

test/persistence/dbschema_test.cpp
sphaerophoria: 🤦 i misread as -> reset

src/persistence/history.cpp
sphaerophoria: Id probably argue that the efficiency improvements in the db arent worth this change. Hex encoding the strings is 2x the size and adds a little computational cost, but i think everything touched in history is unfortunately worse because of it,. Even though conceptually this is more correct, I think real world impact is negligible and maintencne cost is high. My recommendation is to back out the blob changes

anthonybilinski: To summarize some private discussion:
1) I agree the size/efficiency difference is tiny. Keys in general are a tiiiny portion of our rows compared to chat history, and each one is small.
2) Every other binary field in our db is BLOB - so it's a bit unintuitive from History's perspective to know to hex encode only peers/authors/chats.
3) Agreed that still overall, History's usage is pretty nasty because of this. The injection of bound arguments from the generate strings is surprising, and having to copy the same bound argument 9 times is nasty.
4) I think generally, we should favour argument binding rather than putting arguments in the query string i.e. "SQL injection mode" - so generally would like the move to binding more arguments (maybe even ints in the future).
5) I think the best of both would be if we could have argument binding in history be as clean as QString::arg's usage - i.e. we could just tell RawHistory "put this binary argument at these 9 places" and it just worked. If somehow the generateString helper functions could return both a query section with a bound arg too, and History could bind to many places all at once, that would be nice. I'll take a look at implementing this.

anthonybilinski: I've updated the generate.*String functions to take in the current query string and bound arg vector so that they can add onto both at the point in the query string where they belong. So now the query string and bound args get build together in pieces, making each bound arg much easier to follow where it goes.

It's still probably not as nice as before, but I don't think there's a great way to do similar functionality. Any parsing of the query string in rawdatabase to bind single arguments possibly multiple times seems complex and at risk of SQL injection.

src/persistence/history.cpp
anthonybilinski: Should be addressed by https://github.com/qTox/qTox/pull/6583#discussion_r835998682

src/persistence/history.cpp
sphaerophoria: This is so much better!

src/persistence/db/upgrades/dbto11.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+    // splitting peers table into separate chat and authors table.
+    // peers had a dual-meaning of each friend having their own chat, but also each peer
+    // being authors of messages. This wasn't true for self which was in the peers table
+    // but had no related chat. For upcomming group history, the confusion is amplified
+    // since each group is a chat so is in peers, but authors no messages, and each group
+    // member is an author so is in peers, but has no chat.
+    // Splitting peers makes the relation much clearer and the tables have a single meaning.
+    QVector<RawDatabase::Query> upgradeQueries;
+    PeersToAuthors::peersToAuthors(db, upgradeQueries);
```
sphaerophoria: In general there are a lot of missed bool checks in this path. I think we should fail at the top level if we fail to read any of the expected fields

src/persistence/db/upgrades/dbupgrader.cpp
sphaerophoria: It's fine the way it is, I just thought I'd note it.

src/persistence/db/upgrades/dbto11.cpp
```diff
@@ -0,0 +1,290 @@
...
+        "file_path BLOB NOT NULL, "
+        "file_hash BLOB NOT NULL, "
+        "file_size INTEGER NOT NULL, "
+        "direction INTEGER NOT NULL, "
+        "file_state INTEGER NOT NULL, "
+        "FOREIGN KEY (id, message_type) REFERENCES history(id, message_type), "
+        "FOREIGN KEY (sender_alias) REFERENCES aliases_new(id));"));
+}
+
+bool DbTo11::PeersToAuthors::populateAuthors(RawDatabase &db, QVector<RawDatabase::Query> &upgradeQueries)
```
sphaerophoria: nit: I couldn't place why I felt weird reading these last time, but I think it's because these functions are all named like they're executing the transaction now. I think a slight naming improvement would be to name these as "appendPopulateAuthorQueries". Same for the unit tests.

src/persistence/db/upgrades/dbto11.cpp
```diff
@@ -0,0 +1,290 @@
+/*
```
sphaerophoria, line 1: There's a lot of code here, but  the tests look like they ensure that the schema has upgraded successfully and the right fields are in the right spots. I'm going to consider the SQL correct based off that.

sphaerophoria (changes requested):

src/persistence/db/upgrades/dbto11.cpp
anthonybilinski: Agreed that makes more sense, renamed DbTo11 and Test10to11 functions to fit the "append.*Queries" style.

src/persistence/db/upgrades/dbto11.cpp
anthonybilinski: Doh, piped the bools up but then never checked them. Fixed.

sphaerophoria (approved):

PR #6585, opened 2022-03-24
fix(History): Correct SQL query for empty finished file transfer

anthonybilinski:
"finished" is not a field in file_transfers. Bug has been present since
initially introduced in d9b39b3102eff686a072630610c199639f0d8219.

Had no real effect since file transfers are inserted with initial state
CANCELLED, which is usually the case for a 0-length finished file, but
it still logged an SQL error.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6585)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -355,7 +355,7 @@ RawDatabase::Query History::generateFileFinished(RowId id, bool success, const Q
                                   {filePath.toUtf8(), fileHash});
     } else {
         return RawDatabase::Query(QStringLiteral("UPDATE file_transfers "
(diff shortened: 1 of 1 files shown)


anthonybilinski:
Note: testing of this should be done on top of https://github.com/qTox/qTox/pull/6586, although it's independent. The bug there causes corrupt ToxPk's to be used in ChatHistory, which only has the observable effect of self's name not being bolded on chat loaded from history (that I've seen).

sudden6 (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6585)* status: 0 of 1 LGTMs obtained

PR #6586, opened 2022-03-24
fix(History): Deep copy QByteArrays from SQL selects

anthonybilinski:
We already deep copy all other data types. Especially with Qt types like
QByteArray, having a deep copy to start is important since it shallow
copies on any subsequent copy, so the callback provider only gets a
shallow copy in the QByteArray of the temporary data from SQLite.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6586)


sudden6 (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6586)* status: 0 of 1 LGTMs obtained

PR #6591, opened 2022-03-26
refactor(tests): Move database test utilities to their own lib

anthonybilinski:
So that we can have multiple db upgrade tests that can run in parallel,
instead of one monster test that runs independent test cases sequentially.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6591)


anthonybilinski:
Merged as part of https://github.com/qTox/qTox/pull/6583.

PR #6592, opened 2022-03-26
refactor(History): Move history upgrades to new directory

anthonybilinski:
Upcoming upgrade is fairly large, and all the helper functions for each
upgrade are becoming ambiguous.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6592)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -434,6 +434,7 @@ int main(int argc, char* argv[])
     }
 
     uriDialog = std::unique_ptr<ToxURIDialog>(new ToxURIDialog(nullptr, profile->getCore(), *messageBoxManager));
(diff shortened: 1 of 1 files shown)


sphaerophoria:
I don't really think this makes sense as a PR outside the context of #6583. In isolation I would reject this as it's just moving code to move code. I'll look at #6583 instead :D

anthonybilinski:
Fair :D was trying to PR things are individually as possible without tons of merge conflicts to make it easier to review - but yeah this does end up pretty meaningless by itself.

anthonybilinski:
Merged as part of https://github.com/qTox/qTox/pull/6583.

PR #6597, opened 2022-03-29
fix(main): Construct ToxSave event handler

anthonybilinski:
Bug introduced in 0c967725df8a3f55474c59c48a6f40997c1b974e, when the class
was added but never constructed.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6597)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -434,6 +434,7 @@ int main(int argc, char* argv[])
     }
 
     uriDialog = std::unique_ptr<ToxURIDialog>(new ToxURIDialog(nullptr, profile->getCore(), *messageBoxManager));
+    toxSave = std::unique_ptr<ToxSave>(new ToxSave{*settings});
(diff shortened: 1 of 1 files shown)


sphaerophoria (approved):
So I think this fixes the bug, but why are there so many static variables here? I guess initially so that people could connect signals to them without needing class state? Seems kinda whacky. I'll just close my eyes and pretend i didn't see it 🙈

anthonybilinski:
lol agreed, would like to clean that up too. Since (some) of those are freed manually in cleanup() I didn't want to effect destruction order so just kind of rolled with it for now. Planning on moving all of that (and all other >:( ) global state into functions and classes eventually.

anthonybilinski:
Cleanup of static main variables: https://github.com/qTox/qTox/pull/6603.

PR #6598, opened 2022-03-29
fix(UI): Centre message boxes on main window

anthonybilinski:
Setting the parent of MessageBoxManager to Widget has a dual effect of
centreing the QMessageBox's on Widget' window which we want, but also
giving ownership of MessageBoxManager to Widget, which causes Widget to
destroy it in Widget's destructor. Since the original MessageBoxManager
must outlive Widget to be used in Settings and when loading Profiles
before Widget is constructed, we don't want Widget to destroy it.

Instead of juggling MessageBoxManager's parent around dynamically when
Nexus changes the active Window, just construct a second one in Widget
to be used by all its children with the parent set on Widget, centreing
the window and allowing it to have ownership.

Settings and Profile still use the original parent-less
MessageBoxManager since they show popups before Widget is constructed.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6598)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -237,7 +237,7 @@ int main(int argc, char* argv[])
         qWarning() << "Couldn't load font";
     }
 
-    messageBoxManager = std::unique_ptr<MessageBoxManager>(new MessageBoxManager());
+    messageBoxManager = std::unique_ptr<MessageBoxManager>(new MessageBoxManager(nullptr));
     settings = std::unique_ptr<Settings>(new Settings(*messageBoxManager));
 
     QString locale = settings->getTranslation();
--- a/src/nexus.cpp
+++ b/src/nexus.cpp
@@ -235,7 +235,7 @@ void Nexus::showMainGUI()
     assert(profile);
 
     // Create GUI
-    widget = new Widget(*profile, *audioControl, *cameraSource, *settings, *style, *messageBoxManager);
+    widget = new Widget(*profile, *audioControl, *cameraSource, *settings, *style);
 
     // Start GUI
     widget->init();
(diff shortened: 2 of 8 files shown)


robinlinden (approved):

sphaerophoria (approved):
I think this is an improvement from before, however I think ideally _all_ message boxes should be centered on Widget when it's present. It seems like it's inconsistent that some failures are parented to the main window and others are not.

I think some of these can be dodged. Settings does not need to store the MessageBoxManager as a member if it's passed into loadGlobal. This could probably be taken a bit further if settings construction could indicate failure somehow and return it's upgrade failures to main. 

Potential flow that keeps ownership somewhat clear.

* Settings does not take a reference to messagebox manager, instead it exposes an enum of it's status (OK, PeronalSettingsCorrupt, GlobalSettingsCorrupt..).
* Main holds MessageBoxManager as a unique ptr in main, and checks settings status
* Main calls nexus->setMessageBoxManager(manager.release()), this moves ownership to Qt's ownership model with nexus as teh owner
* Nexus calls setParent on widget when it's created.

All that being said, this is kind of nit picky, with what's submitted I think the behavior is consistent. I think it could just be slightly more future proof.

anthonybilinski:
Fair yeah - I agree that UI wise, passing ownership around to ensure message boxes are always centred when available is better.

I'll merge as-is since it's still an improvement over current behaviour. I did originally start implementation by having mixed manual management and Qt management of the same object and calling QWidget::setParent on it when Widget's constructed or destructed, but found that to just be pretty weird memory-wise. It seems like that kind of thing is strongly recommended against in Qt - but I don't know how that meshes with DI. Settings seems inherently not owned by a top level Widget. Might sort this out sometime later.

Returning an enum from Settings seems interesting, when would the error then be shown though? Is this just so that Settings doesn't rely on GUI in any way and still showing the error from main right away, or are you thinking of holding onto it until GUI is constructed so that we can show the Settings error with a centred messagebox?

PR #6602, opened 2022-03-30
refactor(main): Remove unused sodium_init

anthonybilinski:
Was used by auto-updater which was removed in 7475ba4689fe84913d6feb6d8c6fb79b1d39c5f3

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6602)

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -41,7 +41,6 @@
 
 #include <QtWidgets/QMessageBox>
 #include <ctime>
-#include <sodium.h>
 #include <stdio.h>
 
 #if defined(Q_OS_UNIX)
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -290,12 +290,6 @@ int main(int argc, char* argv[])
         qWarning() << "Can't init IPC, maybe we're in a jail? Continuing with reduced multi-client functionality.";
     }
 
-    // For the auto-updater
-    if (sodium_init() < 0) {
```
robinlinden, line 294: I'm guessing you can remove the sodium include as well now? :x

robinlinden (approved):

src/main.cpp
anthonybilinski: Yup, done.

PR #6603, opened 2022-03-30
refactor(main): Move main global state into new AppManager class

anthonybilinski:
main() couldn't hold its own state since cleanup() needed access to it.
Having the state in a class allows QApplication::aboutToQuit to call
into a member function with access to the state.

Data used in logging still needs to be global due to
qInstallMessageHandler not accepting a void* to get back to this.

Set QGuiApplication attributes before constructing AppManager since they
must be set before the QApplication is constructed.

Keep logic largely unchanged, but with a defined destruction order.

Destruction is still abnormal due to QApplication::aboutToQuit forcing
us to do partial cleanup before QApplication returns, since some OSes
will kill qTox before the QApplication returns.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6603)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,6 +171,16 @@ if(NOT "${SMILEYS}" STREQUAL "DISABLED")
 endif()
 
 set(${PROJECT_NAME}_SOURCES
+  src/appmanager.cpp
+  src/appmanager.h
+  src/friendlist.cpp
+  src/friendlist.h
+  src/grouplist.cpp
+  src/grouplist.h
+  src/ipc.cpp
+  src/ipc.h
+  src/nexus.cpp
+  src/nexus.h
   src/chatlog/chatlinecontent.cpp
   src/chatlog/chatlinecontent.h
   src/chatlog/chatlinecontentproxy.cpp
@@ -250,12 +260,6 @@ set(${PROJECT_NAME}_SOURCES
   src/core/chatid.h
   src/core/toxstring.cpp
   src/core/toxstring.h
-  src/friendlist.cpp
-  src/friendlist.h
-  src/grouplist.cpp
-  src/grouplist.h
-  src/ipc.cpp
-  src/ipc.h
   src/model/about/aboutfriend.cpp
   src/model/about/aboutfriend.h
   src/model/about/iaboutfriend.cpp
@@ -315,8 +319,6 @@ set(${PROJECT_NAME}_SOURCES
   src/net/avatarbroadcaster.h
   src/net/toxuri.cpp
   src/net/toxuri.h
-  src/nexus.cpp
-  src/nexus.h
   src/persistence/db/rawdatabase.cpp
   src/persistence/db/rawdatabase.h
   src/persistence/db/upgrades/dbupgrader.cpp
--- a/src/appmanager.cpp
+++ b/src/appmanager.cpp
@@ -1,0 +1,444 @@
+/*
+    Copyright © 2022 by The qTox Project Contributors
+
+    This file is part of qTox, a Qt-based graphical interface for Tox.
+
+    qTox is libre software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    qTox is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with qTox.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include "appmanager.h"
+
+#include "src/widget/tool/messageboxmanager.h"
+#include "src/persistence/settings.h"
+#include "src/persistence/toxsave.h"
+#include "src/persistence/profile.h"
+#include "src/ipc.h"
+#include "src/widget/translator.h"
+#include "src/nexus.h"
+#include "src/net/toxuri.h"
+#include "src/widget/widget.h"
+
+#if defined(Q_OS_UNIX)
+#include "src/platform/posixsignalnotifier.h"
+#endif
+
+#include <QApplication>
+#include <QFontDatabase>
+#include <QCommandLineParser>
+#include <QDir>
+#include <QMessageBox>
+#include <QObject>
+
+namespace
+{
+// logMessageHandler and associated data must be static due to qInstallMessageHandler's
+// inability to register a void* to get back to a class
+#ifdef LOG_TO_FILE
+QAtomicPointer<FILE> logFileFile = nullptr;
+QList<QByteArray>* logBuffer =
+    new QList<QByteArray>(); // Store log messages until log file opened
+QMutex* logBufferMutex = new QMutex();
+#endif
+
+void logMessageHandler(QtMsgType type, const QMessageLogContext& ctxt, const QString& msg)
+{
+    // Silence qWarning spam due to bug in QTextBrowser (trying to open a file for base64 images)
+    if (QString::fromUtf8(ctxt.function) == QString("virtual bool QFSFileEngine::open(QIODevice::OpenMode)")
+        && msg == QString("QFSFileEngine::open: No file name specified")) {
+        return;
+    }
+    if (msg.startsWith("Unable to find any suggestion for")) {
+        // Prevent sonnet's complaints from leaking user chat messages to logs
+        return;
+    }
+
+    if (msg == QString("attempted to send message with network family 10 (probably IPv6) on IPv4 socket")) {
+        // non-stop c-toxcore spam for IPv4 users: https://github.com/TokTok/c-toxcore/issues/1432
+        return;
+    }
+
+    QRegExp snoreFilter{QStringLiteral("Snore::Notification.*was already closed")};
+    if (type == QtWarningMsg
+        && msg.contains(snoreFilter))
+    {
+        // snorenotify logs this when we call requestCloseNotification correctly. The behaviour still works, so we'll
+        // just mask the warning for now. The issue has been reported upstream:
+        // https://github.com/qTox/qTox/pull/6073#pullrequestreview-420748519
+        return;
+    }
+
+    QString file = QString::fromUtf8(ctxt.file);
+    // We're not using QT_MESSAGELOG_FILE here, because that can be 0, NULL, or
+    // nullptr in release builds.
+    QString path = QString(__FILE__);
+    path = path.left(path.lastIndexOf('/') + 1);
+    if (file.startsWith(path)) {
+        file = file.mid(path.length());
+    }
+
+    // Time should be in UTC to save user privacy on log sharing
+    QTime time = QDateTime::currentDateTime().toUTC().time();
+    QString LogMsg =
+        QString("[%1 UTC] %2:%3 : ").arg(time.toString("HH:mm:ss.zzz")).arg(file).arg(ctxt.line);
+    switch (type) {
+    case QtDebugMsg:
+        LogMsg += "Debug";
+        break;
+    case QtInfoMsg:
+        LogMsg += "Info";
+        break;
+    case QtWarningMsg:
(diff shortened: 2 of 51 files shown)


bodwok (commented):
Tested it worked well for me. The more entities encapsulated, the easier it will be to make changes.

> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6603#-:-N-iDqsb531UiS6_gq-O:b-aq9866)* status: 0 of 1 LGTMs obtained

___
*[CMakeLists.txt, line 318 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iEHv9-G-woweclcuc:-N-iEHv9-G-woweclcud:bqxhly5) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/CMakeLists.txt#L318)):*
> ```CMake
>   src/net/toxuri.cpp
>   src/net/toxuri.h
>   src/appmanager.cpp
> ```

Put it in alphabetical order?
___
*[src/appmanager.h, line 22 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iFcvt5xYqpAF9L8f7:-N-iFcvuDHnJ0cF6NcNL:bgo6op8) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.h#L22)):*
> ```objc
> #pragma once
> 
> #include "src/persistence/toxsave.h"
> ```

It looks like an unnecessary inclusion
___
*[src/appmanager.h, line 44 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iGCCrAEepHHRhP0YR:-N-iGCCrAEepHHRhP0YS:bu57gcw) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.h#L44)):*
> ```objc
>     int run();
> 
> public slots:
> ```

Make `cleanup()` private for more encapsulation?



_Code quote:_
```objc
public slots:
    void cleanup();
```
___
*[src/appmanager.cpp, line 165 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iH4lZ8QiXrjDE_ScH:-N-iH4lZ8QiXrjDE_ScI:b-hyhddf) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L165)):*
> ```cpp
> 
> AppManager::AppManager(int argc_, char** argv_)
>     : argc{(qInstallMessageHandler(logMessageHandler), argc_)}
> ```

How does initialization work for `argc` here? 
Is there a reason why `QApplication` is not initialized using `argc_` and `argv_` directly from the `AppManager` arguments?
___
*[src/appmanager.cpp, line 171 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iHeuW1cyOxBGahZKa:-N-iHeuW1cyOxBGahZKb:bqgduzx) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L171)):*
> ```cpp
>     , settings(new Settings{*messageBoxManager})
>     , ipc(new IPC{settings->getCurrentProfileId()})
>     , toxSave(new ToxSave{*settings})
> ```

Do we need different types of brackets here?



_Code quote:_
```cpp
    : argc{(qInstallMessageHandler(logMessageHandler), argc_)}
    , argv{argv_}
    , qapp(new QApplication(argc, argv))
    , messageBoxManager(new MessageBoxManager{nullptr})
    , settings(new Settings{*messageBoxManager})
    , ipc(new IPC{settings->getCurrentProfileId()})
    , toxSave(new ToxSave{*settings})
```
___
*[src/appmanager.cpp, line 180 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iMZFd5kQw7OJV9VuE:-N-iMZFd5kQw7OJV9VuF:bhm5mza) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L180)):*
> ```cpp
>     // PosixSignalNotifier is used only for terminating signals,
>     // so it's connected directly to quit() without any filtering.
>     QObject::connect(&PosixSignalNotifier::globalInstance(), &PosixSignalNotifier::activated,
> ```

`QObject` can be deleted here and in similar ones because we have already inherited from it
___
*[src/main.cpp, line 30 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iN8NLBodALcLJKlGm:-N-iN8NM3GMVVPsj9Biy:bpi21hv) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/main.cpp#L30)):*
> ```cpp
>     QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
>     QGuiApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
> #endif
> ```

Optionally, it may be worth moving the `QGuiApplication` attribute setting to the `AppManager` constructor?



_Code quote:_
```cpp
#if (QT_VERSION >= QT_VERSION_CHECK(5, 6, 0))
    QGuiApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
    QGuiApplication::setAttribute(Qt::AA_UseHighDpiPixmaps);
#endif
```

anthonybilinski (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6603)* status: 0 of 1 LGTMs obtained

___
*[`CMakeLists.txt` line 318 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iEHv9-G-woweclcuc:-N1Y5IMNBNlE_NpbA05e:b-yjwtfw) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/CMakeLists.txt#L318)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

Put it in alphabetical order?
</blockquote></details>

Changed to stick all the top level files together in alphabetical order. Looks like before things were grouped by directory, which is why I just stuck it before nexus.cpp, but ipc.cpp is earlier and part of the same dir.
___
*[`src/appmanager.h` line 22 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iFcvt5xYqpAF9L8f7:-N1Y5xHR90kr8T4DOkq5:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.h#L22)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

It looks like an unnecessary inclusion
</blockquote></details>

Done.
___
*[`src/appmanager.h` line 44 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iGCCrAEepHHRhP0YR:-N1Y6_E53K5sYPqSMwCN:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.h#L44)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

Make `cleanup()` private for more encapsulation?
</blockquote></details>

Done.
___
*[`src/appmanager.cpp` line 165 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iH4lZ8QiXrjDE_ScH:-N1Y6jgH61cX_Jw0nEaa:b-7tlfg8) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L165)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

How does initialization work for `argc` here? 
Is there a reason why `QApplication` is not initialized using `argc_` and `argv_` directly from the `AppManager` arguments?
</blockquote></details>

This is a bit of a hack to still use the member initializer list, but run `qInstallMessageHandler` before completing that so that logs can be printed out from child class constructors.

It's using the C++ comma operator: https://en.cppreference.com/w/cpp/language/operator_other#Built-in_comma_operator. So it's running the log installer, then assigning argc_ to argc. This is needed because code can't be run in/before the member initializer list outside of an initialization.

Re: saving argc and argv to members at all - yeah that seems useless, not sure why I was doing that. I think because I read that argc and argv must outlive the QApplication, but that's already guaranteed since AppManager is only constructed by main. Removed the two members, and pass argc and argv arguments to QApplication directly, though the comma operator is now moved down to the qapp initializer.
___
*[`src/appmanager.cpp` line 171 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iHeuW1cyOxBGahZKa:-N1Y8bUr23GWv9Ixdo3v:bc9s0u2) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L171)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

Do we need different types of brackets here?
</blockquote></details>

Changed to use parentheses in all cases here.
___
*[`src/appmanager.cpp` line 180 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iMZFd5kQw7OJV9VuE:-N1Y8pq10M3D3G-O7oCg:b-896fix) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/appmanager.cpp#L180)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

`QObject` can be deleted here and in similar ones because we have already inherited from it
</blockquote></details>

Done.
___
*[`src/main.cpp` line 30 at r1](https://reviewable.io/reviews/qtox/qtox/6603#-N-iN8NLBodALcLJKlGm:-N1Y8vDG2FUJsN9YB0IS:b-kfmgio) ([raw file](https://github.com/qtox/qtox/blob/cea4c81a369fe72a2d850cb72e093201097cb86e/src/main.cpp#L30)):*
<details><summary><i>Previously, bodwok wrote…</i></summary><blockquote>

Optionally, it may be worth moving the `QGuiApplication` attribute setting to the `AppManager` constructor?
</blockquote></details>

They must be set before the QApplication is constructed, but I've moved those in to be called by the comma operator as well, and moved all three things called into AppManager::preConstructionInitialization to make it more clear

bodwok (commented):
In files with translations, tags are moved to the `<name>AppManager</name>` without changes? I noticed that now `<translation type="unfinished">` is marked everywhere.

> Reviewed 4 of 52 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6603#-:-N1yWSe20sTdWDaOXUfk:by4rsw5)* status: 0 of 1 LGTMs obtained

anthonybilinski:
This is done automatically by Qt's translation manager when we call lupdate: https://github.com/qTox/qTox/blob/6ad92bb63742d9bc6fff421af897a98505cde509/tools/update-translation-files.sh#L30.

The "unfinished" translations are still loaded by qTox, changing qTox to e.g. french and running `qtox -h` shows the "unfinished" IPv6 help option's output, but I'm guessing this would prompt users on weblate to re-confirm that the translations are valid since the context where the string is used in source changed. That's probably reasonable behaviour in general, but is a bit of busy work for the translators. I've marked all move translations in this case that were finished before back to finished for AppManager.

bodwok (commented):
:lgtm_strong:

> Reviewed 11 of 49 files at r3, 38 of 38 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6603#-:-N2ltIExBkf2i1tD6B4f:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
@bodwok our PR checks changed a bit, and now it uses github's own approval check rather than reviewable's LGTM check. Reviewable's LGTM includes an approval of the PR, but github's approval check only allows approving reviews from members of the org. I'll send an invite your way to the "Reviewers" team of qTox, feel free to add me on tox or send an email if you want to discuss (both addresses are on my github profile).

bodwok (approved):
Ok, thanks for the invitation

PR #6605, opened 2022-03-31
refactor(History): Split remove history query into transaction

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6605)

### DIFF
--- a/src/persistence/history.cpp
+++ b/src/persistence/history.cpp
@@ -270,64 +270,86 @@ void History::removeChatHistory(const ChatId& chatId)
         return;
     }
 
+    QVector<RawDatabase::Query> queries;
     QVector<QByteArray> boundParams;
-    QString queryString = QStringLiteral("DELETE FROM faux_offline_pending "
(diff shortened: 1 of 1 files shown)


sudden6:
What's the advantage of this split?

anthonybilinski:
There's not a big functional difference, mostly I think we should prefer transactions over multi-statement queries and binding arguments over inlining them, and this is a small step in that direction.

I think in this case the differences are:
1) better readability of where each query starts and ends, and which bound parameters belong to what query text
2) the error reported from SQL would say what query failed in the transaction, instead of the entire huge query

sudden6 (approved):
:lgtm_strong:

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6605#-:-N2rZgMm32iHJGroHaow:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6607, opened 2022-03-31
fix(chatlog): Display correct system message timestamp in chatlog

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6607)


sphaerophoria (approved):
Did my best to do some archeology here, I think https://github.com/qTox/qTox/pull/6310#pullrequestreview-769462548 is the only reason that it is how it is.

PR #6608, opened 2022-03-31
feat(History): Add a GUI error on database schema mismatch

### LINKED ISSUE #5683: Invalid db schema version doesn't correctly abort
##### Brief Description

OS: Ubuntu 19.04
qTox version: v1.16.3+
toxcore:  v0.2.9
Qt:  5.12.2
Hardware:  Desktop

Reproducible: Always

##### Steps to reproduce

1. Have a db history schema past an upgrade point.
2. Open qTox with that profile with an old qTox that doesn't know about that schema version.

##### Observed Behavior
1. `persistence/history.cpp:115 : Warning: Database version is newer than we currently support. Please upgrade qTox` is logged
2. qTox keeps running normally with the unknown schema.

##### Expected Behavior
1. qTox logs (critically) and refuses to load history, ideally showing a GUI pop-up stating why it's failing to load history.

##### Additional Info
1. Because the shared_ptr is copied into the function, the function is only resetting its own copy of the shared pointer, which worked before when it was a member variable.
2. Bug introduced in 316893ace9e54fcbccdcd9d53a615638a242b36a.

anthonybilinski:
Fix #5683

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6608)

### DIFF
--- a/src/persistence/db/upgrades/dbupgrader.cpp
+++ b/src/persistence/db/upgrades/dbupgrader.cpp
@@ -22,9 +22,11 @@
 #include "src/core/toxpk.h"
 #include "src/persistence/db/rawdatabase.h"
 #include "src/persistence/db/upgrades/dbto11.h"
+#include "src/widget/tool/imessageboxmanager.h"
 
 #include <QDebug>
 #include <QString>
+#include <QTranslator>
 
 namespace {
 constexpr int SCHEMA_VERSION = 11;
@@ -219,7 +221,7 @@ void addForeignKeyToBrokenMessages(QVector<RawDatabase::Query>& queries)
  * @note On future alterations of the database all you have to do is bump the SCHEMA_VERSION
  * variable and add another case to the switch statement below. Make sure to fall through on each case.
  */
-bool DbUpgrader::dbSchemaUpgrade(std::shared_ptr<RawDatabase>& db)
+bool DbUpgrader::dbSchemaUpgrade(std::shared_ptr<RawDatabase>& db, IMessageBoxManager& messageBoxManager)
 {
     // If we're a new dB we can just make a new one and call it a day
     bool success = false;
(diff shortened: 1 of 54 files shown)


sudden6:
Would be good to integrate this into a unit test too. Otherwise it looks good to me.

Chiitoo:
No biggie or anything but I feel like the

```
"Database version (%1) is newer than we currently support (%2). Please upgrade qTox."
```

might be missing a 'what', like so:

```
"Database version (%1) is newer than what we currently support (%2). Please upgrade qTox."
```

anthonybilinski (commented):
@chiitoo it doesn't sound better to me in the second way, and with some quick searching people seem to generally agree, i.e.: https://forum.wordreference.com/threads/better-than-what-we-have-anticipated.3751021/. I'm not sure about the grammar rules though, I learned English as a first language so didn't really learn grammar rules, just kind of absorbed them, so it's definitely possible something sounds right to me but isn't really :S

Do you have a reference to back that up?

@sudden6, I've added TestDbSchema::testNewerDb to cover the general case of db being newer than qTox supports, at least from DbUpgrader's perspective. Including checking that the error message is sent to IMessageBoxManager.

> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6608#-:-N1XztTi8ECNJ9LwCuJS:b8gjvex)* status: 0 of 1 LGTMs obtained

sudden6 (approved):
:lgtm_strong:

> Reviewed 43 of 53 files at r1, 10 of 10 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6608#-:-N1YgHUN3qfKu8rP5dDC:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

anthonybilinski:
I'll merge as-is, but I'm open to rewording in the future.

PR #6614, opened 2022-04-02
fix(FriendList): exclusion of possible undefined behavior

### LINKED ISSUE #6606: TestFriendListManager::testSortByName occasionally fails in CI
##### Brief Description

OS: All
qTox version: v1.17.4+
Commit hash:  0352ada08c7efa4ad6523071bde566f506daeb4d
toxcore:  v0.2.17
Qt:  Multiple

Reproducible: Rarely

##### Steps to reproduce

1. Run test_friendlistmanager

##### Observed Behaviour
Once in a while, this exact comparison fails:
`TestFriendListManager::testSortByName` fails with:
```
   Actual   (fromManager->getNameItem())  : ".test online"
   Expected (fromSortedVec->getNameItem()): "123"
   Loc: [/Users/runner/work/qTox/qTox/test/model/friendlistmanager_test.cpp(348)]
```
It's not easily reproducible, and there's no code changes between passes and fails. Just retrying the job is enough to get it to pass. It fails on any of the CI environments. I haven't seen it locally, but across the different CI jobs I have a lot more test runs in CI.

https://github.com/anthonybilinski/qTox/runs/5764770322?check_suite_focus=true#step:5:776

##### Expected Behaviour
1. Consistent test results.
2. Passing test results.

##### Additional Info
The inconsistent results seem like a race or memory safety issue, but AddressSanitizer finds no issues with the test, and the test seems completely synchronous.. so not sure what's going on here.

bodwok:
Fix #6606
Tested on Ubuntu. For always reproduced failure, I temporarily wrapped the contents of  `TestFriendListManager::testSortByName` in `for (int j = 0; j < 100; ++j) { ... }`

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6614)

### DIFF
--- a/src/model/friendlist/friendlistmanager.cpp
+++ b/src/model/friendlist/friendlistmanager.cpp
@@ -40,6 +40,11 @@ bool FriendListManager::getPositionsChanged() const
     return positionsChanged;
 }
 
+bool FriendListManager::getGroupsOnTop() const
+{
+    return groupsOnTop;
+}
+
 void FriendListManager::addFriendListItem(IFriendListItem *item)
 {
     if (item->isGroup() && item->getWidget() != nullptr) {
--- a/src/model/friendlist/friendlistmanager.h
+++ b/src/model/friendlist/friendlistmanager.h
@@ -38,6 +38,7 @@ public:
     bool needHideCircles() const;
     // If the contact positions have changed, need to redraw view
     bool getPositionsChanged() const;
+    bool getGroupsOnTop() const;
 
     void addFriendListItem(IFriendListItem* item);
     void removeFriendListItem(IFriendListItem* item);
@@ -69,11 +70,11 @@ private:
 
     bool byName = true;
     bool hideCircles = false;
-    bool groupsOnTop;
-    bool positionsChanged;
-    bool needSort;
+    bool groupsOnTop = true;
+    bool positionsChanged = false;
+    bool needSort = false;
     QVector<IFriendListItemPtr> items;
     // At startup, while the size of items is less than countContacts, the view will not be processed to improve performance
-    int countContacts;
+    int countContacts = 0;
 
 };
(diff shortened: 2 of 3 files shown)


anthonybilinski (changes requested):
Nice find and fix.

I looked at adding automatic detection for reads from uninitialized variables through MemorySanitizer, but it gives false positive if all of its dependencies aren't built with the same flag, so it would mean rebuilding all of libstd++ and Qt etc. so not going to go through with that right now.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6614#-:-MzhHtzC-axT_B_fKvM7:b-niu79j)* status: 0 of 1 LGTMs obtained

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/qtox/qtox/6614#-MzhITJY0dSwmdFk-4e5):*
Your commit title is invalid, although it's not caught by CI due to a bug in the CI check which I've fixed here: https://github.com/qTox/qTox/pull/6615

According to the check, the text inside of the brackets should be 12 characters or less, so "FriendListManager" is too long. Could you reword that to avoid it?

bodwok (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6614)* status: 0 of 1 LGTMs obtained

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/qtox/qtox/6614#-MzhITJY0dSwmdFk-4e5):*

Your commit title is invalid, although it's not caught by CI due to a bug in the CI check which I've fixed here: https://github.com/qTox/qTox/pull/6615

According to the check, the text inside of the brackets should be 12 characters or less, so "FriendListManager" is too long. Could you reword that to avoid it?
</blockquote></details>

Done

anthonybilinski (approved):
:lgtm_strong:

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6614#-:-MzifAeA9LaWlgcrVBKe:b-jbsf8z)* status: :shipit: complete! 1 of 1 LGTMs obtained

PR #6621, opened 2022-04-08
fix(contacts): Multiple friends contacts system dependent

### LINKED ISSUE #6412: Segmentation fault when opening a contact
##### Brief Description

OS: Fedora 35
qTox version: qTox v1.17.3-316-g12fc33ee
Commit hash:  c0e9a3b79609681e5b9f6bbf8f9a36cb1993dc5f
toxcore:  toxcore version  0.2.12
Qt:  version: 5.15.2  
…

Reproducible: Always

##### Steps to reproduce

1. Have at least one tox contact in your friends list
2. Click on that contact to open your conversation history with them. The contact doesn't have to be online
3. Close the conversation window with them
4. Click on that same contact again to open your conversation history with them a second time
5. If the conversation history window opens without any issues, go back to step 3 and repeat all the steps from there again
6. Eventually, when clicking on the contact to open their conversation history window, the entire process will crash. If qTox was open from bash in a terminal window in the background, bash will report a segmentation fault when the process crashes

##### Observed Behavior
Process crashes unexpectedly.

##### Expected Behavior
Regardless of how many times I close and re-open the exact same conversation history window with my contact, the qTox process should never crash.

##### Additional Info
Fedora doesn't have a package with the debug symbols for qTox. As such, I cannot provide a reliable backtrace with GDB, sadly.
For what it's worth, as unreliable as it is, the backtrace that I did get can be found at: https://pastebin.com/a5xL4az1

I don't have much technical expertise but this looks like a nullptr de-referencing issue to me. 

The qtox.log can be found at https://pastebin.com/3FGYDCxJ
Please note that both the backtrace and the log are set to be auto-deleted after 1 calendar Year since the moment they were created (around the same time that this issue was created).

bodwok:
Looks like the issue related to graphics processing in Linux. On Fedora 35 it occurs if Wayland is used (setted by default). If switch to Xorg mode or run qTox with `-platform xcb` key on Fedora 35, then this error was not reproduced for me. 
When using Wayland, if the `ContentDialog` closes, its destructor is not called immediately, but manages to initialize the `currentDialog` pointer in `ContentDialogManager::onDialogActivate`. Next the `ContentDialog` destructor is called, and `currentDialog` becomes a dangling pointer. 

To get around this I added a flag that marks `ContentDialog` as invalid. Also another option that works for me too: use `qtox -platform xcb`.

Fix #6412

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6621)

### DIFF
--- a/src/widget/contentdialog.cpp
+++ b/src/widget/contentdialog.cpp
@@ -645,22 +645,6 @@ void ContentDialog::setStatusMessage(const ToxPk& friendPk, const QString& messa
 }
 
 /**
- * @brief Sets flag that means destructor will be called soon
- */
-void ContentDialog::setValid(bool val)
-{
-    valid = val;
-}
-
-/**
- * @brief Flag means destructor will be called soon
- */
-bool ContentDialog::isValid() const
-{
-    return valid;
-}
-
-/**
  * @brief Update friend widget name and position.
  * @param friendId Friend Id.
  * @param alias Alias to display on widget.
--- a/src/widget/contentdialog.cpp
+++ b/src/widget/contentdialog.cpp
@@ -645,6 +645,22 @@ void ContentDialog::setStatusMessage(const ToxPk& friendPk, const QString& messa
 }
 
 /**
+ * @brief Sets flag that means destructor will be called soon
+ */
+void ContentDialog::setValid(bool val)
+{
+    valid = val;
+}
+
+/**
+ * @brief Flag means destructor will be called soon
+ */
+bool ContentDialog::isValid() const
+{
+    return valid;
+}
+
+/**
  * @brief Update friend widget name and position.
  * @param friendId Friend Id.
  * @param alias Alias to display on widget.
--- a/src/widget/contentdialog.h
+++ b/src/widget/contentdialog.h
@@ -85,6 +85,9 @@ public:
 
     void setStatusMessage(const ToxPk& friendPk, const QString& message);
 
+    void setValid(bool val);
+    bool isValid() const;
+
 signals:
     void friendDialogShown(const Friend* f);
     void groupDialogShown(Group* g);
@@ -149,4 +152,6 @@ private:
     FriendList& friendList;
     GroupList& groupList;
     Profile& profile;
+
+    bool valid = true;
 };
--- a/src/widget/contentdialog.h
+++ b/src/widget/contentdialog.h
@@ -85,9 +85,6 @@ public:
 
     void setStatusMessage(const ToxPk& friendPk, const QString& message);
 
-    void setValid(bool val);
-    bool isValid() const;
-
 signals:
     void friendDialogShown(const Friend* f);
     void groupDialogShown(Group* g);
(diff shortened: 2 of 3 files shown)


anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6621)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6621#-N-nvkzfCUJNehqKOOp-:-N-nvkzfCUJNehqKOOp0:bm88gu5) (no related file):*
What's the root cause of the crash here?

It seems like on window close:
* ContentDialog::closeEvent emits willClose then accepts the event
* ContentDialogManager::addContentDialog connects ContentDialog::willClose to ContentDialogManager::onDialogClose
* ContentDialogManager::onDialogClose calls removeDialog
* contentdialogmanager.cpp's removeDialog, which deletes the ContentDialog

At some point(?) during this, the ContentDialog is destroyed by Qt due to `setAttribute(Qt::WA_DeleteOnClose);` in ContentDialog's constructor.

When the second dialog is being opened, `Widget::openDialog` calls `Widget::addFriendDialog` which calls into ContentDialogManager to get the content dialog for the friend.

So is there a race on destruction here, that the content dialog isn't being destroyed by the time Widget tries to get the current ContentDialog? If so I think we should address the race to make sure destruction is complete rather than create invalid objects that are semi-destructed.

Changing `ContentDialogManager::addContentDialog` to connect the willClose signal synchronously: `connect(&dialog, &ContentDialog::willClose, this, &ContentDialogManager::onDialogClose, Qt::QueuedConnection);` seems to stop the problem on Fedora 35 - so that may be an alternative.

---

bodwok (commented):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6621)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6621#-N-nvkzfCUJNehqKOOp-:-N02i4ac-D6hXvDPhWxf:b-i9or6q) (no related file):*
Now I found that the difference between Xorg and Wayland (Fedora) in the return value of the isActiveWindow property:

>* ContentDialog::closeEvent emits willClose then accepts the event
>* ContentDialogManager::addContentDialog connects ContentDialog::willClose to ContentDialogManager::onDialogClose
>* ContentDialogManager::onDialogClose calls removeDialog
>* contentdialogmanager.cpp's removeDialog, which deletes the ContentDialog

If we continue your call list we will get (if using Wayland):

Without this PR:
1. ContentDialog::changeEvent (here `isActiveWindow() == true` and emit the `activated` signal)
2. ContentDialogManager::onDialogActivate (here the `currentDialog` pointer is set to the dialog that will be deleted)
3. ContentDialog::~ContentDialog

With this PR:
1. ContentDialog::changeEvent (here `isActiveWindow() == false`)
2. ContentDialog::~ContentDialog

The fix you suggested works and ContentDialogManager::onDialogActivate is not called before ContentDialog::~ContentDialog

---

PR #6625, opened 2022-04-19
chore(deps): Update c-toxcore to v0.2.18

anthonybilinski:
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6625)


sudden6:
c-toxcore recently introduced a git submodule, which probably needs to be checked out additionally for builds to pass again.

robinlinden:
msgpack-c was also dropped, so unless you're using that outside of c-toxcore, it can be removed.

anthonybilinski:
We download the source bundle rather than git cloning, so the submodule can't be checked out directly, but I've changed the download archive to the newly added source bundle: https://github.com/TokTok/c-toxcore/issues/2192.

Also removed all msgpack-c scripts.

robinlinden (approved):

PR #6640, opened 2022-05-29
chore(Qt): Replace QGuiApplication usage with QApplication

anthonybilinski:
QGuiApplication is for QML projects, QApplication is for Qt Widget projects.

We use QApplication everywhere else, so use here too.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6640)


sudden6 (approved):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6640)* status: 0 of 1 LGTMs obtained

PR #6645, opened 2022-06-22
feat(chatlog): Add custom web handler as hyperlink

dvn0:
Adds support for custom web scheme handlers. See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#permitted_schemes) for more info.

---
This change is [Reviewable](https://reviewable.io/reviews/qtox/qtox/6645)

### DIFF
--- a/src/chatlog/textformatter.cpp
+++ b/src/chatlog/textformatter.cpp
@@ -93,6 +93,7 @@ const QVector<QRegularExpression> URI_WORD_PATTERNS = {
     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(magnet:[?]((xt(.\d)?=urn:)|(mt=)|(kt=)|(tr=)|(dn=)|(xl=)|(xs=)|(as=)|(x.))[\S| ]+))")),
     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(gemini://\S+))")),
     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(ed2k://\|file\|\S+))")),
+    QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(web+\S+:\S+))")),
 };
 
 


Tha14 (approved):
Looks fine

sudden6:
@dvn0 please add at least one positive and possibly one negative test case to: https://github.com/qTox/qTox/blob/master/test/chatlog/textformatter_test.cpp

dvn0:
Probably not going to get around to making tests anytime soon, unfortunately. Sorry.

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6645)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6645#-N6BHNimEzjVknnEPkDi:-N6BHNin5gPsrkuYDWU9:b-yzf2jl) (no related file):*
My understanding of web+ scheme is that a website asks a web browser to ask a user to register the mapping from a web+<> to a web URL. I don't really see where qTox comes in to this - with qTox having no way to register web+ URIs to http(s) URLs, won't a link clicked in qTox just fail to open without ever getting to the web browser to resolve the mapping, even if it exists there?

Have you seen this work with a website that registers a web+ URI? Testing myself by adding a registration in firefox's console makes it so that those links in the browser then prompt to open the link with the registered URL, but from qTox I just get "gio: The specified location is not supported" and similar from command line.

Additionally firefox' general settings under applications shows the new link registration, but gnome's "default applications" is unchanged.

___

dvn0:
@anthonybilinski - sounds like it doesn't work in your environment, but for me, if I open firefox from the command line with a registered `web+[...]` address then it opens as expected. It will open the website which is registered for the given schema. This should work for me from qTox as my `xdg-open` has the `web+[...]` protocol schema assigned to open with firefox.

anthonybilinski (changes requested):
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6645)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6645#-N6BHNimEzjVknnEPkDi:-N6Cc7hS6CBvF7rYo63i:b-9eab8h) (no related file):*

My understanding of web+ scheme is that a website asks a web browser to ask a user to register the mapping from a web+<> to a web URL. I don't really see where qTox comes in to this - with qTox having no way to register web+ URIs to http(s) URLs, won't a link clicked in qTox just fail to open without ever getting to the web browser to resolve the mapping, even if it exists there?

Have you seen this work with a website that registers a web+ URI? Testing myself by adding a registration in firefox's console makes it so that those links in the browser then prompt to open the link with the registered URL, but from qTox I just get "gio: The specified location is not supported" and similar from command line.

Additionally firefox' general settings under applications shows the new link registration, but gnome's "default applications" is unchanged.
</blockquote></details>

Ah I see, thanks for the explanation. Firefox opened with web+[...]: works for me as well, but the piece my environment is missing is the web+[...] -> Firefox link. With that, I can see how qTox would dispatch the link properly and have no issue with this conceptually. Creating that link seems outside of qTox's scope, so not really something we need to worry about. Creating the links in chat still seems reasonable for us to do.

___
___
*[`src/chatlog/textformatter.cpp` line 96 at r1](https://reviewable.io/reviews/qtox/qtox/6645#-N6CoGr9EPsHOwqOlMEP:-N6CoGr9EPsHOwqOlMEQ:b4g47w3) ([raw file](https://github.com/qtox/qtox/blob/2239ebb32ea59561baaa25aa1f97f121f370d6c4/src/chatlog/textformatter.cpp#L96)):*
> ```cpp
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(gemini://\S+))")),
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(ed2k://\|file\|\S+))")),
>     QRegularExpression(QStringLiteral(R"((?<=^|\s)\S*(web+\S+:\S+))")),
> ```

I believe not including an argument in the link is valid in the spec, and is accepted by Firefox testing, i.e. both "web+burger:" as well as "web+burger:foo" are valid. If you agree, the regex is overly restrictive, and the last \S+ could be changed to \S*.

PR #6650, opened 2022-07-04
feat(chatform, audio): Adding shortut SHIFT+` (~ tilde) for muting mic.

fancyfarmer:
OS: Ubuntu
qTox version: master
Commit hash: https://github.com/qTox/qTox/commit/2197bce610f2abf5e561b1ee528313660b0f0fb9
toxcore: 0.2.18
Qt: 5.12.8

https://github.com/qTox/qTox/issues/6574
In this commit unfortunately focus still has to be on qTox window.

### DIFF
--- a/src/widget/form/chatform.cpp
+++ b/src/widget/form/chatform.cpp
@@ -58,6 +58,7 @@
 #include <QScrollBar>
 #include <QSplitter>
 #include <QStringBuilder>
+#include <QShortcut>
 
 #include <cassert>
 
@@ -221,6 +222,9 @@ ChatForm::ChatForm(Profile& profile_, Friend* chatFriend, IChatLog& chatLog_,
     setAcceptDrops(true);
     retranslateUi();
     Translator::registerHandler(std::bind(&ChatForm::retranslateUi, this), this);
+
+    // shortcut for mute (SHIFT+ `), (SHIFT + grave accent, gives tilde)
+    new QShortcut(Qt::SHIFT | 0x60, this, SLOT(onMicMuteShortcutToggle()));
 }
 
 ChatForm::~ChatForm()
@@ -456,6 +460,15 @@ void ChatForm::onMicMuteToggle()
     CoreAV* av = core.getAv();
     av->toggleMuteCallInput(f);
     updateMuteMicButton();
+}
+
+void ChatForm::onMicMuteShortcutToggle()
+{
+    CoreAV* av = core.getAv();
+    if(av->isCallActive(f))
+    {
+    	onMicMuteToggle();
+    }
 }
 
 void ChatForm::onVolMuteToggle()
--- a/src/widget/form/chatform.h
+++ b/src/widget/form/chatform.h
@@ -102,6 +102,7 @@ private slots:
     void onAnswerCallTriggered(bool video);
     void onRejectCallTriggered();
     void onMicMuteToggle();
+    void onMicMuteShortcutToggle();
     void onVolMuteToggle();
 
     void onFriendStatusChanged(const ToxPk& friendPk, Status::Status status);


anthonybilinski (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/qtox/qtox/6650)* status: 0 of 1 LGTMs obtained

*[a discussion](https://reviewable.io/reviews/qtox/qtox/6650#-N6HcWnl222XSH-jUgCK:-N6HcWnl222XSH-jUgCL:b28h1fi) (no related file):*
re: the window requiring focus, it's not simple to have a cross platform way to read shortcuts without focus. It was mostly implemented a while ago here: https://github.com/qTox/qTox/pull/6185 but I need to get around to finish that off. Until then, this is definitely better than nothing.

___
___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/qtox/qtox/6650#-N6HbjuCAcgo8k8NEI2L):*
I've updated the commit message to just start with feat(audio), to pass the commit check which doesn't want a subject longer than 12 characters.
___
*[`-- commits` line 4 at r2](https://reviewable.io/reviews/qtox/qtox/6650#-N6HbkgH3m77QjUhalaK):*
I've combined the two similar commits and third merge commit into a single commit that still includes your white-space fix. I've pushed this to your PR branch.
___
*[`src/widget/form/chatform.cpp` line 227 at r3](https://reviewable.io/reviews/qtox/qtox/6650#-N6Hf_ei0jDJlDDxYMoX:-N6Hf_ei0jDJlDDxYMoY:b-h5lb2e) ([raw file](https://github.com/qtox/qtox/blob/3ecf2e03c44f73fda9789b87dcf28cd74dce474d/src/widget/form/chatform.cpp#L227)):*
> ```cpp
> 
>     // shortcut for mute (SHIFT+ `), (SHIFT + grave accent, gives tilde)
>     new QShortcut(Qt::SHIFT | 0x60, this, SLOT(onMicMuteShortcutToggle()));
> ```

Any reason to register the shortcut using shift + \`, rather than just registering `~` directly?
___
*[`src/widget/form/chatform.cpp` line 465 at r3](https://reviewable.io/reviews/qtox/qtox/6650#-N6HfjX93Qtq709kY5M3:-N6HfjX93Qtq709kY5M4:b-5hb6zc) ([raw file](https://github.com/qtox/qtox/blob/3ecf2e03c44f73fda9789b87dcf28cd74dce474d/src/widget/form/chatform.cpp#L465)):*
> ```cpp
> }
> 
> void ChatForm::onMicMuteShortcutToggle()
> ```

Can this be enabled for group calls as well?

Currently ChatForm (friends) and GroupChatForm (groups) both handle their own mute state and button update, instead of them being handled by the common GenericChatForm. I'd be fine with just duplicating this logic in GroupChatForm rather than refactoring that now, since abstracting signals isn't that straightforward.

Green-Sky:
just want to add that, on a qwertz keyboard SHIFT+´ is ` 
and ALTGR++ is ~

Chiitoo:
Somewhat surprisingly (or maybe not) this does also work with layouts where tilde comes from a dead key (for example press and hold down altgr, press and release tilde, release altgr, press spacebar).

It is slightly cumbersome, but it works, although I could only make it work by having focus on the mute button.  That is, I had to first click on the mute button, and only then it would work.  It would not work if focus was in the typing area for example, but perhaps that to be expected?

Chiitoo:
D'oh, of course it would seem to work if using space to trigger the dead key, when the button has focus, since that's what the space key tends to do with buttons.  (I blame the sleepy.)

With that said I could not make the shortcut to work (so I guess it would have indeed been at least a bit surprising if it did work).

Testing with a US keyboard layout it does work as long as either the button or text area has focus (but not the typing area which I guess is understandable).

